PR #4, opened 2017-04-23
Fixed translate usage doc.

Manouchehri:


aleclearmind:
Thanks for reporting this, it will be fixed in the next push along with some other doc improvements.

PR #23, opened 2020-12-01
Exception helper: add attribute noreturn

antoniofrighetto:
This MR to simply make sort of a promise to the compiler that we will never return from `raise_exception_helper`, and possibly hint it to fire up some more optimization? Can't see any side effects with this change.

### DIFF
--- a/runtime/support.c
+++ b/runtime/support.c
@@ -496,10 +496,10 @@ void exception_warning(Reason Code,
 }
 
 // Helper function used to raise an exception
-void raise_exception_helper(Reason Code,
-                            target_reg Source,
-                            target_reg Target,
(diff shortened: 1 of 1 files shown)


aleclearmind:
Uhm, looks good. Does it fix any bugs or is it just an improvement?

antoniofrighetto:
A bit of an improvement - looks like `rax` is no more popped off the stack - though `_Unwind_RaiseException` is no more a tail call now. Shall I just drop the call to `abort`?

aleclearmind:
It's OK. Let's not optimize at this granularity unless it's a bottleneck.
Will test and merge soon, thanks!

aleclearmind:
Merged in 59b8c6c, thanks!

PR #29, opened 2020-12-16
Add RecursiveCoroutine headers and tests

pfez:
RecursiveCoroutines are a facility intended to be used as-drop in replacement of recursive functions.

They provide the following features.
- They can be written almost as regular recursive functions, with 4 caveats.
  1. A recursive coroutine that returns a type `T`, needs to be declared to return a `RecursiveCoroutine<T>`.
  2. Inside the body of a recursive coroutine, when recursively calling another recursive coroutine, the recursive call needs to be prepended by the new keywork `rc_recur`.
  3. Inside the body of a recursive coroutine, the `return` statement needs to be substituted with `rc_return`.
  4. When launching a recursive coroutine `A` from a function that is not a recursive coroutine, `A` needs to be called with the provided dedicated template wrapper `rc_run`. The syntax is the following `rc_run(A, arg0, arg1, ...)`. This is necessary to enable swapping off recursive coroutine and fall back to regular recursion for debug. See below for how to do it.
- Unlike regular recursive functions, they don't use the system stack for recursion. They use a custom heap-allocated stack to manage recursion. This makes them more robust for implementing recursive functions that manipulate user-defined input, because they are much less likely to trigger stack overflow.
- They can be turned off compiling with `-DDISABLE_RECURSIVE_COROUTINES`, falling back to regular recursion, for debug purposes.
- They support both direct and indirect recursion, i.e. a recursive coroutine A can recursively call itself, or it can recursively call another recursive coroutine B, which in turns recursively calls A.

### DIFF
--- a/include/revng/ADT/RecursiveCoroutine-coroutine.h
+++ b/include/revng/ADT/RecursiveCoroutine-coroutine.h
@@ -1,0 +1,191 @@
+#pragma once
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include <experimental/coroutine>
+#include <optional>
+
+#include "revng/Support/Assert.h"
+
+struct PromiseBase {
+
+  auto initial_suspend() const { return std::experimental::suspend_always(); }
+  auto final_suspend() const noexcept {
+    return std::experimental::suspend_always();
+  }
+
+  [[noreturn]] void unhandled_exception() const { std::terminate(); }
+
+  PromiseBase *Callee = nullptr;
+};
+
+template<typename ReturnT = void>
+struct [[nodiscard("RecursiveCoroutine is discarded without running "
+                   "it")]] RecursiveCoroutine {
+public:
+  struct promise_type;
+
+public:
+  using coro_handle = std::experimental::coroutine_handle<promise_type>;
+
+public:
+  RecursiveCoroutine(coro_handle H) : OwnedHandle(H) {}
+  RecursiveCoroutine() : RecursiveCoroutine(coro_handle{}) {}
+
+  // Not copyable, because we don't want OwnedHandle to be destroyed twice
+  RecursiveCoroutine &operator=(const RecursiveCoroutine &);
+  RecursiveCoroutine(const RecursiveCoroutine &);
+
+  // Movable, but we clean up the OwnedHandle, because we don't want it to be
+  // destroyed twice in the destructor
+  RecursiveCoroutine &operator=(RecursiveCoroutine &&Other) {
+    this->OwnedHandle = Other.OwnedHandle;
+    Other.OwnedHandle = {};
+    return *this;
+  }
+  RecursiveCoroutine(RecursiveCoroutine && Other) { *this = std::move(Other); }
+
+  ~RecursiveCoroutine() {
+    revng_assert(OwnedHandle and OwnedHandle.done());
+    OwnedHandle.destroy();
+    OwnedHandle = {};
+  }
+
+  bool await_ready() const { return false; }
+
+  // TODO: I wanted to do something like the following
+  //
+  // template<typename CallerReturnT>
+  // void await_suspend(std::coroutine_handle<RecursiveCoroutine<CallerReturnT>>
+  //                    CallerHandle)
+  //
+  // but I had to fight the template parameter type inference engine and I gave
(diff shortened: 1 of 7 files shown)


include/revng/ADT/RecursiveCoroutine-coroutine.h
```diff
@@ -0,0 +1,187 @@
+#pragma once
+
```
aleclearmind, line 2: License notice?

tests/unit/recur.cpp
```diff
@@ -0,0 +1,135 @@
...
+    auto End = high_resolution_clock::now();
+
+    if (I != 0) {
+      Average += (duration_cast<microseconds>(End - Start).count() / Repeat);
+    }
+
+    revng_check(X == 34);
+  }
+
+  printf("Average: %lld\n", Average);
```
aleclearmind, line 134: Let's drop `printf`s

aleclearmind (changes requested):

include/revng/ADT/RecursiveCoroutine-coroutine.h
pfez: Added to all the files!

tests/unit/recur.cpp
pfez: Done this as well.

pfez:
If you have no other comments this should be good to go now!

tests/unit/UnitTests.cmake
```diff
@@ -215,3 +215,43 @@ target_link_libraries(test_genericgraph
...
+target_compile_definitions(test_recursive_coroutines_fallback
+  PRIVATE "BOOST_TEST_DYN_LINK=1")
+target_include_directories(test_recursive_coroutines_fallback
+  PRIVATE "${CMAKE_SOURCE_DIR}")
+target_link_libraries(test_recursive_coroutines_fallback
+  revngSupport
+  ${LLVM_LIBRARIES})
+add_test(NAME test_recursive_coroutines_fallback COMMAND ./bin/test_recursive_coroutines_fallback)
+set_tests_properties(test_recursive_coroutines_fallback PROPERTIES LABELS "unit")
+set_target_properties(test_recursive_coroutines_fallback PROPERTIES COMPILE_FLAGS "${COMPILE_FLAGS} -DDISABLE_RECURSIVE_COROUTINES" )
```
aleclearmind, line 245: `${COMPILE_FLAGS}` doesn't do what you think here.

tests/unit/UnitTests.cmake
```diff
@@ -215,3 +215,43 @@ target_link_libraries(test_genericgraph
...
+target_compile_definitions(test_recursive_coroutines_iterative
+  PRIVATE "BOOST_TEST_DYN_LINK=1")
+target_include_directories(test_recursive_coroutines_iterative
+  PRIVATE "${CMAKE_SOURCE_DIR}")
+target_link_libraries(test_recursive_coroutines_iterative
+  revngSupport
+  ${LLVM_LIBRARIES})
+add_test(NAME test_recursive_coroutines_iterative COMMAND ./bin/test_recursive_coroutines_iterative)
+set_tests_properties(test_recursive_coroutines_iterative PROPERTIES LABELS "unit")
+set_target_properties(test_recursive_coroutines_iterative PROPERTIES COMPILE_FLAGS "${COMPILE_FLAGS} -DITERATIVE" )
```
aleclearmind, line 257: Very redundant, I'm factoring it out in a macro

tests/unit/UnitTests.cmake
```diff
@@ -215,3 +215,43 @@ target_link_libraries(test_genericgraph
   ${LLVM_LIBRARIES})
 add_test(NAME test_genericgraph COMMAND ./bin/test_genericgraph)
 set_tests_properties(test_genericgraph PROPERTIES LABELS "unit")
+
+
+#
+# test_recursive_coroutines
+#
+
+revng_add_private_executable(test_recursive_coroutines "${SRC}/recur.cpp")
```
aleclearmind, line 224: `recur.cpp` was a temporary name. I'm renaming it to `RecursiveCoroutine.cpp`.

aleclearmind:
I'm taking ownership of this PR, no further changes required from your side.

aleclearmind:
Merged in e1b283b0b682877fcb6a8a215022520479d005ec. Thanks!

PR #31, opened 2021-01-04
Refactor revng-merge-dynamic

fcremo:
- Split in multiple files the existing revng-merge-dynamic script
- Add the option to merge additional LOAD segments in the resulting binary (needed for CRISPR)
- Align the new .dynstr to a 4 byte boundary
- Ensure the last verneed entry is marked as such (vn_next == 0)
- Ensure LOAD PHDRs are listed in ascending order, as mandated by ELF spec


CMakeLists.txt
```diff
@@ -153,11 +153,13 @@ configure_file(include/revng/Runtime/commonconstants.h "${CMAKE_BINARY_DIR}/shar
 configure_file(runtime/early-linked.c "${CMAKE_BINARY_DIR}/share/revng/early-linked.c" COPYONLY)
 configure_file(scripts/revng "${CMAKE_BINARY_DIR}/bin/revng")
 configure_file(scripts/revng-merge-dynamic "${CMAKE_BINARY_DIR}/bin/revng-merge-dynamic" COPYONLY)
+file(COPY scripts/revng_merge_dynamic DESTINATION "${CMAKE_BINARY_DIR}/bin")
```
fcremo, line 156: Note: this is needed because `orc install --test` is performed before installation, so the package would not be found.
It does not end up in `root/bin` (the `install` directive below copies the module to `lib/python` as agreed)

aleclearmind: Use `configure_file`, `file(COPY ...)` is run at configure-time and never again.

scripts/revng_merge_dynamic/__main__.py
```diff
@@ -0,0 +1,64 @@
+import argparse
+import sys
+
+from .log import set_verbose
+from .merge_dynamic import merge_dynamic
+
+parser = argparse.ArgumentParser(
```
aleclearmind, line 7: Why global? Can we move it in `main`?

scripts/revng_merge_dynamic/__main__.py
```diff
@@ -0,0 +1,64 @@
...
+    "--merge-load-segments",
+    action="store_true",
+    help="Merge the LOADed segments from the source ELF into the output ELF."
+)
+
+def main():
+    args = parser.parse_args()
+    set_verbose(args.verbose)
+
+    with sys.stdout if args.output == "-" else open(args.output, "wb") as output_file, \
```
aleclearmind, line 48: Use a support variable for this, pretty confusing as is.

CMakeLists.txt
aleclearmind: Does this mean that while in the build directory we will use the installed version? Not OK!

scripts/revng_merge_dynamic/__main__.py
```diff
@@ -0,0 +1,64 @@
...
+            open(args.source, "rb") as source_file, \
+            open(args.to_extend, "rb") as to_extend_file:
+        base = int(args.base, base=0)
+        retcode = merge_dynamic(
+            to_extend_file,
+            source_file,
+            output_file,
+            base=base,
+            merge_load_segments=args.merge_load_segments,
+        )
```
aleclearmind, line 58:
```python
        retcode = merge_dynamic(to_extend_file,
                                source_file,
                                ...
                                merge_load_segments=args.merge_load_segments)
```

scripts/revng_merge_dynamic/file_util.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def file_size(file):
+    file.seek(0, 2)
+    return file.tell()
+
+
+def read_at_offset(file, start, size):
+    file.seek(start)
+    return file.read(size)
```
aleclearmind, line 12: Unused?

scripts/revng_merge_dynamic/__main__.py
```diff
@@ -0,0 +1,64 @@
...
+from .merge_dynamic import merge_dynamic
+
+parser = argparse.ArgumentParser(
+    description="Merge the dynamic portions of the translate ELF with the one from the host ELF."
+)
+parser.add_argument(
+    "to_extend",
+    metavar="TO_EXTEND",
+    help="The ELF to extend.",
+)
```
aleclearmind, line 14:
```
parser.add_argument("to_extend",
                    metavar="TO_EXTEND",
                    help="The ELF to extend.")
```
Same everywhere else. Let's be coherent.

aleclearmind:
1. [ ] Follow the `revng` conventions. Indentation is two spaces.
2. [ ] Not sure about helper files with 2-3 functions used only in another file. Do you envision other users? `log.py` might share code with the `revng` script, but the others?
3. [ ] This seems like a big change with a small commit message. In terms of features/bugfix is that all?
4. [ ] Add copyright notice
5. [ ] Does this work even outside the build directory? Using the installed, version?

scripts/revng_merge_dynamic/__main__.py
fcremo: I've pretty much always seen the argument parser as a global variable in all the codebases, in this case since it has no side effects and nobody has any reason to import `__main__.py` anyway there's no downside I can think of (note: all the other modules do not have `parser` in their scope).
But doing it as you sugges has no downsides as well, I guess it's ultimately up to personal preference/conventions so I'll move it into `main`, no problem

aleclearmind: It's a way to prevent people from importing it. In general, let's avoid global variables and let's try to keep the scope of things as short of possible.

scripts/revng_merge_dynamic/file_util.py
fcremo: Yep, I removed it

aleclearmind:
Running on the CI, should be automerged soon!
Thanks!

aleclearmind:
This has been merged in c5680b2120546c9e4379ff49a2a4a02496574088.
Thanks!

PR #35, opened 2021-01-07
Convert GCBI to new PM

antoniofrighetto:
Given the recent improvements to GCBI, this may be useful before merging the new SA.


include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
```diff
@@ -250,41 +243,77 @@ class GeneratedCodeBasicInfo : public llvm::ModulePass {
 
   /// \brief Return the program counter of the next (i.e., fallthrough)
   ///        instruction of \p TheInstruction
-  MetaAddress getNextPC(llvm::Instruction *TheInstruction) const {
+  MetaAddress getNextPC(llvm::Instruction *TheInstruction) {
```
aleclearmind, line 246: Maybe you wanted to make this `static`?

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
```diff
@@ -250,41 +243,77 @@ class GeneratedCodeBasicInfo : public llvm::ModulePass {
...
+      if (auto *Call = getCallTo(Previous, "function_call"))
         return Call;
 
-      if (not isMarker(&*It))
-        return nullptr;
+      Previous = getPrevious(Previous);
     }
 
     return nullptr;
   }
```
aleclearmind, line 269: `getPrevious` might be nicer but it's less efficient.
Did you find the function so unreadable?

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
```diff
@@ -250,41 +243,77 @@ class GeneratedCodeBasicInfo : public llvm::ModulePass {
...
+
+  /// \brief Return the successor basic block given a function_call marker.
+  static llvm::BasicBlock *getSuccessor(llvm::Instruction *T) {
+    if (auto *Call = getFunctionCall(T)) {
+      if (auto *Successor = llvm::dyn_cast<llvm::BlockAddress>(Call->getOperand(0)))
+        return Successor->getBasicBlock();
+    }
+
+    return nullptr;
+  }
```
aleclearmind, line 291: "Successor" does not seem like an appropriate name. What about Callee?
Also, did you see `getFunctionCallCallee`?
There's quite some duplication between GCBI and `revng.h`/`IRHelpers.h`. Let's not make this worse.

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
```diff
@@ -250,41 +243,77 @@ class GeneratedCodeBasicInfo : public llvm::ModulePass {
...
+
+  /// \brief Return the fall-through basic block given a function_call marker.
+  static llvm::BasicBlock *getFallthrough(llvm::Instruction *T) {
+    if (auto *Call = getFunctionCall(T)) {
+      auto *Fallthrough = llvm::cast<llvm::BlockAddress>(Call->getOperand(1));
+      return Fallthrough->getBasicBlock();
+    }
+
+    return nullptr;
+  }
```
aleclearmind, line 305:
```
include/revng/FunctionCallIdentification/FunctionCallIdentification.h:  llvm::BasicBlock *getFallthrough(llvm::Instruction *T) const {
```

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
```diff
@@ -414,3 +443,31 @@ struct BlackListTrait<const GeneratedCodeBasicInfo &, llvm::BasicBlock *>
     return !this->Obj.isTranslated(Value);
   }
 };
+
+/// Analysis to compute GCBI.
```
aleclearmind, line 447: Too tautological, summarize what information GCBI provides briefly.

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
```diff
@@ -414,3 +443,31 @@ struct BlackListTrait<const GeneratedCodeBasicInfo &, llvm::BasicBlock *>
...
+class GeneratedCodeBasicInfoAnalysis
+  : public llvm::AnalysisInfoMixin<GeneratedCodeBasicInfoAnalysis> {
+  friend llvm::AnalysisInfoMixin<GeneratedCodeBasicInfoAnalysis>;
+  static llvm::AnalysisKey Key;
+
+public:
+  using Result = GeneratedCodeBasicInfo;
+  Result run(llvm::Module &M, llvm::ModuleAnalysisManager &);
+  Result run(llvm::Function &F, llvm::FunctionAnalysisManager &);
```
aleclearmind, line 456: Maybe put a `/// \note``` on how to use this, i.e., use a proxy+`run(Module)` if your pass is running in a MPM, or `run(Function)` otherwise.

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
```diff
@@ -414,3 +443,31 @@ struct BlackListTrait<const GeneratedCodeBasicInfo &, llvm::BasicBlock *>
...
+class GeneratedCodeBasicInfoWrapperPass : public llvm::ModulePass {
+  std::unique_ptr<GeneratedCodeBasicInfo> GCBI;
+public:
+  static char ID;
+
+  GeneratedCodeBasicInfoWrapperPass()
+    : llvm::ModulePass(ID) { }
+
+  GeneratedCodeBasicInfo &getGCBI() { return *GCBI; }
+  const GeneratedCodeBasicInfo &getGCBI() const { return *GCBI; }
```
aleclearmind, line 469: Is this needed? Let's put this only if needed.

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
```diff
@@ -414,3 +443,31 @@ struct BlackListTrait<const GeneratedCodeBasicInfo &, llvm::BasicBlock *>
...
+
+  bool runOnModule(llvm::Module &M) override;
+  void getAnalysisUsage(llvm::AnalysisUsage &AU) const override;
```
aleclearmind, line 472: I like to keep `getAnalysisUsage` body in the header so it "documents" the pass dependencies, if any.

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
```diff
@@ -414,3 +443,31 @@ struct BlackListTrait<const GeneratedCodeBasicInfo &, llvm::BasicBlock *>
...
+
+public:
+  using Result = GeneratedCodeBasicInfo;
+  Result run(llvm::Module &M, llvm::ModuleAnalysisManager &);
+  Result run(llvm::Function &F, llvm::FunctionAnalysisManager &);
+};
+
+/// Legacy pass manager pass to access GCBI.
+class GeneratedCodeBasicInfoWrapperPass : public llvm::ModulePass {
+  std::unique_ptr<GeneratedCodeBasicInfo> GCBI;
```
aleclearmind, line 461: Why `unique_ptr`?

aleclearmind (changes requested):

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
antoniofrighetto: I felt like `sizeof(GeneratedCodeBasicInfo)` was big enough to justify to take a pointer and allocate the object on the heap instead of in the stack.

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
antoniofrighetto: That's needed, it actually provides a way to get a reference to the detached `GeneratedCodeBasicInfo` (EnforceABI should take a const ref IIRC).

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
aleclearmind: Passes end up in the heap in any case.

Early optimization...

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
aleclearmind: This doesn't sound right. Do you ever get (from LLVM) a const reference to `GeneratedCodeBasicInfoWrapperPass`?
If not, drop this.

I see no reason why you should pass around `GeneratedCodeBasicInfoWrapperPass`, just pass `GeneratedCodeBasicInfo` to a function that expects a `const &` and it will be converted automatically.

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
antoniofrighetto: Well, that doesn't mean that the objects end up being in the heap. Also, here it may be used to enforce lack of expensive copying. Not aware of any transformation that moves (big) stack objects into the heap. I doubt this is an early optimization.

aleclearmind: What objects?
`GeneratedCodeBasicInfoWrapperPass` will end up on the heap, therefore even without a `unique_ptr` `GeneratedCodeBasicInfo` won't be on the stack for sure.

What is `GCBI` size? I don't think `GCBI` is big *per se*, it holds vectors/maps which might be big, but they end up in the heap in any case. We're not talking about stack overflow or stuff like that in any case.

When is `GeneratedCodeBasicInfoWrapperPass` being copied? I think it's not otherwise the compiler would complain about a missing copy constructor (the default has been deleted due to the `unique_ptr` field).

I think you're being inspired by this pattern by certain passes that hold analysis results in a `unique_ptr`, but they don't do it "because the results are big". They do it either because there's no default constructor for the object (i.e., you need a Module/Function to construct the object) or because they implement the `releaseMemory` method of `llvm::*Pass` where they `reset` the `unique_ptr`.

The second reason could apply here (even if we could also implement a `clear` method in `GCBI`), but then you need to implement the `releaseMemory` method.

antoniofrighetto: I feel like using a `std::unique_ptr` boils down to personal choice in this context. When I looked at `GCBI` the very first time I was like "OK, let's use a `std::unique_ptr` here, it's safe and it doesn't harm". Indeed, I'm taking care of tearing down each time the previous object by calling `reset`. I didn't implement the `releaseMemory` method because I thought there was no really need to, which GCBI's objects `clear` should be called on? Maybe `JumpTargets`? Anyway, no problem to remove the `std::unique_ptr` at all.

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
antoniofrighetto: Removed.

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
aleclearmind: `std::unique_ptr` increases complexity, I've seen many times places where it's abused. Many many times `std::unique_ptr` hides a bad design choice.
Your first choice should always be the simplest. KISS logic >> personal taste. Let's not discuss taste, let's discuss facts.

Let's follow the pattern of other passes: let's keep the `unique_ptr` but add a `releaseMemory`.

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
antoniofrighetto: Yeah, I do remember this. I think we can fix these redundant cases for good, considering also [this](https://github.com/revng/revng/blob/develop/include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h#L262). Shall we keep this (which fits here) and remove `getFallthrough` from `FCI`?

aleclearmind: Works. We'll need to reorganize GCBI. What matters now is to reduce code duplication.

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
antoniofrighetto: Callee sounds better, but I didn't use it because there's already `getCallee` in IRHelpers.h. 
`getFunctionCallCallee` does the same job but it seems to be only used within the StackAnalysis, and since GCBI.h already had `getFunctionCall` implemented, it looked much cleaner to me to have that `getSuccessor` method. Maybe we can drop `getFunctionCallCallee` from IRHelpers.h and move `getSuccessor` to `getCallee` (now that it is static)?

aleclearmind: Let's move as many `static` methods of GCBI into `IRHelpers.h`.
After all they're just helper functions.
One day we'll reorganize IRHelpers.h, but let's not make grow GCBI.

antoniofrighetto: Shouldn't it be better the opposite? `getFunctionCall` was already in GBCI.h, and given that `getFallthrough` is now there, and this point `getSuccessor` should be left there too, since both using `getFunctionCall`. This way everything is more concise and clean (and logically speaking would make sense to keep them there). Or do you want both of them to be in IRHelpers.h?

aleclearmind: Thins were in part in `IRHelpers.h` and in part in `GCBI`. Let's move stuff in `IRHelpers.h`.
Those functions are useful even if you're not in a pass. And if you're not in a pass there's no point in importing GCBI.h.
Anyway, this need further re-organization. But for now, move as much as you can in `IRHelpers.h`.

antoniofrighetto: Alright, sounds good, thanks for clarifying.

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
antoniofrighetto: I didn't, but I remember talking with @pfez about these lines, who also suggested using `isMarker` and `getPrevious` (if I recollect correctly). Looks cleaner to me this way, if it's not too less efficient, but we can keep only `isMarker` if you say.

aleclearmind: Uhm, OK, make sure that this improvement gets its own commit.

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
antoniofrighetto: No need to change this.

include/revng/BasicAnalyses/GeneratedCodeBasicInfo.h
pfez: My point was that the previous version did not work if the argument was the first instruction in a basic block, because it would never have entered the `while (It != End)`. Using `getPrevious()` works as expected all the times.

pfez: You also have `Instruction * Instruction::getPrevNode()`, which is defined by llvm, and not by ourselves, and does exactly the same thing. Maybe we should avoid defining our own helpers that are identical to llvm stuff that is already available.

aleclearmind: I'm not sure why this was changed, the original version looked fine to me.
However, I'd suggest not to make too many unrelated changes to IRHelpers.h in an opportunistic way.
We need to sit down and review in a single shot `revng.h`, `IRHelpers.h`, `GCBI.h` and reorganize them. But this is not the right time.

antoniofrighetto: Noted down to fix in the future.

lib/BasicAnalyses/GeneratedCodeBasicInfo.cpp
```diff
@@ -141,3 +135,34 @@ GeneratedCodeBasicInfo::getSuccessors(BasicBlock *BB) const {
...
+
+bool GeneratedCodeBasicInfoWrapperPass::runOnModule(Module &M) {
+  GCBI.reset(new GeneratedCodeBasicInfo());
+  GCBI->run(M);
+  return false;
+}
+
+void GeneratedCodeBasicInfoWrapperPass::releaseMemory() {
+  GCBI->releaseMemory();
+}
```
aleclearmind, line 162: In `GeneratedCodeBasicInfoWrapperPass::releaseMemory` just call `GCBI.reset(new GeneratedCodeBasicInfo());`.
I don't think we need to implement `releaseMemory` for `GeneratedCodeBasicInfo` at this point. Even because `.clear` doesn't actually release memory, while `reset`ting the `unique_ptr` does.

aleclearmind (changes requested):
Looks very good!
Next time please keep "clean up commits" first and the "main" commit last (the one actually converting to the new PM).

1. [ ] Fix the remaining comment.
2. [ ] Make sure each commit passes check-conventions (`git rebase $START_COMMIT -x 'orc shell check-revng-conventions --force-format'`)
3. [ ] Make sure each commit compile (`git rebase $STAR_COMMIT -x 'orc shell -c revng ninja'` )

Then I'll reorder the commits and merge!

lib/BasicAnalyses/GeneratedCodeBasicInfo.cpp
antoniofrighetto: Well, `releaseMemory` in GCBI is kind of unneeded, but it doesn't harm calling `clear` and `reset`. Same for `GCBI.reset()`, which is automatically done, but yeah, adding it is harmless.

aleclearmind: Harmless is nice way of saying "useless".
There's no point in `releaseMemory` in GCBI, it's a pattern common in old passes, not in analysis results.
Just `GCBI.reset` and drop `GeneratedCodeBasicInfo::releaseMemory`.

aleclearmind:
Merged in 0e3bf6c.
Thanks a lot!
You did a great job.

PR #37, opened 2021-01-13
EnforceABI: allow translated code re-execution

antoniofrighetto:


lib/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ macro(revng_add_library_internal NAME TYPE)
 endmacro()
 
 add_subdirectory(BasicAnalyses)
+add_subdirectory(Debug)
```
aleclearmind, line 21: Move this to `Dump`.

lib/Debug/DebugInfo.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+class DebugInfo : public ModulePass {
+public:
+    static char ID;
+    DebugInfo() : ModulePass(ID) { }
+
+    bool runOnModule(Module &M);
+};
+
+}
```
aleclearmind, line 25: Keeping this in a `.cpp` file forces us to use this pass from `opt` only. Please move this to a public header.

lib/Debug/DebugInfo.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+bool DebugInfo::runOnModule(Module &M) {
+    DebugHelper DBG("debug_info", &M, DebugInfoType::LLVMIR, M.getModuleIdentifier());
```
aleclearmind, line 28: * [ ] Why `getModuleIdentifier`?
* [ ] Is this creating a `debug_info` file in the current directory?

This pass should not create any file, it should just decorate the IR, AFAIU.

lib/FunctionIsolation/EnforceABI.cpp
```diff
@@ -297,6 +309,19 @@ void EnforceABIImpl::run() {
...
 
+  RaiseExc = M.getFunction("raise_exception_helper");
+  revng_assert(RaiseExc != nullptr);
+
+  // Create a bitmask to trace which CSVs need to be saved for the function
+  auto *ConstantZero = ConstantInt::get(Type::getInt64Ty(Context), 0);
+  new GlobalVariable(M,
+                     Type::getInt64Ty(Context),
+                     false,
+                     GlobalValue::ExternalLinkage,
```
aleclearmind, line 320: Why `ExternalLinkage`?

lib/FunctionIsolation/EnforceABI.cpp
```diff
@@ -297,6 +309,19 @@ void EnforceABIImpl::run() {
...
+                     ConstantZero,
+                     "RestoreCSVBitMask");
+  RestoreCSVBitMask = M.getGlobalVariable("RestoreCSVBitMask");
```
aleclearmind, line 323: Why this extra lookup? Just get the pointer from the previous statement.

lib/FunctionIsolation/EnforceABI.cpp
```diff
@@ -159,6 +167,10 @@ class EnforceABIImpl {
   std::map<Function *, Function *> OldToNew;
   Function *FunctionDispatcher;
   Function *OpaquePC;
+  DenseMap<std::pair<GlobalVariable *, Function *>, AllocaInst *> CSVMap;
+  DenseMap<Function *, BasicBlock *> UnwindsMap;
```
aleclearmind, line 171: "Unwinds"?

lib/FunctionIsolation/EnforceABI.cpp
```diff
@@ -380,6 +409,26 @@ void EnforceABIImpl::run() {
   // Promote CSVs to allocas
   replaceCSVsWithAlloca();
 
+  // Replace call to invoke inst in helper functions
+  replaceCallWithInvoke();
+
+  // Zero out RestoreCSVBitMask before calling raise_exception_helper
+  for (Use &U : RaiseExc->uses()) {
+    InvokeInst *II = cast<InvokeInst>(U.getUser());
+    if (!II) continue;
```
aleclearmind, line 418: New line. Did you run `check-conventions.sh --force-format`?

lib/FunctionIsolation/EnforceABI.cpp
```diff
@@ -380,6 +409,26 @@ void EnforceABIImpl::run() {
   // Promote CSVs to allocas
   replaceCSVsWithAlloca();
 
+  // Replace call to invoke inst in helper functions
+  replaceCallWithInvoke();
+
+  // Zero out RestoreCSVBitMask before calling raise_exception_helper
+  for (Use &U : RaiseExc->uses()) {
```
aleclearmind, line 416: You can iterate on `User`s directly.

lib/FunctionIsolation/EnforceABI.cpp
```diff
@@ -159,6 +167,10 @@ class EnforceABIImpl {
   std::map<Function *, Function *> OldToNew;
   Function *FunctionDispatcher;
   Function *OpaquePC;
+  DenseMap<std::pair<GlobalVariable *, Function *>, AllocaInst *> CSVMap;
+  DenseMap<Function *, BasicBlock *> UnwindsMap;
+  Function *RaiseExc;
```
aleclearmind, line 172: Let's not have abbreviate names, unless needed. It makes the code harder to read.

lib/FunctionIsolation/EnforceABI.cpp
```diff
@@ -758,14 +837,17 @@ void EnforceABIImpl::handleRegularFunctionCall(CallInst *Call) {
 
 void EnforceABIImpl::generateCall(IRBuilder<> &Builder,
                                   Function *Callee,
-                                  CallSiteDescription &CallSite) {
+                                  CallSiteDescription &CallSite,
+                                  bool isPartOfFunctionDispatcher) {
```
aleclearmind, line 841: Arguments need to have first letter capital

lib/FunctionIsolation/EnforceABI.cpp
```diff
@@ -758,14 +837,17 @@ void EnforceABIImpl::handleRegularFunctionCall(CallInst *Call) {
...
   revng_assert(Callee != nullptr);
 
   llvm::SmallVector<Type *, 8> ArgumentsTypes;
   llvm::SmallVector<Value *, 8> Arguments;
   llvm::SmallVector<Type *, 8> ReturnTypes;
   llvm::SmallVector<GlobalVariable *, 8> ReturnCSVs;
 
+  Value *Result;
```
aleclearmind, line 849: Initialize to `nullptr`

lib/FunctionIsolation/EnforceABI.cpp
```diff
@@ -801,17 +883,19 @@ void EnforceABIImpl::generateCall(IRBuilder<> &Builder,
...
     // Additional debug checks if we are not emitting an indirect call.
     revng_log(EnforceABILog,
               "Emitting call to " << getName(Callee) << " from "
                                   << getName(CallSite.Call));
 
-    FunctionDescription &Function = FunctionsMap.at(Callee);
-    revng_assert(Function.Function != nullptr
-                 and Function.Function->getName().startswith("bb."));
-    if (GlobalVariable *CSV = CallSite.isCompatibleWith(Function)) {
+    FunctionDescription &Func = FunctionsMap.at(Callee);
```
aleclearmind, line 895: Don't abbreviate

lib/FunctionIsolation/EnforceABI.cpp
```diff
@@ -834,9 +918,58 @@ void EnforceABIImpl::generateCall(IRBuilder<> &Builder,
...
+    Function *Caller = BB->getParent();
+
+    // If we are in a basic block of an isolated function,
+    // we generate an invoke instruction. Do not emit an invoke instruction
+    // if DisableSafetyChecks is on.
+    // TODO: Review second condition
+    if (!DisableSafetyChecks && Caller->getName().startswith("bb.")
+          && !BB->getName().endswith("__continuation")) {
+      BasicBlock *NormalEdge;
+      std::string NormalEdgeName = std::string(BB->getName()) + "__continuation";
```
aleclearmind, line 932: Use a `Twine`

lib/FunctionIsolation/EnforceABI.cpp
```diff
@@ -443,23 +492,52 @@ void EnforceABIImpl::handleRoot() {
...
-    auto *NewInvoke = Builder.CreateInvoke(Callee,
-                                           Invoke->getNormalDest(),
-                                           Invoke->getUnwindDest(),
-                                           Arguments);
-    NewInvoke->setIsNoInline();
+    auto *Result = Builder.CreateInvoke(Callee,
+                                        NormalDest,
+                                        Invoke->getUnwindDest(),
+                                        Arguments);
+    Result->setIsNoInline();
```
aleclearmind, line 522: Is this necessary?

aleclearmind:
Great to have this.

* [ ] Can you please provide a coarse grain summary of what this patchset does? I need a quick summary of the changes in the resulting IR to better digest the patch. Thanks!
* [ ] Let's keep the `SA: intercallsite Yes propagation` for later
* [ ] Use 50/72 rule and run `check-conventions.sh --force-format` on every commit

scripts/revng
```diff
@@ -327,16 +327,22 @@ def run_translate(args):
...
+                      action="store_true",
+                      help="Invoke linker only.")
   parser.add_argument("-i",
                       "--isolate",
                       action="store_true",
                       help="Enable function isolation.")
   parser.add_argument("--base", help="Load address to employ in lifting.")
   parser.add_argument("-o", "--output", metavar="OUTPUT", help="Output path.")
   parser.add_argument("input", metavar="INPUT", help="The input binary.")
+  parser.add_argument("--opt", type=str, help="Pass to LLVM opt the passes to be performed over the IR.")
```
aleclearmind, line 341: We will need something more flexible than what we have now, but `--opt` is kind of an hack.
Let's just have another argument such as `--isolate` that does The Right Thing™ for you, as opposed to add passes by hand to `--opt`.

lib/FunctionIsolation/EnforceABI.cpp
antoniofrighetto: Because originally I declared that variable in `support.c`. Must have added that code at a later time.

lib/FunctionIsolation/EnforceABI.cpp
antoniofrighetto: I abbreviated to `Func`, because `Function` was creating a name collision with `llvm::Function` over there (after the new changes).

aleclearmind: Find a more precise name or go for `TheFunction`. Abbreviation make the code come from the 90 and harder to read.

lib/FunctionIsolation/EnforceABI.cpp
antoniofrighetto: Yes, I believe we always want the call-site in the entry basic blocks in `root`.

aleclearmind: Why? Does it ever happen on any of the pipelines we have that it gets inlined?
If we launch `-O2` to optimized runtime performance I don't see why we should constrain the optimizer in this way.

lib/Debug/DebugInfo.cpp
antoniofrighetto: The pass doesn't create any file, and no `debug_info` file is created, it simply decorates the IR. ` M.getModuleIdentifier()` comes because we want to overwrite the current `.ll` with the new annotations. If `-debug-info` is enabled, I don't think it would make sense to create a separate `.ll.with_debug_decorations` file.

lib/FunctionIsolation/EnforceABI.cpp
antoniofrighetto: I don't see why we should prevent any optimizations there. Note that `setIsNoInline` was already there before such changes and I thought there were other reasons why it was initially introduced.

aleclearmind: Uhm, I thought I checked but you're right, it was there before.
There's probably a reason for that then, let's leave it there. Maybe we can ask @andrealinux1.
Anyway, `Invoke` seems like a better name than `Result`, restore it or change it to something more meaningful.

scripts/revng
antoniofrighetto: For sure specifying every single argument may be more precise, but does `--opt` really look like a hack? To be honest it just seems like kind of a compact way to directly specify the optimizations (e.g., `--opt=isolate,enforce-abi,debug-info`). Maybe it's not well structured right now, but at least is faster than invoking `opt` manually. About to change it though.

lib/FunctionIsolation/EnforceABI.cpp
antoniofrighetto: Changed it to `NewInvoke` since we are replacing the old one.

lib/FunctionIsolation/EnforceABI.cpp
antoniofrighetto: `UnwindEdgesMap` should make more sense.

lib/FunctionIsolation/EnforceABI.cpp
antoniofrighetto: Second.

scripts/revng
aleclearmind: We have a project in mind for having a declarative way to define meta-commands such as `revng translate`, but not today.
`--opt` is adding an extra layer on an already bad design. I don't want to go that direction.

antoniofrighetto: Alright, sure.

lib/Debug/DebugInfo.cpp
aleclearmind: > M.getModuleIdentifier() comes because we want to overwrite the current .ll with the new annotations

I don't understand. This is a pass, it's not supposed to overwrite `opt` output.

```c++
  /// \brief Create a new DebugHelper
  ///
  /// \param Output path where the LLVM IR should be stored.
  /// \param Debug path where the debug output should be stored. If empty, \p
  ///        Output will be used along with a suffix, e.g. `.pts` if \p Type is
  ///        DebugInfoType::PTC, `.S` if it's DebugInfoType::OriginalAssembly or
  ///        will match \p Output if \p Type is DebugInfoType::LLVMIR.
  /// \param TheModule the LLVM module to print out.
  DebugHelper(std::string Output,
              llvm::Module *TheModule,
              DebugInfoType::Values DebugInfo,
              std::string DebugPath);
```
You're passing "debug_info" as `Output`. Are you sure it's not creating a file?

```c++
    std::ofstream Output(DebugPath);
    raw_os_ostream Stream(Output);
    TheModule->print(Stream, annotator(false));
```

Why are you creating a file named `M.getModuleIdentifier()` (`DebugPath` argument)?

I'm confused.

antoniofrighetto: I came to the conclusion that this is not meant to go upstream. First, the user shouldn't likely need to debug the IR. Second, I didn't really bother to tell if the pass should overwrite `opt` output or not. As things stand, the `.ll` happens to be fully overwritten with the addition of the debug information. I agree that the IR should be simply changed with the debug info, without re-creating the entire module but it was not within the goals when I created that pass. Maybe @andrealinux1 has a say in whole of this, since I adapted this pass from some of his old stuff; and by now this dates back to more than a year ago.

aleclearmind: Maybe the right approach would be:

1. Run the pass adding metadata, no file is written.
2. Have another pass that has an extra option for the output path, where the decorated file is dumped.

Sadly, we can't do much more than this, since to decorate `opt` output we'd need our custom annotator.
The feature is useful and interesting, however, if you prefer to remove this from this PR, I agree. It's already quite big.

antoniofrighetto: To be reviewed in a second moment.

andrealinux1: Mmmh, I'm also a bit confused, so forgive me if I say trivial things or I make stupid annotations.

I guess here the problem is that, as things are now, this new pass creates a new `.ll` file named as the module itself (basically, it overwrites it), due to the fact that we pass explicitly the `DebugPath` parameter. If this is not done, what @aleclearmind says it's right, the pass creates a new `debug_info` file.

What @aleclearmind is suggesting, and tell me if I got this right, is that, following the conventions of llvm passes, a pass should not create a new output per se, but should simply modify the module that is working on. Unfortunately, I think that DebugHelper, as it is designed now, it is not suited for doing this (also here, correct me if I'm wrong, I will be happy about that:)).
The point is that, the `Annotator`, which is the class doing the work of attaching the Metadata to the IR, works by knowing the line at which a certain LLVM Instruction will be in the output `.ll` file (so that the debugger will know at which line to look for showing the corresponding IR instruction when stepping). In fact, in the `generateDebugInfo` method you can see that we serialize the module twice, one just to recover the `DILocation`, and the next to actually serialize the module containing the debug infos (the parameter passed to the annotator handle this behavior):
```
    raw_null_ostream NullStream;
    TheModule->print(NullStream, annotator(true /* DebugInfo */));

    std::ofstream Output(DebugPath);
    raw_os_ostream Stream(Output);
    TheModule->print(Stream, annotator(false));
```
Therefore, I do not see how we can adapt this class to not work on streams but to work on the IR itself, taking into consideration that if the module is later modified in the pipeline, I expected the debug infos to go out-of-sync very fast.

lib/FunctionIsolation/EnforceABI.cpp
andrealinux1: Ok, so the journey is a bit convoluted here. I think you (@antoniofrighetto) borrowed these lines creating the invokes from the corresponding ones in the `FunctionIsolation`, specifically where we create the first `invoke`s that are in charge of switching the execution from the non isolated realm to the isolated one, and that constitute the substituted body of each function entry block.

There, we added the `noinline` (commit af6f963d7765bee9aeb3502d63896187e0fdf316), in order to avoid the `prune-eh` optimization pass to remove these `invoke`s and trasforming them into `call`s. From the commit message:
```
Improve exception handling
    
    When optimizing a module to which function isolation has been applied the
    `prune-eh` pass lead to several issues, specifically:
    
    * `support.c` is now compiled with support for exceptions, to prevent
      `raise_exception_helper` from being marked `nounwind` during the
      link phase.
    * Added a fake `ret` at the end of the `catchblock` to avoid promotion
      of `invoke`s to regular `call`s.
    * Marked the `invoke` instructions as `noinline`.
``` 

Now, reading the `prune-eh` pass, it seems to me that maybe that was due to the fact that, before @antoniofrighetto really understood how the personality function works (and fixed it in 76a6403ebdf2acb7fb289159725c0917f5e12f40),  we didn't really use the unwind block (I don't know if you remember the story about the `ExceptionFlag`), and therefore the `prune-eh` interprocedural analysis decided to promote the `invoke`s to `calls`.

I expect that now, the `prune-eh` is doing things correctly, so it may be worth a try to see if the `invoke`s are still disappearing after that change.

lib/Debug/DebugInfo.cpp
aleclearmind: Yeah, you're probably right. It wouldn't make sense to have such pass scheduled, if not last.

lib/FunctionIsolation/EnforceABI.cpp
antoniofrighetto: This last analysis is correct, however, note that the `invoke` creation as well as the `noinline` attributes were already added by @aleclearmind. Our contribution there was simply concerned with extracting the return values out of the call in case we would be returning from an isolated function to `root` (e.g., after an exception has been managed). Now after installing the personality routine and with the removal of `ExceptionFlag`, I don't see why the `noinline` should be kept.

aleclearmind: @andrealinux1 saves the day!
Can you give it a shot in a working example you have @antoniofrighetto?

aleclearmind: OK, please test on a working example and make a new commit to drop this.

Keep in mind that PRs and code review are an opportunity to improve the existing code base.
So even if the bug was already there, it's worth addressing it since we're already working on this.

aleclearmind: (thanks @andrealinux1!)

antoniofrighetto: Fully agree. Btw, I do remember those lines driving us crazy back then – while debugging an enforce'd binary –, so yeah, checking it again surely doesn't harm.

lib/FunctionIsolation/EnforceABI.cpp
antoniofrighetto: Note that `NormalEdgeName` is used twice (there may be two different `NormalEdge` basic blocks depending on if it's part of the `function_dispatcher` or not), so I would need to create two temporary objects here, like `BasicBlock::Create(Context, Twine(BB->getName()) + "__continuation", ...)` and do the same with `BB->splitBasicBlock()`, looks bad style to me (Twine is meant to be used only for passing strings as const refs in parameters). So, either I use a `const std::string &Name = ` or I create a tmp like `(Twine(BB->getName()) + "__continuation").str();`. Not sure if the first allows to avoid a temporary allocation.

antoniofrighetto: Just to conclude that, after experimenting a bit, it turned out that:
• `const std::string Name = BB.getName() + "__cont";` doesn't compile (no implicit conversion from `llvm::Twine` to `std::string`);
• `const auto Name = BB.getName() + "__cont";` works but it is not safe at all (the outer Twine refers to the inner temporary Twine object that has already been destroyed, so dangling pointer);
• `auto Name = (Twine(BB->getName()) + "__cont").str();` works, the temporary object is materialized into a string and is literally the same as the original construction (line 932).

aleclearmind: Gotcha, makes sense to use a `string` then!

PR #38, opened 2021-01-14
ADT: fix neighbor constness for GenericGraph

AlviseDeFaveri:
Problem:
1. `ForwardNode::getConstNeighbor` returns a pointer to non-const object
2. `FowardNode::toNeighborRange` checks if the type of the argument is
const, but since `Successors` is not declared const the resulting
`toNeighborRange(Successors)` is never executed in the const version

Solution:
1. Make `getConstNeighbor()` return a pointer to a const object
2. Explicitly declare a separate `toNeighborRange()` function which
returns a `const_child_iterator` and remove the template param

### DIFF
--- a/include/revng/ADT/GenericGraph.h
+++ b/include/revng/ADT/GenericGraph.h
@@ -92,7 +92,7 @@ public:
 
 public:
   static Inheritor *&getNeighbor(Edge &E) { return E.Neighbor; }
-  static const Inheritor *&getConstNeighbor(const Edge &E) {
+  static const Inheritor *const &getConstNeighbor(const Edge &E) {
     return E.Neighbor;
   }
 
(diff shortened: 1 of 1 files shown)


aleclearmind:
Is this @pfez approved?

pfez:
Yes, it's the result of a debugging session I had with @AlviseDeFaveri 

But the commit message is wrong in the parts related to point 1.

pfez:
I would reword the parts related to point 1 in the commit message. Something along these lines:

> Problem:
> 1. `ForwardNode::getConstNeighbor` returns a reference to a non-`const` pointer, but the referenced pointer is `const` 
...

> Solution:
> 1. Make `getConstNeighbor()` return a reference to a `const` pointer

pfez:
The differences are subtle, but the issue is subtle, and I think it's better to try to be precise.

pfez:
Oh, and finally, I think there's a `clang-format` violation here: https://github.com/revng/revng/blob/fe01bcc469c1115714e752f65d744005687c7d53/include/revng/ADT/GenericGraph.h#L155

The leading whitespaces whould be removed.

Besides these two cosmetics changes it's good to merge.

pfez (approved):
@aleclearmind, I have already expressed a couple of small concerns before you requested my review. @AlviseDeFaveri has already addressed them. To me, it now looks good to merge!

aleclearmind:
Gotcha, will do a final pass soon and then merge.
Thanks.

aleclearmind:
Merged in 9eb9e9b1a121fce7dcec69ad7ff569ee1f1ad39c.
Thanks and congratulations for your first commit in mainline `:)`!

PR #41, opened 2021-01-25
Use debug symbols to handle `noreturn` callsites and calls through `.plt` stubs

andrealinux1:
Modify the SA in order to exploit some information contained in the function debug symbols in order to improve the quality of the function boundaries identification in occurrence of `noreturn` callsites and external calls through the `.plt` stubs.


include/revng/Support/revng.h
```diff
@@ -184,6 +184,118 @@ inline Values fromName(llvm::StringRef Name) {
...
+  case llvm::ELF::STT_SECTION:
+    return SymbolType::Section;
+
+  case llvm::ELF::STT_FILE:
+    return SymbolType::File;
+
+  default:
+    revng_abort("Unexpected symbol type");
+  }
+}
```
aleclearmind, line 273: Can we keep this in a header not so widespread as `revng.h`?
`SymbolType` is supposed to be used in a binary format-agnostic way, so I don't expect `fromELF` to be used outside `BinaryFile.cpp`.

lib/StackAnalysis/StackAnalysis.cpp
```diff
@@ -51,6 +51,12 @@ static opt<std::string> StackAnalysisOutputPath("stack-analysis-output",
                                                 value_desc("path"),
                                                 cat(MainCategory));
 
+static opt<bool> UseFunctionSymbols("use-function-symbols",
+                                    desc("Use function symbols in CFEP "
+                                         "election"),
+                                    cat(MainCategory),
+                                    init(false));
```
aleclearmind, line 58: We already use function symbols. Call this `UseFunctionSymbolsOnly`/`use-function-symbols-only`.

lib/StackAnalysis/StackAnalysis.cpp
```diff
@@ -102,18 +108,28 @@ bool StackAnalysis<AnalyzeABI>::runOnModule(Module &M) {
...
+      } else if (not IsLoadAddress
+                 and (IsUnusedGlobalData
+                      || (IsMemoryStore and not IsPCStore
+                          and not IsReturnAddress))) {
+        // TODO: keep IsReturnAddress?
+        // Consider addresses found in global data that have not been used or
+        // addresses that are not return addresses and do not end up in the PC
+        // directly.
+        Functions.emplace_back(&BB, false);
+      }
```
aleclearmind, line 132: What about the following?
```c++
if (IsFunctionSymbol or (not UseFunctionSymbolsOnly and IsCallee)) {
  Functions.emplace_back(&BB, true);
} else if (UseFunctionSymbolsOnly) {
  // Ignore
} else if (...) {
  // ...
  Functions.emplace_back(&BB, false);
}
```
In general, I feel like sometimes you're overly cautious in touching other code and tend to "add" features making sure you don't break other stuff, which is good in principle. However, it creates a layers and there's only one problem that cannot be solved by adding another layer, having too many layers.

Feel free to move things around, if tests pass your change is good!

aleclearmind (commented):
I need to think more about how to integrate this.

aleclearmind:
Given recent evolution in handling of external functions in my private branch, this is becoming obsolete.
Demoting to draft until we decide what to do.

PR #44, opened 2021-02-01
Various fixes needed by Caliban

carpikes:
This MR:

* extends the support for the `FOR_EACH` macro used in `INTROSPECTION_NS` to allow serialization of structs with more than 5 entries. It now supports up to 16 entries.
* enables quotes around MetaAddress during serialization, because ':' is a valid YAML separator, which causes wrong deserialization when a vector of MetaAddress is serialized as a list.
* add a missing include in `TupleTreeDiff.h`, `TupleTree.h`
* applied clang-format to GenericGraph
* add SerializableGraph

### DIFF
--- a/include/revng/ADT/GenericGraph.h
+++ b/include/revng/ADT/GenericGraph.h
@@ -361,6 +361,7 @@ public:
   static const bool is_generic_graph = true;
   using NodesContainer = llvm::SmallVector<std::unique_ptr<NodeT>, SmallSize>;
   using Node = NodeT;
+  static constexpr bool hasEntryNode = HasEntryNode;
 
 private:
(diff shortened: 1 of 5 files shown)


aleclearmind:
Merged with support for serializing graphs with edge labels and other stuff in 4d3d625957dec5f9298a6796e8e37a33061239f6.

Some notes:
* Use `orc shell check-revng-conventions --force-format` to reformat your code, otherwise you're using `clang-format` 11 which has a slightly different behavior.
* This PR had a good description about what it does, while the commit messages where a bit more laconic. PR go away, commit messages remain in history. Let's try to have more descriptive comments, and if you're confident enough in them and don't think there's other notes the maintainers should be aware of, just reference the git history.

Thanks!

PR #46, opened 2021-02-09
Fix revng_merge_dynamic installation

fcremo:
Fixes https://github.com/revng/help/issues/17

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,13 +162,19 @@ configure_file(runtime/early-linked.c "${CMAKE_BINARY_DIR}/share/revng/early-lin
 
 configure_file(scripts/revng "${CMAKE_BINARY_DIR}/bin/revng")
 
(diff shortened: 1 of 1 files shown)


aleclearmind:
I've merged this with some modifications in f05f7850235254a750275ade66174e9a8aee29a6.

In general:
1. Try to improve the surrounding code when you make these changes. There were several other `copy_file` + `install` pattern that could exploit the new macro. I understand it's a bugfix, but if we don't take the opportunity to improve the codebase with bugfixes too, the quality of the codebase will degrade over time.
2. The `install` directive was not using `install(PROGRAMS ...)` for `scripts/revng_merge_dynamic/merge_dynamic.py`. I guess this is not necessary, but then we should `chmod -x` that file.

Can you please test everything works from your side and close this?

fcremo:
I confirm that f05f785 works for me, and merge-dynamic is found both at build time and when using revng as installed by revng and revng-distributable.

PR #47, opened 2021-02-10
RecursiveCoroutine: fix reference arguments

pfez:
Before this commit, recursive coroutines did not work properly if they had out arguments with reference type.
The reason is that `rc_run` was inferring the type of its arguments from the arguments themselves, not from the prototype of the recursive coroutine.

Hence, code snippets like the following did not work properly, because `rc_run` was taking x by value, not by reference.

```c++
RecursiveCoroutine<void> accumulate_on_i(int &i) {
  // ...
}

int f() {
  int x = 0;
  rc_run(accumulate_on_i, x);
  return x;
}
```

This commit fixes the problem. Now the arguments of `rc_run` are properly forwarded to the recursive coroutine.

### DIFF
--- a/include/revng/ADT/RecursiveCoroutine-coroutine.h
+++ b/include/revng/ADT/RecursiveCoroutine-coroutine.h
@@ -6,6 +6,7 @@
 
 #include <experimental/coroutine>
 #include <optional>
+#include <utility>
 
 #include "revng/Support/Assert.h"
 
@@ -182,8 +183,8 @@ protected:
 };
 
 template<typename CoroutineT, typename... Args>
-auto rc_run(CoroutineT F, Args... Arguments) {
-  return F(Arguments...).run();
+auto rc_run(CoroutineT F, Args &&... args) {
+  return F(std::forward<Args>(args)...).run();
 }
 
 #define rc_return co_return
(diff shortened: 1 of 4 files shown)


include/revng/ADT/RecursiveCoroutine-coroutine.h
```diff
@@ -181,9 +182,9 @@ struct RecursiveCoroutine<ReturnT>::promise_type : public PromiseBase {
   std::optional<ReturnT> CurrValue = std::nullopt;
 };
 
-template<typename CoroutineT, typename... Args>
-auto rc_run(CoroutineT F, Args... Arguments) {
-  return F(Arguments...).run();
+template<typename CoroutineT, typename... FormalArgs, typename... ActualArgs>
```
drblallo, line 185: why is it necessary to take FormalArgs as template arguments?

```
template<typename CoroutineT, typename... ActualArgs>
auto rc_run(CoroutineT F, ActualArgs &&... args) {
￼  return F(std::forward<ActualArgs>(args)...).run();
￼}
```
should be enough?

pfez: Good point! It was just a leftover. I removed it.

aleclearmind (approved):

aleclearmind:
Merged in 8ffcd49dd24706b284e319917bb82de4f4d6e69b.
Thanks!

PR #48, opened 2021-02-16
LoadModelPass: provide public static function to read Model from llvm::Module

pfez:


include/revng/Model/LoadModelPass.h
```diff
@@ -33,4 +35,6 @@ class LoadModelPass : public llvm::ImmutablePass {
     Modified = true;
     return TheBinary;
   }
+
+  static std::unique_ptr<model::Binary> getModel(const llvm::Module &M);
```
aleclearmind, line 39: Why all these `unique_ptr` business? If you're concerned about efficiency just use `std::move`.

pfez: Done!

pfez:
I added two commits to do the following
- add a static method to `SerializeModulePass` to forcibly serialize the Model to a given `llvm::Module`
- use the newly added method to serialize an empty Model to the `llvm::Module`s lifted by `revng-lift`

aleclearmind:
Looks good! Will merge soon!

aleclearmind:
Merged in 10b90d6e155424d58a742c8b79c2042203624167.
Thanks!

PR #49, opened 2021-02-17
[NewMFP] Move monotone framework implementation to revng

chqma-revng:
This commit moves the implementation of the monotone framework
from revng-c to revng

The core part of the maximal fixed point algorithm for the monotone framework is in:
```
include/revng/MFP/MFP.h
```

The last file is a dependency of the MFP implementation that visits the nodes in reverse post order traversal order
```
include/revng/ADT/ReversePostOrderTraversal.h
```

Finally I am adding a parameter `const MFI &I` to `getMaximumFixedPoint` to let us use contextual readonly data, which we cannot easily do if we only allow static methods. 

----

A separate pull request on revng-c will remote these files from the revng-c repo and fix any code that depends on them

### DIFF
--- a/include/revng/MFP/MFP.h
+++ b/include/revng/MFP/MFP.h
@@ -36,7 +36,8 @@ auto successors(typename GT::NodeRef From) {
 }
 
 template<typename MFI>
-concept MonotoneFrameworkInstance = requires(typename MFI::LatticeElement E1,
+concept MonotoneFrameworkInstance = requires(MFI I,
+                                             typename MFI::LatticeElement E1,
                                              typename MFI::LatticeElement E2,
                                              typename MFI::Label L) {
   /// To compute the reverse post order traversal of the graph starting from
@@ -47,9 +48,9 @@ concept MonotoneFrameworkInstance = requires(typename MFI::LatticeElement E1,
           typename llvm::GraphTraits<typename MFI::GraphType>::NodeRef>;
   // Disable clang-format, because it does not handle concepts very well yet
   // clang-format off
-  { MFI::combineValues(E1, E2) } ->same_as<typename MFI::LatticeElement>;
-  { MFI::isLessOrEqual(E1, E2) } ->same_as<bool>;
-  { MFI::applyTransferFunction(L, E2) } ->same_as<typename MFI::LatticeElement>;
+  { I.combineValues(E1, E2) } ->same_as<typename MFI::LatticeElement>;
+  { I.isLessOrEqual(E1, E2) } ->same_as<bool>;
+  { I.applyTransferFunction(L, E2) } ->same_as<typename MFI::LatticeElement>;
   // clang-format on
 };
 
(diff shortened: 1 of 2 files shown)


pfez:
I'm not sure about one thing. Is it really necessary to move `DataFlowGraph` and friends to revng? I think that we only really need to move `MFP.h` and `ReversePostOrderTraversal.h` to revng for now. `DataFlowGraph` and friends are not strictly necessary, and their only users for now are still in revng-c (type shrinking and bit liveness).

I would be happier if `DataFlowGraph` and its users remain in the same repo. We can either leave `DataFlowGraph` and friends in revng-c, or move also TypeShrinking and BitLiveness in revng-c.

@aleclearmind what's your take?

chqma-revng:
The plan is to implement several analyses based on the monotone framework in revng so those will need something like DataFlowGraph

chqma-revng:
> I'm not sure about one thing. Is it really necessary to move `DataFlowGraph` and friends to revng? I think that we only really need to move `MFP.h` and `ReversePostOrderTraversal.h` to revng for now. `DataFlowGraph` and friends are not strictly necessary, and their only users for now are still in revng-c (type shrinking and bit liveness).
> 
> I would be happier if `DataFlowGraph` and its users remain in the same repo. We can either leave `DataFlowGraph` and friends in revng-c, or move also TypeShrinking and BitLiveness in revng-c.
> 
> @aleclearmind what's your take?

The plan is to implement several analyses based on the monotone framework in revng so those will need something like DataFlowGraph

pfez:
> The plan is to implement several analyses based on the monotone framework in revng and those analyses will need something like DataFlowGraph

Will they? If it's something you guys already know for sure, that's fine by me!

aleclearmind:
Mmmh, wait, the analyses will work on `llvm::Function`/`llvm::BasicBlock`, which implement `llvm::GraphTraits`, no need to materialize any other graph data structure.
I'm not sure what `DataFlowGraph` is, but I guess we don't need it.

include/revng/MFP/MFP.h
```diff
@@ -0,0 +1,146 @@
...
+/// GT an instance of llvm::GraphTraits
+template<MonotoneFrameworkInstance MFI,
+         typename GT = llvm::GraphTraits<typename MFI::GraphType>>
+std::map<typename MFI::Label, MFPResult<typename MFI::LatticeElement>>
+getMaximalFixedPoint(const MFI &Instance, const typename MFI::GraphType &Flow,
+                     typename MFI::LatticeElement InitialValue,
+                     typename MFI::LatticeElement ExtremalValue,
+                     const std::vector<typename MFI::Label> &ExtremalLabels) {
+  typedef typename MFI::Label Label;
+  typedef typename MFI::LatticeElement LatticeElement;
```
aleclearmind, line 67: Use `using`

aleclearmind:
Can I have a small test suite in `tests/unit`?

aleclearmind:
This MR has been superseeded by merging 8cee47f8a7f35dac6c6c7d09c84544c64b54d535.
Thanks!

PR #27, opened 2020-12-10
StackAnalysis2

antoniofrighetto:
Importing...

PR #51, opened 2021-02-22
Port ABI analyses to StackAnalysis2 framework

chqma-revng:


include/revng/ABIAnalyses/DeadRegisterArgumentsOfFunction.h
```diff
@@ -0,0 +1,42 @@
...
+  bool isLessOrEqual(const LatticeElement &Lh, const LatticeElement &Rh) const;
+};
+
+llvm::DenseMap<llvm::GlobalVariable *, State>
+analyze(const llvm::Function *Entry,
+        const GeneratedCodeBasicInfo &GCBI,
+        const StackAnalysis::FunctionProperties &FP);
+} // namespace DeadRegisterArgumentsOfFunction
+
+namespace llvm {}; // namespace llvm
```
aleclearmind, line 42: What is this for?

include/revng/ABIAnalyses/UsedReturnValuesOfFunctionCall.h
```diff
@@ -0,0 +1,42 @@
...
+#include "revng/StackAnalysis/StackAnalysis.h"
+#include "revng/Support/revng.h"
+
+namespace UsedReturnValuesOfFunctionCall {
+
+enum State { Unknown, Maybe, Yes };
+
+// special register index, if found inside LatticeElement
+// it means that it is the first time visiting the BasicBlock
+const int32_t InitialRegisterState = -1;
```
aleclearmind, line 24: In a monotone framework, nodes not visited yet are initialized to bottom (unless they're extremal points).
Put bottom in your lattice.

include/revng/MFP/MFP.h
```diff
@@ -125,22 +107,47 @@ getMaximalFixedPoint(const MFI &Instance, const typename MFI::GraphType &Flow,
...
+/// Compute the maximum fixed points of an instance of monotone framework
+/// GT an instance of llvm::GraphTraits
+template <MonotoneFrameworkInstance MFI,
+          typename GT = llvm::GraphTraits<typename MFI::GraphType>>
+std::map<typename MFI::Label, MFPResult<typename MFI::LatticeElement>>
+getMaximalFixedPoint(const MFI &Instance, const typename MFI::GraphType &Flow,
+                     typename MFI::LatticeElement InitialValue,
+                     typename MFI::LatticeElement ExtremalValue,
+                     const std::vector<typename MFI::Label> &ExtremalLabels) {
+  typedef typename MFI::Label Label;
```
aleclearmind, line 137:
```suggestion
  using Label = typename MFI::Label;
```

include/revng/MFP/MFP.h
```diff
@@ -84,32 +84,14 @@ getMaximalFixedPoint(const MFI &Instance, const typename MFI::GraphType &Flow,
...
-      }
-    }
-  }
-  // Start visits for nodes that we still haven't visited
-  // prioritizing extremal nodes
-  std::vector<Label> Nodes(ExtremalLabels);
-  for (Label Node : llvm::nodes(Flow)) {
-    Nodes.push_back(Node);
-  }
-  for (Label Start : Nodes) {
```
aleclearmind: Can you elaborate what this change is doing? I mean, the whole " [WIP][NewMFPABI] Let user specify the entry nodes" commit.

lib/ABIAnalyses/ABIAnalysisPass.cpp
```diff
@@ -0,0 +1,96 @@
+#include "llvm/Support/CommandLine.h"
```
aleclearmind, line 1: Header?

lib/ABIAnalyses/ABIAnalysisPass.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include "revng/ABIAnalyses/ABIAnalysisPass.h"
+#include "revng/ABIAnalyses/DeadReturnValuesOfFunctionCall.h"
+#include "revng/ABIAnalyses/DeadRegisterArgumentsOfFunction.h"
+#include "revng/ABIAnalyses/UsedRegisterArgumentsOfFunction.h"
+#include "revng/ABIAnalyses/UsedReturnValuesOfFunctionCall.h"
+#include "llvm/IR/Function.h"
+#include "llvm/IR/Instruction.h"
+#include "llvm/IR/Instructions.h"
+#include <iostream>
+#include <string>
```
aleclearmind, line 12: Mispositioned

lib/ABIAnalyses/ABIAnalysisPass.cpp
```diff
@@ -0,0 +1,96 @@
...
+using namespace llvm;
+
+static cl::OptionCategory RevNgCategory("revng new-abi-analysis");
```
aleclearmind, line 15: Drop `RevNg`/`revng`

lib/ABIAnalyses/ABIAnalysisPass.cpp
```diff
@@ -0,0 +1,96 @@
...
+             cl::cat(RevNgCategory));
+char ABIAnalysisPass::ID = 0;
+
+static llvm::RegisterPass<ABIAnalysisPass>
+    X("new-abi-analysis", "abianalysis Functions Pass", true, true);
+
+bool ABIAnalysisPass::runOnFunction(llvm::Function &F) {
+  // Retrieve analysis of the GeneratedCodeBasicInfo pass
+  auto &GCBI = getAnalysis<GeneratedCodeBasicInfoWrapperPass>().getGCBI();
+  if (F.getName().str().compare(FilterFunction) != 0) {
```
aleclearmind, line 30:
```suggestion
  if (F.getName() != FilterFunction) {
```

include/revng/ABIAnalyses/UsedReturnValuesOfFunctionCall.h
```diff
@@ -0,0 +1,42 @@
...
+struct MFI : ABIAnalyses::ABIAnalysis {
+  using LatticeElement = llvm::DenseMap<int32_t, State>;
```
aleclearmind, line 26: Don't use `int32_t`, use a proper `enum` (the one from `Binary.h`, unless there's a good reason not to).

lib/ABIAnalyses/DeadReturnValuesOfFunctionCall.cpp
```diff
@@ -0,0 +1,142 @@
+#include <cstdio>
+#include <iostream>
+#include <llvm/IR/Instructions.h>
+#include <llvm/Support/Casting.h>
```
aleclearmind, line 4:
```suggestion
#include "llvm/IR/Instructions.h"
#include "llvm/Support/Casting.h"
```

Are you using check conventions?

lib/ABIAnalyses/DeadReturnValuesOfFunctionCall.cpp
```diff
@@ -0,0 +1,142 @@
...
+    if (Rh.count(E.first) == 0) {
+      return false;
+    }
+    auto RhState = Rh.lookup(E.first);
+    if (getMax(E.second, RhState) != RhState) {
+      return false;
+    }
+  }
+  return true;
+}
```
aleclearmind, line 67: This is quite involved code. Autogenerating this as we where doing still seems like a good idea.
One of the grand-goals of StackAnalysis2 (and ABIAnalysis2) is to be able to easily play around with analyses.
Writing all these analyses by hand is effectively making the code base more rigid than it was before.

lib/ABIAnalyses/DeadReturnValuesOfFunctionCall.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+LatticeElement MFI::applyTransferFunction(Label L,
+                                          const LatticeElement &E) const {
+  LatticeElement New = E;
+  for (auto &I : *L) {
+    switch (classifyInstruction(&I)) {
+    case TheCall: {
+      if (New.count(InitialRegisterState) != 0) {
+        // The first time we hit TheCall is actually the start of our graph
+        New.clear();
```
aleclearmind, line 77: This doesn't seem a good design. This should not be necessary.
Can't you start *right after* the call site (assuming it's a forward analysis)?

aleclearmind (changes requested):

lib/ABIAnalyses/DeadReturnValuesOfFunctionCall.cpp
chqma-revng: not sure if autogenerating code is the solution as it can easily get out of hand (e.g. a generator full of if conditions for each variation of a snippet of code, generated code that is not understandable)

I can refactor out the code that does the key by key comparison in the Maps and that should improve the readability

include/revng/ABIAnalyses/UsedReturnValuesOfFunctionCall.h
chqma-revng: Will do as soon as the `feature/stackanalysis` branch is rebased

include/revng/ADT/ReversePostOrderTraversal.h
```diff
@@ -0,0 +1,39 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "llvm/ADT/PostOrderIterator.h"
+
+namespace llvm {
+
+template <class GraphT, class GT = llvm::GraphTraits<GraphT>,
+          class SetType = std::set<typename llvm::GraphTraits<GraphT>::NodeRef>>
+class ReversePostOrderTraversalExt {
```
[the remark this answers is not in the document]
chqma-revng: This comes from `revng-c`, it returns the nodes in `RPOT` order of a visit that start from a set of nodes instead of a single node which is useful when you have `DAGs` or you only need a subgraph.
It is similar to `ReversePostOrderTraversal` from llvm which works well for a single entry node only

include/revng/MFP/MFP.h
chqma-revng: This is an issue that arises when we want to use certain `Graphs` from `llvm` as is

For example `GraphTraits<BasisBlock*>` does not implement the complete `GraphTraits` so we cannot iterate over the `nodes` inside that graph without doing a visit first.

Since we already need to do a visit of the graph to get the `RPOT` order of the graph, we only need the user to supply the starting nodes of this visit

lib/ABIAnalyses/DeadReturnValuesOfFunctionCall.cpp
chqma-revng: Yes, but it's with the graphs that we are using (`GraphTraits<BasicBlock*>`) 

Some alternatives:
- checking if the current `BasicBlock` is the parent of the `CallSite` and skipping the first instructions
- instead of clearing, assigning a named constant like `InitialValue` or something similar

include/revng/ABIAnalyses/UsedReturnValuesOfFunctionCall.h
chqma-revng: Yes this is a hack to support `GraphTraits` where we need to know if this is really an unvisited node (i.e. we need a value that will never be produced by any transfer function).

Ex. for `GraphTrats<BasicBlock*>` and analyses that start from a `CallSite` (i.e. potentially in the middle of a `BasicBlock`) and have a special transfer function associated to that in case we reach the call again via a back edge

This value was put to keep using a standard container (`Map` like data structures), but turning it to a flag in a `struct` would work too

lib/ABIAnalyses/DeadReturnValuesOfFunctionCall.cpp
aleclearmind: Have you seen how we're doing it now? In `$BUILD_DIR/lib/StackAnalysis/ABIDataFlows.h`.
It's highly readable IMO, despite verbose. Also, generating code is less prone to copy and paste errors/forgetting edge cases.

`scripts/monotone-framework.py` doesn't handle that many special cases. If we have the need for special cases (the `-1` thing?), we should redesign the things in order not to have them.

I'd just like to autogenerate transfer functions and combine (as we're doing right now), nothing more.

lib/ABIAnalyses/DeadReturnValuesOfFunctionCall.cpp
aleclearmind: Mmmmh, I see. But let's strive to avoid special casing at all costs.
Would splitting the basic block make things easier?
It'd be an harmless change. After all we are generating functions on the fly and then discarding them. If you need the IR in some form that does not disrupt what @antoniofrighetto is doing, let's go for it!

chq-matteo: yes having `Calls` on their own `BasicBlock` would be sufficient

aleclearmind: OK, let's go for it! `\o/`

lib/ABIAnalyses/DeadReturnValuesOfFunctionCall.cpp
chq-matteo: if we limit ourselves to simple lattices I agree.

I am in favour of generating the various components of the underlying lattices (the lattice when you consider a single register), just not for the product lattice

chq-matteo: which I think is what is happening right now

chqma-revng: if we limit ourselves to simple lattices I agree.

I am in favour of generating the various components of the underlying lattices (the lattice when you consider a single register), just not for the product lattice

which I think is what we are doing with `scripts/monotone-framework.py`

aleclearmind: Yup, exactly! Generating the product lattice is not even possible since it depends on run-time stuff, I think.

chq-matteo: sgtm

include/revng/ABIAnalyses/UsedRegisterArgumentsOfFunction.h
```diff
@@ -0,0 +1,42 @@
...
+  bool isLessOrEqual(const LatticeElement &Lh, const LatticeElement &Rh) const;
+};
+
+llvm::DenseMap<llvm::GlobalVariable *, State>
+analyze(const llvm::Function *Entry,
+        const GeneratedCodeBasicInfo &GCBI,
+        const StackAnalysis::FunctionProperties &FP);
+} // namespace UsedRegisterArgumentsOfFunction
+
+namespace llvm {}; // namespace llvm
```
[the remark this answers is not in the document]
chqma-revng: removed

include/revng/ABIAnalyses/UsedRegisterArgumentsOfFunction.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include "revng/ABIAnalyses/Common.h"
+#include "revng/BasicAnalyses/GeneratedCodeBasicInfo.h"
+#include "revng/MFP/MFP.h"
+#include "revng/StackAnalysis/StackAnalysis.h"
+#include "revng/Support/revng.h"
+
+namespace UsedRegisterArgumentsOfFunction {
+
+enum State { Unknown, Maybe, Yes };
```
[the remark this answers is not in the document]
chqma-revng: now these constants are automatically generated, the final results are given using the constants in `Binary.h`

include/revng/ABIAnalyses/Common.h
```diff
@@ -0,0 +1,165 @@
...
+
+  llvm::DenseMap<llvm::GlobalVariable *, int32_t> ABIRegisters;
+  llvm::DenseMap<int32_t, llvm::GlobalVariable *> IndexABIRegisters;
+
+  const llvm::Instruction *CallSite;
+  const StackAnalysis::FunctionProperties &Properties;
+  const GeneratedCodeBasicInfo &GCBI;
+
+  bool isABIRegister(const llvm::Value *) const;
+  bool isABIRegister(int32_t RegID) const;
```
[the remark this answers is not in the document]
chqma-revng: In the end I used the CSV variables because the enums in `Binary.h` wouldn't play nice with llvm containers

include/revng/ABIAnalyses/UsedReturnValuesOfFunctionCall.h
chqma-revng: In the end I couldn't use the enums in `Binary.h` as key, I am using the `CSV` variables directly now

lib/ABIAnalyses/DeadReturnValuesOfFunctionCall.cpp
chqma-revng: I wasn't, should be fixed now

lib/ABIAnalyses/DeadReturnValuesOfFunctionCall.cpp
chqma-revng: Now the analysis expects that each callsite is in its own basic block
So `InitialValue` is not needed anymore

include/revng/ABIAnalyses/UsedReturnValuesOfFunctionCall.h
chqma-revng: Resolved requiring `CallSites` to be in their own basic block

lib/ABIAnalyses/ABIAnalysisPass.cpp
```diff
@@ -48,14 +50,47 @@ bool ABIAnalysisPass::runOnFunction(llvm::Function &F) {
   }
   std::cerr << "---------------- END DeadReturnValuesOfFunctionCall -------------------\n";
 
-  std::cerr << "---------------- START DeadRegisterArgumentsOfFunction ----------------\n";
+  std::cerr << "---------------- START UsedReturnValuesOfFunctionCall ----------------\n";
```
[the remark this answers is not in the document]
chqma-revng: Moved to revng_log

include/revng/ABIAnalyses/ABIAnalysis.h
```diff
@@ -0,0 +1,98 @@
...
+/// the outlined function must have all original function calls
+/// replaced with a basic block starting with a call to @precall_hook
+/// followed by a summary of the side effects of the function
+/// followed by a call to @postcall_hook and a basic block terminating
+/// instruction
+inline void
+analyzeOutlinedFunction(Function *F, const GeneratedCodeBasicInfo &GCBI) {
+  // find summary blocks
+  revng_log(ABILogger, "Analyzing function:\n" << F);
+  ABILogger << "------- start UsedArgumentsOfFunction --------\n";
```
aleclearmind, line 35: I don't like `-----`. Use logger indentation instead of `start`/`end`.

include/revng/ABIAnalyses/ABIAnalysis.h
```diff
@@ -0,0 +1,98 @@
...
+using llvm::ReturnInst;
+
+inline Logger<> ABILogger("new-abi");
+
+/// Run all abi analyses on the oulined function F
+/// the outlined function must have all original function calls
+/// replaced with a basic block starting with a call to @precall_hook
+/// followed by a summary of the side effects of the function
+/// followed by a call to @postcall_hook and a basic block terminating
+/// instruction
```
aleclearmind, line 30: Wrap at 80 cols

include/revng/ABIAnalyses/ABIAnalysis.h
```diff
@@ -0,0 +1,98 @@
...
+              << "\n";
+  }
+  ABILogger << "------- end UsedArgumentsOfFunction --------\n";
+  ABILogger << "------- start DeadRegisterArgumentsOfFunction --------\n";
+  for (auto &[GV, State] :
+       DeadRegisterArgumentsOfFunction::analyze(&F->getEntryBlock(), GCBI)) {
+    ABILogger << GV->getName() << " = " << model::RegisterState::getName(State)
+              << "\n";
+  }
+  ABILogger << "------- end DeadRegisterArgumentsOfFunction --------\n";
```
aleclearmind, line 48: Why these prints, can't you collect results in a proper way and return them?

include/revng/ABIAnalyses/ABIAnalysis.h
```diff
@@ -0,0 +1,98 @@
...
+
+  for (auto &BB : *F) {
+    for (auto &I : BB) {
```
aleclearmind, line 51:
```c++
for (Instruction *I : llvm::instructions(F)
```
Use `auto` only if the type is ugly or present on the same line.

include/revng/ABIAnalyses/ABIAnalysis.h
```diff
@@ -0,0 +1,98 @@
...
+      if (auto *C = dyn_cast<CallInst>(&I)) {
+        if (C->getCalledFunction()->getName().contains("precall_hook")) {
```
aleclearmind, line 53: At the beginning of the function do something like `Module->getFunction("precall_hook")` and then compare against that, instead of using strings.

include/revng/ABIAnalyses/Analyses.h
```diff
@@ -0,0 +1,61 @@
...
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "revng/ABIAnalyses/Generated/DeadRegisterArgumentsOfFunction.h"
+#include "revng/ABIAnalyses/Generated/DeadReturnValuesOfFunctionCall.h"
+#include "revng/ABIAnalyses/Generated/RegisterArgumentsOfFunctionCall.h"
+#include "revng/ABIAnalyses/Generated/UsedArgumentsOfFunction.h"
+#include "revng/ABIAnalyses/Generated/UsedReturnValuesOfFunction.h"
+#include "revng/ABIAnalyses/Generated/UsedReturnValuesOfFunctionCall.h"
```
aleclearmind, line 12: Are these generated in the source directory?

include/revng/ABIAnalyses/Analyses.h
```diff
@@ -0,0 +1,61 @@
...
+
+namespace ABIAnalyses {
+namespace DeadRegisterArgumentsOfFunction {
+using namespace llvm;
+
+DenseMap<const GlobalVariable *, State>
```
aleclearmind, line 18: What's `State`?
Consider that `DenseMap` are designed to be used to map register-side stuff to register-stuff (typically, pointers to pointers). If `State` is larger than that, go for a `std::map`.

include/revng/ABIAnalyses/Common.h
```diff
@@ -0,0 +1,213 @@
...
+
+using llvm::BasicBlock;
+using llvm::CallInst;
+using llvm::GlobalVariable;
+using llvm::Instruction;
+using llvm::LoadInst;
+using llvm::SmallSet;
+using llvm::SmallVector;
+using llvm::StoreInst;
+using llvm::Value;
```
aleclearmind, line 32: You're pulling all of these into the `ABIAnalyses` namespace.
People could use `ABIAnalyses::BasicBlock`, not sure this is what we want.

include/revng/ABIAnalyses/Common.h
```diff
@@ -0,0 +1,213 @@
...
+
+struct ABIAnalysis {
+private:
+  SmallSet<GlobalVariable *, 20> ABIRegisters;
+  SmallVector<GlobalVariable *, 20> RegisterList;
+  const Instruction *CallSite;
+
+public:
+  ABIAnalysis(const GeneratedCodeBasicInfo &GCBI) :
+    ABIAnalysis(nullptr, GCBI){};
```
aleclearmind, line 52: Space before `{`. Are you using `clang-format`?

include/revng/ABIAnalyses/Common.h
```diff
@@ -0,0 +1,213 @@
...
+
+  ABIAnalysis(const Instruction *CS, const GeneratedCodeBasicInfo &GCBI) :
+    RegisterList(), CallSite(CS) {
+
+    for (auto *CSV : GCBI.abiRegisters())
+      if (CSV) {
+        ABIRegisters.insert(CSV);
+        RegisterList.emplace_back(CSV);
+      }
```
aleclearmind, line 61: A scope without `{}` should not contain a scope with `{}`

include/revng/ABIAnalyses/Common.h
```diff
@@ -0,0 +1,213 @@
...
+
+template<typename LatticeElementMap, typename CoreLattice>
+LatticeElementMap
+combineValues(const LatticeElementMap &Lh, const LatticeElementMap &Rh) {
+
+  LatticeElementMap New = Lh;
+  for (const auto &[Reg, S] : Rh) {
+    New[Reg] = CoreLattice::
+      combineValues(getOrDefault<LatticeElementMap, CoreLattice>(New, Reg),
+                    getOrDefault<LatticeElementMap, CoreLattice>(Rh, Reg));
```
aleclearmind, line 97:
```suggestion
    auto Lol = getOrDefault<LatticeElementMap, CoreLattice>(New, Reg);
    auto Lel = getOrDefault<LatticeElementMap, CoreLattice>(Rh, Reg);
    New[Reg] = CoreLattice::combineValues(Lol, Lel);
```
Same SLOCs, much more readable.

include/revng/ABIAnalyses/Common.h
```diff
@@ -0,0 +1,213 @@
...
+                                    getOrDefault<LatticeElementMap,
+                                                 CoreLattice>(Rh, Reg))) {
+      return false;
+    }
+  }
+  return true;
+}
+
+inline bool ABIAnalysis::isABIRegister(const Value *V) const {
+  if (const auto &G = dyn_cast<GlobalVariable>(V)) {
```
aleclearmind, line 124: Mmmh, weird. I think you're deducing `const GlobalVariable *&`.

include/revng/ABIAnalyses/Common.h
```diff
@@ -0,0 +1,213 @@
...
+    if (ABIRegisters.count(G) != 0) {
+      return true;
+    }
```
aleclearmind, line 127:
```suggestion
    return ABIRegisters.count(G) != 0;
```

include/revng/ABIAnalyses/Common.h
```diff
@@ -0,0 +1,213 @@
...
+  }
+  return false;
+}
+
+inline bool isCallSiteBlock(const BasicBlock *B) {
+  if (auto *C = dyn_cast<CallInst>(&*B->getFirstInsertionPt())) {
+    if (C->getCalledFunction()->getName().contains("precall_hook")) {
```
aleclearmind, line 134: Minimize comparisons with names.

I'm a bit confused by the fact that you're exclusively using `precall_hook`.
Aren't you supposed to use `postcall_hook` for forward analyses?

include/revng/ABIAnalyses/Common.h
```diff
@@ -0,0 +1,213 @@
...
+    if (isABIRegister(S->getPointerOperand())) {
+      if (isCallSiteBlock(I->getParent())) {
+        return WeakWrite;
+      }
+      return Write;
+    }
+    break;
+  }
+  case Instruction::Load: {
+    auto L = cast<LoadInst>(I);
```
aleclearmind, line 169: Don't let `auto` deduce `*`
```suggestion
    auto *L = cast<LoadInst>(I);
```

include/revng/ABIAnalyses/Common.h
```diff
@@ -0,0 +1,213 @@
...
+    if (isABIRegister(L->getPointerOperand())) {
+      return Read;
+    }
+    break;
+  }
+  case Instruction::Call: {
+    if (I == CallSite) {
+      return TheCall;
+    }
+  }
```
aleclearmind, line 179: Missing `break`. No compiler warning?

include/revng/ABIAnalyses/Common.h
```diff
@@ -0,0 +1,213 @@
...
+inline TransferKind
+ABIAnalysis::classifyInstruction(const Instruction *I) const {
+  switch (I->getOpcode()) {
+  case Instruction::Store: {
+    auto S = cast<StoreInst>(I);
+    if (isABIRegister(S->getPointerOperand())) {
+      if (isCallSiteBlock(I->getParent())) {
+        return WeakWrite;
+      }
+      return Write;
```
aleclearmind, line 164: Consider
```suggestion
      if (isCallSiteBlock(I->getParent())) {
        return WeakWrite;
      } else {
        return Write;
      }
```

Or even better...

```suggestion
      return isCallSiteBlock(I->getParent())) ? WeakWrite : Write;
```

include/revng/ABIAnalyses/Common.h
```diff
@@ -0,0 +1,213 @@
...
+}
+
+inline SmallVector<const GlobalVariable *, 1>
+ABIAnalysis::getRegistersRead(const Instruction *I) const {
+  SmallVector<const GlobalVariable *, 1> Result;
+  switch (I->getOpcode()) {
+  case Instruction::Load: {
+    auto L = cast<LoadInst>(I);
+    if (isABIRegister(L->getPointerOperand())) {
+      Result.push_back(cast<GlobalVariable>(L->getPointerOperand()));
```
aleclearmind, line 206: Declare local variable for `L->getPointerOperand()`

include/revng/ABIAnalyses/Generated/DeadRegisterArgumentsOfFunction.h
```diff
@@ -0,0 +1,169 @@
+#pragma once
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+// This file has been automatically generated from
+// scripts/monotone-framework-lattice.py, please don't change it
```
aleclearmind, line 8: Submitting for review commits with generated code.
What a brave man. I admire you. 😄

include/revng/ABIAnalyses/Generated/DeadRegisterArgumentsOfFunction.h
```diff
@@ -0,0 +1,169 @@
...
+    case TransferFunction::Write:
+      switch (E) {
+      case LatticeElement::Maybe:
+        return LatticeElement::NoOrDead;
+      case LatticeElement::Unknown:
+        return LatticeElement::Unknown;
+      case LatticeElement::NoOrDead:
+        return LatticeElement::NoOrDead;
+      default:
+        return E;
```
aleclearmind, line 99: Shouldn't we abort?

include/revng/ABIAnalyses/Generated/DeadRegisterArgumentsOfFunction.h
```diff
@@ -0,0 +1,169 @@
...
+  using LGT = GraphType;
+
+  LatticeElement
+  combineValues(const LatticeElement &Lh, const LatticeElement &Rh) const {
+    return ABIAnalyses::combineValues<LatticeElement, CoreLattice>(Lh, Rh);
+  };
+
+  bool isLessOrEqual(const LatticeElement &Lh, const LatticeElement &Rh) const {
+    return ABIAnalyses::isLessOrEqual<LatticeElement, CoreLattice>(Lh, Rh);
+  };
```
aleclearmind, line 126: I think all this stuff is basically a literal in the code generator.
Move this into a base class (using template + CRTP).

Let's make generators generate what's strictly required.

include/revng/ABIAnalyses/Generated/DeadRegisterArgumentsOfFunction.h
```diff
@@ -0,0 +1,169 @@
...
+                                                        CoreLattice>(New, Reg));
+        }
+        break;
+      default:
+        break;
+      }
+    }
+    return New;
+  };
+};
```
aleclearmind, line 167: I don't think we need to generated this.
It can be just a template function/method.

include/revng/ABIAnalyses/Generated/DeadReturnValuesOfFunctionCall.h
```diff
@@ -0,0 +1,182 @@
...
+
+namespace ABIAnalyses::DeadReturnValuesOfFunctionCall {
+
+using namespace ABIAnalyses;
+using Register = model::Register::Values;
+using State = model::RegisterState::Values;
+
+struct CoreLattice {
+
+  // using LatticeElement = model::RegisterState::Values;
```
aleclearmind, line 30: Drop?

lib/ABIAnalyses/ABIAnalysis.template
```diff
@@ -0,0 +1,103 @@
...
+        }
+        break;
+      default:
+        break;
+      }
+    }
+    return New;
+  };
+};
+
```
aleclearmind, line 102: Too much literal code. Move in a base class (using CRTP, I think).

lib/ABIAnalyses/DeadReturnValuesOfFunctionCall.cpp
```diff
@@ -0,0 +1,52 @@
...
+  DenseMap<const GlobalVariable *, State> RegNoOrDead{};
+
+  for (auto &[BB, Result] : Results) {
+    for (auto &[GV, RegState] : Result.OutValue) {
+      if (RegState == CoreLattice::NoOrDead) {
+        RegNoOrDead[GV] = State::NoOrDead;
+      }
+    }
+  }
+  return RegNoOrDead;
```
aleclearmind, line 50: Use `Binary.h` enum.
I was under the impression that this branch has been rebased.

lib/StackAnalysis/StackAnalysis.cpp
```diff
@@ -788,7 +787,7 @@ llvm::Function *CFEPAnalyzer<FunctionOracle>::createDisposableFunction(
   CodeExtractorAnalysisCache CEAC(*Root);
   Function *OutlinedFunction = CodeExtractor(BlocksToExtract)
                                  .extractCodeRegion(CEAC);
-
+
```
aleclearmind, line 790: ?

lib/TypeShrinking/BitLiveness.cpp
```diff
@@ -33,17 +33,17 @@ struct BitLivenessAnalysis {
...
+  using MFPResult = MFP::MFPResult<BitLivenessAnalysis::LatticeElement>;
 
-  static uint32_t combineValues(const uint32_t &Lh, const uint32_t &Rh) {
+  uint32_t combineValues(const uint32_t &Lh, const uint32_t &Rh) const {
     return std::max(Lh, Rh);
   }
 
-  static bool isLessOrEqual(const uint32_t &Lh, const uint32_t &Rh) {
+  bool isLessOrEqual(const uint32_t &Lh, const uint32_t &Rh) const {
     return Lh <= Rh;
```
aleclearmind, line 43: `sed s/Lh/LHS/`
`sed s/Rh/RHS/`

For coherence with other parts of the codebase.

scripts/gen_abi_analysis_lattices.sh
```diff
@@ -0,0 +1,6 @@
...
+python3 scripts/monotone-framework-lattice.py lib/ABIAnalyses/ABIAnalysis.template lib/ABIAnalyses/DeadReturnValuesOfFunctionCall.dot > include/revng/ABIAnalyses/Generated/DeadReturnValuesOfFunctionCall.h
+python3 scripts/monotone-framework-lattice.py lib/ABIAnalyses/ABIAnalysis.template lib/ABIAnalyses/UsedArgumentsOfFunction.dot > include/revng/ABIAnalyses/Generated/UsedArgumentsOfFunction.h
+python3 scripts/monotone-framework-lattice.py lib/ABIAnalyses/ABIAnalysis.template lib/ABIAnalyses/UsedReturnValuesOfFunctionCall.dot > include/revng/ABIAnalyses/Generated/UsedReturnValuesOfFunctionCall.h
+python3 scripts/monotone-framework-lattice.py lib/ABIAnalyses/ABIAnalysis.template lib/ABIAnalyses/RegisterArgumentsOfFunctionCall.dot > include/revng/ABIAnalyses/Generated/RegisterArgumentsOfFunctionCall.h
+python3 scripts/monotone-framework-lattice.py lib/ABIAnalyses/ABIAnalysis.template lib/ABIAnalyses/UsedReturnValuesOfFunction.dot > include/revng/ABIAnalyses/Generated/UsedReturnValuesOfFunction.h
```
aleclearmind, line 6: Integrate with CMake. Take a look at how the old script was doing it.

scripts/monotone-framework-lattice.py
```diff
@@ -0,0 +1,230 @@
+import argparse
```
aleclearmind, line 1: Header and shebang (copy from other scripts)

scripts/monotone-framework-lattice.py
```diff
@@ -0,0 +1,230 @@
...
+import sys
+from collections import defaultdict
+from pygraphviz import AGraph
+
+import importlib
+monotone_framework = importlib.import_module("monotone-framework")
```
aleclearmind, line 7: Uh? Why?

scripts/monotone-framework-lattice.py
```diff
@@ -0,0 +1,230 @@
+import argparse
+import sys
+from collections import defaultdict
+from pygraphviz import AGraph
```
aleclearmind, line 4: `pygraphviz` sucks. We're trying to get rid of it. Can you take a look if you can find some other (possibly pure-Python) library able to parse GraphViz stuff (without going full `graph_tool`)?

aleclearmind (changes requested):

scripts/monotone-framework-lattice.py
chqma-revng: you cannot import modules with "-" in the name using normal import statements

maybe it would be better to rename the old script, but I didn't want to mess with the existing stuff

include/revng/ABIAnalyses/Analyses.h
chqma-revng: state is the `model::RegisterState::Values` enum

lib/ABIAnalyses/DeadReturnValuesOfFunctionCall.cpp
chqma-revng: yes, so the intermediate results use `CoreLattice::LatticeElement` enums that are generated from the `.dot` file spec

The result is `State` which is `model::RegisterState::Values` from `Binary.h`

The reason is that the lattice contains states that are not in the model and may not make sense to add to the model (i.e. `Bottom` which is in some analyses)

aleclearmind: Can't you use `Invalid` as `Bottom`?

scripts/monotone-framework-lattice.py
aleclearmind: Go ahead and rename

include/revng/ABIAnalyses/Common.h
chqma-revng: so the assumption is documented in `ABIAnalysis.h`

```
/// Run all abi analyses on the outlined function F
/// the outlined function must have all original function calls
/// replaced with a basic block starting with a call to @precall_hook
/// followed by a summary of the side effects of the function
/// followed by a call to @postcall_hook and a basic block terminating
/// instruction
```

So if the block starts with `@precall_hook` it is enough to identify it as a call site.

Forward analyses that start on a call site with start the analysis from the unique successor of a call site block, likewise backwards analyses will start from the unique predecessor, but this is done in the respective analysis cpp file

`isCallSiteBlock` is used to quickly determine if a `write` is a true write or a `weak write`

chqma-revng: so the assumption is documented in `ABIAnalysis.h`

```
/// Run all abi analyses on the outlined function F
/// the outlined function must have all original function calls
/// replaced with a basic block starting with a call to @precall_hook
/// followed by a summary of the side effects of the function
/// followed by a call to @postcall_hook and a basic block terminating
/// instruction
```

So if the block starts with `@precall_hook` it is enough to identify it as a call site.

Forward analyses that start on a call site with start the analysis from the unique successor of a call site block, likewise backwards analyses will start from the unique predecessor, but this is done in the respective analysis cpp file

`isCallSiteBlock` is used to quickly determine if a `write` is a true write or a `weak write`

include/revng/ABIAnalyses/Common.h
chqma-revng: it's actually `check-revng-conventions` that removes the space

aleclearmind: OK, `check-revng-conventions` is always right.

include/revng/ABIAnalyses/Common.h
aleclearmind: I'm not very familiar with the IR you're describing.
Could you please paste some relevant parts?

chqma-revng: Sure, it was introduced very recently to avoid having a special guard value to identify the `TheCall` transfer function.

Turns out this way of splitting the function summaries to a separate block is also useful to identify `weak writes` (otherwise we would need to scan the whole basic block looking for calls to `[pre/post]call_hooks`

from `lib/StackAnalysis/StackAnalysis.cpp:647`
```c++

template<class FunctionOracle>
llvm::BasicBlock *
CFEPAnalyzer<FunctionOracle>::integrateFunctionCallee(llvm::BasicBlock *BB) {
  using namespace llvm;

  // In case the basic block has a call-site, the function call is
  // replaced with i) hooks that facilitate the ABI analysis ii) summary
  // of the registers clobbered by that function. The newly created basic block
  // is returned so that it can be added to the ValueToValueMap.
  BasicBlock *Split = nullptr;
  auto *Term = BB->getTerminator();
  auto *Call = getFunctionCall(Term);

  // If the successor is null, we are dealing with an indirect call.
  auto *Next = getFunctionCallCallee(Term);

  // What are the registers clobbered by the callee?
  const auto &ClobberedRegs = Oracle.getRegistersClobbered(Next);

  // What is the function type of the callee?
  FunctionTypeValue Ty = Oracle.getFunctionType(Next);

  switch (Ty) {
  case FunctionTypeValue::Regular: {
    BranchInst *Br = BranchInst::Create(getFallthrough(Term));
    ReplaceInstWithInst(Term, Br);

    auto *FTy = llvm::FunctionType::get(llvm::Type::getVoidTy(Context), false);
    auto *PreHookOpqF = OFPHooksFunctionCall.get("__precall_pure_function",
                                                 FTy,
                                                 "precall_hook");
    auto *Last = CallInst::Create(PreHookOpqF, "", Br);

    Split = BB->splitBasicBlock(Last->getPrevNode(),
                                BB->getName() + Twine("__summary"));

    // Prevent the store instructions from being optimized out by storing
    // the rval of a call to an opaque function into the clobbered registers.
    OFPRegistersClobbered.addFnAttribute(Attribute::ReadOnly);
    OFPRegistersClobbered.addFnAttribute(Attribute::NoUnwind);

    for (GlobalVariable *Reg : ClobberedRegs) {
      auto *CSVTy = Reg->getType()->getPointerElementType();
      auto *RegsClobbOpqF = OFPRegistersClobbered
                              .get(BB->getParent()->getName(),
                                   CSVTy,
                                   {},
                                   "registers_clobbered");
      new StoreInst(CallInst::Create(RegsClobbOpqF, "", Br), Reg, Br);
    }

    auto *PostHookOpqF = OFPHooksFunctionCall.get("__postcall_pure_function",
                                                  FTy,
                                                  "postcall_hook");
    CallInst::Create(PostHookOpqF, "", Split->getTerminator());
    break;
  }
....
  Call->eraseFromParent();
  return Split;
}

```

lib/ABIAnalyses/DeadReturnValuesOfFunctionCall.cpp
chqma-revng: makes sense, then, if we are going that way, I propose to use `Invalid` instead of `Bottom` and to add `Unknown` as a register state, would that work?

include/revng/ABIAnalyses/Generated/DeadRegisterArgumentsOfFunction.h
chqma-revng: wouldn't have caught this if I didn't commit the generated code 😆

include/revng/ABIAnalyses/Common.h
aleclearmind: No, I meant the IR, sorry.

lib/ABIAnalyses/DeadReturnValuesOfFunctionCall.cpp
aleclearmind: I forgot about `Unknown`. Mh, I don't want to expose an implementation detail to the Model, which is our long term disk format.
Keep things as they are.

include/revng/ABIAnalyses/Common.h
chqma-revng: strangely no warning, added break

include/revng/ABIAnalyses/Analyses.h
chqma-revng: not anymore

chqma-revng: now using cmake like `monotone-framework.py`

include/revng/ABIAnalyses/Common.h
chqma-revng: So for example this is the IR these analyses are supposed to work on

```llvm
define internal void @root.bb._start__cloned() {
newFuncRoot:
  br label %bb._start__cloned, !dbg !5768

dispatcher.entry.exitStub:                       
  ret void

bb._start__cloned:                                
  call void (%struct.PlainMetaAddress, i64, i32, i8*, ...) 
  .... ; normal block
  br label %bb._start__cloned__summary, !dbg !5796

bb._start__cloned__summary:                       ; <--- special block
  call void @precall_hook.1()                        ; <---- precall hook call
  ; here you would find load / stores normally
  call void @postcall_hook.2()                     ; <---- postcall hook call
  br label %bb._start.0x5__cloned, !dbg !5797

bb._start.0x5__cloned:                            
  call void (%struct.PlainMetaAddress, i64, i32, i8*, ...)
  .... ; normal block
  br label %anypc__cloned, !dbg !5810, !revng.jt.reasons !5811

anypc__cloned:                                    ; preds = %bb._start.0x5__cloned
  br label %dispatcher.entry.exitStub, !dbg !5812, !revng.block.type !5813
}
```

include/revng/ABIAnalyses/Common.h
chqma-revng: definitely not what I wanted, removed and added `using namespace llvm;` in some function bodies when needed

include/revng/ABIAnalyses/ABIAnalysis.h
chqma-revng: returning a struct with the results and added a `dump` method to the struct

include/revng/ABIAnalyses/Generated/DeadRegisterArgumentsOfFunction.h
chqma-revng: added an abort in case we get a `LatticeElement` that is not valid

I kept the default to identity is for unknown transfer functions (for example if we want to add more kinds of transfer functions that are not useful for a particular analysis)

include/revng/ABIAnalyses/Generated/DeadRegisterArgumentsOfFunction.h
chqma-revng: moved out of the generated code

include/revng/ABIAnalyses/Generated/DeadRegisterArgumentsOfFunction.h
chqma-revng: moved out of the generated code

scripts/monotone-framework-lattice.py
chqma-revng: refarctored to use `nextworkx` instead of `pygraphviz`. It uses `pydot` under the hood, which is pure-Python when used for plain dot parsing and not drawing

aleclearmind:
This PR is now superseded by #27.

PR #53, opened 2021-03-10
[WIP] BinaryFile.h: improve error message for unhandled GNU indirect function symbols

pfez:
It is quite common for new users to compile programs with their system compilers and feeding them to revng. Most system compilers link C programs against glibc, which emits GNU indirect function symbols even when compiled with `-static`. Such symbols are not supported by revng yet, so the typical bug report is a crash with the unhelpful message: "Unexpected symbol type". This commits improves the error message to make this frequent troubleshooting easier.

In the error message we could probably also suggest to use our `x86_64-gentoo-linux-musl-gcc` compiler to statically compile binaries, but for now I didn't want to weigh it down to much. @aleclearmind what do you think about it?

### DIFF
--- a/tools/revng-lift/BinaryFile.cpp
+++ b/tools/revng-lift/BinaryFile.cpp
@@ -38,6 +38,43 @@ using LabelList = BinaryFile::LabelList;
 
 static Logger<> EhFrameLog("ehframe");
 static Logger<> LabelsLog("labels");
+static Logger<> SymbolTypeLog("symbol-types");
+
+namespace SymbolType {
+
+static SymbolType::Values fromELF(unsigned char ELFSymbolType) {
+  switch (ELFSymbolType) {
+  case llvm::ELF::STT_GNU_IFUNC:
+    revng_log(SymbolTypeLog,
+              "symbol STT_GNU_IFUNC: GNU indirect functions not supported yet");
+    return SymbolType::Unknown;
+
+  case llvm::ELF::STT_TLS:
+    revng_log(SymbolTypeLog,
+              "symbol STT_TLS: thread-local storage not supported yet");
+    return SymbolType::Unknown;
+
+  case llvm::ELF::STT_NOTYPE:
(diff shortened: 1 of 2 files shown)


aleclearmind:
Mh, I think the right thing to do is just emitting a warning as opposed to abort in case we meet a relocation we're not aware of.
`BinaryFile` is quite strict right now, it has to grow up in a more tolerant adult.

I would avoid suggesting `x86_64-gentoo-linux-musl-gcc`, `revng` should be as little orchestra-aware as possible.

pfez:
> Mh, I think the right thing to do is just emitting a warning as opposed to abort in case we meet a relocation we're not aware of.
> `BinaryFile` is quite strict right now, it has to grow up in a more tolerant adult.

Makes sense, but if you emit a warning without crashing you have to decide what to return. Would `SymbolType::Unknown` be ok?


 
> I would avoid suggesting `x86_64-gentoo-linux-musl-gcc`, `revng` should be as little orchestra-aware as possible.

Totally agree!

pfez:
Also, we need to agree on a policy about emitting warnings and enforce it project-wide...

aleclearmind:
> Makes sense, but if you emit a warning without crashing you have to decide what to return. Would `SymbolType::Unknown` be ok?

Yup!

> Also, we need to agree on a policy about emitting warnings and enforce it project-wide...

We should probably log from the caller of `fromELF` in case it returns Unknown. I'd just use a logger that makes sense in that context.
This is not a particularly serious warning, I'd say.

pfez:
Done!

pfez:
I see that you about to merge in `develop` a new commit handling this in another way. No warnings printed, but always return `SymbolType::Unknown` for unhandled cases. I guess this is not necessary anymore. I'm closing it.

aleclearmind:
Yes, I figured out that was the simplest approach.
Also, having side effects (logging) in an enum-to-enum conversion function didn't seem right.

PR #58, opened 2021-04-14
Skeleton for flatc

pfez:


include/revng/Model/Types.h
```diff
@@ -0,0 +1,335 @@
...
+struct Type {
+  enum TypeKind {
+    TK_Invalid,
+    TK_Primitive,
+    TK_Enum,
+    TK_Typedef,
+    TK_Struct,
+    TK_Union,
+    TK_FunctionPtr,
+  };
```
aleclearmind, line 27: Where's is this enum serialized?
For `enum`s, please adopt the pattern used [elsewhere](https://github.com/revng/revng/blob/develop/include/revng/Model/Binary.h#L546):
* `namespace`
* `enum Values`
* Specialize `llvm::yaml::ScalarEnumerationTraits`
* No prefix

include/revng/Model/Types.h
```diff
@@ -0,0 +1,335 @@
+#pragma once
```
aleclearmind, line 1: What about `TypeSystem.h`? Otherwise `Type.h` is also OK.

include/revng/Model/Types.h
```diff
@@ -0,0 +1,335 @@
...
+#include <algorithm>
+#include <optional>
+
+#include "llvm/ADT/StringRef.h"
+
+#include "revng/ADT/SortedVector.h"
+#include "revng/Support/Assert.h"
+#include "revng/Support/Debug.h"
+
+namespace flatc {
```
aleclearmind, line 16: Is there a real benefit in having a namespace here?
`flatc` is just a part of the model (`model` namespace).
I was thinking about `flatc` more as an informal thing than something widespread in the codebase.

include/revng/Model/Types.h
```diff
@@ -0,0 +1,335 @@
...
+    TK_FunctionPtr,
+  };
+
+  std::string Name; // Key
+
+private:
+  TypeKind Kind;
+
+public:
+  TypeKind getKind() const { return Kind; }
```
aleclearmind, line 35: Please follow the conventions in `Binary.h`, no private fields, no accessors.

include/revng/Model/Types.h
```diff
@@ -0,0 +1,335 @@
...
+
+inline bool isPtrQualifier(const Qualifier &Q) {
+  revng_assert(Q.verify());
+  return Q.Kind == QualifierKind::Const;
+}
+
+using Path = flatc::Type *;
+
+struct QualifiedType {
+  Path ThePath = nullptr;
```
aleclearmind, line 103: What about?
```suggestion
  Path UnqualifiedType = nullptr;
```

include/revng/Model/Types.h
```diff
@@ -0,0 +1,335 @@
...
+  static bool classof(const Type *T) { return T->getKind() == TK_Union; }
+
+  // Kind must be TK_Union, Fields must not be empty, all Fields must verify
+  bool verify() const debug_function;
+};
+
+struct FunctionPtrType : public Type {
+  QualifiedType ReturnType;
+  std::vector<QualifiedType> ArgumentTypes;
+  bool IsVariadic;
```
aleclearmind, line 300: Let's not introduce expressive power we don't have short-term plans to handle.

aleclearmind (changes requested):

include/revng/Model/Types.h
pfez: From a previous in-person discussion I had already did points 1, 2, 4. I just updated the pull request with these changes.

Point 3 is not strictly necessary, but I noticed that you have done so for all the equivalent enums in `Binary.h`. Is there a specific reason why you are doing so and not relying on llvm's yaml stuff to do it for you?

include/revng/Model/Types.h
pfez: The point is, we already have `llvm::Type`, and `clang::Type` in `revng-c`. Given that `revng-c` is going to be one of the heaviest users of this type system, I think we need a way to distinguish `flatc::Type` from others. But I can do with `model::Type` if you like it better.

include/revng/Model/Types.h
pfez: It's already like that in the latest changes I just pushed.

include/revng/Model/Types.h
pfez: So, you want me to drop `IsVariadic` alltogether right now?

I'm not sure we don't need `IsVariadic` already. For instance, last time I checked `newpc` was variadic, and `GeneratedIRReference.rst` still says it is. Isn't that anymore the case? Or are you saying that you never want to describe `newpc` in the model?

And if we never want to describe it in the model, we are also implying that we cannot emit it in C, since `flatc` is meant to eventually become the only way we use to describe function types. Are you ok with this?

include/revng/Model/Types.h
pfez: I'll go for `Type.h`

include/revng/Model/Types.h
pfez: Will do.

include/revng/Model/Types.h
aleclearmind: 3) what do you mean? There's no automagic YAML salsa for enums, you have to explicitly specialize those.

include/revng/Model/Types.h
aleclearmind: Yeah, `flatc` is just `model` after all. Go for `model::Type`.

include/revng/Model/Types.h
pfez: Well, no, my bad. I've already done point 3 as well. That's why it worked.

All you asked in this comment is done. I'm resolving this conversation.

include/revng/Model/Types.h
aleclearmind: `newpc` is not a lifted function so it doesn't need to be represented in the model.

I'm just saying, we have really foggy plans for variadic functions. Let's just postpone this, otherwise we'll have a field no one is using. Just noise.

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+  uint64_t UUID;
+
+  /// \brief The name of this type. This is the unique key to identifiy a type.
+  std::string Name;
+
+  /// \brief The kind of a flatc::Type
+  TypeKind::Values Kind;
+
+public:
+  TypeKind::Values getKind() const { return Kind; }
```
aleclearmind, line 101: Drop

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+
+  Type(llvm::StringRef NameRef) : Type(NameRef, TypeKind::Values::Invalid) {}
+  Type() : Type("") {}
+  ~Type() = default;
+
+  Type(const Type &) = default;
+  Type &operator=(const Type &) = default;
+
+  Type(Type &&) = default;
+  Type &operator=(Type &&) = default;
```
aleclearmind, line 122: Why this again?

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+      return not Size;
+    case QualifierKind::Array:
+      return Size;
+    }
+    return false;
+  }
+
+  static Qualifier createConst() { return { QualifierKind::Const, 0 }; }
+
+  static Qualifier createPtr() { return { QualifierKind::Pointer, 0 }; }
```
aleclearmind, line 183:
```suggestion
  static Qualifier createPointer() { return { QualifierKind::Pointer, 0 }; }
```

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+
+inline bool isArrayQualifier(const Qualifier &Q) {
+  revng_assert(Q.verify());
+  return Q.Kind == QualifierKind::Const;
+}
+
+inline bool isPtrQualifier(const Qualifier &Q) {
+  revng_assert(Q.verify());
+  return Q.Kind == QualifierKind::Const;
+}
```
aleclearmind, line 208: Why not methods?

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+
+using TypesVector = SortedVector<UpcastablePointer<flatc::Type>>;
+
+} // end namespace flatc
+
+struct flatc::TypesTree {
```
aleclearmind, line 214: It is a tree?

```suggestion
struct flatc::TypesList {
```

Couldn't this just be field in `model::Binary`?

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+  TypesVector Types;
+  bool operator==(const TypesTree &) const = default;
+  bool verify() const {
```
aleclearmind, line 217: * I think we're keeping verify methods in `.cpp` files.
* `debug_function`

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+
+RecursiveCoroutine<std::size_t> size(const flatc::Type *T);
+
+RecursiveCoroutine<std::size_t> size(const flatc::QualifiedType &T);
+
+} // end namespace flatc
+
+/// \brief A primitive type in flatc: sized integers, booleans, floats and void.
+struct flatc::PrimitiveType : public flatc::Type {
+  static constexpr const char *Tag = "!flatc::PrimitiveType";
```
aleclearmind, line 393:
```suggestion
  static constexpr const char *Tag = "!PrimitiveType";
```

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+template<>
+struct KeyedObjectTraits<std::string>
+  : public IdentityKeyedObjectTraits<std::string> {};
+
+/// \brief An entry in a flatc enum, with a name and a value.
+struct flatc::EnumEntry {
+
+  std::size_t Value;
+  std::string Name;
+  SortedVector<std::string> Aliases;
```
aleclearmind, line 430: What's your reasoning for not just having a vector of `Names` instead of `Name` + `Aliases`?

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+
+/// \brief KeyedObjectTraits for std::string based on its value
+template<>
+struct KeyedObjectTraits<std::string>
+  : public IdentityKeyedObjectTraits<std::string> {};
+
+/// \brief An entry in a flatc enum, with a name and a value.
+struct flatc::EnumEntry {
+
+  std::size_t Value;
```
aleclearmind, line 428:
```suggestion
  uint64_t Value;
```

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+
+  static flatc::EnumEntry fromKey(std::size_t V) {
+    return flatc::EnumEntry{ V };
+  }
+};
+
+/// \brief An enum type in flatc. Enums are actually typedefs of unnamed
+/// enums.
+struct flatc::EnumType : public flatc::Type {
+  static constexpr const char *Tag = "!flatc::EnumType";
```
aleclearmind, line 474:
```suggestion
  static constexpr const char *Tag = "!Enum";
```

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+
+  static flatc::StructField fromKey(const std::size_t &Offset) {
+    return flatc::StructField(Offset);
+  }
+};
+
+/// \brief A struct type in flatc. Structs are actually typedefs of unnamed
+/// structs in C.
+struct flatc::StructType : public flatc::Type {
+  static constexpr const char *Tag = "!flatc::StructType";
```
aleclearmind, line 624:
```suggestion
  static constexpr const char *Tag = "!Struct";
```

Please use the name of the associated `enum` entry.

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+  //   - this suggests that we want the key for the fields to be an integer
+  //   representing their position in the struct
+  //   - but if we remove a field from the struct after its creation, we end
+  //   up with a missing index, which is a thing that we'd like to avoid
+  //   - this kind of iteration is important for actually computing pointer
+  //   arithmetic and member access expressions when generating C code, and
+  //   it's likely the most common operation
+
+  SortedVector<StructField> Fields;
+  std::size_t Size;
```
aleclearmind, line 655: Let's just use `uint64_t` in place of `size_t` everywhere.

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+  StructType &operator=(const StructType &) = default;
+
+  StructType(StructType &&) = default;
+  StructType &operator=(StructType &&) = default;
+
+  static bool classof(const Type *T) {
+    return T->getKind() == TypeKind::Struct;
+  }
+
+  RecursiveCoroutine<std::size_t> size() const;
```
aleclearmind, line 673: Can we push all the `RecursiveCoroutine` into the `.cpp` file and drop `#include "revng/ADT/RecursiveCoroutine.h"`?

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+  bool verifyDerived() const debug_function;
+
+  friend struct flatc::Type;
+};
+INTROSPECTION_NS(flatc, StructType, UUID, Name, Kind, Fields, Size);
+
+/// \brief A field of a union type in flatc, with position, qualified type, and
+/// name
+struct flatc::UnionField : flatc::AggregateField {
+  std::size_t Position;
```
aleclearmind, line 687: Why is `Position` the key of `union`?

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+
+  ArgumentType() : ArgumentType(0) {}
+
+  ~ArgumentType() = default;
+
+  ArgumentType(const ArgumentType &) = default;
+  ArgumentType &operator=(const ArgumentType &) = default;
+
+  ArgumentType(ArgumentType &&) = default;
+  ArgumentType &operator=(ArgumentType &&) = default;
```
aleclearmind, line 773: Is all of this really necessary?
Ideally in the model we should have more or less plain structs.

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+  static std::size_t key(const flatc::ArgumentType &Val) { return Val.Pos; }
+
+  static flatc::ArgumentType fromKey(const std::size_t &Pos) {
+    return flatc::ArgumentType(Pos);
+  }
+};
+
+/// \brief A function pointer type in flatc. Function pointers are actually
+/// typedefs.
+struct flatc::FunctionPtrType : public flatc::Type {
```
aleclearmind, line 794:
```suggestion
struct flatc::FunctionPointerType : public flatc::Type {
```

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+template<>
+struct llvm::yaml::ScalarEnumerationTraits<flatc::TypeKind::Values> {
+  static void enumeration(IO &io, flatc::TypeKind::Values &Val) {
+    io.enumCase(Val, "TypeKind::Invalid", flatc::TypeKind::Invalid);
+    io.enumCase(Val, "TypeKind::Primitive", flatc::TypeKind::Primitive);
+    io.enumCase(Val, "TypeKind::Enum", flatc::TypeKind::Enum);
+    io.enumCase(Val, "TypeKind::Typedef", flatc::TypeKind::Typedef);
+    io.enumCase(Val, "TypeKind::Struct", flatc::TypeKind::Struct);
+    io.enumCase(Val, "TypeKind::Union", flatc::TypeKind::Union);
+    io.enumCase(Val, "TypeKind::FunctionPtr", flatc::TypeKind::FunctionPtr);
```
aleclearmind, line 849:
```suggestion
    io.enumCase(Val, "Invalid", flatc::TypeKind::Invalid);
    io.enumCase(Val, "Primitive", flatc::TypeKind::Primitive);
    io.enumCase(Val, "Enum", flatc::TypeKind::Enum);
    io.enumCase(Val, "Typedef", flatc::TypeKind::Typedef);
    io.enumCase(Val, "Struct", flatc::TypeKind::Struct);
    io.enumCase(Val, "Union", flatc::TypeKind::Union);
    io.enumCase(Val, "FunctionPtr", flatc::TypeKind::FunctionPtr);
```

Also, can you push this close to the definition of the `enum`?

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+
+// Make flatc::QualifierKind::Values yaml-serializable, required for making
+// flatc::Qualifier yaml-serializable as well
+template<>
+struct llvm::yaml::ScalarEnumerationTraits<flatc::QualifierKind::Values> {
+  static void enumeration(IO &io, flatc::QualifierKind::Values &Val) {
+    io.enumCase(Val, "QualifierKind::Invalid", flatc::QualifierKind::Invalid);
+    io.enumCase(Val, "QualifierKind::Pointer", flatc::QualifierKind::Pointer);
+    io.enumCase(Val, "QualifierKind::Array", flatc::QualifierKind::Array);
+    io.enumCase(Val, "QualifierKind::Const", flatc::QualifierKind::Const);
```
aleclearmind, line 861: Ibidem

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+struct llvm::yaml::MappingTraits<T> : public TupleLikeMappingTraits<T> {};
+
+// Make UpcastablePointer<flatc::Type> yaml-serializable polymorphically
+template<>
+struct llvm::yaml::MappingTraits<UpcastablePointer<flatc::Type>>
+  : public PolymorphicMappingTraits<UpcastablePointer<flatc::Type>> {};
+
+template<>
+struct llvm::yaml::MappingTraits<flatc::TypesTree>
+  : public TupleLikeMappingTraits<flatc::TypesTree> {};
```
aleclearmind, line 903: Keep support data structures such as YAML close the definition of the associated struct.

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+
+namespace flatc {
+
+inline bool isQualified(const QualifiedType &QT) {
+  // FIXME: should we unwrap ThePath and look for nested qualifiers?
+  revng_assert(QT.verify());
+  return not QT.Qualifiers.empty();
+}
+
+inline constexpr std::array<const char *, 93> CReservedKeywords = {
```
aleclearmind, line 257: Let's push all the verification logic in the `.cpp` file.

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+    if (KindCmp != 0)
+      return KindCmp;
+
+    int NameCmp = llvm::StringRef(Name).compare(llvm::StringRef(Other.Name));
+    if (NameCmp < 0)
+      return std::strong_ordering::less;
+    if (NameCmp > 0)
+      return std::strong_ordering::greater;
+    return std::strong_ordering::equal;
+  }
```
aleclearmind, line 146: What's the purpose of this? It's not used anywhere.

aleclearmind (changes requested):

include/revng/Model/Types.h
pfez: Done.

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+    return flatc::StructField(Offset);
+  }
+};
+
+/// \brief A struct type in flatc. Structs are actually typedefs of unnamed
+/// structs in C.
+struct flatc::StructType : public flatc::Type {
+  static constexpr const char *Tag = "!flatc::StructType";
+
+  // StructType is kinda weird, because we have a set of conflicting
```
aleclearmind: This comment has weird indentation

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+// Make flatc::Type derived types usable with UpcastablePointer
+template<>
+struct concrete_types_traits<flatc::Type> {
+  using type = std::tuple<flatc::PrimitiveType,
+                          flatc::EnumType,
+                          flatc::TypedefType,
+                          flatc::StructType,
+                          flatc::UnionType,
+                          flatc::FunctionPtrType,
+                          flatc::Type>;
```
aleclearmind: `flatc::Type` is "abstract".

include/revng/Model/Types.h
pfez: Well, this is not strictly necessary right now. All the hierarchy of the types is defined in a way that the compiler will always be able to automatically default copy/move constructor/assignment and the destructor as well.

I explictly defaulted them mainly as a safeguard during development, when this classes were changing heavily and often. In this way whenever I managed to screw them up to a point where the compiler could not automatically default those, I would get a compilation error.

In the current state, I can drop all the explicitly defaulted destructors, and copy/move constructor/assignments and no functionality will change. We would just lose this safeguard in case of future development. In that case, at a certain point the compiler could stop being able to default those, but we would not get a compilation error. That's basically it.

Should I drop them?

include/revng/Model/Types.h
pfez: Will do.

include/revng/Model/Types.h
pfez: Leftover. Will remove.

include/revng/Model/Types.h
pfez: Done

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+/// \brief Concept to identify all types that are derived from flatc::Type
+template<typename T>
+concept FlatCType = std::is_base_of_v<flatc::Type, T>;
+
+/// \brief Base class of flatc types used for LLVM-style RTTI
+struct flatc::Type {
+  static constexpr const char *Tag = "!flatc::InvalidType";
+
+  // TODO: we should switch to a sane UUID implementation
+  uint64_t UUID;
```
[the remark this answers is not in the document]
pfez: Done

include/revng/Model/Types.h
pfez: Done

include/revng/Model/Types.h
pfez: Done

include/revng/Model/Types.h
pfez: Done

include/revng/Model/Types.h
aleclearmind: Let's drop everything can be dropped.
I favor simplicity and coherence with the status quo right now, if there's a benefit, we can introduce that in a distinct PR.

include/revng/Model/Types.h
pfez: Done

include/revng/Model/Types.h
pfez: Done

include/revng/Model/Types.h
pfez: Done

include/revng/Model/Types.h
pfez: Done

include/revng/Model/Types.h
pfez: Done

include/revng/Model/Types.h
```diff
@@ -0,0 +1,903 @@
...
+                          flatc::Type>;
+};
+
+/// \brief Concept to identify all types that are derived from flatc::Type
+template<typename T>
+concept FlatCType = std::is_base_of_v<flatc::Type, T>;
+
+/// \brief Base class of flatc types used for LLVM-style RTTI
+struct flatc::Type {
+  static constexpr const char *Tag = "!flatc::InvalidType";
```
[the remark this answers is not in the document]
pfez: Well, it's not really "abstract". You can instantiate one. And I expect a model containing an invalid type to not `verify()`. But I need to have a tag for the `InvalidType` to be able to check for it. If we could ensure at compile-time that we never generate invalid types I would agree to drop it, but in the current form I'm afraid we can't.

include/revng/Model/Types.h
pfez: See the conversation above.

include/revng/Model/Types.h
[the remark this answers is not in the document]
aleclearmind: Make the constructor `protected`?
In any case we don't want to serialize/deserialize this.

include/revng/Model/Types.h
pfez: Done

include/revng/Model/Types.h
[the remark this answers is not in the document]
pfez: Can't make the constructor `protected`, because yaml serialization/deserialization does not work.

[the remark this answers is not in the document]
pfez: "Does not work" as-in "fails to compile".

include/revng/Model/Types.h
pfez: It will become a field in `model::Binary`. I left it separated because during development I was testing it in isolation. This will soon go away.

include/revng/Model/Types.h
pfez: Done

include/revng/Model/Type.h
```diff
@@ -0,0 +1,639 @@
...
+
+  static bool classof(const Type *T) { return T->Kind == TypeKind::Enum; }
+
+private:
+  // Kind must be TypeKind::Enum; Underlying type must verify and be an
+  // integral primitive type; there must be at least one entry, and all the
+  // entries must verify.
+  bool verifyDerived() const debug_function;
+
+  friend struct model::Type;
```
aleclearmind, line 355: Why?

include/revng/Model/Types.h
pfez: Done

include/revng/Model/Types.h
pfez: Done

include/revng/Model/Types.h
pfez: Done

include/revng/Model/Types.h
pfez: Dropped

include/revng/Model/Types.h
pfez: Done

include/revng/Model/Type.h
pfez: So that `Type::verify()` can do the same baseline checks for all types, then call `llvm::cast<Derived>->verifyDerived()` for stuff that depends on the single field. 
For now I switched to `friend bool model::Type::verify() cons`, so that only verification is friend.

Any better idea?

include/revng/Model/Types.h
pfez: Done

include/revng/Model/Types.h
pfez: Done

include/revng/Model/Type.h
aleclearmind: Idea:
* `Type::verify()` verifies itself only
* `Derived::verify()` calls `Type::verify()` and verifies `Derived`-specific stuff.
* If you have `Upcastable<Type> A` you do `upcast(A, [] (auto &Upcasted) { Upcasted.verify() })`.

Upcastable pointers are designed to be used with templated lambdas magic.

include/revng/Model/Types.h
pfez: Mainly because we want to be able to rename fields, so the name cannot be the key. It could be some other kind of ID, but in this way it's related to the order of printing. Fields with lower position are printed first. Which makes easier to write a valid union in yaml by hand.

Thinking about this, the position is not necessarily the best solution anyway. If we remove, say, the third field of a union, what are we going to do? Rescale others is something we don't want to do, cause we should change their keys and it would mess up the diff.

So at this point I'm thinking we should relax the `verify()` on `UnionType` so that it allows fields with non-consecutive positions. In this way we don't need to rescale on removal, and we don't break diffs.

Otherwise we could replace the `Position` with some other kind of ID, like a UUID, but I'm not particularly fond of it.

I have no strong feelings here. The point is all the solutions we can devise will be somehow counter-intuitive, because on one hand we'd like to have the name as a key, but that's a no-go if we want to enable renaming, which is something we'll need to do for sure. So it's a matter of choosing a counter-intuitive solution that is less wrong than the others.

include/revng/Model/Types.h
pfez: My main concern was about invalidation. If you decide that the first element of `Alisases` is "the name", you could invalidate it by adding a new element to `Aliases` that is sorted before the old one.

Not sure this is actually a problem. But having `Name` as a separate field you have consistency, which is good for debugging, and in general for whatever is user-facing (like the code generated by the decompiler). The name of an entry cannot change unless you change it explicitly.

But I can just go with `Names` if you like it more.

include/revng/Model/Types.h
aleclearmind: I think that we should deal with the fact that, wherever we want to put a name as key, we have to use a UUID.

include/revng/Model/Types.h
aleclearmind: I've no strong opinions. It's just a bit more cumbersome to use downstream. For instance, in the UI, we'll have to implement the UI to change both Name and Aliases.

On the other hand, if we don't do it like this, the user cannot specify a "favorite" `Name`, since a `SortedVector` is essentialy a `set` (it's ordered by the string content).

I'll leave it to you.

include/revng/Model/Types.h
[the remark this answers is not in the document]
pfez: Ok, I took a closer look. I think there's no workaround for now.
I cannot make `Type` entirely abstract, because `KeyedObjectTraits<UpcastablePointer<Type>>` needs to instantiate a `Type` in `fromKey()` starting from the UUID, but without knowing the derived type. So it has to create a `Type` of the base class. This is just temporary, since it's used for lookup and then discarded, but still the compiler needs to see `Type` in `concrete_type` otherwise it will fail to unroll the template and fail to compile.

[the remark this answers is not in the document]
pfez: For this reason I'm thinking of leaving the `Tag` where it is, leave `Type` in `concrete_types`. And just do the checks at `verify()`-time.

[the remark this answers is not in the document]
aleclearmind: Mmmh, that's because the kind is not part of the key.
Seems like a push from an invisible force to put the kind into the key, which is weird since the ID is supposed to be already unique, but it doesn't sound too bad to me.
The less things we have to check in `verify`, the better.

What do you think?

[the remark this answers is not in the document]
pfez: It does not sound too bad to me either. I'll give it a try.

include/revng/Model/Types.h
pfez: I see you got the point. I thought that leaving to the user the option to set a "favorite" `Name` was good. It's likely a little bit more cumbersome to implement on the UI side, but my feeling is that it has less-unintuitive meaning for the end user. In 95% of the cases I expect the user to want to: "Just rename this entry and be done with it". So the aliases are there mostly for compatibility with parsing C headers. I'll probably leave it like this until a user emerges (and we can start using it) or someone complains (and we can drop it).

include/revng/Model/Type.h
pfez: I don't like the idea. Pretty sure that someone (probably me) is going to forget to `upcast` in a place where calling only `Type::verify()` will give a fake sense of safety, while getting away with a broken derived thing.

I could change `verify` to be `virtual`, but that would be the only reason to have a vtable here.

But I really want to be impossible to use `verify` in the wrong way.

So here's another idea:
- Rename `Type::verify()` to `Type::verifyBase()` and make it `protected`. In this way, `Type` does not have `verify` and it cannot be called wrongly.
- `Derived::verify()` calls `Type::verifyBase()` and verifies `Derived` stuff
- In this way, users cannot do `MyBaseType.verify()` not `MyBaseTypePtr->verify()` because `verify` is not there on the base type. The only thing they can do is to `upcast(A, [](auto &Upcasted){ Upcasted.verify() }), for which we could provide a helper `bool verify(const Upcastable<Type> &U)` that does the upcast internally.

aleclearmind: Works, but at this point you could keep the `Type::verify` as a name and just make it protected.

pfez: Works. I'll do this.

pfez: None of the ideas suggested above works, because they all require you to hold an `UpcastablePointer` in order to `verify()` a `Type`. In particular, in our case this not always true, e.g. when you are traversing a `TupleTreeReference` from one `Type` to another, calling `get()` on the `TupleTreeReference` gives you a `Type *`, not an `UpcastablePointer<Type>`. We could work around this, making `TupleTreeReference` return you an `UpcastablePointer`, not just the naked pointer it wraps. But I'm not sure I like this idea more than the original implementation with `verifyBase()`, `verifyDerived()` and the `friend bool model::Type::verify()` declaration in derived classes.

What do you think?

include/revng/Model/Types.h
pfez: Ok, will do.

include/revng/Model/Types.h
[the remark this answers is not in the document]
pfez: Ok, it took me a while, but it's fixed now. `Type` cannot be instantiated on its own now. Only when creating an instance of a derived type. Let me know what you think.

include/revng/Model/Type.h
pfez: No, ok, I managed to do it. Take a look.

include/revng/Model/Type.h
```diff
@@ -0,0 +1,726 @@
...
+template<typename T>
+concept FlatCType = std::is_base_of_v<model::Type, T>;
+
+/// \brief Make all types derived from model::Type yaml-serializable like tuples
+template<FlatCType T>
+struct llvm::yaml::MappingTraits<T> : public TupleLikeMappingTraits<T> {};
+
+/// \brief Base class of model types used for LLVM-style RTTI
+struct model::Type {
+  // static constexpr const char *Tag = "!Invalid";
```
aleclearmind, line 151: Drop?

include/revng/Model/Type.h
```diff
@@ -0,0 +1,726 @@
...
+
+namespace model {
+
+using UpcastableType = UpcastablePointer<model::Type>;
+
+} // end namespace model
+
+template<>
+struct llvm::yaml::ScalarTraits<model::Type::Key>
+  : CompositeScalar<model::Type::Key, '#'> {};
```
aleclearmind, line 202: Elsewhere we use `-` as a separator. Reasons to deviate?

include/revng/Model/Type.h
```diff
@@ -0,0 +1,726 @@
...
+struct KeyedObjectTraits<model::UpcastableType> {
+  static model::Type::Key key(const model::UpcastableType &Val) {
+    return Val->key();
+  }
+  static model::UpcastableType fromKey(const model::Type::Key &K);
+
+private:
+  template<size_t I = 0>
+  static model::UpcastableType
+  makeFromID(model::TypeKind::Values Kind, uint64_t ID);
```
aleclearmind, line 215: Uhm, if you want to make this part of the trait make it generic with argument pack and squillo di lusso, otherwise I'd go for a regular function.

include/revng/Model/Type.h
aleclearmind: Looks amazing!

include/revng/Model/Type.h
pfez: And regular function it is.

include/revng/Model/Type.h
pfez: Dropped.

include/revng/Model/Type.h
pfez: Yes but nothing major. The idea is that if we ever switch to GUID in future we might end up with the standard-conforming GUID format that contains `-`. So I went for `#`. But if you prefer consistency I'll switch to `-`. In the end, if we switch to real GUID, and we have `-` in the middle, the tests should fail and we can easily replace `-` with `#` or something else at that point. Just let me know what you think and I'll do it.

aleclearmind: Ah, I see. For now let's stick with `-`. Switching to GUID will be a breaking change anyway.

include/revng/Model/Types.h
pfez: I ended up removing the comment entirely. It looked necessary when I was writing down this data structure the first time, but now that we have stabilized it, the comment just adds lots of unnecessary information IMHO.

include/revng/Model/Type.h
pfez: Done.

include/revng/Model/Types.h
pfez: Done

include/revng/Model/Type.h
```diff
@@ -0,0 +1,699 @@
...
+  } else {
+    return TypePath::fromString("/Types/");
+  }
+}
+
+} // end namespace model
+
+/// \brief A qualified version of a model::Type. Can have many nested qualifiers
+struct model::QualifiedType {
+  TypePath UnqualifiedType = model::makeTypePath(nullptr);
```
aleclearmind, line 279: Can we make `TypePath` default constructible and 
```suggestion
  TypePath UnqualifiedType;
```

lib/Model/Type.cpp
```diff
@@ -0,0 +1,769 @@
...
+
+#include "revng/Model/Binary.h"
+#include "revng/Model/Type.h"
+
+using llvm::cast;
+using llvm::dyn_cast_or_null;
+
+namespace model {
+
+static std::set<std::string> CReservedKeywords = {
```
aleclearmind, line 21:
```suggestion
static std::set<llvm::StringRef> CReservedKeywords = {
```

lib/Model/Type.cpp
```diff
@@ -0,0 +1,769 @@
...
+makePrimitiveName(PrimitiveTypeKind::Values PrimKind, uint8_t BS) {
+  using namespace std::string_literals;
+  revng_assert(isValidPrimitiveSize(PrimKind, BS));
+
+  switch (PrimKind) {
+  case PrimitiveTypeKind::Void:
+    return "void";
+
+  case PrimitiveTypeKind::Bool:
+    return "bool"s + std::to_string(BS * 8) + "_t"s;
```
aleclearmind, line 193:
```suggestion
    return (Twine("bool") + Twine(BS * 8) + Twine("_t")).str();
```

lib/Model/Type.cpp
```diff
@@ -0,0 +1,769 @@
...
+    case QualifierKind::Pointer:
+      // If we find a pointer, we're done
+      rc_return PointerSize;
+
+    case QualifierKind::Array: {
+      // The size is equal to (number of elements of the array) * (size of a
+      // single element).
+      const model::QualifiedType ArrayElem{ QT.UnqualifiedType,
+                                            { std::next(QIt), QEnd } };
+      rc_return rc_recur model::getOrComputeSize(ArrayElem, SizeCache) * Q.Size;
```
aleclearmind, line 323: 😮

include/revng/Model/Binary.h
```diff
@@ -912,11 +913,19 @@ static_assert(IsKeyedObjectContainer<MutableSet<model::Function>>);
 class model::Binary {
 public:
   MutableSet<model::Function> Functions;
+  SortedVector<UpcastablePointer<model::Type>> Types;
```
aleclearmind, line 916: What about
```suggestion
  SortedVector<UpcastablePointer<model::Type>> UnqualifiedTypes;
```
?

include/revng/Model/Type.h
```diff
@@ -0,0 +1,699 @@
...
+  UnionField() : UnionField({}, "") {}
+
+private:
+  UnionField(uint64_t TheID, QualifiedType &&QT, llvm::StringRef NameRef) :
+    AggregateField(std::move(QT), NameRef), ID(TheID) {}
+
+public:
+  UnionField(uint64_t TheID) : UnionField(TheID, {}, "") {}
+
+  bool operator==(const UnionField &Other) const { return ID == Other.ID; }
```
aleclearmind, line 529: Why?

pfez: Leftover, it should be `= default` like for `StructType`.

lib/Model/Type.cpp
```diff
@@ -0,0 +1,769 @@
...
+
+      // Cannot have const-qualified void as argument.
+      if (VoidConst.IsConst) {
+        ArgsAreVerified = false;
+        break;
+      }
+    }
+  }
+
+  rc_return ArgsAreVerified;
```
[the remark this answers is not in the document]
pfez: No reason. Could be done that way. For a while I was looking into low level codegen for coroutine and wanted to see if having many or few `rc_return` (which boils down to `co_return`) did change codegen. And then when I stopped looking I just left it like this.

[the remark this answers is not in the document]
pfez: To be clear, I did not see any meaningful change one way or the other.

aleclearmind:
This has been merged in 92c7ed0. Thanks a lot!

PR #61, opened 2021-04-27
[WIP] RecursiveCoroutine: enable direct continuation

pfez:
Before this commit, the execution logic of RecursiveCoroutine used an underlying std::stack allocated on the heap to hold all the coroutine_handles. It also manually managed passing return values from callees to callers.

This commit drops this unnecessary auxiliary stack.
When a RecursiveCoroutine co_awaits another one, the handle of the awaiter is injected into the awaitee, so that when the awaitee is done it can directly execute the remaining part of the awaiter as a continuation.

With these changes, the inliner should by much happier and we should get even better performance, but I haven't measured it yet. I'll ask alvise to try this out on his use case, which at the moment uses these heavily, so we should be able to se a difference.

### DIFF
--- a/include/revng/ADT/RecursiveCoroutine-coroutine.h
+++ b/include/revng/ADT/RecursiveCoroutine-coroutine.h
@@ -6,155 +6,49 @@
 
 #include <experimental/coroutine>
 #include <optional>
+#include <type_traits>
 #include <utility>
 
 #include "revng/Support/Assert.h"
 
-struct PromiseBase {
+template<typename>
+struct RecursiveCoroutine;
 
-  auto initial_suspend() const { return std::experimental::suspend_always(); }
-  auto final_suspend() const noexcept {
-    return std::experimental::suspend_always();
+namespace detail {
+
+template<typename RetT>
+struct ReturnBase {
+
+  void return_value(RetT R) {
+    revng_assert(not CurrValue.has_value());
+    CurrValue = std::move(R);
+    return;
   }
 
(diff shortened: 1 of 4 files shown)


include/revng/ADT/RecursiveCoroutine.h
```diff
@@ -13,3 +15,12 @@
 #include "RecursiveCoroutine-coroutine.h"
 
 #endif
+
+// This could be dropped right now, as RecursiveCoroutines can be called just
+// like regular functions. However, I'll leave it because rc_run is a nice
+// marker to see in code to realize that a call is actually a call to a
+// RecursiveCoroutine.
```
aleclearmind, line 22: `:OOOOOOOOOOO`

I'm honestly in favor of dropping it at this point. One macro less.

include/revng/ADT/RecursiveCoroutine-coroutine.h
```diff
@@ -6,187 +6,255 @@
...
+    // reaches the final suspension point.
+    //
+    // So our final awaitable object should:
+    // - always suspend (await_ready should always return false)
+    // - return the handle to the awaiter continuation, so that if it's a
+    // valid
+    //   handle it will be resumed.
+    // - never be resumed, because it's the final suspension point
+    // (await_resume
+    //   actuall aborts)
```
aleclearmind, line 84: Weird indentation

pfez:
No reason, it just was low priority.

aleclearmind:
Merged in b58693b.
Thanks!

PR #62, opened 2021-04-28
Latest fixes for inlining helpers

pfez:
With these fixes the only issues left in revng-c are the `ExtractValueInst` patterns we found out today, that I'm going to handle soon.

In the meantime, feel free to fixup these commits wherever you think they fit better.

As soon as I'm done with the `ExtractValueInst` thingy we should be good and we should be able finally merge the whole inline-helpers effort.

### DIFF
--- a/include/revng/ADT/GenericGraph.h
+++ b/include/revng/ADT/GenericGraph.h
@@ -427,14 +427,31 @@ private:
   NodesContainer Nodes;
 };
 
+template<typename T>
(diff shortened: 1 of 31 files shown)


pfez:
I'm done with revng-c. The `ExtractValueInst` issue has been taken care of. You can go ahead with this.

aleclearmind:
Merged in with the `feature/extend-model` branch (bca502a).
Thanks!

PR #68, opened 2021-05-07
Improve interactions between `TupleTreeReference`, `UpcastablePointer` and `KeyedObjectTraits`

pfez:
This MR does works on improving interactions between `TupleTreeReference`, `UpcastablePointer`, and `KeyedObjectTraits`. While doing this, it also tries to slightly improve these classes when this is possible without major refactoring.

At the moment, this MR is composed of y commits. Each does what follows (from older to newer):
- Drops template parameter from `KeyedObjectTraits`. This was necessary before we used concepts to play with `enable_if`. Now we can drop it entirely.
- Uses `std::is_void_v` instead of `std::is_same_v(void, T)` in `UpcastablePointer`.
- Fixes a problem causing bad `TupleTree` traversal with `TupleTreeReference`. Specifically, `CallByPathVisitorWithInstance` did not know how to `visitContainerElement` when the element of the container was an `UpcastablePointer`. This caused the visit to always return `nullptr` even when the `TupleTreeReference` was well-formed. This has been fixed by adding an overload with proper concept constraint, that teaches `CallByPathVisitorWithInstance` how to properly traverse `UpcastablePointers`.
- Requires `IsKeyedObjectContainer` in `TupleTree.h` wherever possible, instead of `IsContainer`.
- 2 commits improving `operator==` for `TupleTreeReference` and `TupleTreePath`
- Fix serialization of `TupleTree`. The changed overload was not used anywhere, and actually failed to compile if someone tried to use it. Now it's fixed, and it enables serializing to a `std::string` buffer, instead of requiring a stream.
- Fix copy/move constructor/assignment in `TupleTreePath.h`. Before this commit, copy/move assignments did not check for `this != &Other` and copy/move constructors were not defined starting from the associated assignments. Both these things are best practices for maintainability, that are fixed by this commit.
- Define `operator<=>` for `TupleTreeReference`
- Add concepts to `YAMLTraits.h`. This considerably helps when adding new yaml-serializable things, because compilation error messages become much more readable
- Cleanup `UpcastablePointer` copy/move constructor and assignments operators. They were a little bit inconsistent before, now they are conforming with typical best practices.
- Move `CompositeScalar` to `YAMLTraits.h`, so that it can be used outside `Binary.h`. We'll need it for the type system.
- Move `IsYamlizable` generic lambda to `TupleTree.h`, so that it can be used outside `Binary.h`. We'll need it for the type system as well
- Add `TupleTreeCompatible` constraint to `TupleTree` type template parameter. It does not make sense to say `TupleTree<T>` is `T` is not `TupleTreeCompatible`.

### DIFF
--- a/include/revng/Model/TupleTree.h
+++ b/include/revng/Model/TupleTree.h
@@ -22,10 +22,13 @@
 
 // clang-format off
 template<typename T>
-concept NotTupleTreeCompatible = (not IsContainer<T>
-                                  and not HasTupleSize<T>
-                                  and not IsUpcastablePointer<T>);
+concept TupleTreeCompatible = IsKeyedObjectContainer<T>
+                                or HasTupleSize<T>
+                                or IsUpcastablePointer<T>;
 // clang-format on
+
+template<typename T>
+concept NotTupleTreeCompatible = not TupleTreeCompatible<T>;
 
 // clang-format off
 template<typename T>
@@ -147,7 +150,7 @@ void visitTupleTree(Visitor &V, T &Obj) {
 }
 
 // Container-like
-template<typename Visitor, IsContainer T>
+template<typename Visitor, IsKeyedObjectContainer T>
 void visitTupleTree(Visitor &V, T &Obj) {
   V.preVisit(Obj);
   using value_type = typename T::value_type;
@@ -235,7 +238,7 @@ ResultT getByKey(RootT &M, KeyT Key) {
   return tupletree::detail::getByKeyTuple<ResultT>(M, Key);
 }
 
-template<typename ResultT, IsContainer RootT, typename KeyT>
+template<typename ResultT, IsKeyedObjectContainer RootT, typename KeyT>
 ResultT *getByKey(RootT &M, KeyT Key) {
   for (auto &Element : M) {
     using KOT = KeyedObjectTraits<std::remove_reference_t<decltype(Element)>>;
@@ -262,7 +265,7 @@ bool callOnPathSteps(Visitor &V, llvm::ArrayRef<TupleTreeKeyWrapper> Path) {
   return callOnPathStepsTuple<element_type>(V, Path);
 }
 
-template<IsContainer RootT, typename Visitor>
+template<IsKeyedObjectContainer RootT, typename Visitor>
 bool callOnPathSteps(Visitor &V, llvm::ArrayRef<TupleTreeKeyWrapper> Path) {
   using value_type = typename RootT::value_type;
   using KOT = KeyedObjectTraits<value_type>;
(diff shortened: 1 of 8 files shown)


include/revng/Model/TupleTree.h
```diff
@@ -448,6 +457,15 @@ struct CallByPathVisitorWithInstance {
   }
 
   template<typename T, typename K, typename KeyT>
+  requires IsUpcastablePointer<K>
```
aleclearmind, line 460:
```suggestion
  template<typename T, IsUpcastablePointer K, typename KeyT>
```

include/revng/Model/TupleTree.h
```diff
@@ -997,6 +1013,16 @@ class TupleTreeReference {
...
 
+  std::weak_ordering operator<=>(const TupleTreeReference &Other) const {
+    // The paths are the same even if they are referred to different roots
+    return Path <=> Other.Path;
+  }
+
+  bool operator==(const TupleTreeReference &Other) const {
+    // The paths are the same even if they are referred to different roots
+    return Path == Other.Path;
+  }
```
aleclearmind, line 1024: Isn't this superfluous given `operator<=>`?

include/revng/Model/TupleTree.h
pfez: Done. I also added another `concept NotIsUpcastablePointer` to use for consistency below, where I previously had `requires (not IsUpcastablePointer<K>)`.

include/revng/Model/TupleTree.h
pfez: Nice catch. The point here is that I started adding `operator<=>` and `operator==` during development because I needed them to get the first implementation through. So, when I added `operator<=>`, `operator==` was already there, and in the rush I did not remove it. Also, notice that `operator<=>` returns `weak_ordering` here, which is why removing `operator==` would not be possible, because AFAIU `operator==` cannot be synthesized starting from `operator<=>` when the latter return `weak_ordering`. However, because of how the two operators are defined  here, the sane thing to do would have been to only leave `operator<=>` but changing it to return `strong_ordering`, so that `operator==` could be synthesized from it.

Anyway, in the final implementation `operator<=>` is not needed anymore, at least for now. So I dropped it entirely. Now we only have `operator==`.

include/revng/Model/TupleTree.h
aleclearmind: `s/NotIsUpcastablePointer/IsNotUpcastablePointer/` maybe

pfez: Ok. Done.

aleclearmind:
Merged in 1c403b5. Thanks!

PR #73, opened 2021-05-19
Lifter + PageBuster

zTehRyaN:
Some additions to the Lifter logic to handle PageBuster custom ELF file.

### DIFF
--- a/tools/revng-lift/BinaryFile.cpp
+++ b/tools/revng-lift/BinaryFile.cpp
@@ -39,6 +39,10 @@ using LabelList = BinaryFile::LabelList;
 static Logger<> EhFrameLog("ehframe");
 static Logger<> LabelsLog("labels");
 
+// Hack-ish custom PT
+// TODO find a finer way
+constexpr int PT_PAGEBUSTER { 0x8 };
+
 const unsigned char R_MIPS_IMPLICIT_RELATIVE = 255;
 
 namespace nooverflow {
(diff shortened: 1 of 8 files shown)


tools/revng-lift/BinaryFile.cpp
```diff
@@ -39,6 +39,12 @@ using LabelList = BinaryFile::LabelList;
 static Logger<> EhFrameLog("ehframe");
 static Logger<> LabelsLog("labels");
 
+// Hack-ish custom PT
+// TODO find a finer way
+constexpr int PT_PAGEBUSTER { 0x8 };
+
+static int g_epoch_counter { 0 };
```
aleclearmind, line 46: Do not use global variables, this can be a local variable just outside the loop.
Also make sure you follow the [LLVM Coding Conventions](https://llvm.org/docs/CodingStandards.html).

```suggestion
static int EpocCounter { 0 };
```

tools/revng-lift/BinaryFile.cpp
```diff
@@ -957,10 +963,13 @@ void BinaryFile::parseELF(object::ObjectFile *TheBinary,
 
   for (Elf_Phdr &ProgramHeader : *ProgHeaders) {
     switch (ProgramHeader.p_type) {
+    case PT_PAGEBUSTER:
     case ELF::PT_LOAD: {
+      bool pagebuster { ProgramHeader.p_type == PT_PAGEBUSTER };
```
aleclearmind, line 968: Just use regular assignments unless there's a reason to.
```suggestion
      bool IsPageBuster = (ProgramHeader.p_type == PT_PAGEBUSTER);
```

tools/revng-lift/BinaryFile.cpp
```diff
@@ -1002,7 +1011,13 @@ void BinaryFile::parseELF(object::ObjectFile *TheBinary,
                                    - u64(ProgramHeader.p_offset));
         ProgramHeaders.Address = PhdrAddress;
       }
+
+      if (pagebuster) {
+        g_epoch_counter++;
+      }
+
     } break;
+    [[fallthrough]];
```
aleclearmind, line 1020: Why? It's not falling through, there's a `break`.

tools/revng-lift/BinaryFile.cpp
```diff
@@ -1002,7 +1011,13 @@ void BinaryFile::parseELF(object::ObjectFile *TheBinary,
                                    - u64(ProgramHeader.p_offset));
         ProgramHeaders.Address = PhdrAddress;
       }
+
+      if (pagebuster) {
+        g_epoch_counter++;
+      }
```
aleclearmind, line 1017: Maybe we should outline the whole body of this `case` into a function and call it differently depending on whether we're in `case PT_PAGEBUSTER` or `case PT_LOAD`.

PR #76, opened 2021-05-27
Update revng to build against llvm-12

pfez:
This PR is still a draft, since there are runtime errors in some of revng tests.

This is expected to build fine with `orc install --test revng`, with the following pre-requisites:
- `orchestra` is based on: https://github.com/revng/orchestra/pull/36
- `llvm` is based on: https://github.com/revng/llvm-project/pull/5

@aleclearmind can you take a look at the failing tests?

### DIFF
--- a/include/revng/BasicAnalyses/MaterializedValue.h
+++ b/include/revng/BasicAnalyses/MaterializedValue.h
@@ -31,13 +31,17 @@ public:
 
 public:
   bool operator==(const MaterializedValue &Other) const {
-    auto This = std::tie(IsValid, SymbolName, Value);
-    auto That = std::tie(Other.IsValid, Other.SymbolName, Other.Value);
+    auto MaxBitWidth = std::max(Value.getBitWidth(), Other.Value.getBitWidth());
+    auto ZExtValue = Value.zextOrSelf(MaxBitWidth);
+    auto ZExtOther = Other.Value.zextOrSelf(MaxBitWidth);
+    auto This = std::tie(IsValid, SymbolName, ZExtValue);
+    auto That = std::tie(Other.IsValid, Other.SymbolName, ZExtOther);
     return This == That;
   }
 
   bool operator<(const MaterializedValue &Other) const {
-    if (Value.ult(Other.Value))
+    auto MaxBitWidth = std::max(Value.getBitWidth(), Other.Value.getBitWidth());
+    if (Value.zextOrSelf(MaxBitWidth).ult(Other.Value.zextOrSelf(MaxBitWidth)))
       return true;
     auto This = std::tie(IsValid, SymbolName);
     auto That = std::tie(Other.IsValid, Other.SymbolName);
--- a/lib/FunctionIsolation/IsolateFunctions.cpp
+++ b/lib/FunctionIsolation/IsolateFunctions.cpp
@@ -190,14 +190,20 @@ private:
   void createFunctionCall(IRBuilder<> &Builder,
                           MetaAddress ExpectedCallee,
                           const Boundary &TheBoundary,
-                          FunctionBlocks &ClonedBlocks);
+                          FunctionBlocks &ClonedBlocks,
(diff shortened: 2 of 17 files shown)


lib/FunctionIsolation/IsolateFunctions.cpp
```diff
@@ -491,18 +501,43 @@ bool IFI::handleIndirectBoundary(const std::vector<Boundary> &Boundaries,
...
+      BasicBlock::iterator InsertPoint = Builder.GetInsertPoint();
+      revng_assert(not Builder.GetInsertBlock()->empty());
+      Instruction *Old = InsertPoint == Builder.GetInsertBlock()->end() ?
+                           &*Builder.GetInsertBlock()->rbegin() :
+                           &*InsertPoint;
+      throwException(Builder,
+                     "An unrechable instruction has been "
+                     "reached",
+                     Old->getDebugLoc());
+    } break;
```
aleclearmind, line 540: Too much code repetition for something of little relevance.
Why don't you push the boring code into `throwException` and add an argument for the `Instruction` to copy from?

include/revng/BasicAnalyses/MaterializedValue.h
```diff
@@ -31,13 +31,17 @@ class MaterializedValue {
...
   bool operator==(const MaterializedValue &Other) const {
-    auto This = std::tie(IsValid, SymbolName, Value);
-    auto That = std::tie(Other.IsValid, Other.SymbolName, Other.Value);
+    auto MaxBitWidth = std::max(Value.getBitWidth(), Other.Value.getBitWidth());
+    auto ZExtValue = Value.zextOrSelf(MaxBitWidth);
+    auto ZExtOther = Other.Value.zextOrSelf(MaxBitWidth);
+    auto This = std::tie(IsValid, SymbolName, ZExtValue);
+    auto That = std::tie(Other.IsValid, Other.SymbolName, ZExtOther);
     return This == That;
   }
```
aleclearmind, line 40: Maybe we should `assert` here and let the caller take care of this.
I can look into this.

aleclearmind (changes requested):

include/revng/BasicAnalyses/MaterializedValue.h
pfez: Ok, just be aware that llvm does the assertion on its own already. I decided to fix it because it triggered failures at runtime in `revng` tests.

IMHO `operator==` is tricky, because it might be used under the hood in some algorithm that uses `MaterializedValue`s. My feeling was that in this case it was easier to fix `operator==` than force all users to be aware of this. For instance, what if someone puts a bunch of `MaterializedValue`s into a `std::vector` and calls `std::find` on it? It looks an innocuous operation, but if `operator==` calls `assert` under the hood it might crash.

Just my 2 cents. If you come up with a different solution let me know.

lib/FunctionIsolation/IsolateFunctions.cpp
pfez: I'll take a look into this.

pfez: Done. Let me know what you think.

lib/FunctionIsolation/IsolateFunctions.cpp
```diff
@@ -207,25 +213,51 @@ class IsolateFunctionsImpl {
...
+  } else {
+
+    revng_assert(not InsertBlock->empty());
+    BasicBlock::iterator InsertPoint = Builder.GetInsertPoint();
+    if (InsertPoint != InsertBlock->end())
+      DbgLoc = InsertPoint->getDebugLoc();
+    else
+      DbgLoc = InsertBlock->back().getDebugLoc();
+  }
+  revng_assert(DbgLoc);
```
aleclearmind, line 250: This code is very similar to the one in `createFunctionCall`. Can you factor it out in a function?

aleclearmind:
Merged in 952602c. Thanks!

PR #77, opened 2021-06-03
Add an include category for Qt headers.

Cvelth:
Since `cold-revng` extensively uses Qt headers it'd be nice to group them separately from all the others.

It also attempts to prevent clang-format from selecting a "main" header and putting it into a '0'-category.

### DIFF
--- a/lib/StackAnalysis/InterproceduralAnalysis.cpp
+++ b/lib/StackAnalysis/InterproceduralAnalysis.cpp
@@ -9,9 +9,8 @@
 
 #include "revng/Support/Statistics.h"
 
-#include "InterproceduralAnalysis.h"
-
 #include "Cache.h"
+#include "InterproceduralAnalysis.h"
 
 using llvm::BasicBlock;
 using llvm::GlobalVariable;
--- a/scripts/clang-format-style-file
+++ b/scripts/clang-format-style-file
@@ -82,12 +82,16 @@ IncludeCategories:
     Priority:     -5,
   }
 - {
-    Regex:        '^"boost/',
+    Regex:        '^"Qt[^/]+/',
     Priority:     -6,
   }
 - {
-    Regex:        '^<',
+    Regex:        '^"boost/',
     Priority:     -7,
+  }
+- {
+    Regex:        '^<',
+    Priority:     -8,
   }
 IndentCaseLabels: false,
 IndentPPDirectives: None,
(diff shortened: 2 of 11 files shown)


scripts/clang-format-style-file
```diff
@@ -82,13 +82,17 @@ IncludeCategories:
     Priority:     -5,
   }
 - {
-    Regex:        '^"boost/',
+    Regex:        '^"Qt[A-Za-z]+/Q[A-Za-z]+"$',
```
aleclearmind, line 85:
```suggestion
    Regex:        '^"Qt[^/]+/',
```

Cvelth: I wanted to be extra explicit to minimize possibility of false positives. Those filenames shouldn't use anything but letters if I understand correctly.

aleclearmind: `root/include/QtQuick3DAssetImport`

Cvelth: I'm stupid, sorry :)

Cvelth: Done.

aleclearmind:
```
find root/include/Qt* -type f -not -name "Q*"
```
I think just matching `Qt*/` is OK.

aleclearmind:
Can you make two distinct commits? The two changes are unrelated.

scripts/clang-format-style-file
Cvelth: `IncludeCategories` regex is case-insensitive by default, that leaves us with `QtBodymovin/bmglobal.h`, `QtZlib/zconf.h` and `QtZlib/zlib.h` which are not meant to be included explicitly.

But ok, I'll remove the Q just in case someone trying to include `zlib` in this weird fashion.

lib/StackAnalysis/InterproceduralAnalysis.cpp
```diff
@@ -9,9 +9,8 @@
 
 #include "revng/Support/Statistics.h"
 
-#include "InterproceduralAnalysis.h"
-
 #include "Cache.h"
+#include "InterproceduralAnalysis.h"
```
aleclearmind, line 13: Nice!

Cvelth:
I hope I didn't screw up anything, it was my first in-place commit edit ever.

aleclearmind:
LGTM

aleclearmind:
Merged in 2e214bc6616e132a32699615871266ae09eeb863. Thanks!

PR #64, opened 2021-05-05
Fixed typos in GeneratedIRReference.rst

zTehRyaN:

PR #80, opened 2021-06-17
Actualize GeneratedIRReference.rst

andrealinux1:
This PR revives #64.

The goal is to actualize the instructions contained in the `GeneratedIRReference.rst` document reflecting the latest changes.

### DIFF
--- a/docs/GeneratedIRReference.rst
+++ b/docs/GeneratedIRReference.rst
@@ -27,7 +27,7 @@ program compiled for x86-64:
       return a + myfunction();
     }
 
-The program has been compiled as follows:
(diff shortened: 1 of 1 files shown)


andrealinux1:
The only thing I am still unsure of is the fact that in the documentation we still heavily mention the `func.entry` and `func.member.of` metadata. They are still present, but the information is now contained in the model, if I got that correctly. Should we mention something about it or leave it to later?

aleclearmind:
Merged in fe7d0b256f5d5fe2951179ae5f1e94bafd8ac721. Thanks!

PR #81, opened 2021-06-23
`revng::SmallMap::contains`

Cvelth:
Now that we updated our active `llvm` branch to 12, we can actually add this method I've greatly missed.
For the context, `llvm::SmallSet::contains` was finally added in llvm-12.

### DIFF
--- a/include/revng/ADT/SmallMap.h
+++ b/include/revng/ADT/SmallMap.h
@@ -352,6 +352,10 @@ public:
       return const_iterator(Map.lower_bound(Key));
   }
 
+  bool contains(const K &Key) const {
+    return find(Key) != end();
+  }
+
   void clear() {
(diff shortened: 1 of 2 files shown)


aleclearmind:
ACK. Seems OK, maybe throw a little test in there.

Cvelth:
Test for what? Whether `revng::SmallMap::find` works? Whether `revng::SmallMap::end` works? Or whether `revng::SmallMap::const_iterator::operator!=(revng::SmallMap::const_iterator const &)` works? xDD

aleclearmind:
Just a quick test for the method you're adding.

Cvelth:
Something like this?

aleclearmind:
Finally merged in fbf90f8. Thanks a lot! :)

PR #83, opened 2021-07-27
Fix: `ManagedStatic` `IDGenerator` was not `static`

pfez:
This caused failures in `llvm::shutdown` in downstream libraries that depended on the revngModel library, such as upcoming changes in `revng-c`.


pfez:
Closing this, superseeded by https://github.com/revng/revng/pull/84

PR #88, opened 2021-08-26
Allow lines to start with `.`

Cvelth:
This rule, even if it prevents weird line-breaks, does not allow us to take full advantage of both designated initializers and method-chaining idiom.


aleclearmind:
Please use your full name and respect the 50/72 rule for commits.
Fixed it for you in this case.

PR #91, opened 2021-09-07
Fix `arm-thumb` disassembly

Cvelth:
Because of how `MetaAddressType` is converted to `llvm::Tripple::ArchType`, the `arm-thumb` disassembly doesn't work properly (since it's treated as normal `arm` instructions`).

### DIFF
--- a/include/revng/Support/MetaAddress.h
+++ b/include/revng/Support/MetaAddress.h
@@ -155,8 +155,9 @@ inline const llvm::Optional<llvm::Triple::ArchType> arch(Values V) {
   case Code_mipsel:
     return llvm::Triple::mipsel;
   case Code_arm:
-  case Code_arm_thumb:
     return llvm::Triple::arm;
+  case Code_arm_thumb:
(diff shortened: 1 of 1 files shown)


aleclearmind:
Uhm, this doesn't seem right. These function are supposed to give the CPU type, not the ISA flavor.
Maybe `llvm::Triple` is not the best, but LLVM does not offer something to purely represent the CPU.
Maybe we should have our own enum and push these functions closer to where they're used depending on the situation, but at the current stage, this patch violates the semantics the functions were designed to have.

If you want this in revng (as opposed to handling it on your side), we should have a new function such as `toLLVMISA`.

Cvelth:
If you think, it's better I handle this on Caliban side, I'm fine with that.

aleclearmind:
> Either way, we're ending up with this `toLLVMISA` function and I don't really see any difference as to where to put it - so the lowest scope (Caliban) works.

Go for it, we'll need to review this part and its usages eventually. Then, we might decide to merge back.

PR #96, opened 2021-10-01
Model JSON schema generation

fcremo:
Implements the `revng-dump-schema` command, which dumps the JSON-schema for the rev.ng model.

There are some weird parentheses, those are there to make clang-format and revng-check-conventions happy as they don't play well with concepts.

Note: This PR is based on 5f9fdd73fc7ae118c263005891f2e891058ffaf9 from the current state of feature/dynamic-functions (its current HEAD is broken).


aleclearmind:
I reset'd `feature/dynamic-functions` to that commit.
Please make the PR target that branch so it can be reviewed it without noise.

fcremo:
Ack! I have to close & open another PR as AFAIK there's no way to change the target branch for a PR

PR #97, opened 2021-10-04
Autogenerated model JSON schema and python classes, `revng-convert-idb` tool

fcremo:
This PR:

- moves revng python classes to `/python`
- introduces python classes to represent, serialize and deserialize rev.ng models, autogenerated from the schema
- introduces the `revng-verify-model` command, which verifies that a model is valid
  - base tool was taken from @pfez, I added the capability to load a list of invalid types to be removed from file
- introduces the `revng-convert-idb` command, which can be used to convert IDB files to revng models

Other minor changes worth noting:
- two commits relaxing some validation constraints, not sure if we still want them or not
- one commit (Handle function prototypes cleanup) needed to cleanup function prototypes referencing invalid types
- Minor improvement to `revng-dump-schema`, to not print an exception if a SIGINT (CTRL+C) is raised while it is processing (or waiting to read from stdin, as it is the default behavior if invoked without parameters)


aleclearmind:
Please undraft once this is compatible with #109.

fcremo:
This PR is now based on fcremo:feature/model-classes-generator (from PR #109)

fcremo:
Splitted into #128 and #129, closing

PR #99, opened 2021-10-06
Bug in `revng --prefix`

andrealinux1:
Fix a operator precedence bug that caused an incorrect override of the search path for libraries when using the `revng` script

### DIFF
--- a/scripts/revng
+++ b/scripts/revng
@@ -134,7 +134,7 @@ def get_command(command):
 
 def build_opt_args(args):
   analysis_libraries = []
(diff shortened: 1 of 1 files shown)


andrealinux1:
@pfez just spotted a further issue with this piece of code (local libraries not having precedence over root ones), let's wait until also this is fixed before merging this.

aleclearmind:
A superset of this fix has been merged in 93b93a3908eea9af7c16ed21760c007e970c5b93.
Thanks!

PR #100, opened 2021-10-06
Feature/debug info pass

andrealinux1:
Add a pass which which invokes DebugHelper in order to decorate an LLVM
IR module with the debug information that can be later exploited by gdb
in order to perform src stepping when executing a translated binary.

Add also an option in revng translate to automatically schedule the
pass.

I also addressed some doubts expressed in a previous MR that later dropped this functionality, [here](https://github.com/revng/revng/pull/37#discussion_r562720680) and [here](https://github.com/revng/revng/pull/37#discussion_r562720680), specifically the extraction of the class definition in an header, and the addition of a cl::opt that enables to specify a name for the newly created module, while the default behavior is to overwrite the input module in place.

### DIFF
--- a/include/revng/Support/DebugInfo.h
+++ b/include/revng/Support/DebugInfo.h
@@ -1,0 +1,17 @@
+#pragma once
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "llvm/Pass.h"
+
+class DebugInfo : public llvm::ModulePass {
+public:
+  static char ID;
+
+public:
+  DebugInfo() : ModulePass(ID) {}
+
+  bool runOnModule(llvm::Module &M) override;
+};
--- a/lib/Support/CMakeLists.txt
+++ b/lib/Support/CMakeLists.txt
@@ -7,6 +7,7 @@ revng_add_library_internal(revngSupport SHARED
   CommandLine.cpp
   Debug.cpp
   DebugHelper.cpp
+  DebugInfo.cpp
   ExampleAnalysis.cpp
   FunctionTags.cpp
   IRHelpers.cpp
--- a/lib/Support/DebugInfo.cpp
+++ b/lib/Support/DebugInfo.cpp
@@ -1,0 +1,34 @@
+/// \file DebugInfo.cpp
+/// \brief Run an instance of DebugHelper over the IR.
+
+#include "llvm/IR/AssemblyAnnotationWriter.h"
+#include "llvm/IR/IRBuilder.h"
+
+#include "revng/Support/CommandLine.h"
+#include "revng/Support/Debug.h"
+#include "revng/Support/DebugHelper.h"
+#include "revng/Support/DebugInfo.h"
+#include "revng/Support/revng.h"
+
+using namespace llvm;
+
+char DebugInfo::ID = 0;
+static RegisterPass<DebugInfo> X("debug-info", "Add debug metadata");
+
+cl::opt<std::string> DebugPath("debug-info-path",
+                               cl::desc("Filename produced as output"),
+                               cl::value_desc("filename"),
+                               cl::NumOccurrencesFlag::Optional);
+
+bool DebugInfo::runOnModule(Module &M) {
+  std::string DecoratedModulePath;
+  if (DebugPath.size() == 0) {
+    DecoratedModulePath = M.getModuleIdentifier();
+  } else {
+    DecoratedModulePath = DebugPath;
+  }
+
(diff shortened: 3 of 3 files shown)


aleclearmind:
Please don't stack PRs if not strictly necessary.

lib/DebugInfo/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+revng_add_analyses_library_internal(revngDebugInfo
+  DebugInfo.cpp)
+
+target_link_libraries(revngDebugInfo
+  revngSupport
+  ${LLVM_LIBRARIES})
```
aleclearmind, line 10: I think this can be part of `librevngSupport`, where `DebugHelper` is defined.

include/revng/DebugInfo/DebugInfo.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include <memory>
```
aleclearmind, line 7: Drop?

scripts/revng
```diff
@@ -420,6 +424,14 @@ If you're just interested in lifting to LLVM IR, simply use `revng lift`.
...
   output = linked
 
+  # Emit debug info with the ad-hoc pass.
+  if args.debuginfo:
+    opt_invocation = build_opt_args(["-S",
+                                    "-debug-info",
+                                    relative(output),
+                                    "-o", "/dev/null"])
+    run(opt_invocation)
+
```
aleclearmind, line 434: Is this feature so important that we want to expose it through `revng-translate`?
This seems to me more of a debug tool to use when needed.
Also, the new autoenforcer would have to implement this, which is a bit annoying.

lib/DebugInfo/DebugInfo.cpp
```diff
@@ -0,0 +1,35 @@
...
+  if (DebugPath.size() == 0) {
+    DebugHelper DBG("debug_info",
+                    &M,
+                    DebugInfoType::LLVMIR,
+                    M.getModuleIdentifier());
+    DBG.generateDebugInfo();
+  } else {
+    DebugHelper DBG("debug_info", &M, DebugInfoType::LLVMIR, DebugPath);
+    DBG.generateDebugInfo();
+  }
```
aleclearmind, line 33: * [ ] Use the `if` for the path only, the rest is identical.
* [ ] It makes me a bit sad that we have to specify a path. It'd be nice to be able to run `opt --debug-info` and just get the final output decorated as appropriately. However, I think is not possible due to the fact that debug info need to explicitly reference a path.
* [ ] Why not have a switch for the other `DebugInfoType`s?

aleclearmind (changes requested):

scripts/revng
andrealinux1: I have to admit it is very useful, especially because if used in conjunction with the `--skip` flag, it enables you basically to take any `.ll`, and apply the debug info emission on it, but I get this may not be an universal need, I'll keep this on my local branch. Changed in 7920dab4d3755c4e576b8de736184a8ce4ddac8e

lib/DebugInfo/DebugInfo.cpp
andrealinux1: Improved the `if` in 7920dab4d3755c4e576b8de736184a8ce4ddac8e.

For the second point, yeah I'm sad too but unfortunately, as you pointed out, due to the fact that we are actually abusing a bit the `DILocation` mechanism the standard way to interact with `opt` is not adequate (e.g. the double serialization pass just to collect the line at which a certain instruction will be serialized in a `.ll` so that we can fix the correct `DILocation`, and therefore the fact that this pass does not make sense unless it is executed as the last one (we discussed more on this aspects in this old discussion on the old PR https://github.com/revng/revng/pull/37#discussion_r564695242).

For the third point, yes in general I guess we can have the switch, for the moment I just focused on decorating llvm ir beacuse I never used it for anything else. 
Practically, how would you takle the task? I see two main problems:
- The first, how to specify the `DebugInfoType` wanted? Another `cl::option`?
- The second, if we implement the switch we also need to take care that the name of the produced module is coherent with the argument (simplifying, `.ll` for LLVMIR, `.S` etc...). Do we just rely on the fact that the user specify coherent things across the `DebugInfoType` and extension, or we manually add the coherent extension ourselves?

PythonSHA:
@antoniofrighetto

PR #104, opened 2021-10-23
Fix memory leak in TupleTreePath::operator=

AlviseDeFaveri:
Operator= should cleanup the object that is being assigned before resizing it, otherwise the TupleTreeKeyWrapper destructor is never called and, as a consequence, the object pointed by the Pointer member is never freed.


aleclearmind:
Merged in 2e19fad. Thanks!

PR #107, opened 2021-11-15
Fixes for `dla-pointer-edges`

AlviseDeFaveri:
This MR contains some commits that are needed by `revng-c`'s `feature/dla-pointer-edges` branch before it can be merged.

### DIFF
--- a/include/revng/Model/Binary.h
+++ b/include/revng/Model/Binary.h
@@ -614,6 +614,7 @@ public:
   bool verify(bool Assert) const debug_function;
   bool verify(VerifyHelper &VH) const;
   void dump() const debug_function;
+  std::string toString() const debug_function;
 };
 INTROSPECTION_NS(model,
                  Binary,
--- a/include/revng/Support/Debug.h
+++ b/include/revng/Support/Debug.h
@@ -282,3 +282,5 @@ extern Logger<> NRALog;
 extern Logger<> PassesLog;
 extern Logger<> ReleaseLog;
 extern Logger<> VerifyLog;
+
+void writeToFile(llvm::StringRef What, llvm::StringRef Path) debug_function;
--- a/lib/FunctionIsolation/EnforceABI.cpp
+++ b/lib/FunctionIsolation/EnforceABI.cpp
@@ -115,6 +115,7 @@ void EnforceABIImpl::run() {
                               "opaque_pc",
                               M);
   OpaquePC->addFnAttr(Attribute::NoUnwind);
+  OpaquePC->addFnAttr(Attribute::WillReturn);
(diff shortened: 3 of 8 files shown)


include/revng/Model/Binary.h
```diff
@@ -614,6 +614,7 @@ class model::Binary {
   bool verify(bool Assert) const debug_function;
   bool verify(VerifyHelper &VH) const;
   void dump() const debug_function;
+  void dumpOnFile(const char *FileName) const debug_function;
```
aleclearmind, line 617:
```suggestion
  template<typename T>
  void dump(T &Stream) const;
  void dump() const debug_function { dump(dbg); }
```

AlviseDeFaveri: Can I use this inside `gdb` to dump the model at arbitrary points on a file for later inspection?

aleclearmind: Mmmh, we have a `dumpModule` function for `llvm::Module` for `gdb`.
Not sure, I don't want this on every possible part of the model, but maybe we can have it on the root.

Maybe we can have a `std::string toString()` method and then a global function `writeToFile(StringRef Path, StringRef What)`.
```
(gdb) p writeToFile(Model.toString(), "/tmp/asd.yml");
```

What do you think?

AlviseDeFaveri: That would be nice

AlviseDeFaveri: Where is the best place to put `writeToFile()`?

aleclearmind: `include/revng/Support/Debug.h` for now.

AlviseDeFaveri: I noticed that the order you suggested was different in the gdb example and in the description of the to-be function signature. I went for the former (`writeToFile(StringRef What, StringRef Path)`).

include/revng/Support/Debug.h
```diff
@@ -282,3 +282,6 @@ extern Logger<> NRALog;
 extern Logger<> PassesLog;
 extern Logger<> ReleaseLog;
 extern Logger<> VerifyLog;
+
+void writeToFile(const llvm::StringRef What,
+                 const llvm::StringRef Path) debug_function;
```
aleclearmind, line 287: Passing `const` by value does not make a lot of sense.
Also, keep in mind that `StringRef` and `ArrayRef` already represent read-only references (you can't make changes to the pointed data).

AlviseDeFaveri: oops, fixed removing const.

aleclearmind:
Merged in 2f4b716. Thanks!

PR #109, opened 2021-11-18
Model classes generator

fcremo:
Switch to generating model classes from their YAML description.


include/revng/Model/Binary.h
```diff
@@ -6,593 +6,16 @@
...
-class Function;
-class DynamicFunction;
-class Binary;
-class FunctionEdge;
-class CallEdge;
-class BasicBlock;
-class Segment;
-} // namespace model
+#include "revng/Model/GeneratedBinary.h"
+#include "revng/Model/GeneratedPrimitiveTypeKind.h"
```
aleclearmind, line 10: Why do we need this? Can't we generate this include in `revng/Model/GeneratedBinary.h`?

include/revng/Model/Binary.h
```diff
@@ -6,593 +6,16 @@
...
 
 // TODO: Prevent changing the keys. Currently we need them to be public and
 //       non-const for serialization purposes.
```
aleclearmind, line 13: Soon...

include/revng/Model/CallEdge.h
```diff
@@ -0,0 +1,28 @@
...
+
+public:
+  bool verify() const debug_function;
+  bool verify(bool Assert) const debug_function;
+  bool verify(VerifyHelper &VH) const;
+  void dump() const debug_function;
+
+public:
+  // TODO: ideally we should not need to define this here
+  static constexpr const char *Tag = "!CallEdge";
```
aleclearmind, line 25: Didn't we decide for autogenerating a `getTag` trait for this?

include/revng/Model/CallEdge.h
```diff
@@ -0,0 +1,28 @@
...
+
+/// A CFG edge to represent function calls (direct, indirect and tail calls)
+class model::CallEdge : public model::generated::CallEdge {
+public:
+  using generated::CallEdge::CallEdge;
+
+public:
+  static bool classof(const FunctionEdge *A) {
+    return FunctionEdgeType::isCall(A->Type);
+  }
```
aleclearmind, line 15: Can you make a brief markdown document describing how to create a new class?

How to include, who you need to inherit from, how to import constructors, what typically you have to write by hand (`verify`, `dump`, `classof`), how inheritance works.

include/revng/Model/DynamicFunction.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include "revng/Model/GeneratedDynamicFunction.h"
+
+/// Function defined in a dynamic library
+class model::DynamicFunction : public model::generated::DynamicFunction {
```
aleclearmind, line 6: FYI: what have does not seem to be a [mixin](https://en.wikipedia.org/wiki/Mixin).

include/revng/Model/DynamicFunction.h
```diff
@@ -0,0 +1,22 @@
...
+public:
+  bool verify() const debug_function;
+  bool verify(bool Assert) const debug_function;
+  bool verify(VerifyHelper &VH) const;
+  void dump() const debug_function;
+};
+
+#include "revng/Model/GeneratedDynamicFunction_late.h"
+
+static_assert(validateTupleTree<model::DynamicFunction>(IsYamlizable));
```
aleclearmind, line 22: Can we push this in the generated code?
Should always be true.

include/revng/Model/FunctionEdge.h
```diff
@@ -0,0 +1,58 @@
...
+class model::FunctionEdge : public model::generated::FunctionEdge {
+public:
+  using generated::FunctionEdge::FunctionEdge;
+
+public:
+  bool operator<(const FunctionEdge &Other) const {
+    auto ThisTie = std::tie(Destination, Type);
+    auto OtherTie = std::tie(Other.Destination, Other.Type);
+    return ThisTie < OtherTie;
+  }
```
aleclearmind, line 16: Can you try to generate the `<=>` operator?

include/revng/Model/FunctionEdge.h
```diff
@@ -0,0 +1,58 @@
...
+
+public:
+  // TODO: ideally we should not need to define this here
+  static constexpr const char *Tag = "!FunctionEdge";
+};
+
+// This must be included first
+#include "revng/Model/GeneratedFunctionEdge_late.h"
+// And this second. The comments don't let clang-format reorder them.
+#include "revng/Model/CallEdge.h"
```
aleclearmind, line 36: Split the file. No includes in the middle of the file.

include/revng/Model/FunctionEdge.h
```diff
@@ -0,0 +1,58 @@
...
+      return ResultType(new CallEdge(std::get<0>(Obj), std::get<1>(Obj)));
+    } else {
+      return ResultType(new FunctionEdge(std::get<0>(Obj), std::get<1>(Obj)));
+    }
+  }
+};
+
+template<>
+class llvm::yaml::MappingTraits<UpcastablePointer<model::FunctionEdge>>
+  : public PolymorphicMappingTraits<UpcastablePointer<model::FunctionEdge>> {};
```
aleclearmind, line 58: I'd like this to be generated too.

include/revng/Model/Qualifier.h
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  bool isPointerQualifier() const {
+    revng_assert(verify(true));
+    return Kind == QualifierKind::Pointer;
+  }
+};
+
+// Make std::vector<model::Qualifier> yaml-serializable as a sequence
+LLVM_YAML_IS_SEQUENCE_VECTOR(model::Qualifier)
```
aleclearmind, line 46: Let's push this in generated code, if we want to customize this we can add an option in the YAML.

include/revng/Model/Register.h
```diff
@@ -8,688 +8,12 @@
 
 #include "revng/ADT/KeyedObjectTraits.h"
 #include "revng/Model/Architecture.h"
+#include "revng/Model/GeneratedRegister.h"
+#include "revng/Model/GeneratedRegisterState.h"
```
aleclearmind, line 12: As previously, why do we need this?
Also, missing `_late` inclusion.

include/revng/Model/Type/BaseType.h
```diff
@@ -0,0 +1,111 @@
+#pragma once
```
aleclearmind, line 1: The name of the file should reflect the name of the class it defines.

include/revng/Model/Type/BaseType.h
```diff
@@ -0,0 +1,111 @@
...
+  static bool classof(const Type *T) { return false; }
+
+  Key key() const { return Key{ Kind, ID }; }
+
+  Identifier name() const;
+
+public:
+  bool operator==(const Type &Other) const { return key() == Other.key(); }
+
+  bool operator<(const Type &Other) const { return key() < Other.key(); }
```
aleclearmind, line 28: Needs to be generated

include/revng/Model/Type/BaseType.h
```diff
@@ -0,0 +1,111 @@
...
+  // TODO: Constructors cannot be inherited, since the default one is
+  //  manually implemented in order to generate a random ID
+  Type();
+  Type(TypeKind::Values TK);
+  Type(TypeKind::Values Kind, uint64_t ID) : model::generated::Type(Kind, ID) {}
+
+public:
+  static bool classof(const Type *T) { return false; }
+
+  Key key() const { return Key{ Kind, ID }; }
```
aleclearmind, line 21: Is the `Key` type generated in the base class?

include/revng/Model/Type/BaseType.h
```diff
@@ -0,0 +1,111 @@
...
+  // TODO: we should not need to define this here
+  static constexpr const char *Tag = "!Type";
+  static constexpr const auto AssociatedKind = TypeKind::Invalid;
+};
+
+namespace model {
+using TypePath = TupleTreeReference<model::Type, model::Binary>;
+} // end namespace model
+
+#include "revng/Model/GeneratedType_late.h"
```
aleclearmind, line 50: No interleaved includes.

include/revng/Model/Type/BaseType.h
```diff
@@ -0,0 +1,111 @@
...
+public:
+  bool verify() const debug_function;
+  bool verify(bool Assert) const debug_function;
+  RecursiveCoroutine<bool> verify(VerifyHelper &VH) const;
+  void dump() const debug_function;
+
+public:
+  // TODO: we should not need to define this here
+  static constexpr const char *Tag = "!Type";
+  static constexpr const auto AssociatedKind = TypeKind::Invalid;
```
aleclearmind, line 43: One day we might want to generate `classof` and `AssociatedKind`. But at this stage, I'm not sure it's easy due to `FunctionEdge`/`CallEdge`.

include/revng/Model/Type/Type.h
```diff
@@ -0,0 +1,38 @@
+#pragma once
```
aleclearmind, line 1: Maybe rename this to `include/revng/Model/AllTypes.h`?
`Type.h` should be for the `model::Type` class.

include/revng/Model/schema.yml
```diff
@@ -0,0 +1,701 @@
...
+    members:
+      # x86 registers
+      - name: eax_x86
+      - name: ebx_x86
+      - name: ecx_x86
+      - name: edx_x86
+      - name: esi_x86
+      - name: edi_x86
+      - name: ebp_x86
+      - name: esp_x86
```
aleclearmind, line 43: One day, `ytt`. Not today.

include/revng/Model/schema.yml
```diff
@@ -0,0 +1,701 @@
...
+  #include <optional>
+
+  #include "llvm/ADT/StringRef.h"
+  #include "llvm/Support/YAMLTraits.h"
+  #include "revng/ADT/RecursiveCoroutine.h"
+  #include "revng/Model/Identifier.h"
+  #include "revng/Model/TupleTree.h"
+  #include "revng/Model/VerifyHelper.h"
+  #include "revng/Support/MetaAddress.h"
+
```
aleclearmind, line 14: This YAML file should strive to be language agnostic. One day we could drop `SortedVector` in favor of some other syntax.

For sure, no includes.
These can be pushed in a `.h` file somewhere.

Why do you need to include `VerifyHelper.h` and ` RecursiveCoroutine.h` everywhere?
The others seem fine.

include/revng/Model/schema.yml
```diff
@@ -0,0 +1,701 @@
...
+types:
+  # Enums
+  - name: ABI
+    type: enum
+    members:
+      - name: SystemV_x86_64
```
aleclearmind, line 20: Do enums have something else except the name?
If not, make it a list of strings.
I guess they `doc`.

include/revng/Model/schema.yml
```diff
@@ -0,0 +1,701 @@
...
+      - name: Pointer
+      - name: Array
+      - name: Const
+
+  - name: FunctionEdgeType
+    doc: Type of edge on the CFG
+    type: enum
+    members:
+      - doc: Branch due to function-local CFG (a regular branch)
+        name: DirectBranch
```
aleclearmind, line 224: `name` first

include/revng/Model/schema.yml
```diff
@@ -0,0 +1,701 @@
...
+        name: Fake
+
+  # Structs
+  - name: Type
+    doc: Base class of model types used for LLVM-style RTTI
+    type: struct
+    fields:
+      - name: Kind
+        type: model::TypeKind::Values
+        default: TypeKind::Invalid
```
aleclearmind, line 263: Didn't we say that `Invalid` is not implicit?

include/revng/Model/schema.yml
```diff
@@ -0,0 +1,701 @@
...
+      - name: ID
+        type: uint64_t
+        default: "0"
```
aleclearmind, line 266: Is there a single situation where `default` is not `0`/`Invalid`?
I'd like to drop `default` or at least make it optional as much as possible.
Initializing fields with `{}` should do the trick.

include/revng/Model/schema.yml
```diff
@@ -0,0 +1,701 @@
...
+    key:
+      - Kind
+      - ID
+    default_comparable: false
```
aleclearmind, line 270: What's this?

include/revng/Model/schema.yml
```diff
@@ -0,0 +1,701 @@
...
+    virtual: true
```
aleclearmind, line 271:
```suggestion
    abstract: true
```

include/revng/Model/schema.yml
```diff
@@ -0,0 +1,701 @@
...
+      - name: PrimitiveKind
+        type: model::PrimitiveTypeKind::Values
+        default: PrimitiveTypeKind::Invalid
+      - name: Size
+        doc: Size in bytes
+        type: uint8_t
+        default: "0"
+    preamble: |
+      #include "revng/Model/Type/BaseType.h"
+
```
aleclearmind, line 287: I really don't think this belongs here. Come over and let's discuss this.

include/revng/Model/schema.yml
fcremo: Yes, enum members can have a `doc` (see `FunctionEdgeType` for instance)

include/revng/Model/Architecture.h
```diff
@@ -6,34 +6,9 @@
 
 #include "llvm/ADT/Triple.h"
 
-#include "revng/Model/TupleTree.h"
-#include "revng/Support/YAMLTraits.h"
+#include "revng/Model/GeneratedArchitecture.h"
```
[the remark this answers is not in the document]
fcremo: Enums do not need the `_late` include. Initially I generated files in `generated/XXX.h` and `generated/XXX_late.h`, I like your proposal better, I'll make the change.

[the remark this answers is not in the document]
aleclearmind: > Enums do not need the `_late` include.

OK, but that's not obvious to me and in the future they might be needed.
I'd prefer to include an empty file than having to remember when I don't have to include it.
I'm dumb and I like predictability.

include/revng/ADT/UpcastablePointer/YAMLTraits.h
```diff
@@ -51,3 +51,6 @@ struct PolymorphicMappingTraits {
     dispatchMappingTraits(io, Obj);
   }
 };
+
+template<typename T>
+const char *getTag();
```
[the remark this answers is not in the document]
fcremo: Removed

include/revng/Model/Architecture.h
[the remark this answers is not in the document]
fcremo: I made the change, there's an unfortunate side effect, namely an `#include "../ForwardDecls.h"` in the autogenerated early files. We should discuss a more elegant solution.

include/revng/Model/Argument.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
```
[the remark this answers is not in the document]
fcremo: Oops, sorry, fixed!

include/revng/Model/Argument.h
```diff
@@ -0,0 +1,21 @@
...
+
+#include "revng/Model/GeneratedArgument.h"
+
+/// \brief The argument of a function type
+///
+/// It features an argument index (the key), a type and an optional name
```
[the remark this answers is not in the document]
fcremo: Indeed, I dropped the redundant docs

include/revng/Model/Binary.h
fcremo: Yep, there's no need to include `PrimitiveTypeKind.h`, I dropped it.

include/revng/Model/DynamicFunction.h
fcremo: There already is such an assert in the generated code :) I just missed it from the cleanup, removed!

include/revng/Model/Register.h
fcremo: Well, we need `GeneratedRegister.h` (now `Generated/Early/Register.h`) because `inline llvm::StringRef getRegisterName(Values V)` refers to `model::Register::Values`.
Same goes for `isYesOrDead`, it refers to `model::RegisterState::Values`. I can create a separate `RegisterState.h` if you prefer.
As I was saying in the other reply, enums do not need (nor have) a `_late` file to include.

include/revng/Model/Type/BaseType.h
fcremo: Yes

include/revng/Model/Binary.h
```diff
@@ -615,17 +38,8 @@ class model::Binary {
...
-                 Functions,
-                 ImportedDynamicFunctions,
-                 Types,
-                 Architecture,
-                 Segments)
 
-template<>
-struct llvm::yaml::MappingTraits<model::Binary>
-  : public TupleLikeMappingTraits<model::Binary> {};
+#include "revng/Model/GeneratedBinary_late.h"
```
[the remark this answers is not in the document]
aleclearmind: Can we push this in the generated file then?

include/revng/Model/Register.h
aleclearmind: Can't you include `revng/Model/RegisterState.h`?
Maybe `revng/Model/GeneratedRegister.h` should do that.

include/revng/Model/Type/BaseType.h
aleclearmind: Who is using this method? If it's used only by generated code, I think we can drop it.

lib/Model/CMakeLists.txt
```diff
@@ -11,3 +11,17 @@ revng_add_analyses_library_internal(revngModel
...
+
+set(OUTPUT_DIR "${CMAKE_BINARY_DIR}/include/revng/Model")
+set(MAIN_OUTPUT "${GENERATED_DIR}/Generated.h")
+set(MODEL_YAML "${CMAKE_SOURCE_DIR}/include/revng/Model/schema.yml")
+
+add_custom_command(
+    OUTPUT "${MAIN_OUTPUT}"
+    COMMAND "${CMAKE_SOURCE_DIR}/scripts/model_gen_ng/main.py"
+    ARGS "${MODEL_YAML}" "${OUTPUT_DIR}"
+)
```
aleclearmind, line 23: OK, let's see. I've to say that, while all of this is super cool, when looking at the hand-written files, I'm like "wait, what fields this class has again?".

Maybe it's better to disaggregate the YAML file in the various header files.

```c++
#pragma once

#include "revng/Model/GeneratedSegment.h"

class model::Segment : public model::generated::Segment {
//+ - name: Segment
//+   type: struct
//+   fields:
//+     - name: StartAddress
//+       type: MetaAddress
//+     - name: EndAddress
//+       type: MetaAddress
//+     - name: StartOffset
//+       type: uint64_t
//+     - name: EndOffset
//+       type: uint64_t
//+     - name: IsReadable
//+       type: bool
//+       default: "false"
//+     - name: IsWriteable
//+       type: bool
//+       default: "false"
//+     - name: IsExecutable
//+       type: bool
//+       default: "false"
//+     - name: CustomName
//+       type: Identifier
//+       optional: true
//+
//+   key:
//+     - StartAddress
//+     - EndAddress
public:
  using generated::Segment::Segment;

public:
  Identifier name() const;

public:
  bool verify() const debug_function;
  bool verify(VerifyHelper &VH) const;
};

#include "revng/Model/GeneratedSegment_late.h"

static_assert(validateTupleTree<model::Segment>(IsYamlizable));
```

Then, in CMake, we explicitly list the header files, we have a script that basically `grep //+` and assembles the YAML file in the build directory, and then the script we already have uses it.

It's not ideal, but overall I think it's better.

lib/Model/Type.cpp
```diff
@@ -180,6 +179,10 @@ class RNG {
 
 static llvm::ManagedStatic<RNG> IDGenerator;
 
+model::Type::Type() : model::generated::Type() {
+  ID = IDGenerator->get();
+};
```
aleclearmind, line 184:
```suggestion
model::Type::Type() : model::generated::Type(), ID(IDGenerator->get()) { };
```

lib/Model/Type.cpp
```diff
@@ -1056,20 +1061,54 @@ bool NamedTypedRegister::verify(bool Assert) const {
...
+
+  // Zero-sized types are not allowed
+  if (not MaybeTypeSize)
+    rc_return VH.fail();
+
+  size_t RegisterSize = model::Register::getSize(Location);
+  if (*MaybeTypeSize > RegisterSize)
+    rc_return VH.fail();
+
+  rc_return VH.maybeFail(rc_recur Type.verify(VH));
```
aleclearmind, line 1086: Uhm, I'm a bit confused, why this pops up here now?

scripts/model_gen_ng/metaschema.yml
```diff
@@ -0,0 +1,168 @@
...
+    additionalItems: false
+
+  StructKey:
+    type: array
+    items:
+      type: string
+
+    additionalItems: false
+
+  BinaryArchive:
```
aleclearmind, line 108: Wat?

scripts/model_gen_ng/main.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python3
```
aleclearmind, line 1: Let's not couple this project to the model, but to tuple tree.
```
s/model_gen_ng/tuple_tree_generator/
```

scripts/model_gen_ng/model_gen_ng/clangfmt.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+def reformat(source: str) -> str:
+    """Applies clang-format to the given source"""
+    return subprocess.check_output(
+        [
+            "clang-format",
+        ],
+        input=source,
+        encoding="utf-8",
```
aleclearmind, line 11: * [ ] Let's make this soft-fail if `clang-format` is not available.
* [ ] Is this worth a separate file?

scripts/model_gen_ng/model_gen_ng/jinja_utils.py
```diff
@@ -0,0 +1,33 @@
+from pathlib import Path
+
+import jinja2
```
aleclearmind, line 3: Can you please make a list of all the Python deps we have in `revng`?
I'd like to have in CMake a `python -c 'import $PACKAGE'` to detect a configure time the missing dep.
Also, we need to update the orchestra config.

scripts/model_gen_ng/model_gen_ng/struct.py
```diff
@@ -0,0 +1,188 @@
+import re
```
aleclearmind, line 1: Copyright notice as all the Python files.
If we don't have it, please add it to the others too.

scripts/model_gen_ng/model_gen_ng/templates/class_forward_decls.tpl
```diff
@@ -0,0 +1,8 @@
+// Forward declarations
```
aleclearmind, line 1: All files, should always have the copyright notice.

scripts/model_gen_ng/model_gen_ng/templates/class_forward_decls.tpl
```diff
@@ -0,0 +1,8 @@
...
+[% for ns, names in ns_to_names.items() %]
+namespace [[ ns ]] {
+[%- for name in names|sort %]
+class [[ name ]];
+[%- endfor %]
+} // namespace [[ ns ]]
+[% endfor %]
```
aleclearmind, line 8: Testing highlighting:
```c++
/* for ns, names in ns_to_names.items() */
namespace /* ns */ {
/*- for name in names|sort */
class /* name */;
/*- endfor */
} // namespace /* ns */
/* endfor */
```
This is kinda nice.

scripts/model_gen_ng/model_gen_ng/templates/enum.tpl
```diff
@@ -0,0 +1,60 @@
...
+template<>
+struct KeyedObjectTraits<[[ enum.fullname ]]>
+  : public IdentityKeyedObjectTraits<[[ enum.fullname ]]> {};
+
+namespace llvm::yaml {
+template<>
+struct ScalarEnumerationTraits<[[ enum.fullname ]]>
+  : public NamedEnumScalarTraits<[[ enum.fullname ]]> {
+  static constexpr const char Name[] = "[[ enum.name ]]";
+  [# TODO: do we want to include the Invalid member in the count or not? -#]
```
aleclearmind, line 56: Yes.

scripts/model_gen_ng/model_gen_ng/templates/enum.tpl
```diff
@@ -0,0 +1,60 @@
...
+
+template<>
+struct KeyedObjectTraits<[[ enum.fullname ]]>
+  : public IdentityKeyedObjectTraits<[[ enum.fullname ]]> {};
+
+namespace llvm::yaml {
+template<>
+struct ScalarEnumerationTraits<[[ enum.fullname ]]>
+  : public NamedEnumScalarTraits<[[ enum.fullname ]]> {
+  static constexpr const char Name[] = "[[ enum.name ]]";
```
aleclearmind, line 55: We already have this in `TupleLikeTraits`.

scripts/model_gen_ng/model_gen_ng/templates/enum.tpl
```diff
@@ -0,0 +1,60 @@
...
+  [# TODO: do we want to include the Invalid member in the count or not? -#]
+  static constexpr const unsigned int NumMembers = [[ enum.members|length ]];
```
aleclearmind, line 57: Let's not abuse `llvm::yaml::*Traits`.
Make a new trait.

```c++
template<typename T>
constexpr size_t EnumElementsCount;

template<>
constexpr size_t EnumElementsCount<model::Architecture::Values> = 42;
```

scripts/model_gen_ng/model_gen_ng/templates/struct_late.tpl
```diff
@@ -0,0 +1,79 @@
+#pragma once
+
```
aleclearmind, line 2: * [ ] Copyright notice
* [ ] All generated files should clearly state they are generated. I can see people getting frustrated changing them.

scripts/model_gen_ng/model_gen_ng/templates/struct.tpl
```diff
@@ -0,0 +1,93 @@
...
+#include "[[ header ]]"
+[% endfor %]
+
+[[ struct.doc | docstring -]]
+struct [[ struct.fullname ]]
+        [%- if struct.inherits %] : public [[ struct.inherits.user_fullname ]] [% endif -%]
+        {
+    [#- --- Member list --- #]
+    [%- for field in struct.fields %]
+    [[ field.doc | docstring ]]
```
aleclearmind, line 13: Don't we need `///` here?

scripts/model_gen_ng/model_gen_ng/templates/enum.tpl
```diff
@@ -0,0 +1,60 @@
...
+            return [[ member.name ]];
+        }
+        [%- endif -%]
+        [%- endfor -%]
+        else {
+            return Invalid;
+        }
+    }
+
+    [[ enum.customcode ]]
```
aleclearmind, line 44: Drop

scripts/model_gen_ng/model_gen_ng/templates/struct.tpl
```diff
@@ -0,0 +1,93 @@
...
+        [%- endfor -%]
+        >;
+    [% endif %]
+
+    [# --- Operators --- #]
+    [% if struct.default_comparable %]
+    bool operator==(const [[ struct.name ]] &Other) const = default;
+    [% endif %]
+
+    [[ struct.customcode ]]
```
aleclearmind, line 91: Drop

scripts/model_gen_ng/model_gen_ng/struct.py
```diff
@@ -0,0 +1,188 @@
...
+            user_namespace,
+            name,
+            f"{namespace}::{name}",
+            f"{user_namespace}::{name}",
+            doc=doc,
+            tag=tag,
+        )
+        self.fields: List[StructField] = fields
+        self.default_comparable = default_comparable
+        self.virtual = virtual
```
aleclearmind, line 56: `s/virtual/abstract/` everywhere.

scripts/model_gen_ng/model_gen_ng/struct.py
```diff
@@ -0,0 +1,188 @@
...
+from .definition import Definition
+from .jinja_utils import environment
+
+struct_template = environment.get_template("struct.tpl")
+struct_late_template = environment.get_template("struct_late.tpl")
+struct_forward_decls_template = environment.get_template("struct_forward_decls.tpl")
+class_forward_decls_template = environment.get_template("class_forward_decls.tpl")
+
+
+class StructField:
```
aleclearmind, line 15: There is quite a bit of code for this. I wonder how much we could avoid this by going either schemaless and use `struct["name"]` from jinja or autogenerate these classes from the JSON schema.

Not saying we need to do this now, but this quite a bit of boilerplate code that needs to stay in sync with the schema.

aleclearmind (changes requested):

scripts/model_gen_ng/model_gen_ng/templates/struct_late.tpl
```diff
@@ -0,0 +1,79 @@
...
+[# --- YAML traits --- -#]
+INTROSPECTION_NS([[- struct.user_namespace ]], [[ struct.name ]],
+                [%- for field in struct.all_fields -%]
+                [[- field.name ]][% if not loop.last %], [% endif -%]
+                [%- endfor %])
```
aleclearmind, line 7: OK, I think it's now a good to time unwrap the macro castle in here.

include/revng/Model/Architecture.h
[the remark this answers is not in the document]
fcremo: There are enums for which we don't define any additional method in their namespace -- therefore there's no associated manually written source file. Should I create sources for those enums as well with a content along these lines? It seems a bit silly.
```
#include "eary/thing.h"
#include "late/thing.h"
```

include/revng/Model/CallEdge.h
fcremo: Actually, after we changed the way we handle inheritance, there's no need anymore, it can directly be defined in the autogenerated classes :)

include/revng/Model/FunctionEdge.h
fcremo: Is it possible to generate `fromKey`? I guess I could instanciate a `FunctionEdge` from the key and then see if it is a `CallEdge` or a `FunctionEdge` using `classof`

include/revng/Model/Register.h
fcremo: I created a `RegisterState.h`, now `Register.h` has nothing related to `RegisterState` (both generated and manually written)

include/revng/Model/Type/BaseType.h
fcremo: Not only by generated code, it's used by `model::Binary::getTypePath` and multiple places in `KeyedObjectContainers.ccp`

include/revng/Model/FunctionEdge.h
aleclearmind: Yeah, `classof` enables you to decouple from `isCall`.

scripts/model_gen_ng/metaschema.yml
fcremo: Ops, I copypasted the jsonschema from orchestra to create this one and I accidentally left this snippet. Removed.

scripts/model_gen_ng/model_gen_ng/templates/struct.tpl
fcremo: `docstring` takes care of that, also because `///` has to be injected for each line of the doc

scripts/model_gen_ng/model_gen_ng/clangfmt.py
fcremo: Moved this to `main.py` and implemented soft-fail

scripts/model_gen_ng/model_gen_ng/templates/class_forward_decls.tpl
fcremo: I'm evil-laughing for this terrible but good idea I had. I changed the tokens to
```cpp
/*# This is a comment #*/
// Forward declarations
/** for ns, names in ns_to_names.items() **/
namespace /*= ns =*/ {
/**- for name in names|sort **/
class /*= name =*/;
/**- endfor **/
} // namespace /*= ns =*/
/** endfor **/
```
The tokens for delimiting a "block" and a variable to print have to be different. Also, while just `/*` for output (and `/**` for blocks) works, I find it confusing, so I used `/*=`

scripts/model_gen_ng/model_gen_ng/templates/enum.tpl
fcremo: But enums are not TupleLikes?

scripts/model_gen_ng/model_gen_ng/templates/enum.tpl
fcremo: Premise: as of now, nobody is using `NumMembers`, I created it anticipating that sometime we might need it.
That said, I created a `revng/Model/EnumTraits.h` file with the template definition and instantiated the template for each enum, but isn't wouldn't it be more appropriate to add Name and NumMembers to our own NamedScalarEnumTraits?

scripts/model_gen_ng/model_gen_ng/templates/struct_late.tpl
fcremo: Very good point, added copyright notice and "this is an autogenerated file" notice

lib/Model/Type.cpp
fcremo: Because now NamedTypedRegister is not a subclass of TypedRegister anymore, so the `verify` method body had to be duplicated. Not sure if that was the question, the other interpretation I have is "why is this defined here", but the `verify` method for `TypedRegister` was already defined in `Type.cpp` (and still is).

lib/Model/Type.cpp
fcremo: Changed to
```cpp
model::Type::Type()
  : model::generated::Type(model::TypeKind::Invalid, IDGenerator->get()) { };
```
as agreed OOB

lib/Model/CMakeLists.txt
fcremo: Having it inlined with the C++ is cool, I only have one thing to note, I would not require prefixing every line with `//+`, it's tedious. We could delimit the YAML with something like `/*<marker> <multiline yaml> <marker>*/`, it can be extracted with a regex just as easily.

include/revng/Model/schema.yml
fcremo: The compiler failed to emit `operator==` for some classes, this is not the case anymore, I'll drop this option

include/revng/Model/schema.yml
fcremo: There is no need to include `VerifyHelper.h` and `RecursiveCoroutine.h` anymore (it used to be because I used to declare `verify` in the autogenerated classes). I removed them.
I'm not taking action re: the other includes, because they might become redundant once we move the YAML definitions to their corresponding files (we could use the same includes found in the file or maybe no includes at all as all the necessary headers would be included by the manually written file)

include/revng/Model/schema.yml
fcremo: I managed to remove all defaults from the yaml!
Turns out it possible to initialize all fields like `FieldType Name = FieldType{}`. Unfortunately, some classes (e.g. `MetaAddress`) have an explicit default constructor, so it's not possible to just write `FieldType Name = {}`, hence the requirement to write the field name twice. I could use `auto`, I don't know if you like it or not.
Since it is autogenerated code it does not matter that much one way or the other IMO, and I find the repetition better than using auto, at least all type names can be found in the code at the same column, I find it easier to read.

include/revng/Model/schema.yml
fcremo: I guess you meant **is implicit**; in that case, yes, the `Invalid` field is implicitly created in all enums and now there's no need to specify it as the default value for struct members anymore (see https://github.com/revng/revng/pull/109#discussion_r752350403)

include/revng/Model/schema.yml
aleclearmind: The following should work:
```c++
MetaAddress Asd = MetaAddress {};
```

lib/Model/CMakeLists.txt
aleclearmind: OK, let's give that a shot. `// TupleTree: Model.yml` and `// End TupleTree`? Or maybe no terminator and just keep eating comments until you find a blank line?

lib/Model/Type.cpp
aleclearmind: Do not duplicate, use a common function.

scripts/model_gen_ng/model_gen_ng/templates/enum.tpl
aleclearmind: Yeah, but they're kinda part of the family. The alternative is to make a new trait, these are for YAML stuff.

scripts/model_gen_ng/model_gen_ng/templates/enum.tpl
aleclearmind: Remember that `NamedScalarEnumTraits` are used by the YAML I/O which does not consume these new fields. Not the best place to put them.

include/revng/Model/Binary.h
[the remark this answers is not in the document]
fcremo: The `IsYamlizable` assert can be moved to the autogenerated file, but
```
static_assert(validateTupleTree<model::Binary>(OnlyKOC),
              "Only SortedVectors and MutableSets with YAMLizable keys are "
              "allowed");
```
cannot, because `OnlyKOC` does not apply to all classes. I can add an option to the model YAML, but it's not ideal, what do you think?

include/revng/Model/FunctionEdge.h
fcremo: I tried (with Pietro help) but failed because (likely) our libcpp does not implement `operator<=>` for `std::tuple`

include/revng/Model/schema.yml
fcremo: That's what I did :)

include/revng/Model/schema.yml
fcremo: There are no preambles anymore! 🎉

scripts/model_gen_ng/model_gen_ng/templates/struct_late.tpl
fcremo: Done!

scripts/model_gen_ng/model_gen_ng/struct.py
fcremo: Yeah the boilerplate classes are a bit tedious to write, but the templates are already not  so readable, adding `["propname"]` to each access would only worsen things. Besides, the classes also compute various properties (again, to avoid having logic in the template), and that will need to stay somewhere anyway...

include/revng/Model/Architecture.h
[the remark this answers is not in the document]
fcremo: The YAML is there now, closing

include/revng/Model/FunctionEdge.h
fcremo: Done

include/revng/Model/CallEdge.h
fcremo: Done!

include/revng/Model/CMakeLists.txt
```diff
@@ -0,0 +1,89 @@
...
+    OUTPUT ${GENERATED_HEADERS} ${GENERATED_IMPLS}
+    DEPENDS "${COLLECTED_YAML_PATH}" ${MODEL_TEMPLATES}
+)
+
+add_custom_target(
+    generated-model-classes
+    DEPENDS ${GENERATED_HEADERS} ${GENERATED_IMPLS}
+)
+
+revng_add_analyses_library(revngModelGenerated revng ${GENERATED_IMPLS})
```
aleclearmind, line 89: Why have a distinct library?
Let's put this in the `revngModel` library.

include/revng/Model/CMakeLists.txt
```diff
@@ -0,0 +1,89 @@
...
+)
+
+set(MODEL_TEMPLATES
+    "${CMAKE_SOURCE_DIR}/scripts/tuple_tree_generator/tuple_tree_generator/templates/class_forward_decls.tpl.hpp"
+    "${CMAKE_SOURCE_DIR}/scripts/tuple_tree_generator/tuple_tree_generator/templates/enum.tpl.hpp"
+    "${CMAKE_SOURCE_DIR}/scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.tpl.hpp"
+    "${CMAKE_SOURCE_DIR}/scripts/tuple_tree_generator/tuple_tree_generator/templates/struct_forward_decls.tpl.hpp"
+    "${CMAKE_SOURCE_DIR}/scripts/tuple_tree_generator/tuple_tree_generator/templates/struct_late.tpl.hpp"
+    "${CMAKE_SOURCE_DIR}/scripts/tuple_tree_generator/tuple_tree_generator/templates/struct_impl.tpl.cpp"
+)
```
aleclearmind, line 49:
```suggestion
set(TEMPLATES_DIR "${CMAKE_SOURCE_DIR}/scripts/tuple_tree_generator/tuple_tree_generator/templates")
set(MODEL_TEMPLATES
    "${TEMPLATES_DIR}/class_forward_decls.tpl.hpp"
    "${TEMPLATES_DIR}/enum.tpl.hpp"
...
)
```

include/revng/Model/CMakeLists.txt
```diff
@@ -0,0 +1,89 @@
...
+
+# -- Collect YAML from manually written headers
+set(COLLECTED_YAML_PATH "${CMAKE_BINARY_DIR}/include/revng/Model/model_schema.yml")
+
+add_custom_command(
+    OUTPUT "${COLLECTED_YAML_PATH}"
+    COMMAND "${CMAKE_SOURCE_DIR}/scripts/tuple_tree_generator/extract_yaml.py"
+    ARGS "MODEL-YAML" ${MODEL_HEADERS} > "${COLLECTED_YAML_PATH}"
+    WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
```
aleclearmind, line 58: Needed?

include/revng/Model/CMakeLists.txt
```diff
@@ -0,0 +1,89 @@
...
+    "${CMAKE_SOURCE_DIR}/scripts/tuple_tree_generator/tuple_tree_generator/templates/struct_impl.tpl.cpp"
+)
+
+# -- Collect YAML from manually written headers
+set(COLLECTED_YAML_PATH "${CMAKE_BINARY_DIR}/include/revng/Model/model_schema.yml")
+
+add_custom_command(
+    OUTPUT "${COLLECTED_YAML_PATH}"
+    COMMAND "${CMAKE_SOURCE_DIR}/scripts/tuple_tree_generator/extract_yaml.py"
+    ARGS "MODEL-YAML" ${MODEL_HEADERS} > "${COLLECTED_YAML_PATH}"
```
aleclearmind, line 57:
```suggestion
    "MODEL-YAML" ${MODEL_HEADERS} > "${COLLECTED_YAML_PATH}"
```

include/revng/Model/CMakeLists.txt
```diff
@@ -0,0 +1,89 @@
...
+  set(IMPL_OUTPUT "${OUTPUT_DIR}/Impl/${HEADER_FILENAME_WE}.cpp")
+  list(APPEND GENERATED_HEADERS "${EARLY_OUTPUT}")
+  list(APPEND GENERATED_HEADERS "${LATE_OUTPUT}")
+  list(APPEND GENERATED_IMPLS "${IMPL_OUTPUT}")
+endforeach()
+list(APPEND GENERATED_HEADERS "${OUTPUT_DIR}/ForwardDecls.h")
+
+add_custom_command(
+    COMMAND "${CMAKE_SOURCE_DIR}/scripts/tuple_tree_generator/main.py"
+    ARGS --cpp-headers "${COLLECTED_YAML_PATH}" "${OUTPUT_DIR}"
```
aleclearmind, line 79: Would `python ${CMAKE_SOURCE_DIR}/scripts/tuple_tree_generator` work?

include/revng/Model/CallEdge.h
```diff
@@ -0,0 +1,54 @@
...
+doc: A CFG edge to represent function calls (direct, indirect and tail calls)
+type: struct
+inherits: FunctionEdge
+fields:
+  - name: Prototype
+    doc: |
+      Path to the prototype for this call site
+
+      In case of a direct function call, it has to be invalid.
+    # TODO: how can we handle this?
```
aleclearmind, line 21: Handle what?

include/revng/Model/CallEdge.h
```diff
@@ -0,0 +1,54 @@
...
+//
+
+#include "revng/Model/AbstractFunctionEdge.h"
+#include "revng/Model/VerifyHelper.h"
+
+/* MODEL-YAML
+name: CallEdge
+doc: A CFG edge to represent function calls (direct, indirect and tail calls)
+type: struct
+inherits: FunctionEdge
```
aleclearmind, line 14: This document has two `inherits` entries

include/revng/Model/CallEdge.h
```diff
@@ -0,0 +1,54 @@
...
+    # TODO: how can we handle this?
+    reference:
+      pointeeType: "model::Type"
+      rootType: "model::Binary"
+  - name: DynamicFunction
+    doc: |
+      Name of the dynamic function being called, or empty if not a dynamic call
+    type: "std::string"
+    optional: true
+inherits: AbstractFunctionEdge
```
aleclearmind, line 30: Move up

include/revng/Model/FunctionEdge.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include "revng/Model/AbstractFunctionEdge.h"
+#include "revng/Model/VerifyHelper.h"
+
+/* MODEL-YAML
+name: FunctionEdge
+doc: An edge on the CFG
+type: struct
+fields: []
+inherits: AbstractFunctionEdge
```
aleclearmind, line 15: Let's keep `inherits`s to the top, it's important.

include/revng/Model/RegisterState.h
```diff
@@ -0,0 +1,34 @@
...
+//
+
+/* MODEL-YAML
+name: RegisterState
+type: enum
+members:
+  - name: "No"
+  - name: NoOrDead
+  - name: Dead
+  - name: "Yes"
```
aleclearmind, line 14: `No` and `Yes` being booleans is the dumbest thing in the world.

include/revng/Support/revng.h
```diff
@@ -14,7 +14,7 @@
 #include "llvm/ADT/Triple.h"
 #include "llvm/BinaryFormat/ELF.h"
 
-#include "revng/Model/ABI.h"
+#include "revng/Model/Generated/Early/ABI.h"
```
aleclearmind, line 17: Mmmmmmmmh

lib/DwarfImporter/DwarfImporter.cpp
```diff
@@ -745,7 +746,7 @@ class DwarfToModelConverter {
     using namespace model;
 
     // Create function type
-    UpcastableType NewType = makeType<CABIFunctionType>();
+    UpcastablePointer<model::Type> NewType = makeType<CABIFunctionType>();
```
aleclearmind, line 749: Why?

scripts/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+# Check that required python dependencies are available
```
aleclearmind, line 1: License notice?

scripts/requirements.txt
```diff
@@ -0,0 +1,11 @@
+# -- Requirements for tuple_tree_generator
```
aleclearmind, line 1: License notice

scripts/tuple_tree_generator/metaschema.yml
```diff
@@ -0,0 +1,100 @@
+# This file is distributed under the MIT License. See LICENSE.md for details.
```
aleclearmind, line 1:
```suggestion
#
# This file is distributed under the MIT License. See LICENSE.md for details.
#
```

scripts/tuple_tree_generator/metaschema.yml
```diff
@@ -0,0 +1,100 @@
...
+
+"$schema": http://json-schema.org/draft-07/schema#
+"$ref": "#/definitions/Schema"
+definitions:
```
aleclearmind, line 5: Can we enforce the order of entries?

scripts/tuple_tree_generator/tuple_tree_generator/definition/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .definition import Definition
```
aleclearmind, line 1: Please put license notice *everywhere*.

```
#
#
```

aleclearmind (changes requested):

scripts/tuple_tree_generator/metaschema.yml
fcremo: You mean the order in which fields are declared? I don't think it's possible to do that, no

scripts/tuple_tree_generator/tuple_tree_generator/definition/__init__.py
fcremo: We really need a check for the license notice in `check-revng-conventions`

include/revng/Model/Binary.h
```diff
@@ -6,601 +6,67 @@
...
+    doc: List of segments in the original binary
+    array:
+      type: SortedVector
+      elementType: model::Segment
+  - name: EntryPoint
+    doc: Program entry point
+    type: MetaAddress
+  - name: Types
+    doc: The type system
+    # TODO: recognize this is upcastable
```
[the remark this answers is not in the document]
fcremo: I meant that it is possible to parse `model::Type` and drop the need to specify `upcastable: true`. I'm not sure we want to do it, because maybe you do want a list of `model::Type` and not a list of `UpcastablePointer<model::Type>`.

include/revng/Model/CMakeLists.txt
fcremo: Because `revng_add_analyses_library` must be invoked only once, it cannot add files incrementally to a given library (it's not a limitation of `revng_add_analyses_library` itself, it uses CMake builtin `add_library`, which has this limitation).
There are two workarounds as far as I know:
- having a global list of files which have to be added to the model
- moving the CMake stuff for the autogenerated model where the "main" model library is defined. I didn't do this because I found having the CMake that generates stuff from the model header files in their folder much clearer

Which option do you prefer? Personally I'm for option 2 if we must do away with the second library.

fcremo: (settled IRL for option 2)

include/revng/Model/CMakeLists.txt
fcremo: I have to rework the package a bit (to turn `main.py` into `__main__.py`) but yeah, it can work

fcremo: Scratch that, we would need to add the module to PYTHONPATH, which is even uglier. I'd leave it the way it is.

include/revng/Model/CallEdge.h
fcremo: It's a leftover TODO, marking that I needed to figure out how to handle `TupleTreeReference`s

include/revng/Model/CallEdge.h
fcremo: And python YAML parser apparently has no issue with it, I'm mad

include/revng/Model/RegisterState.h
fcremo: Agree :(

include/revng/Support/revng.h
fcremo: Ops, leftover from early times when enums did not have a manually written associated class, replaced with `#include "revng/Model/ABI.h"`

lib/DwarfImporter/DwarfImporter.cpp
fcremo: At some point it was not possible to have a `using UpcastableType = UpcastablePointer<model::Type>` in a place which made sense, so I expanded it manually. I'm not 100% sure what the problem was, but now it's solved

include/revng/Model/CMakeLists.txt
fcremo: Yes, first I need to collect all the YAML definitions from the headers, then those get processed to generate the C++ and the JSON schema.

include/revng/Model/AllTypes.h
```diff
@@ -0,0 +1,38 @@
...
+} // end namespace model
+
+template<typename K, typename V>
+V getOrDefault(const std::map<K, V> &Map, const K &Key, const V &Default) {
+  auto It = Map.find(Key);
+  if (It == Map.end())
+    return Default;
+  else
+    return It->second;
+}
```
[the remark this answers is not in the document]
fcremo: Before it was in `Type.h` which contained all definitions for `Type` and derived types, now it only contains the definition for `Type` and `AllTypes.h` is the one including them all, so I thought it was more appropriate to move it in `AllTypes.h`

scripts/tuple_tree_generator/tuple_tree_generator/definition/__init__.py
aleclearmind: Makes sense, but maybe not now.
Feel free to open an issue for this.

include/revng/Model/Binary.h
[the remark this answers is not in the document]
aleclearmind: Mmmh, having `SortedVector<UpcastablePointer<model::Type>>` vs `SortedVector<model::Type>` is an orthogonal problem.

I think yo use `upcastable: true` to emit some code to make this usable into an `UpcastablePointer`, isn't it?
If so, we could as well infer `upcastable: true` automatically if there's at least a `struct` inheriting it.
What do you think?

include/revng/Model/CMakeLists.txt
aleclearmind: I think I meant `WORKING_DIRECTORY`. Let's avoid using the cwd unless necessary.

include/revng/Model/AllTypes.h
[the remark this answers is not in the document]
aleclearmind: Who is using it? Push it as close as possible to its users.
This is a very generic method, not specific to the model at all, could stay in `STLExtras.h` as well.

scripts/tuple_tree_generator/metaschema.yml
fcremo: I didn't know about `propertiesOrder`, json-schema.org does not seem to mention it, neither do a number of python jsonschema validators. Seems it was not implemented. I still added it to the schema, so if one day it will be implemented we'll have this additional guarantee.

aleclearmind: Uhm, if it's not enforceable maybe it's better to leave it out, what do you say?
If you want to keep it, make sure things are in a sensible order (e.g., inherit is important and should stay on top).

include/revng/Model/Binary.h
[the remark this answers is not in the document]
fcremo: Wait, `upcastable: true` is a property of the list
```yaml
  - name: Types
    doc: The type system
    # TODO: recognize this is upcastable
    list:
      type: SortedVector
      upcastable: true
      elementType: model::Type
```
which I use to tell wether the type of an element of the list is `UpcastablePointer<T>` or just `T`

include/revng/Model/CMakeLists.txt
fcremo: Fair, removed

include/revng/Model/Binary.h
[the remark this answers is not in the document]
aleclearmind: Ah sorry, my bad, ENOTENOUGHCONTEXT.
OK, then I'm happy with `upcastable` here, just drop the TODO.
Maybe we could make `upcastable: true` less C++-specific and say `accept_inheritors: true` or `accept_derived_structs: true`.

scripts/model_gen_ng/model_gen_ng/jinja_utils.py
fcremo: See the `python/CMakeLists.txt` (which uses a `requirements.txt` in the same folder)

scripts/tuple_tree_generator/metaschema.yml
fcremo: On second thought, it does not make a lot of sense to have something we don't enforce, I'm going to drop it

include/revng/Model/Binary.h
```diff
@@ -6,601 +6,67 @@
...
-class Segment;
-} // namespace model
+/* MODEL-YAML
+name: Binary
+doc: Data structure representing the whole binary
+type: struct
+fields:
+  - name: Functions
+    doc: List of the functions within the binary
+    array:
```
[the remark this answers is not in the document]
fcremo: I propose `collection` instead, because we also have sets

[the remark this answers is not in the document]
aleclearmind: YAML calls them [sequence](https://yaml.org/spec/1.2.2/#sequence). Shall we go for that?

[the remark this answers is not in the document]
fcremo: I like it less, because sequence makes me think there's a defined iteration order, which is not necessarily true for sets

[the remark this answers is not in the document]
aleclearmind: Mmmh, actually in model data structures the order is *always* well defined.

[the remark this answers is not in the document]
fcremo: Renamed to `sequence`

tests/unit/ModelType.cpp
```diff
@@ -4,13 +4,13 @@
...
 #include <bit>
 
-#include "revng/Model/ABI.h"
 #define BOOST_TEST_MODULE ModelType
 bool init_unit_test();
 #include "boost/test/unit_test.hpp"
 
 #include "revng/Model/Binary.h"
-#include "revng/Model/Type.h"
+#include "revng/Model/Generated/Early/ABI.h"
```
aleclearmind, line 12: Drop `/Generated`

aleclearmind:
The "Generate C++ model" commit message is very laconic, can you expand on what's going on?

scripts/tuple_tree_generator/tuple_tree_generator/__init__.py
```diff
@@ -0,0 +1 @@
+# This file is distributed under the MIT License. See LICENSE.md for details.
```
aleclearmind, line 1: Everywhere
```suggestion
#
# This file is distributed under the MIT License. See LICENSE.md for details.
#
```

scripts/requirements.txt
```diff
@@ -0,0 +1,12 @@
...
+PyYAML
+Jinja2
+# Needed for python < 3.9
+graphlib-backport
+
+# -- Requirements for revng_merge_dynamic
+pyelftools
+
+# -- Requirements for monotone-framework
+pygraphviz
```
aleclearmind, line 12: Due to SA2:
```suggestion
networkx
pydot
```

Can you please make a companion PR for orchestra adding jinja2 as a dependency?

scripts/tuple_tree_generator/tuple_tree_generator/definition/defaults.py
```diff
@@ -0,0 +1,5 @@
+# This file is distributed under the MIT License. See LICENSE.md for details.
+
+DEFAULT_STRUCT_NAMESPACE = "model::generated"
+DEFAULT_ENUM_NAMESPACE = "model"
+DEFAULT_USER_NAMESPACE = "model"
```
aleclearmind, line 5: Can we have a test for a non-model tuple-tree-like data structure?
We have plans to take certain things out of the model, and having a simple example on how to create such a thing would help a lot.

aleclearmind (changes requested):

scripts/requirements.txt
fcremo: https://github.com/revng/orchestra/pull/62/

scripts/tuple_tree_generator/tuple_tree_generator/definition/defaults.py
fcremo: I refactored the cmake invoking `tuple_tree_generator`, added a test that does a round-trip (object1 -> yaml -> object2, assert object1 == object2) and can serve as an example for most things `tuple_tree_generator` can do.
Also, now `tuple_tree_generator` takes the namespace to use as a parameter.

scripts/model_gen_ng/model_gen_ng/templates/enum.tpl
aleclearmind: Updates on this matter?

scripts/tuple_tree_generator/tuple_tree_generator/templates/enum.tpl.hpp
```diff
@@ -0,0 +1,75 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+// This file is autogenerated! Do not edit it directly
+
+#include "revng/Model/EnumTraits.h"
+#include "revng/TupleTree/TupleTree.h"
+
+namespace /*= enum.namespace =*/ {
+    /*= enum.doc | docstring =*/
```
aleclearmind, line 17: * Do not indent for namespaces.
* Indent with two spaces only.

scripts/model_gen_ng/model_gen_ng/templates/enum.tpl
aleclearmind: I'd say just add a `constexpr StringRef name() { return "asdf"; }` in the `enum` namespace and use that.

We can switch to a trait if we need to. It'd look something like:

```c++
// In some header
template<typename Enum>
StringRef getEnumName();
```
```c++
// In the generated code for the enum
template<>
StringRef getEnumName<MyEnum>() { return "MyEnum"; }
```

aleclearmind:
Merged in b92d99e! Thanks!

PR #110, opened 2021-11-22
TupleTree: make Root a `std::variant<RootT *, const RootT*>`

pfez:
This enables holding `TupleTreeReference`s to immutable model objects.

The patchset includes a few other changes that are necessary for enabling `TupleTreeReference`s to immutable model objects, or for improving their ergonomics.

This branch has already been tested, and checked for style issues.

### DIFF
--- a/include/revng/TupleTree/TupleTree.h
+++ b/include/revng/TupleTree/TupleTree.h
@@ -7,6 +7,7 @@
 #include <array>
 #include <set>
 #include <type_traits>
+#include <variant>
 #include <vector>
 
 #include "llvm/ADT/ArrayRef.h"
@@ -401,7 +402,7 @@ bool callOnPathSteps(Visitor &V,
   if (It == M.end())
     return false;
 
-  value_type *Matching = &*It;
+  auto *Matching = &*It;
 
   V.template visitContainerElement<RootT>(TargetKey, *Matching);
   if (Path.size() > 1) {
@@ -1009,24 +1010,29 @@ struct std::tuple_size<T>
 template<TupleTreeCompatible T>
 class TupleTree;
 
-template<typename T, typename RootT>
+template<typename T, typename R>
+concept ConstOrNot = std::is_same_v<R, T> or std::is_same_v<const R, T>;
+
+template<typename T, typename RootType>
 class TupleTreeReference {
 public:
-  using pointee = T;
+  using RootT = RootType;
+  using RootVariant = std::variant<RootT *, const RootT *>;
 
 public:
-  RootT *Root = nullptr;
+  RootVariant Root = static_cast<RootT *>(nullptr);
   TupleTreePath Path;
 
 public:
-  static TupleTreeReference fromPath(RootT *Root, const TupleTreePath &Path) {
-    TupleTreeReference Result;
-    Result.Root = Root;
-    Result.Path = Path;
-    return Result;
+  static TupleTreeReference
+  fromPath(ConstOrNot<TupleTreeReference::RootT> auto *Root,
+           const TupleTreePath &Path) {
+    return TupleTreeReference{ .Root = RootVariant{ Root }, .Path = Path };
   }
 
-  static TupleTreeReference fromString(RootT *Root, llvm::StringRef Path) {
+  static TupleTreeReference
+  fromString(ConstOrNot<TupleTreeReference::RootT> auto *Root,
+             llvm::StringRef Path) {
     std::optional<TupleTreePath> OptionalPath = stringAsPath<RootT>(Path);
     if (not OptionalPath.has_value())
       return TupleTreeReference{};
@@ -1043,26 +1049,43 @@ public:
 
   const TupleTreePath &path() const { return Path; }
 
-  T *get() {
-    revng_check(Root != nullptr);
-
-    if (Path.size() == 0)
-      return nullptr;
-
-    return getByPath<T>(Path, *Root);
+private:
+  bool hasNullRoot() const {
+    const auto IsNullVisitor = [](const auto &Pointer) {
+      return Pointer == nullptr;
+    };
+    return std::visit(IsNullVisitor, Root);
   }
 
+  bool canGet() const {
+    return Root.index() != std::variant_npos and not hasNullRoot();
+  }
+
+public:
   const T *get() const {
-    revng_check(Root != nullptr);
+    revng_assert(canGet());
 
     if (Path.size() == 0)
       return nullptr;
 
-    return getByPath<T>(Path, *Root);
+    const auto GetByPathVisitor = [&Path = Path](const auto &RootPointer) {
+      return getByPath<T>(Path, *RootPointer);
+    };
+
+    return std::visit(GetByPathVisitor, Root);
+  }
+
+  T *get() {
+    revng_assert(canGet());
+
+    if (Path.size() == 0)
+      return nullptr;
+
+    return getByPath<T>(Path, *std::get<RootT *>(Root));
   }
 
   bool isValid() const debug_function {
-    return (*this != TupleTreeReference() and get() != nullptr);
+    return not hasNullRoot() and not Path.empty() and get() != nullptr;
   }
 };
 
@@ -1079,7 +1102,7 @@ struct llvm::yaml::ScalarTraits<T> {
   static llvm::StringRef input(llvm::StringRef Path, void *, T &Obj) {
     // We temporarily initialize Root to nullptr, a post-processing phase will
     // take care of fixup these
-    Obj = T::fromString(nullptr, Path);
+    Obj = T::fromString(static_cast<typename T::RootT *>(nullptr), Path);
     return {};
   }
 
@@ -1205,7 +1228,13 @@ private:
     bool Result = true;
 
     visitReferences([&Result, this](const auto &Element) {
-      Result = Result and (Element.Root == Root.get());
+      const auto SameRoot = [&]() {
+        const auto GetPtrToConstRoot =
+          [](const auto &RootPointer) -> const T * { return RootPointer; };
+
+        return std::visit(GetPtrToConstRoot, Element.Root) == Root.get();
+      };
+      Result = Result and SameRoot();
     });
 
     return Result;
@@ -1244,3 +1273,16 @@ struct NamedEnumScalarTraits {
     }
   }
 };
+
+/// \brief Specialization for the std::variant we have in TupleTreeReference
+template<bool X, typename T>
+inline void writeToLog(Logger<X> &This,
+                       const std::variant<T *, const T *> &Var,
+                       int Ignored) {
+  if (Var.index() == std::variant_npos)
+    writeToLog(This, llvm::StringRef("std::variant_npos"), Ignored);
+  else if (std::holds_alternative<T *>(Var))
+    writeToLog(This, std::get<T *>(Var), Ignored);
+  else if (std::holds_alternative<const T *>(Var))
+    writeToLog(This, std::get<const T *>(Var), Ignored);
+}
--- a/include/revng/TupleTree/TupleTree.h
+++ b/include/revng/TupleTree/TupleTree.h
@@ -25,7 +25,7 @@
 template<typename T>
 concept TupleTreeCompatible = IsKeyedObjectContainer<T>
                                 or HasTupleSize<T>
-                                or IsUpcastablePointer<T>;
+                                or UpcastablePointerLike<T>;
 // clang-format on
 
 template<typename T>
--- a/include/revng/TupleTree/TupleTree.h
+++ b/include/revng/TupleTree/TupleTree.h
@@ -1084,6 +1084,8 @@ public:
     return getByPath<T>(Path, *std::get<RootT *>(Root));
   }
 
+  const T *getConst() const { return get(); }
+
   bool isValid() const debug_function {
     return not hasNullRoot() and not Path.empty() and get() != nullptr;
   }
--- a/lib/Model/Binary.cpp
+++ b/lib/Model/Binary.cpp
@@ -131,6 +131,13 @@ Binary::getPrimitiveType(PrimitiveTypeKind::Values V, uint8_t ByteSize) {
   return getTypePath(It->get());
 }
 
+model::TypePath
+Binary::getPrimitiveType(PrimitiveTypeKind::Values V, uint8_t ByteSize) const {
+  PrimitiveType Temporary(V, ByteSize);
+  Type::Key PrimitiveKey{ TypeKind::Primitive, Temporary.ID };
+  return getTypePath(Types.at(PrimitiveKey).get());
+}
+
 TypePath Binary::recordNewType(UpcastablePointer<Type> &&T) {
   auto It = Types.insert(T).first;
   return getTypePath(It->get());
--- a/lib/Model/Type.cpp
+++ b/lib/Model/Type.cpp
@@ -728,9 +728,6 @@ verifyImpl(VerifyHelper &VH, const StructType *T) {
   if (T->Size == 0)
     rc_return VH.fail("Struct type has zero size", *T);
 
-  if (T->Fields.empty())
-    rc_return VH.fail("Struct has no fields", *T);
-
   size_t Index = 0;
   llvm::SmallSet<llvm::StringRef, 8> Names;
   auto FieldIt = T->Fields.begin();
--- a/lib/Model/Type.cpp
+++ b/lib/Model/Type.cpp
@@ -443,8 +443,7 @@ static VoidConstResult isVoidConst(const QualifiedType *QualType) {
       // We know that it's const-qualified here, and it only has one
       // qualifier, hence we can skip the const-qualifier.
       Result.IsConst = true;
-      if (not QualType->UnqualifiedType.Root)
-        return Result;
+      return Result;
     }
 
     UnqualType = QualType->UnqualifiedType.get();
--- a/lib/Model/Type.cpp
+++ b/lib/Model/Type.cpp
@@ -945,8 +945,8 @@ RecursiveCoroutine<bool> Type::verify(VerifyHelper &VH) const {
   rc_return VH.maybeFail(Result);
 }
 
-void QualifiedType::dump() const {
-  serialize(dbg, *this);
+void QualifiedType::dump(llvm::raw_ostream &OS) const {
+  serialize(OS, *this);
 }
 
 bool QualifiedType::verify() const {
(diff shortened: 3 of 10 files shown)


include/revng/Model/TupleTree.h
```diff
@@ -1205,7 +1228,14 @@ class TupleTree {
     bool Result = true;
 
     visitReferences([&Result, this](const auto &Element) {
-      Result = Result and (Element.Root == Root.get());
+      const auto SameRoot = [&]() {
+        const auto GetPtrToConstRoot = [](const auto &RootPointer) {
+          return const_cast<const T *>(RootPointer);
```
aleclearmind, line 1233: I don't think you need a `const_cast` to *add* const-ness.
Let's not raise unnecessary suspicion to the reader with a `const_cast`.

lib/Model/Type.cpp
```diff
@@ -443,7 +443,7 @@ static VoidConstResult isVoidConst(const QualifiedType *QualType) {
       // We know that it's const-qualified here, and it only has one
       // qualifier, hence we can skip the const-qualifier.
       Result.IsConst = true;
-      if (not QualType->UnqualifiedType.Root)
+      if (QualType->UnqualifiedType.hasNullRoot())
         return Result;
```
aleclearmind, line 447: Can we do something like `.get() == nullptr`?

lib/Model/Binary.cpp
```diff
@@ -86,6 +86,16 @@ Binary::getPrimitiveType(PrimitiveTypeKind::Values V, uint8_t ByteSize) {
   return getTypePath(It->get());
 }
 
+model::TypePath
+Binary::getPrimitiveType(PrimitiveTypeKind::Values V, uint8_t ByteSize) const {
+  PrimitiveType Temporary(V, ByteSize);
+  Type::Key PrimitiveKey{ TypeKind::Primitive, Temporary.ID };
+  auto It = Types.find(PrimitiveKey);
```
aleclearmind, line 93: I think you can use `Types.at`

include/revng/Model/TupleTree.h
```diff
@@ -25,7 +26,7 @@
 template<typename T>
 concept TupleTreeCompatible = IsKeyedObjectContainer<T>
                                 or HasTupleSize<T>
-                                or IsUpcastablePointer<T>;
+                                or UpcastablePointerLike<T>;
```
aleclearmind, line 29: Why?
Is this a self-contained change that we can isolate in a commit?

aleclearmind (changes requested):

include/revng/Model/TupleTree.h
```diff
@@ -1043,22 +1056,32 @@ class TupleTreeReference {
...
-    revng_check(Root != nullptr);
+    revng_check(Root.index() != std::variant_npos);
+    revng_check(not hasNullRoot());
 
     if (Path.size() == 0)
       return nullptr;
 
-    return getByPath<T>(Path, *Root);
+    const auto GetByPathVisitor =
+      [&Path = std::as_const(Path)](const auto &RootPointer) {
```
aleclearmind, line 1072: `std::as_const`, interesting. But is this really necessary or is it just for the sake of it?

include/revng/Model/TupleTree.h
pfez: Yup, we can. Before this change it was too strict (`IsUpcastablePointer` is stricter than `UpcastablePointerLike`), but it never failed nevertheless, because we didn't have any instance that was `UpcastablePointerLike` but not `IsUpcastablePointer`. Now it's needed but I'll move it into a separate commit.

include/revng/Model/TupleTree.h
```diff
@@ -1043,22 +1056,32 @@ class TupleTreeReference {
...
-    revng_check(Root != nullptr);
-
-    if (Path.size() == 0)
-      return nullptr;
-
-    return getByPath<T>(Path, *Root);
+  bool hasNullRoot() const {
+    const auto IsNullVisitor = [](const auto &Ptr) { return Ptr == nullptr; };
+    return std::visit(IsNullVisitor, Root);
   }
```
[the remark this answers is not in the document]
pfez: I wanted to make it `private`, but at the moment it's used in `lib/Model/Type.cpp`, like you noted in another comment below. If I can fix the other comment I'd be more than happy to make this `private`.

lib/Model/Type.cpp
pfez: Eh, I'd love to, but the previous implementation of `get()` asserts that `Root` is not null. So we'd get an error on `get()` here.

pfez: I could change `get()` not to assert, though.

include/revng/Model/TupleTree.h
```diff
@@ -1043,22 +1056,32 @@ class TupleTreeReference {
...
+
+    return std::visit(GetByPathVisitor, Root);
+  }
+
+  T *get() {
+    // Strict check that we're not holding the const alternative
+    revng_check(not std::holds_alternative<const RootT *>(Root));
+    // Then just call the const version and cast constness away.
+    auto *const_this = const_cast<const TupleTreeReference<T, RootT> *>(this);
+    return const_cast<T *>(const_this->get());
```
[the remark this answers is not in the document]
pfez: Not really. If anything, we could do:
```C++
return getByPath<T>(Path, std::get<RootT *>(Root));
```

But in this way we wouldn't be sharing the implementation with the other `get()` method, which I thought was kinda nice because it guarantees they do the same thing.

But yes, the `const_cast` dance is not really nice to see. I can go with the other version up here if you prefer it.

include/revng/Model/TupleTree.h
```diff
@@ -1013,17 +1014,29 @@ template<typename T, typename RootT>
 class TupleTreeReference {
 public:
   using pointee = T;
+  using root_t = RootT;
```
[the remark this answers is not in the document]
pfez: Done

include/revng/Model/TupleTree.h
```diff
@@ -1043,22 +1056,32 @@ class TupleTreeReference {
...
 
-  T *get() {
-    revng_check(Root != nullptr);
-
-    if (Path.size() == 0)
-      return nullptr;
-
-    return getByPath<T>(Path, *Root);
+  bool hasNullRoot() const {
+    const auto IsNullVisitor = [](const auto &Ptr) { return Ptr == nullptr; };
```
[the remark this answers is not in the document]
pfez: You mean you don't like `Ptr`? I changed it to `Pointer`. Anything else?

include/revng/Model/TupleTree.h
```diff
@@ -1079,7 +1102,7 @@ struct llvm::yaml::ScalarTraits<T> {
   static llvm::StringRef input(llvm::StringRef Path, void *, T &Obj) {
     // We temporarily initialize Root to nullptr, a post-processing phase will
     // take care of fixup these
-    Obj = T::fromString(nullptr, Path);
+    Obj = T::fromString((typename T::root_t *)nullptr, Path);
```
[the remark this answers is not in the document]
pfez: Done, switch to `static_cast`.

lib/Model/Binary.cpp
pfez: Done.

include/revng/Model/TupleTree.h
pfez: Done

include/revng/Model/TupleTree.h
```diff
@@ -1013,17 +1014,29 @@ template<typename T, typename RootT>
 class TupleTreeReference {
 public:
   using pointee = T;
+  using root_t = RootT;
+  using root_variant = std::variant<RootT *, const RootT *>;
 
 public:
-  RootT *Root = nullptr;
+  root_variant Root = (RootT *) nullptr;
```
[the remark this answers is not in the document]
pfez: Done.

include/revng/Model/TupleTree.h
pfez: If you don't explicitly use `const_cast` here, the `GetPtrToConstRoot` lambda will not have a unique return type for the various types of `RootPointer`. Here's why.

The generic lambda here, without the `const_cast`, expands to something like:
```c++
template<typename T>
auto GetPtrToConstRoot(const T& RootPointer) {
  return RootPointer;
}
```

Given that we are using this to visit a `std::variant<RootT *, const Root *T>`, when substituting the two different types of the variant, it will return a `RootT *` in a case and a `const RootT *` in another case.

This does not work with `std::visit` because it requires the visitor to have a single return type. Hence I needed to explicitly add the `const_cast`, so that all the instantiations of the generic lambda on the various types in the variant will return the same type.

If I don't do that, the code does not compile:
```
error: static_assert failed due to requirement 'integral_constant<bool, false>::value' "`std::visit` requires the visitor to have a single return type."
```

Maybe there's a cleaner way out that does not require `const_cast`. If so, feel free to suggest it!

include/revng/Model/TupleTree.h
pfez: This was just for the sake of it. Should I drop it?

include/revng/Model/TupleTree.h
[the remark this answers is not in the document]
pfez: In the meantime, to explore our options, I tried to go also with this:
```c++
    T *get() {
      revng_check(Root.index() != std::variant_npos);
      revng_check(not hasNullRoot());

      if (Path.size() == 0)
        return nullptr;

      return getByPath<T>(Path, std::get<RootT *>(Root));
    }
```
and I get assertions from `DwarfImporter` for trying to use `llvm::isa` on a null pointer.

This is slightly puzzling, I'm investigating it.

include/revng/Model/TupleTree.h
aleclearmind: I was just suggesting to use `static_cast`.
But you can also specify the return type of the lambda with something like `-> const decltype(RootPointer)`.
You can also make add an explicit template parameter list to the lambda if you prefer, but I'd just go with a `static_cast`.

lib/Model/Type.cpp
aleclearmind: Mmmh, I'm not sure why here we're checking if `Root` is `nullptr`.
`Root` is `nullptr` only during deserialization, in all other cases it's not a valid state.
I think we can just drop this check, it's probably a leftover of a darker past.

include/revng/Model/TupleTree.h
aleclearmind: I'd keep it simple and drop it.

lib/Model/Type.cpp
pfez: Could be. I'm going to drop this, and make `hasNullRoot` private then.

include/revng/Model/TupleTree.h
pfez: Oh, I see what you mean. I'm going for `-> const T *` then, which is more concise and exactly what we want.

include/revng/Model/TupleTree.h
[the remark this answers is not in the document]
pfez: Ok it was just a bug of mine. I'll leave the `const_cast` dance out then, and duplicate a little bit of code.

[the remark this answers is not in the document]
pfez: Done, take a look!

include/revng/Model/TupleTree.h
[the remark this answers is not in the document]
pfez: Done

include/revng/Model/TupleTree.h
pfez: Done, take a look

lib/Model/Type.cpp
pfez: Done

include/revng/Model/TupleTree.h
pfez: Done

include/revng/Model/TupleTree.h
aleclearmind: Looks great!

lib/Model/Type.cpp
aleclearmind: > there are valid situations where we need to have some `QualifiedTypes` with null roots for a while

What are those? `TupleTreeReference`s were not designed for this.

AlviseDeFaveri: Since the type system can have loops (consider pointers), the type creation pass does the following: first, it generates all types except pointers, then it updates all the fields corresponding to pointer with the right type. Between the two phases, the pointer fields exist in the model, but they do not have a valid `UnqualifiedType` yet, and I want to check that that is the case. In some cases, an additional wrapper might have been generated for the pointer, so I need to descend the wrapper structures until I find one with an empty `UnqualifiedType`, but that is an implementation detail that can be worked around.

AlviseDeFaveri: TL;DR I want to check if a field of a struct/union has a type that has yet to be decided, so I can use whatever mechanism is exposed at a higher level (e.g. maybe check if the Type is Invalid?)

aleclearmind: Sounds like you can use `TupleTreeReference::isValid`.
Having a valid path but not a valid root (which is what we're discussing here) makes sense only if upon creation of the `TupleTreeRereference` the root (`model::Binary`) has not been instantiated yet.
But this doesn't seem to be the case.

AlviseDeFaveri: Agreed

pfez: I tried this with `isValid` and it works. Thanks @AlviseDeFaveri for the clarification. I'm going to close this conversation.

lib/Model/Type.cpp
```diff
@@ -786,28 +782,44 @@ verifyImpl(VerifyHelper &VH, const StructType *T) {
...
   for (auto &Group : llvm::enumerate(T->Fields)) {
     auto &Field = Group.value();
     uint64_t ExpectedIndex = Group.index();
 
-    if (Field.Index != ExpectedIndex)
-      rc_return VH.fail();
+    if (Field.Index != ExpectedIndex) {
+      using llvm::Twine;
+      rc_return VH.fail(Twine("Union type is missing field ")
+                        + Twine(ExpectedIndex));
```
aleclearmind, line 801:
```suggestion
      rc_return VH.fail(Twine("Union type is missing field ")
                        + Twine(ExpectedIndex), *T);
```

lib/Model/Type.cpp
```diff
@@ -786,28 +782,44 @@ verifyImpl(VerifyHelper &VH, const StructType *T) {
...
 
   llvm::SmallSet<llvm::StringRef, 8> Names;
   for (auto &Group : llvm::enumerate(T->Fields)) {
     auto &Field = Group.value();
     uint64_t ExpectedIndex = Group.index();
 
-    if (Field.Index != ExpectedIndex)
-      rc_return VH.fail();
+    if (Field.Index != ExpectedIndex) {
+      using llvm::Twine;
```
aleclearmind, line 799: This is used a lot, please put it at the top of the file once.

lib/Model/Type.cpp
```diff
@@ -786,28 +782,44 @@ verifyImpl(VerifyHelper &VH, const StructType *T) {
...
-    if (isVoidConst(&Field.Type).IsVoid)
-      rc_return VH.fail();
+    if (isVoidConst(&Field.Type).IsVoid) {
+      using llvm::Twine;
+      rc_return VH.fail("Field " + Twine(Field.Index) + " is void", *T);
+    }
+
+    if (not Field.CustomName.empty()
+        and not Names.insert(Field.CustomName).second)
+      rc_return VH.fail("Collision in union fields names", *T);
```
aleclearmind, line 822: This kind of verbose messages are so much needed. We need to expand a lot.

include/revng/Model/VerifyHelper.h
```diff
@@ -59,6 +61,12 @@ class VerifyHelper {
   }
 
 public:
+  void evict(const model::Type *T) {
```
aleclearmind, line 64: The commit message says:
> ...a performance hit that cannot be taken...

Is it actually the case?
Such an API induces the user to try and keep up-to-date an inherently temporary data structure, which is a delicate thing to do. Let's do this only if necessary, otherwise let's start over with a new `VerifyHelper` and be done with it.

aleclearmind (changes requested):

lib/Model/Type.cpp
pfez: Ok. Any action needed here?

include/revng/Model/VerifyHelper.h
pfez: I'm measuring the performance penalty of avoiding this in the only place I need it. Will report back. I'd be happy to avoid introducing this.

lib/Model/Type.cpp
aleclearmind: Nope.

include/revng/TupleTree/TupleTree.h
```diff
@@ -1013,17 +1014,29 @@ template<typename T, typename RootT>
...
+
+  static TupleTreeReference
+  fromString(const RootT *Root, llvm::StringRef Path) {
+    std::optional<TupleTreePath> OptionalPath = stringAsPath<RootT>(Path);
+    if (not OptionalPath.has_value())
+      return TupleTreeReference{};
+    return fromPath(Root, *OptionalPath);
   }
 
   static TupleTreeReference fromString(RootT *Root, llvm::StringRef Path) {
```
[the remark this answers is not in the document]
pfez: Done. How do you like it now?

include/revng/TupleTree/TupleTree.h
```diff
@@ -1043,26 +1056,46 @@ class TupleTreeReference {
...
+    const auto IsNullVisitor = [](const auto &Pointer) {
+      return Pointer == nullptr;
+    };
+    return std::visit(IsNullVisitor, Root);
+  }
+
+  void doGetChecks() const {
+    revng_check(Root.index() != std::variant_npos);
+    revng_check(not hasNullRoot());
+  }
```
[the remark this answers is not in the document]
pfez: Done

include/revng/Model/VerifyHelper.h
pfez: Cool I managed to drop this.

lib/Model/Type.cpp
pfez: Done

include/revng/TupleTree/TupleTree.h
```diff
@@ -1013,17 +1014,29 @@ template<typename T, typename RootT>
 class TupleTreeReference {
 public:
   using pointee = T;
+  using root = RootT;
+  using root_variant = std::variant<RootT *, const RootT *>;
```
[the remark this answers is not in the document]
pfez: Done

aleclearmind:
Merged in 41008d9. Thanks!

I've made some minor changes. Most significant one: I moved the zero-sizedness check from `StructType::verify` and `UnionType::verify` into `AggregateField::verify`.

PR #115, opened 2021-12-22
EarlyFunctionAnalysis: import from model

antoniofrighetto:
Allow EarlyFunctionAnalysis to import from model in order to pre-populate the cache.


lib/EarlyFunctionAnalysis/ForcedCFEPCollectionPass.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+using namespace llvm;
+
+char ForcedCFEPCollectionWrapperPass::ID = 0;
+
+using Register = RegisterPass<ForcedCFEPCollectionWrapperPass>;
+static Register
+  Y("collect-forced-cfep", "Forced CFEP Collection Pass", true, true);
+
+static Logger<> CFEPLog("forced-cfep");
```
aleclearmind, line 20: Can we get rid of CFEP entirely? It's a leftover from an older design.

lib/EarlyFunctionAnalysis/ForcedCFEPCollectionPass.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+  Function &Root = *M.getFunction("root");
+  // Static symbols have already been registered during lifting phase. Now
+  // register all the other candidate entry points.
+  for (BasicBlock &BB : Root) {
+    if (GCBI.getType(&BB) != BlockType::JumpTargetBlock)
+      continue;
+
+    auto It = llvm::find_if(Functions,
+                            [&BB](const auto &E) { return E == &BB; });
```
aleclearmind, line 37: Just use `llvm:find`?

lib/EarlyFunctionAnalysis/NonForcedCFEPCollectionPass.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+    if (not IsLoadAddress
+        and (IsUnusedGlobalData
+             || (IsMemoryStore and not IsPCStore and not IsReturnAddress))) {
+      // TODO: keep IsReturnAddress?
+      // Consider addresses found in global data that have not been used or
+      // addresses that are not return addresses and do not end up in the PC
+      // directly.
+      MetaAddress Entry = GCBI.getJumpTarget(&BB);
+      Binary.Functions[Entry].CustomName = BB.getName().str();
```
aleclearmind, line 48:
```suggestion
```

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -169,7 +168,7 @@ struct FunctionSummary {
   }
 
   FunctionSummary() = delete;
-  FunctionSummary(const FunctionSummary &) = delete;
+  FunctionSummary(const FunctionSummary &) = default;
```
aleclearmind, line 171: Why?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -464,6 +475,60 @@ CFEPAnalyzer<FunctionOracle>::CFEPAnalyzer(llvm::Module &M,
...
 
+static void
+initializeCache(llvm::Module &M,
+                FunctionAnalysisResults &Properties,
+                const std::vector<llvm::GlobalVariable *> &ABIRegisters,
+                model::Binary &Binary) {
+  using namespace model;
+  using RegisterState = model::RegisterState::Values;
+
+  for (const auto &F : Binary.Functions) {
```
aleclearmind, line 486: Do not abbreviate ([B.6](https://pad.rev.ng/#B-CC)) and avoid `auto` ([B.15](https://pad.rev.ng/#B-CC)).

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -464,6 +475,60 @@ CFEPAnalyzer<FunctionOracle>::CFEPAnalyzer(llvm::Module &M,
...
+    if (F.Type == model::FunctionType::Invalid)
+      continue;
+
+    MetaAddress EntryPC = F.Entry;
+    auto &Summary = Properties[EntryPC];
+    Summary.Type = F.Type;
+
+    const model::Type *PrototypeType = F.Prototype.get();
+    if (!llvm::isa<RawFunctionType>(PrototypeType))
```
aleclearmind, line 495: Everywhere.
```suggestion
    if (!isa<RawFunctionType>(PrototypeType))
```

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -658,8 +722,9 @@ finalizeModel(GeneratedCodeBasicInfo &GCBI,
     }
 
     Function.CFG = Summary.CFG;
-    revng_check(Function.verify(true));
   }
+
+  revng_check(Binary.verify());
```
aleclearmind, line 727:
```suggestion
  revng_check(Binary.verify(true));
```

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -602,14 +666,14 @@ finalizeModel(GeneratedCodeBasicInfo &GCBI,
   }
 
   // Finish up the CFG
-  for (const auto &F : Functions) {
-    MetaAddress EntryPC = getBasicBlockPC(F.Entry);
-    model::Function &Function = Binary.Functions[EntryPC];
-    auto &Summary = Properties.at(EntryPC);
+  for (model::Function &Function : Binary.Functions) {
+    if (Function.CFG.size() != 0)
+      continue;
```
aleclearmind, line 671: Why?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -683,12 +748,13 @@ combineCrossCallSites(MetaAddress EntryPC, FunctionSummary &Summary) {
 }
 
 /// Perform cross-call site propagation
-static void interproceduralPropagation(const std::vector<CFEP> &Functions,
-                                       FunctionAnalysisResults &Properties) {
+static void
+interproceduralPropagation(FunctionAnalysisResults &Properties,
+                           SortedVector<model::Function> &Functions) {
   for (auto &J : Functions) {
```
aleclearmind, line 754: `J` is  bad name.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -1828,9 +1859,16 @@ bool EarlyFunctionAnalysis::runOnModule(Module &M) {
...
+  // which have `Invalid` and `Fake` as type.
   UniquedQueue<BasicBlockNode *> CFEPQueue;
   for (auto *Node : llvm::post_order(&CG)) {
-    if (Node != RootNode)
+    if (Node == RootNode)
+      continue;
+
+    auto &Function = Binary.Functions[getBasicBlockPC(Node->BB)];
+    if (Function.Type == model::FunctionType::Invalid
+        || Function.Type == model::FunctionType::Fake)
```
aleclearmind, line 1871: As I mentioned, I think we need to ensure that `Fake` functions are analyzed first.
Also, we probably don't need to analyze them, we could find out that they are not fake, but we don't want that.
Basically, we just want to obtain an `llvm::Function` we can inline. This might to reorganize some stuff.

lib/FunctionIsolation/IsolateFunctions.cpp
```diff
@@ -637,7 +644,9 @@ bool IFI::handleDirectBoundary(const Boundary &TheBoundary,
     }
   }
 
-  revng_assert(Consumed == TheBoundary.Successors.Addresses.size());
+  // TODO: we may need to review how extra successors
+  // (not present within the model) are handled.
+  revng_assert(Consumed <= TheBoundary.Successors.Addresses.size());
```
aleclearmind, line 649: Note to self: think about this.

aleclearmind (changes requested):

include/revng/EarlyFunctionAnalysis/ForcedCFEPCollectionPass.h
```diff
@@ -0,0 +1,34 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "llvm/Pass.h"
+
+#include "revng/BasicAnalyses/GeneratedCodeBasicInfo.h"
+#include "revng/Model/Binary.h"
+#include "revng/Model/LoadModelPass.h"
+
+class ForcedCFEPCollectionWrapperPass : public llvm::ModulePass {
```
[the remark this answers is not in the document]
antoniofrighetto: I think `CollectFunctionsFromCallees` and `CollectFunctionsFromUnusedAddresses `, is that OK?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -464,6 +475,60 @@ CFEPAnalyzer<FunctionOracle>::CFEPAnalyzer(llvm::Module &M,
...
+                model::Binary &Binary) {
+  using namespace model;
+  using RegisterState = model::RegisterState::Values;
+
+  for (const auto &F : Binary.Functions) {
+    if (F.Type == model::FunctionType::Invalid)
+      continue;
+
+    MetaAddress EntryPC = F.Entry;
+    auto &Summary = Properties[EntryPC];
```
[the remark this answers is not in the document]
antoniofrighetto: Precisely because there's no `Summary` for that PC, we need to create a new entry in `Properties`. The `at` method just performs a look up, and will throw an out-of-range exception if the PC is not found (which is exactly this case).

include/revng/EarlyFunctionAnalysis/ForcedCFEPCollectionPass.h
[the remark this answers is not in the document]
aleclearmind: Works.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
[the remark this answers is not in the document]
antoniofrighetto: Also, we still don't want to make the map `FunctionsBucket` field public, that's the reason of the introduction of `operator[]`, which adds a new element in case it doesn't exists.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: As much as we want to skip the commit of the prototype to the model if the original function has type different from `model::FunctionType::Invalid`, we also want to the skip the commit of its CFG.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: While we basically only want the `llvm::Function`, I thought it was not too much of a problem if fake functions are wholly recomputed too. But it's true, we need to make sure they are recomputed first. Initially I kinda thought that this was what was actually happening since fake functions should be leaf functions too (?), but I guess we want to make completely sure that fake functions are pushed onto the queue first.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: I guess `K` (in the inner loop) is also a bad one. Do you have any suggestions by any chance? `First` and `Second` could sounds good?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -219,6 +218,14 @@ class FunctionAnalysisResults {
...
     DefaultSummary(std::move(DefaultSummary)) {}
 
+  FunctionSummary &operator[](const MetaAddress &PC) {
+    auto It = FunctionsBucket.find(PC);
+    if (It != FunctionsBucket.end())
+      return It->second;
+
+    return FunctionsBucket.try_emplace(PC, DefaultSummary).first->second;
+  }
+
```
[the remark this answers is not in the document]
antoniofrighetto: (See below)

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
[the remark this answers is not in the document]
aleclearmind: I understand, but the alternative to `operator[]` is something like `insert`/`emplace`.
`operator[]` is a more sophisticated tool than we need and it pulls in `DefaultSummary` which scares me a bit.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
aleclearmind: Why use a different criterion here?

I think you're checking `Funciton.CFG.size()` because you can no longer rely on `.Type` being `Invalid`, is that right?

However, you can simply create a `set` of `model::Function *` that need to work on and iterate over those. No need to come up with "alternative ways".

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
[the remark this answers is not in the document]
antoniofrighetto: I see, but then it just a matters of changing name, since the `insert` method would still need to `try_emplace` a `DefaultSummary`, no? Also, now that we moved `initializeCache` as a member of the class, we may want to move `operator[]` / `insert` as a private member function.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
aleclearmind: If you do a full blown analysis, EFA could determine that it's not fake and that's not what we want.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Sounds good.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
aleclearmind: Now that I look at this piece of code it seems quite suspicious.
Why are you iterating over all the functions for each function?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: OK, trying to see if we can extract the code responsible for creating the fake function for the `llvm::Function*`.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Well, if we need to cross-merge the results, this must be done for each call-site of a function (inner loop) for all the functions (outer loop). We keep fixed one, and in the meantime iterate over all the others (when the functions are the same we don't care).

antoniofrighetto: Uhmmm, I don't seem to be able to easily get the CallSites though, especially right about now that we are going to take the CFG out of the model?

aleclearmind: We're well before the model is needed here.

Can't you ask `FunctionAnalysisResults` to give all the functions it has?
Maybe, we don't need to pull in the model at all for this part.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Fix this by introducing `CFEPAnalyzer::initializeFakeFunctions` which re-create an `llvm::Function *` for fake functions; no need to put them onto the queue / ensure they're analyzed first at this point. I think it should be OK now.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Separate commit for this.

aleclearmind: Uh? Is it fixed? If so, mark as resolved.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -1163,6 +1227,17 @@ llvm::Function *CFEPAnalyzer<FO>::createFakeFunction(llvm::BasicBlock *Entry) {
...
 }
 
+template<class FO>
+void CFEPAnalyzer<FO>::initializeFakeFunctions() {
+  for (const auto &Function : Binary.Functions) {
+    if (Function.Type != model::FunctionType::Fake)
+      continue;
+
+    auto &Summary = Oracle.at(Function.Entry);
+    Summary.FakeFunction = createFakeFunction(GCBI->getBlockAt(Function.Entry));
```
aleclearmind, line 1237: Is this all you need to do?
Don't you at least need to set the function type to `Fake`?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
aleclearmind: That sounds nice!

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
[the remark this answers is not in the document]
aleclearmind: Why do you need `DefaultSummary`? Isn't a default-constructed summary enough?
`DefaultSummary` role is to give a "default" analysis results for the situations where the analysis has not been performed yet, which doesn't seem to be the case here.

`(*this)[...]` is not very nice :P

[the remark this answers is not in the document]
aleclearmind: Why do you need `DefaultSummary`? Isn't a default-constructed summary enough?
`DefaultSummary` role is to give a "default" analysis results for the situations where the analysis has not been performed yet, which doesn't seem to be the case here.

`(*this)[...]` is not very nice :P

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Right, fixed! Not really needed for the fake function itself, but rather for its callers. As concerns the FSO / Preserved Registers, we should not care.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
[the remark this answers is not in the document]
antoniofrighetto: The thing with the whole `FunctionSummary` was to allow to select it at runtime, thus creating a default-constructed summary makes no sense from a s semantic point of view; indeed we had decided to delete it. As far as I see it, each entry PC should have associated a `DefaultSummary` or a summary, which is a more refined version of the default one, and never a default-constructed one. Even if the analysis won't be performed here, `initializeCache` will still bring `DefaultSummary` to a more accurate summary.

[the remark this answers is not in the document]
antoniofrighetto: More specifically, I reckon that, for the semantic the summary has (i.e., either a `DefaultSummary`, or an object returned from `milkInfo` using the parametrized constructor), one should never be allowed to default-instantiate a `FunctionSummary` object.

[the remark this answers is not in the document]
aleclearmind: Why do we care at all about `DefaultSummary` here? Is there something relevant we are inheriting from it?
If not, just use the constructor of `FunctionSummary`, even if it's not the default one.
What's the point of copying `DefaultSummary` just to clobber it immediately after everything was in it?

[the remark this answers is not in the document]
antoniofrighetto: Well, actually I just take advantage of the fact that `Summary.ClobberedRegisters` has been fulfilled with all the ABI regs (as is the case for `DefaultSummary`), and then we erase from the clobbered ones, those which are preserved.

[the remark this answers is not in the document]
aleclearmind: I think we are not on the same page on the role of `DefaultSummary`.
`DefaultSummary` role is to have a conservative, ABI-specific, summary for functions we know nothing about. It's not some sort of "starting point" for all the other summaries.

Right now, `DefaultSummary` it's being populated in a certain way, but it will become ABI-specific and out of your control.
You should not rely on its list of clobbered registers to be all the ABI registers.
If other parts of the code do that too, that's not nice either and it should be changed.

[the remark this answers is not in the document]
antoniofrighetto: Makes sense. What do you think now?

aleclearmind:
Closing this, I'll go with #125 directly.

PR #117, opened 2021-12-23
Preliminary changes for the introduction of revng::pipeline

drblallo:


include/revng/Model/ModelHelpers.h
```diff
@@ -0,0 +1,43 @@
...
+
+#include "llvm/IR/Function.h"
+
+#include "revng/Model/Binary.h"
+#include "revng/Support/Assert.h"
+#include "revng/Support/IRHelpers.h"
+#include "revng/Support/MetaAddress.h"
+
+namespace model {
+inline MetaAddress getMetaAddressOfIsolatedFunction(const llvm::Function &F) {
```
aleclearmind, line 15: We recently introduced a file for model-related helpers that deal with LLVM stuff: `include/revng/Model/IRHelpers.h`

lib/Support/ProgramRunner.cpp
```diff
@@ -0,0 +1,38 @@
...
+                        std::vector<std::string> &Args) {
+  auto MaybeProgramPath = findProgramByName(ProgramName, Paths);
+  revng_assert(MaybeProgramPath);
+
+  // Prepare actual arguments
+  std::vector<StringRef> StringRefs{ *MaybeProgramPath };
+  for (std::string &Arg : Args)
+    StringRefs.push_back(Arg);
+
+  // Invoke linker
```
aleclearmind, line 35: Linker?

lib/Support/ProgramRunner.cpp
```diff
@@ -0,0 +1,38 @@
+#include "llvm/ADT/SmallVector.h"
```
aleclearmind, line 1: Header?

lib/Support/ProgramRunner.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+  // Append PATH
+  char *Path = getenv("PATH");
+  if (Path == nullptr)
+    return;
+  StringRef(Path).split(Paths, ":");
+}
+
+void ProgramRunner::run(llvm::StringRef ProgramName,
+                        std::vector<std::string> &Args) {
```
aleclearmind, line 26:
```suggestion
                        ArrayRef<std::string> Args) {
```

lib/Support/ProgramRunner.cpp
```diff
@@ -0,0 +1,38 @@
...
+  Paths = { { path::parent_path(MainExecutable) } };
+
+  // Append PATH
+  char *Path = getenv("PATH");
+  if (Path == nullptr)
+    return;
+  StringRef(Path).split(Paths, ":");
+}
+
+void ProgramRunner::run(llvm::StringRef ProgramName,
```
aleclearmind, line 25: Isn't this going to be used by people that shouldn't assert? Shall we return an `llvm::Error`?

include/revng/Model/LoadModelPass.h
```diff
@@ -43,17 +47,28 @@ class LoadModelWrapperPass : public llvm::ImmutablePass {
...
 public:
   bool doInitialization(llvm::Module &M) override final;
   bool doFinalization(llvm::Module &M) override final;
 
 public:
-  ModelWrapper &get() { return *Wrapper; }
+  ModelWrapper &get() {
+    if (isModelExternal())
+      return *std::get<ModelWrapper *>(Wrapper);
+    return std::get<ModelWrapper>(Wrapper);
```
aleclearmind, line 65:
```suggestion
    else
      return std::get<ModelWrapper>(Wrapper);
```

include/revng/Model/IRHelpers.h
```diff
@@ -33,3 +35,32 @@ getCallSitePrototype(const model::Binary &Binary,
...
+  if (auto It = Binary.Functions.find(MaybeMetaAddres);
+      It != Binary.Functions.end())
+    return &(*It);
+
+  return nullptr;
+}
+
+inline const model::Function *
+llvmToModelFunction(const model::Binary &Binary, const llvm::Function &F) {
+  auto MaybeMetaAddres = getMetaAddressMetadata(&F, FunctionEntryMDNName);
```
aleclearmind, line 58:
```suggestion
  auto MaybeMetaAddress = getMetaAddressMetadata(&F, FunctionEntryMDNName);
```

include/revng/ADT/STLExtras.h
```diff
@@ -103,4 +103,29 @@ template<class ContainerTy, class FuncTy>
...
 }
+
+auto dereferenceIterator(auto Iter) {
+  return llvm::map_iterator(Iter, [](const auto &Ptr) -> decltype(*Ptr) & {
+    return *Ptr;
+  });
+}
+
+template<typename T>
+using DereferenceIteratorType = decltype(dereferenceIterator(
```
aleclearmind, line 114: * `check-revng-conventions`?
* Are you sure this is going to be needed?

include/revng/ADT/STLExtras.h
```diff
@@ -103,4 +103,29 @@ template<class ContainerTy, class FuncTy>
...
+  std::declval<T>()));
+
+auto dereferenceRange(auto &&Range) {
+  return llvm::make_range(derefereceIterator(Range.begin()),
+                          derefereceIterator(Range.end()));
```
aleclearmind, line 119:
```suggestion
  return llvm::make_range(dereferenceIterator(Range.begin()),
                          dereferenceIterator(Range.end()));
```

include/revng/Model/LoadModelPass.h
```diff
@@ -22,6 +24,8 @@ class ModelWrapper {
   ModelWrapper(TupleTree<model::Binary> &&TheBinary) :
     TheBinary(std::move(TheBinary)) {}
 
+  ModelWrapper() = default;
```
aleclearmind, line 27: Why?

include/revng/TupleTree/TupleTree.h
```diff
@@ -1176,6 +1193,14 @@ class TupleTree {
     return Result;
   }
 
+  static llvm::ErrorOr<TupleTree> loadFromFile(const llvm::StringRef &Path) {
```
aleclearmind, line 1196:
```suggestion
  static llvm::ErrorOr<TupleTree> fromFile(const llvm::StringRef &Path) {
```

include/revng/TupleTree/TupleTree.h
```diff
@@ -1129,6 +1130,22 @@ void serialize(S &Stream, const T &Element) {
   serialize(Stream, const_cast<T &>(Element));
 }
 
+template<Yamlizable TupleLike>
+llvm::Error
+serializeToFile(const TupleLike &Tuple, const llvm::StringRef &Path) {
```
aleclearmind, line 1135: Why `loadFromFile` is a method and this is not? What are these going to be used for?

aleclearmind (changes requested):

lib/Support/ProgramRunner.cpp
drblallo: do you mean that the header file of this cpp file should first or that small vector is not used?

drblallo: ah no the license ok

lib/Support/ProgramRunner.cpp
drblallo: it seems that the underlyin api returns
```
  /// -1 indicates failure to execute
  /// -2 indicates a crash during execution or timeout
  /// everything else is the return code of the invoked program
```
i guess assert that is positive and return a error if not zero?

aleclearmind: Up to you, we could even return error even if < 0.

include/revng/ADT/STLExtras.h
drblallo: yeah,  the alterinative is to dig into llvm implementation to figure out what is the exact type of a map iterator but i would prefer this way

include/revng/Model/LoadModelPass.h
drblallo: if it is default constructed the optional is empty and it will try to load the model from the module itself.

include/revng/TupleTree/TupleTree.h
drblallo: i saw that the serialize routine was not already here and i put serializedToFile here and i saw the fromString method in the other place and i added fromFile there, i can move them around tho.

they will be used to load and save the model on disk

aleclearmind: What I meant is: if you don't see a use case for `serializeToFile` `TupleLike`s in general, but you'll be using it on `TupleTree<T>` only, make this function a method of `TupleTree<T>`, just as `fromFile`.

lib/Support/ProgramRunner.cpp
drblallo: i will assert that the return code is not -1 or -2 to avoid an potential attacker to figure out that they have some control over an external program invocation

include/revng/TupleTree/TupleTree.h
drblallo: a issue with this is that if someone holds a reference to a binary instead of a reference to a tuple tree binary it will not be able to serialize it, while with the current version it will be able to

drblallo: and due to compatibility with load model pass we were returning a const binary& to people that were requiring a readonly model

aleclearmind (changes requested):

include/revng/Support/IRHelpers.h
```diff
@@ -1268,6 +1271,17 @@ MetaAddress getMetaAddressMetadata(T *U, llvm::StringRef Name) {
   return MetaAddress::invalid();
 }
 
+template<HasMetadata T>
+MetaAddress getMetaAddressMetadata(const T *U, llvm::StringRef Name) {
```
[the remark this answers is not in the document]
drblallo: uhm, why would one that? if one is ok with accepting a HasMetadata type T and only needs to read it, it can accept const T& like i do here

include/revng/Support/TupleIntrospection.h
```diff
@@ -0,0 +1,36 @@
...
+
+template<typename T, typename Tuple_t>
+struct index_in_tuple {
+  static constexpr size_t value = index_in_tuple_fn<0, T, Tuple_t>();
+};
+
+template<typename T, typename Tuple_t>
+constexpr size_t index_in_tuple_v() {
+  return index_in_tuple<T, Tuple_t>::value;
+}
```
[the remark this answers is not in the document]
drblallo: i find out the entire file was no longer needed

include/revng/Support/IRHelpers.h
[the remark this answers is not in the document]
aleclearmind: If `HasMetadata` capture `const T` too you don't need to duplicate this function.
Maybe you can leave it as it is and we can dedup this using [a solution we devised with @pfez](https://github.com/revng/revng/pull/110/files#diff-500497276c0f1d9cd36809234e2956b70c306cdcc1dddd9485ed97cc1402dbeeL1029).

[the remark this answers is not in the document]
drblallo: ah wait i am stupid, the non const version is not needed, we are returning by value

aleclearmind:
Merged in b8f6303! Thanks!

I made the following changes:
* Rewrote a bit the commit messages to be more explicit
* I dropped the `Drop Add TypeWithID` and `Drop Add indent`, please push them in the other MR. They were too much out of context.
* I dropped `revng_assert(ExitCode >= 0)`. If a subprogram crashes or does not exists I prefer not to have any special behavior w.r.t. failure exit code. Let's if that works out!

PR #119, opened 2022-01-10
Feature/revng pipes

drblallo:


lib/Pipes/Lift.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+  llvm::Optional<uint64_t> EntryPointAddressOptional;
+  if (EntryPointAddress.getNumOccurrences() != 0)
+    EntryPointAddressOptional = EntryPointAddress;
+  Generator.translate(EntryPointAddressOptional);
+}
+
+using namespace pipeline;
+
+static RegisterPipe<LiftPipe> E1("Lift");
```
aleclearmind, line 158: I suggest that we:
1. Rename `LiftPipe::Name` to `::Description`
2. Move the registered name to `LiftPipe::Name`
3. Use lower-case [Kebab-case](https://en.wikipedia.org/wiki/Letter_case#Kebab_case) for registered names, as LLVM does for passes.

lib/Pipes/Lift.cpp
```diff
@@ -0,0 +1,158 @@
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
```
aleclearmind, line 3: Fix all headers of `.cpp` files.

include/revng/Pipes/LinkForTranslation.h
```diff
@@ -0,0 +1,44 @@
+#pragma once
```
aleclearmind, line 1: `LinkForTranslationPipe.h`

lib/Pipes/LinkForTranslation.cpp
```diff
@@ -0,0 +1,118 @@
...
+    "-o",
+    LinkerOutput.getOrCreatePath().str(),
+    ("-Wl,-z,max-page-size=" + Twine(PageSize)).str(),
+    "-fuse-ld=bfd"
+  };
+
+  revng_assert(Model.Segments.size() > 0);
+  uint64_t Min = Model.Segments.begin()->StartAddress.address();
+  uint64_t Max = Model.Segments.begin()->EndAddress.address();
+  ;
```
aleclearmind, line 67:
```suggestion

```

include/revng/Pipes/LinkSupport.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
```
aleclearmind, line 1: `LinkSupportPipe.h`

include/revng/Pipes/RevngPipes.h
```diff
@@ -0,0 +1,59 @@
...
+extern Kind Binary;
+
+extern Kind Object;
+extern Kind Translated;
+
+extern Kind ABIEnforced;
+
+extern Kind Dead;
+
+class LLVMContextWrapper : public SavableObject<LLVMContextWrapper> {
```
aleclearmind, line 39: Move to its own header.

include/revng/Pipes/RevngPipes.h
```diff
@@ -0,0 +1,59 @@
+#pragma once
```
aleclearmind, line 1: Please split everything in here to its own header and call this one `AllPipes.h`.

include/revng/Pipes/TemporaryFile.h
```diff
@@ -0,0 +1,71 @@
...
+#include "revng/Support/Assert.h"
+
+namespace pipeline {
+
+/// A temporary file is a container that wraps a single temporary file on disk.
+/// The enumeration of this container will yield the empty set if there is
+/// currently no file associated to a instance of Temporary file, and will
+/// return The target ("root", K) otherwise, where K is the kind provided at
+/// construction time.
+class TemporaryFile : public Container<TemporaryFile> {
```
aleclearmind, line 20: Is it correct to say that the main use case for this is materializing files for command line tools we launch (e.g., linker) that are not in memory yet?

Once all of these are in-memory, we wouldn't have much use for this, is this correct?

Or is this to arbitrarily serialize generic files?

include/revng/Pipes/CompileModule.h
```diff
@@ -0,0 +1,58 @@
...
+class CompileModulePipe {
+public:
+  static constexpr auto Name = "Compile Module Pipe";
+  std::array<ContractGroup, 1> getContract() const {
+    return { ContractGroup(Root,
+                           Exactness::DerivedFrom,
+                           0,
+                           Object,
+                           1,
+                           InputPreservation::Preserve) };
```
aleclearmind, line 30: So compiling a LLVM module to an object file is not destructive?

include/revng/Pipes/CompileModule.h
```diff
@@ -0,0 +1,58 @@
...
+  }
+  void
+  run(const Context &, LLVMContainer &TargetsList, TemporaryFile &TargetBinary);
+};
+
+class CompileModulePipeIsolated {
```
aleclearmind, line 36:
```suggestion
class CompileIsolatedModulePipe {
```

include/revng/Pipes/CompileModule.h
```diff
@@ -0,0 +1,58 @@
...
+  std::array<ContractGroup, 1> getContract() const {
+    return { ContractGroup(Root,
+                           Exactness::DerivedFrom,
+                           0,
+                           Object,
+                           1,
+                           InputPreservation::Preserve) };
+  }
+  void
+  run(const Context &, LLVMContainer &TargetsList, TemporaryFile &TargetBinary);
```
aleclearmind, line 33: * `TargetsList` seems a bad name
* Would it make sense for `TargetsList` to be `const`?

include/revng/Pipes/Isolate.h
```diff
@@ -0,0 +1,54 @@
...
+#include "revng/FunctionIsolation/IsolateFunctions.h"
+#include "revng/Pipes/RevngPipes.h"
+#include "revng/Pipes/RootKind.h"
+
+namespace pipeline {
+
+struct ABIDetectPipe {
+  static constexpr auto Name = "ABIDetect";
+
+  std::vector<ContractGroup> getContract() const { return {}; }
```
aleclearmind, line 18: Is there a reason to define this? Can't the user use `PureLLVMPass`?
We need to decide what to do in these cases. I'm not sure we want to put the burden on the user to define a class for each pass.

include/revng/Pipes/Isolate.h
```diff
@@ -0,0 +1,54 @@
...
+struct IsolatePipe {
+  static constexpr auto Name = "Isolate";
+
+  std::vector<ContractGroup> getContract() const {
+    return { ContractGroup(Root,
+                           Exactness::Exact,
+                           0,
+                           Isolated,
+                           0,
+                           InputPreservation::Preserve) };
```
aleclearmind, line 34: This is not very nice, the indices (`0` and `0`) refer to nothing immediately visible. This is very obscure.
Can we make a `LLVMContractGroup` that has a constructor with those values hardcoded so we can hide this from the user?

include/revng/Pipes/LinkForTranslation.h
```diff
@@ -0,0 +1,44 @@
...
+                        2,
+                        InputPreservation::Preserve);
+    Contract ObjectPart(Object,
+                        Exactness::Exact,
+                        1,
+                        Translated,
+                        2,
+                        InputPreservation::Preserve);
+    return { ContractGroup({ BinaryPart, ObjectPart }) };
+  }
```
aleclearmind, line 36: Given a pipeline, I'd be very happy to dump a graph of the contracts.
But I guess this can wait.

include/revng/Pipes/LinkSupport.h
```diff
@@ -0,0 +1,30 @@
...
+
+class LinkSupportPipe {
+public:
+  static constexpr auto Name = "LLVM Link Support Pipe";
+
+  std::array<ContractGroup, 1> getContract() const {
+    return { ContractGroup(Root,
+                           Exactness::DerivedFrom,
+                           0,
+                           InputPreservation::Preserve) };
```
aleclearmind, line 24: Is this actually necessary? Can't this be a `PureLLVMPipe`?

lib/Pipes/PipelineManager.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+const char LLVMContextWrapper::ID = '0';
+
+class LoadModelPipePass {
+private:
+  ModelWrapper *Wrapper;
+
+public:
+  static constexpr auto Name = "Load Model";
+  std::vector<ContractGroup> getContract() const { return {}; }
```
aleclearmind, line 32: `PureLLVMPipe`?

aleclearmind (changes requested):

include/revng/Pipes/Helpers.h
```diff
@@ -0,0 +1,51 @@
...
+  ModelPipelineWrapper() = default;
+
+  const ModelWrapper &getModelWrapper() const { return Model; }
+
+  ModelWrapper &getModelWrapper() { return Model; }
+};
+
+inline const model::Binary &getModelFromContext(const Context &Ctx) {
+  using Wrapper = ModelPipelineWrapper;
+  const auto &Model = llvm::cantFail(Ctx.getGlobal<Wrapper>(Wrapper::Name));
```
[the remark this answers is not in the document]
drblallo: nah, it just because otherwise it would be more than 80 characters

include/revng/Pipes/Helpers.h
```diff
@@ -0,0 +1,51 @@
...
+#include "revng/Pipeline/Context.h"
+#include "revng/Pipeline/Contract.h"
+#include "revng/Pipeline/GenericLLVMPipe.h"
+#include "revng/Pipeline/LLVMContainer.h"
+#include "revng/Pipeline/Loader.h"
+#include "revng/Pipeline/SavableObject.h"
+#include "revng/Pipeline/Target.h"
+#include "revng/TupleTree/TupleTree.h"
+
+namespace pipeline {
```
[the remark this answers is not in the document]
drblallo: i put everything in revng::pipes

include/revng/Pipes/IsolatedKind.h
```diff
@@ -0,0 +1,83 @@
...
+    if (not Tag->isTagOf(&Symbol) or Symbol.isDeclaration())
+      return std::nullopt;
+
+    auto MaybeAddress = getMetaAddressOfIsolatedFunction(Symbol);
+    revng_assert(MaybeAddress.isValid());
+    return Target({ "root", MaybeAddress.toString() }, *this);
+  }
+};
+
+extern TaggedFunctionKind Isolated;
```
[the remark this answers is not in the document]
drblallo: i put everything in revng::pipes

include/revng/Pipes/TemporaryFile.h
drblallo: more like, some stuff we do uses executables that we cannot use as a library, (linker, merge dynamic) and they need a file do to stuff on them. furthermore it's used by the input and output too.

include/revng/Pipes/CompileModule.h
drblallo: the compilation is operated as a llvm passmanager and that requires a non cost reference.

include/revng/Pipes/Isolate.h
drblallo: the problem would be that at the moment the parser rules to load purellvmpasses or non pure ones are unrelated, that is the names do not pollute each other, so you could not load a pure llvm pass inside a non pure llvm pipe. it would require some working to make sure that they work together.

include/revng/Pipes/LinkForTranslation.h
drblallo: yeah, i guess it count as a new feature.

include/revng/Pipes/LinkSupport.h
drblallo: same issue as other one above.

lib/Pipes/PipelineManager.cpp
drblallo: same issue as other one above.

lib/Pipes/Lift.cpp
aleclearmind: I see, but it'd be so much better to have a the pipe name in the class. Otherwise one has to dig into the `.cpp` files. It's what happens with LLVM too, but at least there we have `--help` to list all passes.

What if we informally move it in the class and 
```suggestion
static RegisterPipe<LiftPipe> E1(LiftPipe::Name);
```

Those that need to be registered multiples times simply don't have `::Name`.

include/revng/Pipes/TemporaryFile.h
aleclearmind: Mmh, then it's more `File` than `TemporaryFile`. Maybe `FileContainer`? The input and the output are not *temporary*.

include/revng/Pipes/CompileModule.h
aleclearmind: Try to `verify` and `dump` when you get the chance. If it's destructive, it's a subtle bug.

include/revng/Pipes/Isolate.h
aleclearmind: Mmmmh, this is rather unexpected. I'm afraid we need to fix this. In particular approaching the back of the decompilation pipeline we mix heavy LLVM analysis and our passes, it's not very nice to discard all the analysis results for infra limitations.

Can you see how much effort it'd be and decide whether it fits within the scope of this PR=

include/revng/Pipes/Helpers.h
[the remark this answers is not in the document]
aleclearmind: I mean
```suggestion
  const auto &Model = llvm::cantFail(Ctx.getGlobal<Wrapper>());
```
And `getGlobal` fetches the name from `Wrapper::Name` on its own.

lib/Pipes/Lift.cpp
drblallo: what i did was to add a overload to the various register things that does not take the string and deduces it from the registerd type. So one can overload it if he wishes

include/revng/Pipes/TemporaryFile.h
drblallo: they are temporary in the sense that while executing are temporary, the input, output and those written into steps are the serialized non temporary version,

include/revng/Pipes/Helpers.h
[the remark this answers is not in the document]
drblallo: this is different than the various register things, since the various user may want to register the same class multiple times, and fetch the same object type multiple times with different name. Here i would force the user to be always explicit about which one it was so that he does not forget to write T::Name and ends up fetching a different one but with the same type.

include/revng/Pipes/TemporaryFile.h
aleclearmind: Mmmh, intuitively if I serialize a *temporary file* I still get a temporary file.
Just as if I serialize an LLVM module I still get an LLVM module.
I feel like it gives the wrong intuition, but I also understand the parallel with the fact that `llvm::Module` in memory is "temporary" too.

include/revng/Pipes/Helpers.h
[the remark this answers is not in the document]
aleclearmind: Uhm OK.

include/revng/Pipes/Isolate.h
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  void registerPasses(llvm::legacy::PassManager &Manager) {
+    Manager.add(new IsolateFunctions());
+  }
+};
```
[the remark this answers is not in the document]
drblallo: i am really considering this, the issue is that it would invert the dependencies and have isolation depend on the pipes library, which is really strange

include/revng/Pipes/LinkSupport.h
drblallo: ah this one cannot because it's not a pass.it invokes llvm::linkModules

lib/Pipes/PipelineManager.cpp
drblallo: this one is magical and must be scheduled in front of every llvm pipeline so that they can access the model

include/revng/Pipes/Isolate.h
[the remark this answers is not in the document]
aleclearmind: Is it? Isolation depends on LLVM too.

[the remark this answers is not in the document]
drblallo: yeah, then what would happen is that each library will end up depending on pipes, pretty much

include/revng/Pipes/PipelineManager.h
```diff
@@ -0,0 +1,101 @@
...
+                   llvm::ArrayRef<std::string> EnablingFlags,
+                   llvm::StringRef ExecutionDirectory);
+
+  llvm::Error overrideContainer(llvm::StringRef PipelineFileMapping);
+  llvm::Error overrideContainer(pipeline::PipelineFileMapping Mapping);
+  llvm::Error overrideModel(llvm::StringRef Path);
+
+  llvm::Error storeOverrides(llvm::ArrayRef<std::string> StoresOverrides);
+  llvm::Error
+  storeOverride(const pipeline::PipelineFileMapping &StoresOverride);
```
aleclearmind, line 71: You sure this isn't a `revng-pipeline` feature? Maybe it should be moved there.

include/revng/Pipes/PipelineManager.h
```diff
@@ -0,0 +1,101 @@
...
+#include "revng/Pipes/ModelGlobal.h"
+
+namespace revng::pipes {
+
+/// This is a "God Object" that can be used when there will be exactly one
+/// Pipeline spawned by this process.
+///
+/// It will care of all configurations including llvm ones and provide a runner
+/// in a usable state.
+class PipelineManager {
```
aleclearmind, line 25: Document all methods.
* What's the difference between `create` and `createFromMemory`? One might think the second option does not use the disk at all.
* What to all the `recalculate*` methods do?
* What's the difference between the *current* state and the *last* state? What are the costs involved in calling `getCurrentState` vs `getLastPipelineState`?

include/revng/Pipes/PipelineManager.h
```diff
@@ -0,0 +1,101 @@
...
+
+  const pipeline::Context &context() const { return *PipelineContext; }
+
+  void recalculateAllPossibleTargets();
+  void recalculateCurrentState();
+  void recalculateCache();
+  void getAllPossibleTargets(pipeline::Runner::State &State) const;
+  void getCurrentState(pipeline::Runner::State &State) const;
+  const pipeline::Runner::State &getLastPipelineState() const { return State; }
+  llvm::Error produceAllPossibleTargets();
```
aleclearmind, line 82: What's the use case for this? Is it like `make all`?

include/revng/Pipes/PipelineManager.h
```diff
@@ -0,0 +1,101 @@
...
+
+  const pipeline::TargetsList *
+  getAviableTargetsOf(const pipeline::ContainerSet::value_type &Container) {
```
aleclearmind, line 85:
```suggestion
  getTargetsAvailableFor(const pipeline::ContainerSet::value_type &Container) {
```

include/revng/Pipes/PipelineManager.h
```diff
@@ -0,0 +1,101 @@
...
+    else
+      return Iter->second;
+  }
+
+  void dump() const { Pipeline->dump(); }
+
+  const pipeline::Runner &getRunner() const { return *Pipeline; }
+  pipeline::Runner &getRunner() { return *Pipeline; }
+
+  void dumpPossibleTargets() const;
```
aleclearmind, line 97:
```suggestion
  void dumpAllPossibleTargets() const;
```
`private`?

include/revng/Pipes/PipelineManager.h
```diff
@@ -0,0 +1,101 @@
...
+    if (auto Iter = LookUpMap.find(&Container); Iter == LookUpMap.end())
+      return nullptr;
+    else
+      return Iter->second;
+  }
+
+  void dump() const { Pipeline->dump(); }
+
+  const pipeline::Runner &getRunner() const { return *Pipeline; }
+  pipeline::Runner &getRunner() { return *Pipeline; }
```
aleclearmind, line 95: * Do you see the possibility of *not* exposing the `Runner` to the user? If not, please consider dropping the `getKind` method.
* Can we rename the `Pipeline` field to `Runner`?

include/revng/Pipes/PipelineManager.h
```diff
@@ -0,0 +1,101 @@
...
+  llvm::Error storeToDisk();
+
+  const pipeline::Context &context() const { return *PipelineContext; }
+
+  void recalculateAllPossibleTargets();
+  void recalculateCurrentState();
+  void recalculateCache();
+  void getAllPossibleTargets(pipeline::Runner::State &State) const;
+  void getCurrentState(pipeline::Runner::State &State) const;
+  const pipeline::Runner::State &getLastPipelineState() const { return State; }
```
aleclearmind, line 81:
```suggestion
  const pipeline::Runner::State &getLastState() const { return State; }
```

lib/Pipes/PipelineManager.cpp
```diff
@@ -0,0 +1,302 @@
...
+  PipelineManager Manager;
+  Manager.ExecutionDirectory = ExecutionDirectory.str();
+  Manager.Context = std::make_unique<LLVMContextWrapper>();
+
+  Manager.ModelWrapper = make_unique<ModelGlobal>();
+
+  auto Ctx = setUpContext(*Manager.Context, *Manager.ModelWrapper);
+  Manager.PipelineContext = make_unique<pipeline::Context>(move(Ctx));
+
+  auto Loader = setUpLoader(*Manager.PipelineContext, EnablingFlags);
```
aleclearmind, line 167:
```suggestion
  auto Loader = setupLoader(*Manager.PipelineContext, EnablingFlags);
```

scripts/revng_translate/translate.py
```diff
@@ -0,0 +1,110 @@
...
+        command.append("--silent")
+
+    if args.O1:
+        command.append("--Compile-OptLevel=1")
+    elif args.O2:
+        command.append("--Compile-OptLevel=2")
+        command.append("-f")
+        command.append("O2")
+    else:
+        command.append("--Compile-OptLevel=0")
```
aleclearmind, line 89: Kebab case for options (all lowercase).

scripts/revng_translate/translate.py
```diff
@@ -0,0 +1,110 @@
...
+    return run(command, command_prefix)
+    
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="The rev.ng translator.")
+    subparsers = parser.add_subparsers(dest="command_name",
+                                       help='sub-commands help')
+    register_translate(subparsers)
+    args = parser.parse_args()
+    sys.exit(run_translate(args, [], []))
```
aleclearmind, line 110: Define a `main` function (see other scripts).

scripts/revng
```diff
@@ -268,249 +174,6 @@ def get_stderr(args):
...
-                        "--isolate",
-                        action="store_true",
-                        help="Enable function isolation.")
-    parser.add_argument("--base", help="Load address to employ in lifting.")
-    parser.add_argument("-o", "--output", metavar="OUTPUT",
-                        help="Output path.")
-    parser.add_argument("input", metavar="INPUT", help="The input binary.")
-
-
-def run_translate(args, post_dash_dash):
```
aleclearmind, line 359: I'm so happy all this logic has to go.

scripts/revng
```diff
@@ -224,35 +159,6 @@ def split_dash_dash(args):
     return list(extra_args), args
 
 
-def get_architecture(path):
```
aleclearmind, line 227: Who does this now?

scripts/revng
```diff
@@ -657,10 +317,10 @@ def main():
         sys.stdout.write("rev.ng version @VERSION@\n")
         return 0
 
-    return run_command(args, unknown_args, post_dash_dash)
+    return run_command(args, unknown_args, post_dash_dash, search_path, command_prefix)
```
aleclearmind, line 320: What is this `command_prefix` thing? Seems like some standalone change has slipped in.
If this is the case, please make a standalone commit.

scripts/revng_translate/translate.py
```diff
@@ -0,0 +1,110 @@
...
+
+    if args.trace:
+        command.append("--link-trace")
+
+    if args.showCommand:
+        print(" ".join(command))
+        return 0
+
+    return run(command, command_prefix)
+
```
aleclearmind, line 102: You're leaving whitespaces at the end of line.
Please configure your editor to automatically purge those and double check all the patchset.
`git diff develop`, `/ $` to see all the occurrences.

scripts/revng_translate/translate.py
```diff
@@ -0,0 +1,110 @@
...
+    else:
+        command.append("--Compile-OptLevel=0")
+
+    if args.base:
+        command.append("--base={}".format(args.base))
+
+    if args.trace:
+        command.append("--link-trace")
+
+    if args.showCommand:
```
aleclearmind, line 97: snake_case

include/revng/PipelineC/PipelineC.h
```diff
@@ -0,0 +1,378 @@
...
+ * Every pointer returned by this library is owned by the library and must not
+ * be deleted, except for pointers returned by functions containing the word
+ * "create". Objects returned by functions named "create" are owned by the
+ * caller and must be destroyed invoking the appropriate function called
+ * rp_TYPE_destroy.
+ */
+
+/**
+ * Must be invoked before any other rp_* function is used,
+ * must be invoked exactly once.
```
aleclearmind, line 77: Can we assert if any other API function is called before `rp_init`?

include/revng/PipelineC/PipelineC.h
```diff
@@ -0,0 +1,378 @@
...
+ * This will take care of initializing all llvm related
+ * stuff needed by the revng pipeline to operate.
+ *
+ * Do not initialize all stuff on your own.
+ */
+bool rp_init(int argc,
```
aleclearmind, line 83:
```suggestion
bool rp_initialize(int argc,
```

include/revng/PipelineC/PipelineC.h
```diff
@@ -0,0 +1,378 @@
...
+#include "revng/Pipeline/Target.h"
+#include "revng/Pipes/PipelineManager.h"
+#else
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#endif
+
+#ifdef __cplusplus
+typedef revng::pipes::PipelineManager rp_runner;
```
aleclearmind, line 22: This is confusing. Shall we have `rp_pipeline_manager` or simply `rp_pipeline`?

Can you review the name of the C stuff to be as coherent as possible with C++ stuff? Otherwise we're making the life of newcomers approaching the C API harder for no reason.

In particular, this is not great:
```
 typedef const pipeline::TargetsList rp_container_status;
```

include/revng/PipelineC/PipelineC.h
```diff
@@ -0,0 +1,378 @@
...
+ * pipeline_flags can be empty, pipeline_flags_count must be the size of the
+ * pipeline_flags array.
+ *
+ * returns the created rp_runner if no error happened, nullptr otherwise.
+ *
+ * this function can be called only once, since it will take take of the
+ * initialization of all dynamically loaded libraries.
+ *
+ * this must be invoked after rp_init and all invocations of
+ * rp_load_library_permanently
```
aleclearmind, line 101: Can we assert this?

include/revng/PipelineC/PipelineC.h
```diff
@@ -0,0 +1,378 @@
...
+ *
+ * runner cannot be null, index must be less than
+ * rp_runner_containers_count(runner)
+ */
+rp_container *rp_runner_get_container(rp_runner *runner, uint64_t index);
+
+/**
+ * returns the name of the nth container registered inside the runner
+ */
+const char *rp_container_get_name(rp_container *container);
```
aleclearmind, line 125: Can you group and clearly identify functions that are *class methods* of the same *class*?
Consider using [Doxygen grouping](https://www.doxygen.nl/manual/grouping.html).

include/revng/PipelineC/PipelineC.h
```diff
@@ -0,0 +1,378 @@
...
+ *
+ * Do not initialize all stuff on your own.
+ */
+bool rp_init(int argc,
+             char *argv[],
+             int libraries_count,
+             const char *libraries_path[]);
+
+/**
+ * pipeline_path, flags, execution_directory cannot be null.
```
aleclearmind, line 89: Use Doxygen way of documenting individual arguments.

Also, describe what this does first. Afterwards you  have alternatives (`rp_runner_create_from_string` and `rp_runner_create_memory_only`) that say do the some thing as this except..., but you don't say what *this function* does.

include/revng/PipelineC/PipelineC.h
```diff
@@ -0,0 +1,378 @@
...
+void rp_runner_destroy(rp_runner *runner);
+
+/**
+ * returns the amount of steps present in the runner.
+ *
+ * runner cannot be null
+ */
+uint64_t rp_runner_steps_count(rp_runner *runner);
+
+const inline uint64_t NO_KNOWN_STEP = UINT64_MAX;
```
aleclearmind, line 169:
```suggestion
const inline uint64_t RP_STEP_NOT_FOUND = UINT64_MAX;
```

include/revng/PipelineC/PipelineC.h
```diff
@@ -0,0 +1,378 @@
...
+
+/**
+ * returns the index of the step with the provided name.
+ * returns NO_KNOWN_STEP if no step with such name existed.
+ *
+ * runner cannot be null
+ */
+uint64_t rp_runner_step_name_to_index(rp_runner *runner, const char *name);
```
aleclearmind, line 177: Either keep this and drop `rp_runner_get_step_from_name` or viceversa.
Let's keep the API minimal and offer only one way and one way only do to things.
If the users wants, he can easily write wrappers.

include/revng/PipelineC/PipelineC.h
```diff
@@ -0,0 +1,378 @@
...
+
+/**
+ *
+ * returns the container checkpoint associated to the provided container,
+ *
+ * step cannot be null
+ *
+ */
+rp_container_checkpoint *
+rp_step_get_checkpoint(rp_step *step, rp_container *container);
```
aleclearmind, line 202: Mmmmh, `checkpoint`. We never speak about checkpoints anywhere else.
I know I might have been pushing from this, but at this stage I think we'd like someone reading the `revng-pipeline` doc not to be puzzled by the C API.
Let's reconsider `checkpoint` for something closer to what the rest of the world calls them.

include/revng/PipelineC/PipelineC.h
```diff
@@ -0,0 +1,378 @@
...
+ */
+rp_target *rp_create_target(rp_kind *kind,
+                            int is_exact,
+                            uint64_t path_components_count,
+                            const char *path_components[]);
+
+/**
+ * Deletes the provided target, (THIS TARGET MUST BE ONE OBTAINED FROM
+ * rp_create_target)
+ */
```
aleclearmind, line 247:
```suggestion
/**
 * Deletes the provided target
 *
 * \note WARNING: run this only on targets created through rp_create_target.
 */
```

Do we have a way to `assert` this?

include/revng/PipelineC/PipelineC.h
```diff
@@ -0,0 +1,378 @@
...
+/**
+ * requests the production of the provided targets in a particular container
+ *
+ * returns 0 if an error was encountered, 1 otherwise
+ *
+ * runner cannot be null
+ * target cannot be null
+ * container_name cannot be null
+ */
+bool rp_produce_targets(rp_runner *runner,
```
aleclearmind, line 259:
```suggestion
bool rp_runner_produce_targets(rp_runner *runner,
```

include/revng/PipelineC/PipelineC.h
```diff
@@ -0,0 +1,378 @@
...
+ *
+ * no argument can be null
+ */
+rp_container_status *
+rp_get_container_status(rp_runner *runner, rp_container_checkpoint *container);
+
+/**
+ * returns the amount of target inside the provided container statuses
+ */
+uint64_t rp_container_status_count(rp_container_status *status);
```
aleclearmind, line 306:
```suggestion
uint64_t rp_container_status_targets_count(rp_container_status *status);
```

include/revng/PipelineC/PipelineC.h
```diff
@@ -0,0 +1,378 @@
...
+void rp_runner_recompute_all_available_targets(rp_runner *runner);
+
+/**
+ * returns the container status associated to the provided checkpoint
+ * or null if no status is associated to the provided checkpoint
+ *
+ * no argument can be null
+ */
+rp_container_status *
+rp_get_container_status(rp_runner *runner, rp_container_checkpoint *container);
```
aleclearmind, line 301:
```suggestion
rp_runner_get_container_status(rp_runner *runner, rp_container_checkpoint *container);
```

include/revng/PipelineC/PipelineC.h
```diff
@@ -0,0 +1,378 @@
...
+ * target cannot be null
+ */
+rp_kind *rp_target_get_kind(rp_target *target);
+
+/**
+ *
+ * Serializes target into a string, target cannot be null.
+ *
+ */
+char *rp_create_target_serialization(rp_target *target);
```
aleclearmind, line 329:
```suggestion
char *rp_create_target_serialize(rp_target *target);
```
or
```suggestion
char *rp_create_target_to_string(rp_target *target);
```
* Who owns the return value? Should it be `const char *`?
* This function starts with `rp_create_` but has no `rp_destroy_` counterpart. Let's be strict on this.

include/revng/PipelineC/PipelineC.h
```diff
@@ -0,0 +1,378 @@
...
+
+/**
+ *
+ * Deserialize a target from a string, arguments cannot be null.
+ *
+ * Returns null if the input string was malformed.
+ *
+ */
+rp_target *
+rp_create_target_from_serialization(rp_runner *runner, const char *serialized);
```
aleclearmind, line 339:
```suggestion
rp_create_target_from_string(rp_runner *runner, const char *serialized);
```
```suggestion
rp_create_target_deserialize(rp_runner *runner, const char *serialized);
```

include/revng/PipelineC/PipelineC.h
```diff
@@ -0,0 +1,378 @@
...
+ * possible targets. returns null if index is out of bound
+ *
+ * target cannot be null
+ */
+const char *rp_target_get_path_component(rp_target *target, uint64_t index);
+
+/**
+ * returns the path of the provided container at the provided step
+ */
+char *rp_create_container_path(rp_runner *runner,
```
aleclearmind, line 367: Why not `const char *`?

include/revng/PipelineC/PipelineC.h
```diff
@@ -0,0 +1,378 @@
...
+                               const char *step_name,
+                               const char *container_name);
+
+/**
+ * frees the string returned by rp_create_container_path
+ */
+void destroy_container_path(char *container_path);
```
aleclearmind, line 374:
```suggestion
void rp_destroy_container_path(char *container_path);
```

include/revng/Pipes/Isolate.h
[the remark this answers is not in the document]
aleclearmind: I see what you mean, it's not great and it will likely increase compile time. However, adding a new pass it's already quite a pain in the ass. I don't really want to make it worse having to add other files in other directories.
Also, if we want to rename things it sucks.

tools/revng-lift/Main.cpp
```diff
@@ -203,52 +111,63 @@ int main(int argc, const char *argv[]) {
...
+  // Establish goals
+  ContainerToTargetsMap Targets;
+  Targets.add(Lifted, Target({ PathComponent("root") }, Root));
 
-  llvm::Optional<uint64_t> EntryPointAddressOptional;
-  if (EntryPointAddress.getNumOccurrences() != 0)
-    EntryPointAddressOptional = EntryPointAddress;
-  Generator.translate(EntryPointAddressOptional);
+  // Run the pipeline
+  exitOnError(Pipeline.run("end", Targets));
```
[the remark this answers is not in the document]
drblallo: revng lift was where we tried for the first time to implement the pipeline. There was no really "big plan" behind it, we can go back to the original lift version and not tie this to the pipeline, if you want

tools/revng-pipeline/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+
+revng_add_executable(revng-pipeline
+  Main.cpp)
+
+target_link_libraries(revng-pipeline
+  revngPipes
+  revngPipeline)
```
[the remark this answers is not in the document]
drblallo: yeah, but why being implicit, when you can be explicit

tools/revng-lift/Main.cpp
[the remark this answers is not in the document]
aleclearmind: Yes please.

tools/revng-pipeline/Main.cpp
```diff
@@ -0,0 +1,168 @@
...
+                                                          &Msg))
+      exitOnError(llvm::createStringError(llvm::inconvertibleErrorCode(), Msg));
+  }
+
+  Registry::runAllInitializationRoutines();
+
+  auto Manager = exitOnError(makeManager());
+
+  for (const auto &Override : ContainerOverrides)
+    exitOnError(Manager.overrideContainer(Override));
```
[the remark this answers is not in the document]
drblallo: i tried and it's even worst looking. Every line is expanded from
```
exitOnError(Manager.doSomething())
```
into 
```
if (auto Error  = Manager.doSothing(); Error)
  return Error;
```

which is as ugly as exit on error, only on two lines, and it becomes much worse when there is a Expected<Something> which ends up taking 4 lines
```
Something Obj = exitOnError(manager.tryGetSomething())
```
to
```
auto MaybeSomething = Manager.tryGetSomething();
if (not MaybeSomething)
  return MaybeSomething.getError();
 Something auto& Obj = *MaybeSomething;
```
i would keep this as it is now

include/revng/Pipes/PipelineManager.h
```diff
@@ -0,0 +1,101 @@
...
+class PipelineManager {
+private:
+  explicit PipelineManager() = default;
+
+  std::string ExecutionDirectory;
+  std::unique_ptr<LLVMContextWrapper> Context;
+  std::unique_ptr<ModelGlobal> ModelWrapper;
+  std::unique_ptr<pipeline::Context> PipelineContext;
+  std::unique_ptr<pipeline::Loader> Loader;
+  std::unique_ptr<pipeline::Runner> Pipeline;
```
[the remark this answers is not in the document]
drblallo: this is done to keep pointers stable most of all, and to delay the creation of the object since they require complex initialization. It is not a issue because this class is just  a thin wrapper that all it does is to keep everything used by the c api in a single place.

include/revng/Pipes/PipelineManager.h
```diff
@@ -0,0 +1,101 @@
...
+  pipeline::Runner::State State;
```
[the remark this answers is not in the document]
drblallo: it's just the place where we throw the full state enumeration requested by the c api. the only purpose is to take ownership of the memory so that the user of the c api does not have to free it themself.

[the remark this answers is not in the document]
aleclearmind: OK, `LastState`, `CachedState`? State is too general and misleading.

include/revng/Pipes/PipelineManager.h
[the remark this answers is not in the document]
aleclearmind: > this is done to keep pointers stable most of all

Is this FUD or is it actually necessary?

> to delay the creation of the object since they require complex initialization

If you're using `std::unique_ptr` for delayed initialization purposes, you could use `std::optional` as well, but please try to minimize this (and maybe even reconsider the various things to have a default constructor, if they don't have it).

tools/revng-pipeline/Main.cpp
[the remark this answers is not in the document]
aleclearmind: This makes me a bit sad.
What do you think of having a dumb class that has `operator=(const llvm::Expected<T> &E) { exitOnError(E); }` and then go for:

```c++
ExitIfErrorType ExitIfError;
for (const auto &Override : ContainerOverrides)
  ExitIfError = Manager.overrideContainer(Override);
```

I'm not excited, if you don't like, go ahead and mark this as resolved.

include/revng/Pipes/PipelineManager.h
drblallo: eh it's used in main

include/revng/Pipes/PipelineManager.h
drblallo: yeah, no the runner has too many functionalities to not be exposed.

scripts/revng
drblallo: in liftPipe.cpp
```
  BinaryFile TheBinary(SourceBinary.path()->str(), BaseAddress);
  TheBinary.architecture().name()
```
and then it's later read from the model when used
lift should store it in the model i think

scripts/revng_translate/translate.py
drblallo: wait, i am confused, i added the newline there because in some other file where there was not you asked me to add it

include/revng/PipelineC/PipelineC.h
drblallo: because this will free the container path returned by rp_create_container_path. All functions named rp_create_x return stuff that must be freed by the user.

include/revng/PipelineC/PipelineC.h
drblallo: not really, it's a random heap object.

scripts/revng
drblallo: ok i figure it out, there were two global variables, search_path and command prefix.
search_path was the place were programs were searched for, and command_prefix is used to enable stuff like gdb --args. This was needed because now some of the user of those two variables are in another module.

I guess i can split this into two commits and have the one moving to not having those global variables. Not really sure why split them toh. The changes regarding the pipeline require the searchj_path scope reduction

include/revng/Pipes/PipelineManager.h
drblallo: it's used for debug purposes to require the production of everything that can be produced and detect crashes in a reliable way

include/revng/Pipes/PipelineManager.h
```diff
@@ -0,0 +1,101 @@
...
+         llvm::StringRef ExecutionDirectory);
+
+  static llvm::Expected<PipelineManager>
+  createFromMemory(llvm::ArrayRef<std::string> InMemoryPipeline,
+                   llvm::ArrayRef<std::string> EnablingFlags,
+                   llvm::StringRef ExecutionDirectory);
+
+  llvm::Error overrideContainer(llvm::StringRef PipelineFileMapping);
+  llvm::Error overrideContainer(pipeline::PipelineFileMapping Mapping);
+  llvm::Error overrideModel(llvm::StringRef Path);
```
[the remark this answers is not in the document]
drblallo: it's not really the path toh, it fully loads the content of a particular container with the serialized content of a particular file

include/revng/Pipes/PipelineManager.h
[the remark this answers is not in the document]
drblallo: i can't do this simply. The construction of this object are failable, so there is a constructor method
```
static llvm::Expected<PipelineManager> Pipeline::create(llvm::ArrayRef<std::string> PipelinePath, llvm::ArrayRef<std::string> EnablingFlags, llvm::StringRef ExecutionDirectory);
```
since expected is allocated on the stack, the pointers contained would be moved and thus references would break. I can return a Expected<unique_ptr> to avoid the issue, but the current solution is a bit more complex internally to be less complex outside.

scripts/revng
aleclearmind: If you think you can have a commit that's generally beneficial, create a separate commit.
If this is something conceptually part of another commit, leave it there (and maybe mention it in the commit message). Just make sure that this is not some completely unrelated change slipping in.

include/revng/Pipes/PipelineManager.h
aleclearmind: OK, document this. The public API of this class otherwise is quite confusing.

include/revng/Pipes/PipelineManager.h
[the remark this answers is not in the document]
aleclearmind: Then a better name would be `overwrite...`?
However I'm confused about this whole thing, can you explain how this works?
In particular in combination with `storeOverrides`.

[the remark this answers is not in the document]
drblallo: they are called overrides because it's the way one thinks of them when using the executable interaface

```
revng pipeline x y z -i file:step:container -o file2:step2:contaner2 
```
the triple file:step:container is a file mapping and they operate a load override after the normal load but before the execution and operate the store override after the execution

include/revng/Pipes/PipelineManager.h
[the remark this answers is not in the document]
drblallo: that code snippet is still UB i think. 
rvo does not guarantee that the move will not happen, i think. not sure toh

include/revng/Pipes/PipelineManager.h
```diff
@@ -0,0 +1,131 @@
...
+  llvm::Error overrideContainer(pipeline::PipelineFileMapping Mapping);
+  /// The same as the previous overload except the mapping is provided as a
+  /// triple to be parsed is the usual way.
+  llvm::Error overrideContainer(llvm::StringRef PipelineFileMapping);
+
+  /// Entirelly replaces the model with the content of the file indicated by
+  /// Path
+  llvm::Error overrideModel(llvm::StringRef Path);
+
+  /// Stores the content of the of the
```
aleclearmind, line 81: `of the of the`?
Stress this is saving only the thing you tell it to.

include/revng/Pipes/PipelineManager.h
aleclearmind: Document `writeAllPossibleTargets`, `getTargetsAvailableFor`, `create`, `createFromMemory`.

aleclearmind (changes requested):

aleclearmind:
Merged in d9c8935. Thanks a lot!

PR #121, opened 2022-01-17
Workaround: renvg.h should not use model headers

drblallo:
Before this commit the bug can be found by
orc clean revng
cd /root/of/orchestra
rm -rf ./root
orc configure revng
cd build/revng/optimized
ninja renvgSupport

the error should manifest itself as some generated headers missing

revng.h is currently using some headers belonging to model, while not
expressing a dependency to the model target. Such dependency cannot be
expressed because revngModel already depends on revngSupport, it would
generate cyclic dependencies.

The proper fix would be to rewrite revng.h so that it does not uses the
model.


aleclearmind:
Merged in 6421aeed3ec6d3c05343fef48bf60f342f33ac05. Thanks!

PR #125, opened 2022-01-21
EarlyFunctionAnalysis: take the control-flow graph out of the model

antoniofrighetto:
CFG out of the model.


lib/FunctionIsolation/IsolateFunctions.cpp
```diff
@@ -667,7 +674,9 @@ bool IFI::handleDirectBoundary(const Boundary &TheBoundary,
     }
   }
 
-  revng_assert(Consumed == TheBoundary.Successors.Addresses.size());
+  // TODO: we may need to review how extra successors
+  // (not present within the model) are handled.
+  revng_assert(Consumed <= TheBoundary.Successors.Addresses.size());
```
aleclearmind, line 677: MEMOTOSELF: revert this change and investigate the situation.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -1882,7 +1836,7 @@ bool EarlyFunctionAnalysis::runOnModule(Module &M) {
 
   // Instantiate a CFEPAnalyzer object
   using CFEPA = CFEPAnalyzer<FunctionAnalysisResults>;
```
aleclearmind, line 2029: Can we make disappear the `CFEP` wording entirely. At this point they are no longer "candidates".

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -666,36 +666,29 @@ finalizeModel(GeneratedCodeBasicInfo &GCBI,
...
-           zipmap_range(Summary.ABIResults.ArgumentsRegisters,
-                        CallSites.ArgumentsRegisters)) {
-        auto *CSV = FuncArg == nullptr ? CSArg->first : FuncArg->first;
-        auto RSFArg = FuncArg == nullptr ? RegState::Maybe : FuncArg->second;
-        auto RSCSArg = CSArg == nullptr ? RegState::Maybe : CSArg->second;
+  for (auto &[FuncArg, CSArg] :
+       zipmap_range(Callee.ArgumentsRegisters, CallSite.ArgumentsRegisters)) {
+    auto *CSV = FuncArg == nullptr ? CSArg->first : FuncArg->first;
+    auto RSFArg = FuncArg == nullptr ? RegState::Maybe : FuncArg->second;
+    auto RSCSArg = CSArg == nullptr ? RegState::Maybe : CSArg->second;
```
aleclearmind, line 796: Can you elaborate (in the commit message) what type of improvement is going on here?
Were we not propagating information correctly across call sites?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -1788,9 +1884,15 @@ bool EarlyFunctionAnalysis::runOnModule(Module &M) {
...
 
+  // The intraprocedural analysis will be scheduled only for those functions
+  // which have `Invalid` as type.
   UniquedQueue<BasicBlockNode *> CFEPQueue;
   for (auto *Node : llvm::post_order(&CG)) {
-    if (Node != RootNode)
+    if (Node == RootNode)
+      continue;
+
+    auto &Function = Binary.Functions[getBasicBlockPC(Node->BB)];
```
aleclearmind, line 1987: `operator[]` will create the entry if absent, which is not what we want. Switch to `at()`.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -461,6 +473,74 @@ CFEPAnalyzer<FunctionOracle>::CFEPAnalyzer(llvm::Module &M,
   }
 }
 
+static FunctionSummary
+initializeSummary(llvm::Module &M,
+                  const std::vector<GlobalVariable *> &ABIRegs,
```
aleclearmind, line 512: Use `ArrayRef<GlobalVariable>`

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -461,6 +473,74 @@ CFEPAnalyzer<FunctionOracle>::CFEPAnalyzer(llvm::Module &M,
...
+  const auto *Prototype = cast<RawFunctionType>(Function.Prototype.get());
+
+  for (auto *CSV : ABIRegs) {
+    Summary.ABIResults.ArgumentsRegisters[CSV] = RegisterState::No;
+    Summary.ABIResults.FinalReturnValuesRegisters[CSV] = RegisterState::No;
+  }
+
+  for (const auto &NTR : Prototype->Arguments) {
+    auto Name = ABIRegister::toCSVName(NTR.Location);
+    auto *CSV = cast<GlobalVariable>(M.getGlobalVariable(Name, true));
```
aleclearmind, line 534: `assert(CSV != nullptr)`

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -461,6 +473,74 @@ CFEPAnalyzer<FunctionOracle>::CFEPAnalyzer(llvm::Module &M,
...
+    Summary.ABIResults.ArgumentsRegisters[CSV] = RegisterState::Yes;
```
aleclearmind, line 535: Use `at()`

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -461,6 +473,74 @@ CFEPAnalyzer<FunctionOracle>::CFEPAnalyzer(llvm::Module &M,
   }
 }
 
+static FunctionSummary
+initializeSummary(llvm::Module &M,
```
aleclearmind, line 511: Rename to `importModelFunction`

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -461,6 +473,74 @@ CFEPAnalyzer<FunctionOracle>::CFEPAnalyzer(llvm::Module &M,
...
+                          model::Binary &Binary) {
+  using namespace llvm;
+
+  for (const model::Function &Function : Binary.Functions) {
+    if (Function.Type == model::FunctionType::Invalid)
+      continue;
+
+    // TODO: handle CABIFunctionType
+    if (!isa<model::RawFunctionType>(Function.Prototype.get()))
+      continue;
```
aleclearmind, line 572: Mmmh, what happens if we don't insert it?
Will we assume later on a default summary?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -1827,10 +1929,16 @@ bool EarlyFunctionAnalysis::runOnModule(Module &M) {
                                  nullptr);
   FunctionAnalysisResults Properties(std::move(DefaultSummary));
 
+  // Cache pre-population
+  Properties.initializeCache(M, ABIRegisters, Binary);
```
aleclearmind, line 2026: We don't call it cache anywhere else (I think).
What about simply `importModel`?
We could also consider embedding this functionality in the constructor.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -1827,10 +1929,16 @@ bool EarlyFunctionAnalysis::runOnModule(Module &M) {
...
+
   // Instantiate a CFEPAnalyzer object
   using CFEPA = CFEPAnalyzer<FunctionAnalysisResults>;
   CFEPA Analyzer(M, &GCBI, ABIRegisters, Properties, Binary);
 
+  // Re-create fake functions, should they exist
+  Analyzer.initializeFakeFunctions();
```
aleclearmind, line 2033: Why let the user call this? Make the function `private` and call it in the constructor.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -1156,6 +1239,18 @@ llvm::Function *CFEPAnalyzer<FO>::createFakeFunction(llvm::BasicBlock *Entry) {
   return FakeFunction.extractFunction();
 }
 
+template<class FO>
+void CFEPAnalyzer<FO>::initializeFakeFunctions() {
+  for (const auto &Function : Binary.Functions) {
+    if (Function.Type != model::FunctionType::Fake)
+      continue;
+
+    auto &Summary = Oracle.at(Function.Entry);
```
aleclearmind, line 1285: Uhm, I'd expect the Oracle not to have an entry associated to `Function.Entry` and `at()` should assert in such case.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Can `EntrypointsAnalyzer` work instead?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Correct, forgot this.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Nope, no assumption on a default summary for that specific `Entry`; if it is not inserted the corresponding `Summary` will be entry.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Why do you say? The Oracle should have an entry associated to that `Function.Entry` for sure at this point since the model has already been imported.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -719,36 +748,38 @@ finalizeModel(GeneratedCodeBasicInfo &GCBI,
...
+            auto TypedPrototype = Binary.recordNewType(std::move(Prototype));
+            Function->CallSitePrototype.insert({ Block.Start, TypedPrototype });
           }
         }
       }
     }
 
-    Function->CFG = Summary.CFG;
+    efa::FunctionMetadata FM(Function->Entry, Summary.CFG);
+    FM.verify(Binary);
```
aleclearmind, line 787: You're ignoring the return value.
Usually we have an overload with a `bool` stating whether you should assert, if you use that this is fine.

lib/EarlyFunctionAnalysis/FunctionMetadata.cpp
```diff
@@ -0,0 +1,382 @@
+/// \file ControlFlowGraph.cpp
```
aleclearmind, line 1: * Wrong name
* Any idea why github doesn't detect a moved file? It's hard to check for changes.

include/revng/EarlyFunctionAnalysis/CFGCollectionPass.h
```diff
@@ -0,0 +1,29 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "llvm/Pass.h"
+
+#include "revng/EarlyFunctionAnalysis/EarlyFunctionAnalysis.h"
+
+namespace EarlyFunctionAnalysis {
+
+class CFGCollectionPass : public llvm::ModulePass {
```
aleclearmind, line 13: Let's go for:

```c++
using CFGCollectionPass = EarlyFunctionAnalysis<false>;
using DetectABIPass = EarlyFunctionAnalysis<true>;
```

No need to have a separate passes.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -1973,54 +2034,30 @@ bool EarlyFunctionAnalysis::runOnModule(Module &M) {
...
-        }
-      }
-    }
+    // Finalize model
+    finalizeModel(GCBI, ABIRegisters, Properties, Binary);
   }
 
+  // Serialize function metadata, including the CFG, to the entry basic blocks
+  // in `root`.
+  serializeFunctionMetadata(GCBI, Properties, Binary);
```
aleclearmind, line 2055: I feel like `runInterproceduralAnalysis`, `interproceduralPropagation`, `finalizeModel` and `serializeFunctionMetadata` could become `Analyzer`'s methods and have no arguments at all.

aleclearmind (changes requested):

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
aleclearmind: Mmmh, so we assume no arguments, no return values, no clobbered registers?
Are you sure we won't get a default summary at a certain point?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
aleclearmind: ACK. If it's already there you don't need to set `Summary.Type = model::FunctionType::Fake`.
I'd avoid touching the model entirely in this function, you can iterate over each function present in the `Oracle` and, if it has `.Type == Fake`, then call `createFakeFunction`.

Otherwise this a bit confusing.

include/revng/EarlyFunctionAnalysis/BasicBlock.h
```diff
@@ -26,32 +26,32 @@ type: struct
     type: MetaAddress
   - name: CustomName
     doc: Optional custom name
-    type: Identifier
+    type: model::Identifier
     optional: true
```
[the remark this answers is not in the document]
antoniofrighetto: OK, dropping `CustomName` as well.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: I think that setting `Summary.Type` (within the local cache of EFA, not the model) is necessary here. In fact, we need to query the Oracle (via `isFakeFunction`, which checks `Summary.Type`) when re-scheduling an analysis for the fake function callers, if we got improved results for a specific function (recall that, when inserting the functions to be analyzed into the queue, we need to recompute its callers, if a caller turns out to be fake).

antoniofrighetto: Removed, fixed in `importModel`.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Note: this is being done.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: [commit [tmp](https://github.com/revng/revng/pull/125/commits/09ada2570036f6e8f3097118fe573320d8392e82)]: I had mixed feelings about making `finalizeModel` (formerly, `commitToModel`) an `Analyzer`'s method, since it is not completely related to the analysis per se. Nonetheless, given that it was using several common fields, I moved `finalizeModel`, together with `runInterproceduralAnalysis` and `interproceduralPropagation` as methods of `Analyzer`. Conversely, I felt more like that `serializeFunctionMetadata` should not be part of them (not related to the analysis at all). Also, I took the chance to simplify and refactor `EarlyFunctionAnalysis::runOnModule` and create a static function `createApproximateCallGraph` that returns the entry-points queue to be analyzed (decided not to bring it into `Analyzer` though, since it sounds more as a pre-phase to be done before instantiating the `Analyzer` object). What do you think?

antoniofrighetto: Removed the last part, since the lifetime of different objects is compromised if I move the creation of the call graph out in a static function. Need to think about this.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -2048,13 +2065,13 @@ bool EarlyFunctionAnalysis<ShouldAnalyzeABI>::runOnModule(Module &M) {
   } else {
     // Interprocedural analysis over the collected functions in post-order
     // traversal (leafs first).
-    runInterproceduralAnalysis(CFEPQueue, RootNode, Analyzer, Properties);
+    Analyzer.runInterproceduralAnalysis(CFEPQueue);
```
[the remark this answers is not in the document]
antoniofrighetto: - Sorry, as I have written in review [above](https://github.com/revng/revng/pull/125#discussion_r808053183), the process of dropping CFEP is being done. I have this in a local commit, haven't pushed it yet. I should have specified it here as well, I apologize.
- The queue is only basically being popped off, it contains all the functions collected ordered in post order traversal. I need it to pass it to the `Analyzer` somehow. Maybe it sounds better to move it as argument of the constructor of  `Analyzer`.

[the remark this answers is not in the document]
antoniofrighetto: OK, finalized first point in newly-pushed [cfep -> entryanalyzer](https://github.com/revng/revng/pull/125/commits/0610c1639cddbfc85018479fdc9770611d186d07) commit.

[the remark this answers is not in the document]
antoniofrighetto: Mind clarifying the second point, please? It seems that if we don't want to pass it as argument, all the call-graph creation needs to be done within `runInterproceduralAnalysis` method.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
aleclearmind: Ping?

antoniofrighetto: Preferred to double-check this. We won't get a default summary if nothing is imported from the model, as it happens to be the case here. However, the function exists in the model, so an EarlyFunctionAnalysis will be scheduled, therefore, after the analysis has run, such functions (not previously imported) will have their own `Summary`.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -1727,85 +1890,73 @@ OutlinedFunction CFEPAnalyzer<FO>::outlineFunction(llvm::BasicBlock *Entry) {
   return OutlinedFunction;
 }
 
-bool EarlyFunctionAnalysis::runOnModule(Module &M) {
-  Function &F = *M.getFunction("root");
+using BasicBlockQueueTy = UniquedQueue<BasicBlockNode *>;
```
aleclearmind, line 1893: Drop `Ty` suffix.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -1841,10 +1994,18 @@ bool EarlyFunctionAnalysis::runOnModule(Module &M) {
...
-      CFEPQueue.insert(Node);
+    if (Node == RootNode)
+      continue;
+
+    // The intraprocedural analysis will be scheduled only for those functions
+    // which have `Invalid` as type.
+    auto &Function = Binary.Functions.at(getBasicBlockPC(Node->BB));
+    if (Function.Type == model::FunctionType::Invalid)
+      FunctionEntrypointsQueue.insert(Node);
   }
```
aleclearmind, line 2009: `FunctionEntrypointsQueue` is used only by `Analyzer`. I think you can push it to become a member of its class and initialize it in the constructor or somewhere.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
aleclearmind: Why do you say that we'll analyze this function? If it has a non-`RawFunctionType` prototype (i.e., `CABIFunctionType`) but its type is not `Invalid`, we will not analyze it and we'll use the default summary for it, isn't it?

I'm not saying we need to change this, I'm trying to understand what the code does.

antoniofrighetto: I stand corrected, if the function's prototype is non-`RawFunctionType` and the type is not `Invalid`, the function will not be analyzed. If the `Oracle` is questioned on this function – for which no summary has been cached –, then the [default summary](https://github.com/revng/revng/blob/develop/lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp#L268) is used.

include/revng/Model/CallSitePrototype.h
```diff
@@ -0,0 +1,35 @@
...
+key:
+  - CallerBlockAddress
+TUPLE-TREE-YAML */
+
+#include "revng/Model/Generated/Early/CallSitePrototype.h"
+
+class model::CallSitePrototype : public model::generated::CallSitePrototype {
+public:
+  using generated::CallSitePrototype::CallSitePrototype;
+};
```
aleclearmind, line 33: Add `dump` and `verify` just as all the other parts of the model.
`verify` must be called by `Binary::verify`.

include/revng/EarlyFunctionAnalysis/FunctionMetadata.h
```diff
@@ -0,0 +1,46 @@
...
+class efa::FunctionMetadata : public efa::generated::FunctionMetadata {
+public:
+  using generated::FunctionMetadata::FunctionMetadata;
+
+public:
+  bool verify(model::Binary &Binary) const debug_function;
+  bool verify(model::Binary &Binary, bool Assert) const debug_function;
+  bool verify(model::Binary &Binary, model::VerifyHelper &VH) const;
+  void dumpCFG(model::Binary &Binary) const debug_function;
+};
```
aleclearmind, line 44: Add `dump` method.

lib/EarlyFunctionAnalysis/CollectFunctionsFromCalleesPass.cpp
```diff
@@ -0,0 +1,72 @@
...
+    auto It = llvm::find(Functions, &BB);
+    if (It != Functions.end())
+      continue;
+
+    uint32_t Reasons = GCBI.getJTReasons(&BB);
+    bool IsCallee = hasReason(Reasons, JTReason::Callee);
+
+    if (IsCallee) {
+      MetaAddress Entry = GCBI.getJumpTarget(&BB);
+      Binary.Functions[Entry].Type = model::FunctionType::Invalid;
```
aleclearmind, line 48: You're overriding what's already there. Introduce the entry only if not already there.

lib/EarlyFunctionAnalysis/CollectFunctionsFromUnusedAddressesPass.cpp
```diff
@@ -0,0 +1,124 @@
...
+    bool IsNotPartOfOtherCFG = VisitedBlocks.find(Entry) == VisitedBlocks.end();
+
+    if (not IsLoadAddress
+        and (IsUnusedGlobalData
+             || (IsMemoryStore and not IsPCStore and not IsReturnAddress))) {
+      // TODO: keep IsReturnAddress?
+      // Consider addresses found in global data that have not been used or
+      // addresses that are not return addresses and do not end up in the PC
+      // directly.
+      if (IsNotPartOfOtherCFG) {
```
aleclearmind, line 86: Squash this `if` in the parent `if`.

lib/EarlyFunctionAnalysis/FunctionMetadata.cpp
```diff
@@ -0,0 +1,382 @@
...
+        revng_abort();
+        break;
+      }
+    }
+  }
+
+  return Graph;
+}
+
+bool FunctionMetadata::verify(model::Binary &Binary) const {
```
aleclearmind, line 120: Take `const model::Binary &` in all `verify` methods.

include/revng/EarlyFunctionAnalysis/IRHelpers.h
```diff
@@ -0,0 +1,74 @@
...
+  using namespace llvm;
+
+  efa::FunctionMetadata FM;
+  const MDOperand &Op = MD->getOperand(0);
+  revng_assert(MD != nullptr && isa<MDString>(Op));
+
+  StringRef YAMLString = cast<MDString>(Op)->getString();
+  auto MaybeParsed = TupleTree<efa::FunctionMetadata>::deserialize(YAMLString);
+  revng_assert(MaybeParsed);
+  return std::move(MaybeParsed.get());
```
aleclearmind, line 29: `MaybeParsed->verify(true)`

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -1880,54 +2050,38 @@ bool EarlyFunctionAnalysis::runOnModule(Module &M) {
...
-        }
-      }
-    }
+    // Commit the results onto the model
+    Analyzer.finalizeModel();
   }
 
+  // Serialize function metadata, including the CFG, to the entry basic blocks
+  // in `root`.
+  serializeFunctionMetadata(M.getContext(), GCBI, Properties, Binary);
```
aleclearmind, line 2079: Consider making this a method of `Analyzer`.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -464,13 +493,117 @@ CFEPAnalyzer<FunctionOracle>::CFEPAnalyzer(llvm::Module &M,
...
+
+    BasicBlock *BB = GCBI.getBlockAt(Function.Entry);
+    std::string Buffer;
+    {
+      efa::FunctionMetadata FM(Function.Entry);
+      raw_string_ostream Stream(Buffer);
+      for (efa::BasicBlock Edge : CFG)
+        FM.ControlFlowGraph.insert(Edge);
+
+      serialize(Stream, FM);
```
aleclearmind, line 524: `FM.verify(true)`

aleclearmind (changes requested):

include/revng/Model/CallSitePrototype.h
aleclearmind: `FunctionMetadata::verify` can assume the whoel `model::Binary` verifies correctly.
`CallSitePrototype::verify` must not rely on `FunctionMetadata`, that's out of the model.
I think the logic you're describing is fine in `CallEdge::verify`.

antoniofrighetto: I'm not sure I've understood correctly. `Prototype` has been removed from `CallEdge::verify`, therefore now checking for prototype's validity ought to be done in `CallSitePrototype::verify`. Though, we need the `Start` MetaAddress of each `efa::BasicBlock` that is performing a direct/indirect call and the type of `Edge` for `CallSitePrototype::verify`, so I assume this may be done in `FunctionMetadata::verify` since in `model:Binary` we do not have the CFG anymore.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Took the freedom not to move this method this time, I felt like the serialization of metadata is completely independent from the whole analysis of the function. Lemme know if you have strong opinions on this though, and I'll move it.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Closing this, as `CABIFunctionType`'s prototype is now supported.

python/revng/cli/translate/pipelines/isolate-translate.yml
```diff
@@ -5,7 +5,12 @@ Steps:
    Pipes:
      - Name:             LLVMPipe
        UsedContainers: [module]
-       Passes: [detect-abi, isolate, invoke-isolated-functions]
+       Passes: [collect-functions-from-callees,
+                detect-abi,
+                collect-functions-from-unused-addresses,
+                detect-abi,
```
Cvelth, line 11: I noticed that abi detection pass is called twice. Is that an intentional decision? Why can’t we just detect all the functions before running it only once? Does the second level of detection depends on abi or something?

cc: @aleclearmind

include/revng/EarlyFunctionAnalysis/BasicBlock.h
```diff
@@ -24,34 +24,30 @@ type: struct
       End address of the basic block, i.e., the address where the last
       instruction ends
     type: MetaAddress
-  - name: CustomName
```
Cvelth, line 27: So, I guess we're getting rid of basic block names. I feel like there might be a subset of users who'd prefer being able to "mark" some of them for future reference, etc?

Maybe something like having a separate `sequence` in the model specifying "labels" for specific addresses (at that point, maybe we shouldn't even limit it to basic blocks only) worth considering in the future?

P. S. Going even crazier, it might be really cool to carry those "labels" on through the decompiler in some capacity and preserve them in like c-style `label:`s or even just comments - might make working with the decompiled code easier.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -612,98 +794,90 @@ finalizeModel(GeneratedCodeBasicInfo &GCBI,
...
-  for (auto &[PC, CallSites] : Summary.ABIResults.CallSites) {
-    if (PC == EntryPC) {
-      for (auto &[FuncArg, CSArg] :
-           zipmap_range(Summary.ABIResults.ArgumentsRegisters,
-                        CallSites.ArgumentsRegisters)) {
-        auto *CSV = FuncArg == nullptr ? CSArg->first : FuncArg->first;
-        auto RSFArg = FuncArg == nullptr ? RegState::Maybe : FuncArg->second;
-        auto RSCSArg = CSArg == nullptr ? RegState::Maybe : CSArg->second;
+  for (auto &[FuncArg, CSArg] :
+       zipmap_range(Callee.ArgumentsRegisters, CallSite.ArgumentsRegisters)) {
```
Cvelth, line 865: Maybe it's not strictly related to this PR (or more likely I'm just being way too paranoid), but the trauma of having unexpected problems with `zip`-ers is still way too fresh (there's a discussion with @drblallo somewhere in `~libsdg`). What I'm leading towards, maybe use cases like this should actually assert that both `lhs` and `rhs` containers have the same size - just to be double-safe.

include/revng/Model/Binary.h
```diff
@@ -59,6 +59,11 @@ type: struct
       type: SortedVector
       upcastable: true
       elementType: model::Type
+  - name: DefaultPrototype
```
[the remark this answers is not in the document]
aleclearmind: The default prototype needs to be materialized in the model. If we have a default, but it's not in the model, passes with read-only access to the model won't be able to use it.

include/revng/Model/Function.h
```diff
@@ -62,6 +56,11 @@ type: struct
       type: MutableSet
       elementType: model::FunctionAttribute::Values
     optional: true
+  - name: CallSitePrototypes
+    sequence:
```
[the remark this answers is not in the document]
aleclearmind: Uhm, the role of `CallSitePrototypes` is to tell you the prototype of the call site at a certain address (actually, the start address of the basic block ending with the call we're considering).
You can't have more than one prototype for that address.

If this does not answer your question, could you please rephrase it?

include/revng/Model/Binary.h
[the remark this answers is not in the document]
Cvelth: So, in essence, this is just `abi::defaultPrototype(MyCoolBinary.DefaultABI)` but cached?

include/revng/Model/Function.h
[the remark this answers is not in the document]
Cvelth: In my understanding, any function takes a predefined number of parameters (it uses whatever registers and stack part it uses - never more) - so it shouldn't technically matter where this address is called from - it should always have the same prototype, does it not?

python/revng/cli/translate/pipelines/isolate-translate.yml
aleclearmind: `detect-abi` goals is to obtain a prototype for all the function in the model that have an `Invalid` function type.
The idea is that you:
1. do a first pass (`collect-functions-from-callees`) collecting function looking at `call` instructions (which is a very reliable way to identify functions) ;
2. use `detect-abi` to discover their prototype (and their CFG)
3. then you do another pass (`collect-functions-from-unused-addresses`) which tries to identify other functions that are not target of direct calls (e.g., addresses we found in global data, which is typical for e.g., virtual tables). The reason for splitting these two collections is that, in this second phase, we ignore all the function entry points the we have previously detect to be part of another function.

Basically the idea is: try to find addresses that we're highly confident represent a function entry point, identify their CFG, and then have a less reliable source of function entry points which we filter in order to exclude entry points that, at the previous iteration, we identified as part of another function (and therefore it's likely to be a false positive).

Makes sense?

include/revng/EarlyFunctionAnalysis/BasicBlock.h
aleclearmind: Eh, you're right. We'd need something like `CallSitePrototypes` to specify these optional kind of information.
But I'd postpone this.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
aleclearmind: I'm all for extra assertions! :)

include/revng/Model/Binary.h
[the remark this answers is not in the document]
aleclearmind: Yeah, but the user can customize it, in case.

python/revng/cli/translate/pipelines/isolate-translate.yml
Cvelth: So what you are saying is that my suspicion about
```
Passes: [collect-functions-from-callees, collect-functions-from-unused-addresses, detect-abi]
```
begin equivalent to
```
Passes: [collect-functions-from-callees, detect-abi, collect-functions-from-unused-addresses, detect-abi]
```
is not wrong - it's just that we want to "mark" them different for now so that at some point we might want to process them differently. Am I understanding it right?

include/revng/Model/Binary.h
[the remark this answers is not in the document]
Cvelth: I see, closing this thread then. Thanks for the explanation.

include/revng/Model/Function.h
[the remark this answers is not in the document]
aleclearmind: Ah, this is only for *indirect function calls*, where you don't know the callee but you still need to know the prototype.

In the future, we might use this for variadic function calls too, where the callee is variadic but the call site has an explicit prototype.

In the future, we might also want to let the user override on the call site the prototype and emit a cast.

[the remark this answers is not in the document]
Cvelth: A wait, I was being stupid, right, we don't always have access to the callee! Thanks for the explanation!

python/revng/cli/translate/pipelines/isolate-translate.yml
aleclearmind: If you run `detect-abi` only once at the end, `collect-functions-from-unused-addresses` will not see the CFGs of the functions identified by `collect-functions-from-callees` and will not be able to blacklist any candidate, even if it will later be detect as part of another function.

But now that I think about it, since we only need the CFG, instead of running `detect-abi`, we could run only the part recovering the CFG.

Cvelth: Ah, wait, CFG was only recovered as a part of abi-detection? I had no idea :D

include/revng/EarlyFunctionAnalysis/CallEdge.h
```diff
@@ -0,0 +1,118 @@
...
+
+public:
+  bool verify() const debug_function;
+  bool verify(bool Assert) const debug_function;
+  bool verify(model::VerifyHelper &VH) const;
+  void dump() const debug_function;
+};
+
+inline model::TypePath getPrototype(const model::Binary &Binary,
+                                    MetaAddress FunctionAddress,
```
aleclearmind, line 64: `CallerFunctionAddress`?

include/revng/EarlyFunctionAnalysis/IRHelpers.h
```diff
@@ -0,0 +1,99 @@
...
+inline TupleTree<efa::FunctionMetadata>
+extractFunctionMetadata(llvm::BasicBlock *BB) {
+  auto *MDNode = BB->getTerminator()->getMetadata(FunctionMetadataMDName);
+  return detail::extractFunctionMetadata(MDNode);
+}
+
+/// \brief Given a Call instruction and the model type of its parent function,
+///        return the edge on the model that represents that call, or nullptr if
+///        this doesn't exist.
+inline std::pair<efa::CallEdge *, MetaAddress>
```
aleclearmind, line 51: Maybe return a `const efa::CallEdge &`

include/revng/EarlyFunctionAnalysis/IRHelpers.h
```diff
@@ -0,0 +1,99 @@
...
+    ParentFunction = llvmToModelFunction(Binary, *Call->getFunction());
+
+  if (not ParentFunction)
+    return nullptr;
+
+  const auto &[Edge, BlockAddress] = getCallEdge(Binary, Call);
+  if (not Edge)
+    return nullptr;
+
+  return getPrototype(Binary, ParentFunction->Entry, BlockAddress, *Edge).get();
```
aleclearmind, line 98: Uhm, I'd drop `.get()`.
You can return an invalid `TupleTreeReference` in case.

include/revng/Model/CallSitePrototype.h
```diff
@@ -0,0 +1,44 @@
...
+fields:
+  - name: CallerBlockAddress
+    doc: Address of the basic block of the call
+    type: MetaAddress
+  - name: Prototype
+    doc: Prototype
+    reference:
+      pointeeType: model::Type
+      rootType: model::Binary
+    optional: true
```
aleclearmind, line 25: I wouldn't make it `optional`.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -490,6 +491,39 @@ FEA::FunctionEntrypointAnalyzer(llvm::Module &M,
...
+  using namespace llvm;
+
+  for (const auto &Function : Binary->Functions) {
+    if (Function.Type == FunctionTypeValue::Fake)
+      continue;
+
+    auto &CFG = Properties.at(Function.Entry).CFG;
+    // Already serialized?
+    if (CFG.size() == 0)
+      continue;
```
aleclearmind, line 532: How is this possible?

aleclearmind (changes requested):

lib/FunctionIsolation/IsolateFunctions.cpp
antoniofrighetto: Could `FunctionIsolation: handle extra successors` work as commit message?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -207,14 +208,30 @@ struct FunctionSummary {
 /// methods to retrieve the stored information, acting as a oracle.
 class FunctionAnalysisResults {
 private:
+  llvm::Module &M;
+  ArrayRef<GlobalVariable *> ABIRegisters;
+  const TupleTree<model::Binary> &Binary;
+
   /// For each function, the result of the intraprocedural analysis
   std::map<MetaAddress, FunctionSummary> FunctionsBucket;
   FunctionSummary DefaultSummary;
```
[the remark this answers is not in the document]
antoniofrighetto: You mean to document where `TupleTree<model::Binary>` is used, right?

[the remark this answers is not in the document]
aleclearmind: No, `DefaultSummary`.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -1933,7 +2006,7 @@ bool EarlyFunctionAnalysis::runOnModule(Module &M) {
 
   // Create an over-approximated call graph of the program. A queue of all the
   // function entrypoints is maintained.
-  UniquedQueue<BasicBlockNode *> EntrypointsQueue;
+  BasicBlockQueue EntrypointsQueue;
```
aleclearmind, line 2069: I think we should push the ownership of `EntrypointsQueue` into `Analyzer`. The queue is a concept that only makes sense *within* the analyzer.

So, either push the initialization of the queue in the analyzer too, or make the analyzer accept a `set` instead of a queue.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -433,12 +453,14 @@ using FEA = FunctionEntrypointAnalyzer;
 FEA::FunctionEntrypointAnalyzer(llvm::Module &M,
                                 GeneratedCodeBasicInfo *GCBI,
                                 ArrayRef<GlobalVariable *> ABIRegs,
+                                BasicBlockQueue *EntrypointsQueue,
                                 FunctionAnalysisResults &Oracle,
-                                const TupleTree<model::Binary> &Binary) :
+                                TupleTree<model::Binary> &Binary) :
```
aleclearmind, line 461: `-collect-cfg` must work with a read-only model.
Maybe you can make this `const` and add a non-`const` `TupleTree<model::Binary> &Binary` argument to `finalizeModel` (which should probably be renamed to `commitToModel`)?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -1990,56 +2063,33 @@ bool EarlyFunctionAnalysis::runOnModule(Module &M) {
...
-      }
-    }
+    // Commit the results onto the model
+    Analyzer.finalizeModel();
   }
 
+  // Serialize function metadata, including the CFG, to the entry basic blocks
+  // in `root`.
+  serializeFunctionMetadata(M.getContext(), GCBI, Properties, Binary);
+
```
aleclearmind, line 2151: I think we should move this into the `if (!ShouldAnalyzeABI)` branch and run `opt -detect-abi -collect-cfg`.
We're duplicating some work, but we have a nice property:
* `-detect-abi` writes the model but not the IR.
* `-collect-cfg` writes the IR but not the model.
Would this work?

aleclearmind (changes requested):

include/revng/Support/IRHelpers.h
```diff
@@ -1363,6 +1363,14 @@ inline unsigned getMemoryAccessSize(llvm::Instruction *I) {
   return getPointeeSize(getPointer(I));
 }
 
+inline llvm::Value *hasDynamicSymbol(llvm::BasicBlock *BB) {
```
aleclearmind, line 1366: * Rename to `getDynamicSymbol`.
* Can we return directly the `StringRef` to the symbol name? Seems like the right thing to do.

One day, we'll need to split off `IRHelpers.h`.

lib/EarlyFunctionAnalysis/CollectFunctionsFromCalleesPass.cpp
```diff
@@ -36,6 +36,9 @@ static void collectFunctionsFromCallees(Module &M,
     if (Binary.Functions.find(Entry) != Binary.Functions.end())
       continue;
 
+    if (hasDynamicSymbol(&BB) != nullptr)
+      continue;
```
aleclearmind, line 40: Why? Comment.

lib/EarlyFunctionAnalysis/CollectFunctionsFromUnusedAddressesPass.cpp
```diff
@@ -71,6 +71,9 @@ void CFFUAImpl::collectFunctionsFromUnusedAddresses() {
     if (Binary.Functions.find(Entry) != Binary.Functions.end())
       continue;
 
+    if (hasDynamicSymbol(&BB) != nullptr)
+      continue;
```
aleclearmind, line 75: Why? Comment.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -2062,6 +2122,9 @@ bool EarlyFunctionAnalysis<ShouldAnalyzeABI>::runOnModule(Module &M) {
   // Cache pre-population of existing functions in the model
   Properties.importModel();
 
+  // Cache pre-population of imported dynamic symbols
+  Properties.importDynamicSymbols();
```
aleclearmind, line 2126: This is part of `importModel`, no reason to split it off.

lib/Lift/JumpTargetManager.cpp
```diff
@@ -393,7 +393,7 @@ JumpTargetManager::readFromPointer(Constant *Pointer, bool IsLittleEndian) {
       auto RelocationSize = model::RelocationType::getSize(Relocation.Type);
       if (LoadAddress == Relocation.Address and LoadSize == RelocationSize) {
         revng_assert(not StringRef(Function.name()).contains('\0'));
-        Result = { Function.name().str().str(), NewAPInt(Addend) };
+        Result = { Function.OriginalName, NewAPInt(Addend) };
```
aleclearmind, line 396: Make an isolated commit for this change. It's self-contained and beneficial *per se*.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -613,6 +616,44 @@ void FunctionAnalysisResults::importModel() {
...
+    auto [Entry, _] = getPC(&*BB.begin());
+
+    // Already imported?
+    if (FunctionsBucket.find(Entry) != FunctionsBucket.end())
+      continue;
+
+    llvm::StringRef SymbolName = extractFromConstantStringPtr(SymbolNameString);
+    revng_assert(SymbolName.size() != 0);
+
+    // TODO: change this to assert, dynamic function must have been imported
```
aleclearmind, line 639: I don't think we'll ever want to assert this. I foresee the possibility of having a symbol associated to `newpc` even without an entry in `ImportedDynamicFunctions`.
At very least, turn this is into a question.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -613,6 +616,44 @@ void FunctionAnalysisResults::importModel() {
...
+
+    auto *SymbolNameString = hasDynamicSymbol(&BB);
+    if (not SymbolNameString)
+      continue;
+
+    auto [Entry, _] = getPC(&*BB.begin());
+
+    // Already imported?
+    if (FunctionsBucket.find(Entry) != FunctionsBucket.end())
+      continue;
```
aleclearmind, line 634: Mmmh, so we're giving precedence to locally defined functions uh?
Might make sense, but please put a comment that clearly states this.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -613,6 +616,44 @@ void FunctionAnalysisResults::importModel() {
...
+    llvm::StringRef SymbolName = extractFromConstantStringPtr(SymbolNameString);
+    revng_assert(SymbolName.size() != 0);
+
+    // TODO: change this to assert, dynamic function must have been imported
+    auto It = Binary->ImportedDynamicFunctions.find(SymbolName.str());
+    if (It == Binary->ImportedDynamicFunctions.end())
+      continue;
+
+    if (not It->Prototype.isValid())
+      It->Prototype = Binary->DefaultPrototype;
```
aleclearmind, line 645: Can you make a function in `Binary.h`.
```
getPrototype(const model::Binary &Binary, const model::ImportedFunction &)
```
?
Otherwise this code will end up duplicated in a lot of places. If we have a central place, the day we decide to refactor this, it will be easy.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -613,6 +616,44 @@ void FunctionAnalysisResults::importModel() {
...
+
+    model::FunctionType::Values FunctionType;
+    if (It->Attributes.count(model::FunctionAttribute::NoReturn) != 0)
+      FunctionType = model::FunctionType::NoReturn;
+    else
+      FunctionType = model::FunctionType::Regular;
```
aleclearmind, line 651: We really need to drop `model::FunctionType` entirely.
Please put it up next in the task list.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -878,6 +914,54 @@ static MetaAddress getFinalAddressOfBasicBlock(llvm::BasicBlock *BB) {
...
+        // Indirect call
+        Destination = MetaAddress::invalid();
+        Type = efa::FunctionEdgeType::IndirectCall;
+      }
+    }
+  } else {
+    // Indirect call
+    Destination = MetaAddress::invalid();
+    Type = efa::FunctionEdgeType::IndirectCall;
+  }
```
aleclearmind, line 953:
```c++
 // Direct or indirect call?
  MetaAddress AddressPC = MetaAddress::invalid()
  auto FunctionIt = Binary->Functions.end();
  Something *SymbolNameString = nullptr;
  if (isa<ConstantStruct>(CalleePC)) {
    auto AddressPC = MetaAddress::fromConstant(CalleePC);
    FunctionIt = Binary->Functions.find(AddressPC);
    BasicBlock *OriginalEntry = GCBI->getBlockAt(AddressPC);
    SymbolNameString = hasDynamicSymbol(OriginalEntry);
  }

  if (FunctionIt != Binary->Functions.end())
    // The function exist within the model
    Destination = AddressPC;
    Type = efa::FunctionEdgeType::FunctionCall;
  } else if (SymbolNameString) {
    // The function exist within the imported functions
    // Dynamic function call. We set `Edge.Destination` to the callee's
    // MetaAddress, and fix it to `Invalid` later in `finalizeModel`.
    Destination = AddressPC;
    Type = efa::FunctionEdgeType::FunctionCall;
  } else {
    // Indirect call
    Destination = MetaAddress::invalid();
    Type = efa::FunctionEdgeType::IndirectCall;
  }
```

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -878,6 +914,54 @@ static MetaAddress getFinalAddressOfBasicBlock(llvm::BasicBlock *BB) {
...
+      Type = efa::FunctionEdgeType::FunctionCall;
+    } else {
+      // Does the function exist within the imported functions?
+      auto *OriginalEntry = GCBI->getBlockAt(AddressPC);
+      SymbolNameString = hasDynamicSymbol(OriginalEntry);
+      if (SymbolNameString) {
+        // Dynamic function call. We set `Edge.Destination` to the callee's
+        // MetaAddress, and fix it to `Invalid` later in `finalizeModel`.
+        Destination = AddressPC;
+        Type = efa::FunctionEdgeType::FunctionCall;
```
aleclearmind, line 942: This seems plain wrong. You should not set `CallEdge::Destination` but `CallEdge::DynamicFunction`.
This is quite fundamental.

aleclearmind (changes requested):

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Uhmm, I think that's not exactly what I wanted to achieve here. I drop it, but now that I think about it, does it make sense to serialize function metadata only if `Function.Type != FunctionTypeValue::Fake` and  `Function.Type != FunctionTypeValue::Invalid`? If for some reason is `Invalid`, we should not want to serialize the CFG, should we?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -1162,6 +1258,17 @@ FunctionEntrypointAnalyzer::createFakeFunction(llvm::BasicBlock *Entry) {
...
+  for (const auto &Function : Binary->Functions) {
+    if (Function.Type != FunctionTypeValue::Fake)
+      continue;
+
+    auto &Summary = Oracle.at(Function.Entry);
+    revng_assert(Summary.Type == FunctionTypeValue::Fake);
+    Summary.FakeFunction = createFakeFunction(GCBI->getBlockAt(Function.Entry));
+  }
+}
+
```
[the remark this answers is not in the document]
aleclearmind: You have a `model::Function`, you get its type and its address and create an entry into the EFA cache.
This really sounds like importing to me.
Please merge them.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Is it really a demanding property that we want to have? I ask this because this would imply to always run `--detect-abi` alongside `--collect-cfg`, if we want to write both on the IR and on the model, and I don't think we should want to have this distinction at this granularity. AFAIC, EFA has been designed to do both, but doing so, we would split the two things at the extent in which part of the work would be duplicated (as you noted). Furthermore, as far as we conceived this, CFG collection-only should be a subset of what `--detect-abi` does, thus --detect-abi` should be allowed to write the IR as well.

include/revng/EarlyFunctionAnalysis/IRHelpers.h
aleclearmind: This is just a wrapper around `getPrototype`. For API coherency, return whatever `getPrototype` returns (`TypePath`).
In case of failure, you can return an empty one.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: OK, then, since `importDynamicSymbols` iterates over the BBs in `root`, does it make sense to still have its own function `importDynamicSymbols`, but as private member called by `importModel`?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
aleclearmind: > Is it really a demanding property that we want to have?

Yes. I really want a clear distinction between those writing the model and those writing the IR.

> I ask this because this would imply to always run --detect-abi alongside --collect-cfg

No, you're still thinking in terms of one linear pipeline. But `--detect-abi` could be run on its own just to produce effects on the model and then not run anything after that.

> part of the work would be duplicated

This is just a performance issue that we can tackle in case it becomes an actual bottleneck.

> CFG collection-only should be a subset of what --detect-abi does, thus --detect-abi` should be allowed to write the IR as well

It's a subset only in terms of the computations it makes, not in terms of where it writes its results.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
aleclearmind: You can merge them. You have a bit this tendency to create a new function for each new feature you implement (`importDynamicSymbols` and `initializeFakeFunctions`) but there's not much point to it.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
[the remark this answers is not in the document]
antoniofrighetto: Sorry, I recalled only now – while changing this – the actual reason behind this. `createFakeFunction` is part of `Analyzer` method. Originally, I had tried to isolate this and make `createFakeFunction` a separate static function, but no success on it (it requires too much stuff that is part of `FunctionEntrypointAnalyzer` object), that's why I eventually split the two things.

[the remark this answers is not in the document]
aleclearmind: I don't understand.
Can't you move the body of `initializeFakeFunctions` at the end of `importModel`?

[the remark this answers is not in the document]
antoniofrighetto: I cannot. `importModel` is a public method of the local EFA cache (`FunctionAnalysisResults`), whereas `initializeFakeFunctions` needs to be a method of `FunctionEntrypointAnalyzer`, since it needs to call `FunctionEntrypointAnalyzer::createFakeFunction` to create the fake function.

include/revng/EarlyFunctionAnalysis/IRHelpers.h
antoniofrighetto: Alright, fixed it.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Merged!

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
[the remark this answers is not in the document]
aleclearmind: This design is a bit smelly.
Anyway, do not read the model then, read the list of functions tracked by the oracle and create a function for each fake function in there.
Otherwise we're importing the model in two steps, which is confusing.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Now that you make me think about it, I'm not sure about this decision and why I exactly I decided to do so. Might make more sense removing this check. Works?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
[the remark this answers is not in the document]
antoniofrighetto: Uhmmm, no, I need to read the model. I don't have a list of functions tracked by the oracle. Note that this is actually done also in `FunctionEntrypointAnalyzer::recoverCFG`, `FunctionEntrypointAnalyzer::interproceduralPropagation` and `FunctionEntrypointAnalyzer::finalizeModel`.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -461,6 +485,76 @@ FEA::FunctionEntrypointAnalyzer(llvm::Module &M,
...
+  using RegisterState = abi::RegisterState::Values;
+
+  FunctionSummary Summary(Type,
+                          { ABIRegisters.begin(), ABIRegisters.end() },
+                          ABIAnalyses::ABIAnalysesResults(),
+                          {},
+                          0,
+                          nullptr);
+
+  auto Layout = abi::FunctionType::Layout::make(Prototype);
```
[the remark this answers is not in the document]
antoniofrighetto: Uhmmm, no, I need to read the model. I don't have a list of functions tracked by oracle, there's `EntrypointsQueue` but we only enqueue functions with `Invalid` FunctionType. Note that, reading from model is also done in `recoverCFG`, `interproceduralPropagation`, and `finalizeModel` (all methods of `Analyzer` object).

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
aleclearmind: Who takes precedence? Local functions or dynamic functions?
That's the decision to make.

include/revng/Support/IRHelpers.h
antoniofrighetto: 1) OK.
2) I think it makes more sense to return a `Value*`, easier to operate with, and there is already a dedicated method `extractFromConstantStringPtr` in `IRHelpers.h` responsible for returning the symbol name, if `Value*` exists.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -207,14 +208,30 @@ struct FunctionSummary {
...
+  const TupleTree<model::Binary> &Binary;
+
   /// For each function, the result of the intraprocedural analysis
   std::map<MetaAddress, FunctionSummary> FunctionsBucket;
   FunctionSummary DefaultSummary;
 
 public:
-  FunctionAnalysisResults(FunctionSummary DefaultSummary) :
+  FunctionAnalysisResults(llvm::Module &M,
+                          ArrayRef<GlobalVariable *> ABIRegisters,
```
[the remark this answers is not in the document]
antoniofrighetto: I would actually leave `ABIRegisters`. We are using `ABIRegisters` throughout the whole EFA codebase (not just in this case above), in GCBI as well, and it was previously used by SA too.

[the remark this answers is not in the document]
aleclearmind: It's not an array of registers, it's an array of CSVs. If it's broken elsewhere those parts needs to be fixed too. If it was broken in the past, it's not relevant.

However, we can postpone this if you feel it makes rebase hard.

include/revng/Support/IRHelpers.h
aleclearmind: There's only one usage of this function and the first thing you do is extracting the string. How is returning a `Value *` easier to operate?
And I think this is basically what every user of this function will want: get the name so they can go to the model and extract information about the symbol.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
aleclearmind: Makes sense, then please make sure the code reflects this and add a comment about this so we know it was intentional and not something we overlooked.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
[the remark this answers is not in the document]
antoniofrighetto: My previous comment sounded actually more, say, historically-oriented, and I realize it is pretty much pointless, considering that we have re-written SA for a reason! The only thing that I'd like to point out is that, IMHO, *Registers* (its broader concept) sounds more descriptive, and makes it more easily understandable even by those who are not familiar w/ revng concepts and wish to read the code (not even among all of us – at least from what I could see – the concept of CSVs is fully known!). But I also agree that, to be strict to the concept, CSV is clearly more correct than Registers. Not a problem to change and for the rebase at all!

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Sounds good!

include/revng/Support/IRHelpers.h
antoniofrighetto: OK, so you're saying to make `getDynamicSymbol` directly call `extractFromConstantStringPtr`, right? Should `extractFromConstantStringPtr` still exist at this point or should we drop it?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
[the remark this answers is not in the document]
aleclearmind: I'm pretty sure before approaching reading `EFA` you need *at least* to understand what a CSV is.
We're not writing code for the benefit of bystanderds `:)`

include/revng/Support/IRHelpers.h
aleclearmind: I think that's either used or can be used in other situations where we have constant strings and we want to get to the data easily (e.g., assembly). Let's keep it.

include/revng/EarlyFunctionAnalysis/IRHelpers.h
antoniofrighetto: There's a case (line 58) in which I'm returning `return { nullptr, BlockAddress };`. I'm not sure, is there a way to return a CallEdge invalid if we want to return a reference?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Fixed this by anticipating `CE->DynamicFunction = SymbolName` from `finalizeModel` to `collectDirectCFG`. `finalizeModel`  accomplishes the creation of the prototype for indirect calls (that ought to be done there).

include/revng/Support/IRHelpers.h
antoniofrighetto: In `include/revng/EarlyFunctionAnalysis/CallEdge.h`, now there's `inline model::TypePath getPrototype(const model::Binary &Binary, MetaAddress CallerFunctionAddress, MetaAddress CallerBlockAddress, const efa::CallEdge &Edge);` (was previously [this one](https://github.com/revng/revng/blob/develop/include/revng/Model/Binary.h#L131), now moved in `efa` namespace). Are we sure we still want to do another specification of this function?

aleclearmind: Please don't mark as resolved discussions where you're asking a question.

If someones uses `hasDynamicSymbol`, it needs to have the characteristics we discussed... if no one uses it, I guess it can be dropped.

antoniofrighetto: Sorry, there must have been a mistake, this issue has been fixed. The question I asked is for the review below, not for this one. I copy and paste my comment below.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: In `include/revng/EarlyFunctionAnalysis/CallEdge.h`, now there's `inline model::TypePath getPrototype(const model::Binary &Binary, MetaAddress CallerFunctionAddress, MetaAddress CallerBlockAddress, const efa::CallEdge &Edge);` (was previously [this one](https://github.com/revng/revng/blob/develop/include/revng/Model/Binary.h#L131), now moved in `efa` namespace). Are we sure we still want to do another specification of this function?

include/revng/Support/IRHelpers.h
aleclearmind: ACK

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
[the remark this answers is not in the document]
antoniofrighetto: This has been fixed by making `importModel` a method of `FunctionEntrypointAnalyzer` (and not of the cache). As a result:
1) Now the local EFA cache is constructed as "dumb", like empty (as it is being done in the current version of EFA in develop).
2) `Analyzer.importModel` is a public method called in order to import prototype of existing functions from model, re-create fake functions (`initializeFakeFunctions` has been dropped, no need for a new function) and to import dynamic functions as well.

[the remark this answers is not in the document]
aleclearmind: Sounds good!

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: @aleclearmind, what shall we do with this? I think the change is really very minor to move `It->Prototype = Binary->DefaultPrototype;` into a function `getPrototype` (which already exists, btw). Also, there's no `model::ImportedFunction`, I guess you meant `model::DynamicFunction`, how would it be used inside the function? I'm just retrieving the prototype of a dynamic function by its name here.

aleclearmind: I think we need a function to make sure we don't duplicate code.

antoniofrighetto: OK, since we need to query `Binary` by Name, and we are also using the `model::DynamicFunction` (for its attributes), if the prototype is valid, I came up with this:
```cpp
inline std::pair<model::TypePath, const model::DynamicFunction &>
getPrototype(const model::Binary &Binary, std::string Name) {
  model::TypePath Result;

  auto It = Binary.ImportedDynamicFunctions.find(Name);
  if (It != Binary.ImportedDynamicFunctions.end()) {
    Result = It->Prototype;
    if (not Result.isValid())
      Result = Binary.DefaultPrototype;
  }

  return { Result, *It };
}
```
Not sure if returning a pair as such is the best approach though.

aleclearmind: I wouldn't push the look up of the dynamic function by name here.
I'd just do `getPrototype(const model::Binary &Binary, model::DynamicFunction &)`.

antoniofrighetto: Then, we `getPrototype` would simply return `Binary.DefaultPrototype`?

aleclearmind:
```c++
model::TypePath
getPrototype(const model::Binary &Binary, model::DynamicFunction &X) {
  if  (X.Prototype.isValid())
    return X.Prototype;
  else
    return Binary.DefaultPrototype;
}
```

antoniofrighetto: LG, fixed.

lib/FunctionIsolation/IsolateFunctions.cpp
aleclearmind: Please don't mark resolved notes for myself.
I need to check this before merging.

aleclearmind:
Merged in b495499. Thanks!

Issue #124, opened 2022-01-21
Build failure with pyparsing version >2.4.7


dramthy:
(Py3.revng) vscode ➜ ~/orchestra (master) $ orc install revng
[+] INFO - Executing Action install (build) of revng@optimized
[+] INFO - The following script is going to be executed:
cmake --build "$BUILD_DIR" ${JOBS:+-j$JOBS}
cmake --install "$BUILD_DIR"
if test "$RUN_TESTS" -eq 1; then
  cd "$BUILD_DIR"
  ctest --output-on-failure ${JOBS:+-j$JOBS}
  cd -
fi
[59/413] Generating DeadRegisterArgumentsOfFunction.h
FAILED: lib/EarlyFunctionAnalysis/ABIAnalyses/MFIGraphs/DeadRegisterArgumentsOfFunction.h 
cd /home/vscode/orchestra/build/revng/optimized/lib/EarlyFunctionAnalysis/ABIAnalyses/MFIGraphs && /home/vscode/orchestra/sources/revng/scripts/monotone_framework_lattice.py /home/vscode/orchestra/sources/revng/lib/EarlyFunctionAnalysis/ABIAnalyses/MFIGraphs/ABIAnalysis.template /home/vscode/orchestra/sources/revng/lib/EarlyFunctionAnalysis/ABIAnalyses/MFIGraphs/DeadRegisterArgumentsOfFunction.dot > DeadRegisterArgumentsOfFunction.h
Traceback (most recent call last):
  File "/home/vscode/orchestra/sources/revng/scripts/monotone_framework_lattice.py", line 242, in <module>
    sys.exit(main())
  File "/home/vscode/orchestra/sources/revng/scripts/monotone_framework_lattice.py", line 226, in main
    generated_code = process_graph(
  File "/home/vscode/orchestra/sources/revng/scripts/monotone_framework_lattice.py", line 165, in process_graph
    _ = monotone_framework.check_lattice(lattice)
  File "/home/vscode/orchestra/sources/revng/scripts/monotone_framework.py", line 109, in check_lattice
    assert bottom is None
AssertionError
[60/413] Generating DeadReturnValuesOfFunctionCall.h
FAILED: lib/EarlyFunctionAnalysis/ABIAnalyses/MFIGraphs/DeadReturnValuesOfFunctionCall.h 
[...]

PR #126, opened 2022-01-21
Check pyparsing version at build time

fcremo:
Ensures that pyparsing 2.4.7 is installed when building to avoid issues like #124


aleclearmind:
Merged in 31354a4cd683e6f6207c4d8b7da7d61513383daa. Thanks!

PR #127, opened 2022-02-01
Model: reserve names defined in stdint.h

pfez:
This commit adds a bunch of type names and macro names to the set of reserved keywords that cannot be used for names in the revng Model. These are the names defined in the stdint.h standard C header, that is included automatically by the decompiler in decompiled code.

We need to reserve these names, otherwise it would be possible to add things to the revng Model with names that clash with the names defined in stdint.h, which would in turn break recompilation of decompiled code because of conflicting definitions.


aleclearmind:
Merged in 29c50fa. Thanks!

PR #129, opened 2022-02-02
IDB importer

fcremo:
**Note**: this PR needs our fork of python-idb:
```
python3 -m pip install git+https://github.com/revng/python-idb.git@master
```

### DIFF
--- a/cmake/Common.cmake
+++ b/cmake/Common.cmake
@@ -188,7 +188,7 @@ include("${CMAKE_CURRENT_LIST_DIR}/TupleTreeGenerator.cmake")
 
 function(check_python_requirements)
   execute_process(
-    COMMAND pip3 freeze
+    COMMAND pip3 list --format freeze
     RESULT_VARIABLE RETURNCODE
     OUTPUT_VARIABLE PYTHON_PACKAGES
     ERROR_FILE "/dev/null")
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -130,3 +130,9 @@ python_module(TARGET_NAME revng-python-commands_registry MODULE_FILES
 #
 python_module(TARGET_NAME revng-python-llvm_pipeline MODULE_FILES
               "revng/cli/llvm_pipeline.py")
+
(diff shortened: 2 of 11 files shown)


aleclearmind:
As per our discussion yesterday:
* [ ] Let's switch to an approach more focused on `model::QualifiedType`
* [ ] Let's have a dictionary that can be queried through ordinal or name (pay attention not to have distinct entries for an ordinal and a name that are actually the same)
* [ ] Factor out building the `IDBPython` object
* [ ] Please investigate the `enumerate_entrypoints` situation. Try to do a run asserting it returns either zero or one results. Hardcoding `_start` is not great. If they return more than one results, we need to look into them, decide what's going on and properly comment the code about that.
* [ ] Yesterday we made a change that seemed beneficial/innocuous (i.e., dropping `idb_types_to_revng_types`) but it was actually bad! We need a test ensuring that we properly handle recursive types. A linked list example should do, we already have such an example in revng-qa ([linked_list.c](https://github.com/revng/revng-qa/blob/master/tests/analysis/Decompilation/dynamic_native/linked_list.c#L8)):
    1. import it in IDA with debug info;
    2. make sure it properly recognizes the linked list;
    3. make sure we test the corresponding IDB;
* [ ] Integrate `convert-idb` into the new organization of python tools. This means it must be `revng model import idb`. So you need to do something [like this](https://github.com/revng/revng/blob/develop/python/revng/cli/llvm_pipeline.py#L10):
    ```python
    super().__init__(("model", "import", "idb"), "...")
    ```

fcremo:
Re: bullet point 2, I don't think it's possible, we thought it would be because we only looked at `TypedefTypeData`, but other types don't have `ord` or `name`.
I have a commit ready to push that addresses the other points.

aleclearmind:
Converting this to draft. Will be handed over to @dwarfyresearch once he's done with PE/COFF.

dwarfyresearch:
Switching to the https://github.com/revng/revng/pull/199.

aleclearmind:
Closing this, superseeded by https://github.com/revng/revng/pull/199

PR #130, opened 2022-02-07
Promote PC CSVs

AlviseDeFaveri:
We can set the PC to the expected value after each call insteadof creating an `opaquepc()` call.

### DIFF
--- a/lib/FunctionIsolation/PromoteCSVs.cpp
+++ b/lib/FunctionIsolation/PromoteCSVs.cpp
@@ -5,6 +5,8 @@
 // This file is distributed under the MIT License. See LICENSE.md for details.
 //
 
+#include "llvm/ADT/STLExtras.h"
+
 #include "revng/ADT/GenericGraph.h"
 #include "revng/FunctionIsolation/PromoteCSVs.h"
 #include "revng/FunctionIsolation/StructInitializers.h"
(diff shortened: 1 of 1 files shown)


AlviseDeFaveri:
NOTE: this breaks things for decompilation, we are investigating the problem. Please don't merge.

lib/FunctionIsolation/EnforceABI.cpp
```diff
@@ -343,13 +328,9 @@ void EnforceABIImpl::handleRegularFunctionCall(CallInst *Call) {
 
   // Generate the call
   IRBuilder<> Builder(Call);
-  CallInst *NewCall = generateCall(Builder, Callee, Block, *CallSite);
```
andrealinux1, line 346: This is an erroneous removal of the call to the isolated function, therefore they disappear from the IR and therefore in the decompiled code

AlviseDeFaveri: gotcha!

aleclearmind: Undraft and ping me once this is fixed.

lib/FunctionIsolation/PromoteCSVs.cpp
```diff
@@ -89,15 +99,16 @@ PromoteCSVs::PromoteCSVs(Module *M, const GeneratedCodeBasicInfo &GCBI) :
   // Record existing initializers
   for (GlobalVariable *CSV : GCBI.csvs()) {
```
aleclearmind, line 100: Try `llvm::concat_range`.

aleclearmind:
Rebase and update PR title please.

lib/FunctionIsolation/PromoteCSVs.cpp
AlviseDeFaveri: fixed

aleclearmind:
Merged in 7593d4d. Thanks!

PR #131, opened 2022-02-09
Implement unified function type description system

Cvelth:


lib/ABI/FunctionType.cpp
```diff
@@ -613,12 +647,60 @@ convertToCABIFunctionType(const model::RawFunctionType &Function,
...
+  for (const model::StructField &Field : StackStruct->Fields) {
+    typename Layout::Argument::StackSpan StackSpan;
+    StackSpan.Offset = Field.Offset;
+
+    auto MaybeSize = Field.Type.size();
+    revng_assert(MaybeSize.has_value());
+    StackSpan.Size = MaybeSize.value();
+
+    ResultArguments.emplace_back().Stack = std::move(StackSpan);
+  }
```
aleclearmind, line 688: Can we have a single argument for all of these?

Cvelth: What do you mean? This creates `StackStruct->Fields.size()` new arguments.

Cvelth: Done: https://github.com/revng/revng/pull/131/commits/7d430547c41935a010d2ff9cc708475f1f2590d8.

include/revng/ABI/FunctionType.h
```diff
@@ -0,0 +1,94 @@
...
+
+struct Layout;
+
+/// Extracts the information about argument and return value location layout
+/// from the \param Function.
+Layout layout(const model::RawFunctionType &Function);
+
+/// Extracts the information about argument and return value location layout
+/// from the \param Function.
+Layout layout(const model::CABIFunctionType &Function);
```
aleclearmind, line 36: Promote to `static` "factory" methods. This also saves us from `friend`.

include/revng/ABI/FunctionType.h
```diff
@@ -0,0 +1,94 @@
...
+public:
+  struct ReturnValueRegisters {
+    llvm::SmallVector<model::Register::Values, 2> Registers;
+  };
+
+  struct Argument : public ReturnValueRegisters {
+  public:
+    struct StackSpan {
+      size_t Offset;
+      size_t Size;
```
aleclearmind, line 51: Let's use `uint64_t` just to be sure.

include/revng/ABI/FunctionType.h
```diff
@@ -0,0 +1,94 @@
...
+public:
+  bool verify() const;
+  size_t argumentRegisterCount() const;
+  size_t returnValueRegisterCount() const;
+
+public:
+  const llvm::SmallVector<Argument, 4> Arguments;
+  const ReturnValueRegisters ReturnValue;
+  const llvm::SmallVector<model::Register::Values, 24> CalleeSavedRegisters;
+  const uint64_t FinalStackOffset;
```
aleclearmind, line 80: Using `const` makes the object non-assignable, you sure you don't want to pass around `const` references instead?
Or maybe go for data hiding.

include/revng/ABI/FunctionType.h
```diff
@@ -0,0 +1,94 @@
...
+namespace abi::FunctionType {
+
+/// Best effort `CABIFunctionType` to `RawFunctionType` conversion.
+///
+/// If `ABI` is not specified, `TheBinaryToAddSupportTypesTo.DefaultABI`
+/// is used instead.
+std::optional<model::CABIFunctionType>
+tryConvertToCABI(const model::RawFunctionType &Function,
+                 model::Binary &TheBinaryToAddSupportTypesTo,
+                 std::optional<model::ABI::Values> ABI = std::nullopt);
```
aleclearmind, line 18: The fact that we return a `optional<CABIFunctionType>` but we need to get a non-`const` `model::Binary` is suspicious.
It seems to me we're risking to modify `Binary` but then the user might decide to discard the returned function type but it will still have changes on `model::Binary`.

I'd just return a `TypePath` or a `model::CABIFunctionType *` to a type that has been inserted.

Also:
```suggestion
std::optional<model::CABIFunctionType>
tryConvertToCABI(const model::RawFunctionType &Function,
                 model::Binary &TheBinaryToAddSupportTypesTo,
                 model::ABI::Values ABI = model::ABI::Invalid);
```

include/revng/ABI/FunctionType.h
```diff
@@ -0,0 +1,94 @@
...
+
+/// Best effort `RawFunctionType` to `CABIFunctionType` conversion.
+///
+/// \note: this convertion is loosy since there's no way to represent some types
+/// in `RawFunctionType` in a reversible manner.
+std::optional<model::RawFunctionType>
+tryConvertToRaw(const model::CABIFunctionType &Function,
+                model::Binary &TheBinaryToAddSupportTypesTo);
```
aleclearmind, line 26: Unlike `tryConvertToCABI`, I'd expect this to always succeed. In what situation it doesn't?
Maybe we need to turn that into an assertion.

lib/ABI/FunctionType.cpp
```diff
@@ -95,6 +98,13 @@ static model::QualifiedType getTypeOrDefault(const model::QualifiedType &Type,
     return buildType(Register, Binary);
 }
 
+struct InternalLayout {
+  std::decay<decltype(Layout::Arguments)>::type Arguments;
+  std::decay<decltype(Layout::ReturnValue)>::type ReturnValue;
+  std::decay<decltype(Layout::CalleeSavedRegisters)>::type CalleeSavedRegisters;
+  std::decay<decltype(Layout::FinalStackOffset)>::type FinalStackOffset;
+};
```
aleclearmind, line 106: Why do we need this instead of just using `Layout`?

lib/ABI/FunctionType.cpp
```diff
@@ -278,6 +285,36 @@ class ConvertionHelper {
...
+    }
+
+    DistributedArgument RetValue = distributeReturnValue(Function.ReturnType);
+    revng_assert(RetValue.SizeOnStack == 0);
+    Result.ReturnValue.Registers = std::move(RetValue.Registers);
+
+    Result.CalleeSavedRegisters.resize(AT::CalleeSavedRegisters.size());
+    llvm::copy(AT::CalleeSavedRegisters, Result.CalleeSavedRegisters.begin());
+
+    Result.FinalStackOffset = 0; // TODO: calculate this offset properly.
```
aleclearmind, line 313: Mmmh, is this because we don't handle stack arguments yet?
At least set it to `model::Architecture::getCallPushSize`, which is the right thing to do for the ABIs where stack arguments cleanup responsibility is left to the caller.

lib/ABI/FunctionType.cpp
```diff
@@ -613,12 +650,98 @@ convertToCABIFunctionType(const model::RawFunctionType &Function,
...
+  });
+  return Layout(std::move(R.Arguments),
+                std::move(R.ReturnValue),
+                std::move(R.CalleeSavedRegisters),
+                std::move(R.FinalStackOffset));
+}
+
+Layout layout(const model::RawFunctionType &Function) {
+  // Lay register arguments out.
+  std::decay<decltype(Layout::Arguments)>::type ResultArguments;
```
aleclearmind, line 673: This function heavily uses `std::decay` and `typename` for local vars, not very readable.
Either put the actual type or put some `using`s at the beginning.

lib/ABI/FunctionType.cpp
```diff
@@ -613,12 +650,98 @@ convertToCABIFunctionType(const model::RawFunctionType &Function,
...
+        return false;
+
+  // Verify return values
+  LookupHelper.clear();
+  for (model::Register::Values Register : ReturnValue.Registers)
+    if (!VerificationHelper(Register))
+      return false;
+
+  if (CalleeSavedRegisters.empty())
+    return false;
```
aleclearmind, line 729: Why?

lib/ABI/FunctionType.cpp
```diff
@@ -613,12 +650,98 @@ convertToCABIFunctionType(const model::RawFunctionType &Function,
...
+bool Layout::verify() const {
+  model::Architecture::Values ExpectedArch = model::Architecture::Invalid;
+  std::unordered_set<model::Register::Values> LookupHelper;
+  auto VerificationHelper = [&](model::Register::Values Register) -> bool {
+    if (!LookupHelper.emplace(Register).second)
+      return false;
+    if (ExpectedArch == model::Architecture::Invalid)
+      ExpectedArch = model::Register::getArchitecture(Register);
+    else if (ExpectedArch != model::Register::getArchitecture(Register))
+      return false;
```
aleclearmind, line 710:
```suggestion
    // Ensure each register is present only once
    if (!LookupHelper.emplace(Register).second)
      return false;
      
    // Ensure all the registers belong to the same architecture
    if (ExpectedArch == model::Architecture::Invalid)
      ExpectedArch = model::Register::getArchitecture(Register);
    else if (ExpectedArch != model::Register::getArchitecture(Register))
      return false;
```

lib/ABI/FunctionType.cpp
```diff
@@ -278,6 +285,36 @@ class ConvertionHelper {
...
+    InternalLayout Result;
+
+    size_t CurrentOffset = 0;
+    DistributedArguments Arguments = distributeArguments(Function.Arguments);
+    revng_assert(Arguments.size() == Function.Arguments.size());
+    for (size_t Index = 0; Index < Arguments.size(); ++Index) {
+      auto &Current = Result.Arguments.emplace_back();
+      Current.Registers = std::move(Arguments[Index].Registers);
+      if (Arguments[Index].SizeOnStack != 0) {
+        /// \todo maybe some kind of alignment considerations are needed here.
```
aleclearmind, line 298: We do not use doxygen within function bodies

```suggestion
        // TODO: maybe some kind of alignment considerations are needed here.
```

aleclearmind (changes requested):

include/revng/ABI/FunctionType.h
Cvelth: Actually, they can just be constructors at this point! They used to return an `std::optional` but since we got rid of that, no real reason for them to be free functions.

include/revng/ABI/FunctionType.h
Cvelth: 1. Is a really good point I haven't thought about earlier, I feel like that can be fixed by temporarily storing "support types" internally before adding them to the binary at the very end. Should I apply the same approach to the primitive types, or is it fine if this registers a new primitive type even when failing?

2. Yeah, that's a good point and an easy fix, will do.

3. I guess I like the extra explicitness `std::optional` provides, this works too, though.

include/revng/ABI/FunctionType.h
Cvelth: 1. An ABI doesn't provide a register for return values but function returns something (can be an assert)
2. `model::type::size()` returns `std::nullopt` (also an easy assert)
3. The result of `distributeReturnValue` call providing number of registers insufficient to fit whatever type this function returs (pretty much, when `distributeReturnValue` fails in an unclear manner), (I guess this can also be an assert, since it should never happen, save for some bug with the distribution)
4. Some potential issues with stack conversions that are not implemented yet (?)

Except for the forth one I see no opposition to making this function return a value.

lib/ABI/FunctionType.cpp
Cvelth: Because `Layout` is immutable ;)

lib/ABI/FunctionType.cpp
Cvelth: Yeah, but if we go that far, we might as well implement that properly (since we ~~have~~ had (it's easy to reintroduce) an `abi::Trait::CalleeIsResponsibleForStackCleanup` flag) and set them properly.
So, it's `getCallPushSize` + the combined size of stack arguments + (some extra stuff abi wants to write to the stack here ???) for caller-cleaned-case and what was it for callee-cleaned-case?

lib/ABI/FunctionType.cpp
Cvelth: I can't imagine an ABI without at least one of those. Might be a good indication that the layout is not doing right???

No big deal thought, I'm fine with getting rid of it.

include/revng/ABI/FunctionType.h
aleclearmind: 1. In order not to increase complexity for this "small" problem, the ideal solution would be to first perform all the checks to verify whether we can return a result or not, and *then* modify `Binary`, once you're sure you'll be successful.
  For what concerns `PrimitiveType`s: feel free to add them, at one point we'll offer some guarantees on what `PrimitiveType`s you can assume will be already available. Put a TODO hinting at this and we'll fix this later.
2. :+1:
3. Yeah, extra explicit but now you have two distinct way to represent similar concepts. Either we drop `Invalid` everywhere in favor of `optional<myenum>` or we use `Invalid`.

Cvelth: 1. This would require a clear "condition" of what `CABI` function can't in fact be represented as a `Raw` function. Which is nice in it's own right, but this might cause a bit of a rewrite of the underlying implementation.

include/revng/ABI/FunctionType.h
aleclearmind: 1. Ah, but in that case the problem is not the conversion to `RawFunctionType`, the problems is that the `CABIFunctionType` contains information incompatible with its `ABI` field. If you want to check this (if that ever happens), it's `CABIFunctionType::verify` business.
2. Under the assumption (which can make) that the input is verifies, you can assert that.
3. `assert` FTW.
4. If there will be such failures, that will call for a review of our design. As for now, I think we can assume going to raw never fails, just as `layout` should never fail (they serve similar purpose).

lib/ABI/FunctionType.cpp
aleclearmind: So if no longer necessary we can elide it?

include/revng/ABI/FunctionType.h
Cvelth: Sounds good, one less `optional`, ya-a-ay :D

lib/ABI/FunctionType.cpp
Cvelth: It's already so in my working branch.

lib/ABI/FunctionType.cpp
aleclearmind: Mmh, I'm thinking the user might to something stupid with `RawFunctionType` (no specific ABI involved).
Up to you.

include/revng/ABI/FunctionType.h
aleclearmind: 1. We can have a function with a `bool` (template?) that when it's `false` "tries" and reports success/failure and when it's `true` actually does the thing? Not excited about this idea.

include/revng/ABI/FunctionType.h
Cvelth: Done: https://github.com/revng/revng/pull/131/commits/505909f2fca23b0a400bb6c17031fac8481b8472.

lib/ABI/FunctionType.cpp
Cvelth: Gone: https://github.com/revng/revng/pull/131/commits/505909f2fca23b0a400bb6c17031fac8481b8472.

lib/ABI/FunctionType.cpp
Cvelth: Gone: https://github.com/revng/revng/pull/131/commits/505909f2fca23b0a400bb6c17031fac8481b8472.

lib/ABI/FunctionType.cpp
Cvelth: Gone: https://github.com/revng/revng/pull/131/commits/0bc0bc7396467e8916fd0f908d4c77ddbdf198fc.

lib/ABI/FunctionType.cpp
Cvelth: Done: https://github.com/revng/revng/pull/131/commits/0bc0bc7396467e8916fd0f908d4c77ddbdf198fc.

include/revng/ABI/FunctionType.h
Cvelth: Done: https://github.com/revng/revng/pull/131/commits/45ef15cef4660f04950930cfcb50171079ac55b5

lib/ABI/FunctionType.cpp
Cvelth: Done: https://github.com/revng/revng/pull/131/commits/ebf296a3672b71363d27ecfc68cfcbfbe43a1bf2.

aleclearmind:
This has been merged in 24e6873 a while ago. Thanks!

PR #132, opened 2022-02-10
GenericLLVMPipe: fix `O2` pipe name

antoniofrighetto:


aleclearmind:
What does this commit do? Does it fix a bug? How?

antoniofrighetto:
Sorry, forgot to leave basic explanation. When running `revng --verbose translate -i -O2 input_binary`, `revng-pipeline` is called with flag `-f O2`, but it fails with error `Could not load llvm pass O2`. This commit addresses the issue by changing the name of the pass to `O2` (it is not made as default option though).

aleclearmind:
Please do not use `_` in branch names.
Merged in e798131. Thanks!

PR #133, opened 2022-02-10
Add translate arguments propagation

drblallo:
now arguments provided after -- when invoking revng translate are forwarded to the pipeline


aleclearmind:
Merged in ef1b591. Thanks!

PR #134, opened 2022-02-11
Introduce invalidation support

drblallo:


include/revng/Pipeline/LLVMContainer.h
```diff
@@ -196,21 +202,25 @@ class LLVMContainerBase
...
-    bool Result = TheLinker.linkInModule(std::move(Module),
-                                         llvm::Linker::OverrideFromSrc);
+    bool Failure = TheLinker.linkInModule(std::move(Module),
+                                          llvm::Linker::OverrideFromSrc);
 
-    Result = Result
-             and TheLinker.linkInModule(std::move(ToMerge.Module),
+    Failure = Failure
+              or TheLinker.linkInModule(std::move(ToMerge.Module),
                                         llvm::Linker::OverrideFromSrc);
```
aleclearmind, line 204: What was the bug?
What's the fix?

Try to be a bit more verbose in commit messages `:P`

aleclearmind (changes requested):

include/revng/Pipeline/InvalidationEvent.h
```diff
@@ -0,0 +1,19 @@
...
+
+#include "revng/Pipeline/Runner.h"
+
+namespace pipeline {
+
+class InvalidationEvent {
+public:
+  llvm::Error run(Runner &Runner) const;
+  virtual void getInvalidations(const Runner &Runner,
+                                Runner::InvalidationMap &Out) const = 0;
```
aleclearmind, line 15: * `computeInvalidations`?
* Doc
* Why not returning it? Always in order to accumulate? Probably more efficient but it seems coupled to its users.
* Do we need a `Runner` reference here? Maybe there's a more specific object we can pass in?

include/revng/Pipes/ModelInvalidationEvent.h
```diff
@@ -0,0 +1,33 @@
...
+namespace revng::pipes {
+
+class ModelInvalidationEvent : public pipeline::InvalidationEvent {
+public:
+  explicit ModelInvalidationEvent(TupleTreeDiff<model::Binary> Diff) :
+    Diff(std::move(Diff)) {}
+
+  template<typename... Args>
+  explicit ModelInvalidationEvent(Args &&...Arguments) :
+    Diff(std::forward<Args>(Arguments)...) {}
```
aleclearmind, line 20: Can we drop this?

include/revng/Pipes/ModelInvalidationEvent.h
```diff
@@ -0,0 +1,33 @@
...
+#include "revng/Model/Binary.h"
+#include "revng/Pipeline/InvalidationEvent.h"
+#include "revng/TupleTree/TupleTreeDiff.h"
+
+namespace revng::pipes {
+
+class ModelInvalidationEvent : public pipeline::InvalidationEvent {
+public:
+  explicit ModelInvalidationEvent(TupleTreeDiff<model::Binary> Diff) :
+    Diff(std::move(Diff)) {}
```
aleclearmind, line 16: Why not `const &`?

include/revng/Pipes/ModelInvalidationRule.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include "llvm/ADT/ArrayRef.h"
+
+#include "revng/Pipeline/Container.h"
+#include "revng/Pipeline/Runner.h"
+#include "revng/Pipeline/Target.h"
+
+namespace revng::pipes {
+
+class ModelInvalidationRule {
```
aleclearmind, line 15: * Doc
* How is this supposed to be used? What are some examples of invalidation rules you foresee?
  I was expecting we had some form of `Kind::invalidate(TupleTreeDiff)` method.
  I see many abstractions (`InvalidationEvent`, `ModelInvalidationEvent`, `ModelInvalidationRule`) but I fail to see the endgame, which might not need all these intermediate abstractions.

lib/Pipeline/InvalidationEvent.cpp
```diff
@@ -0,0 +1,18 @@
...
+//
+
+#include "revng/Pipeline/InvalidationEvent.h"
+
+using namespace pipeline;
+
+llvm::Error InvalidationEvent::run(Runner &Runner) const {
+  Runner::InvalidationMap Map;
+  getInvalidations(Runner, Map);
+  if (auto Error = Runner.getInvalidations(Map); Error)
```
aleclearmind, line 15: * `getInvalidations` is not a great name. `deduceDerivedInvalidations`?
* This whole method makes me a little suspicious. Why should an invalidation event extend its results?
  In case of multiple invalidation events this would also be more efficient.
  These all seem to be responsibilities of the caller of this method. Can't we have the invalidation event *exclusively* compute the set of invalidate target?

include/revng/Pipes/ModelInvalidationRule.h
```diff
@@ -0,0 +1,39 @@
...
+  registerAllInvalidations(const pipeline::Runner &Runner,
+                           pipeline::Runner::InvalidationMap &Invalidations);
+
+  virtual ~ModelInvalidationRule() = default;
+
+private:
+  static std::vector<ModelInvalidationRule *> &getRulesRegistry() {
+    static std::vector<ModelInvalidationRule *> Rules;
+    return Rules;
+  }
```
aleclearmind, line 36: Can you promote this to a `static` field?

include/revng/Pipes/ModelInvalidationRule.h
```diff
@@ -0,0 +1,39 @@
...
+  static llvm::ArrayRef<ModelInvalidationRule *> getRules() {
+    return getRulesRegistry();
+  }
+
+  virtual void registerInvalidations(const pipeline::ContainerBase &Container,
+                                     pipeline::TargetsList &ToRemove) const = 0;
+
+  static void
+  registerAllInvalidations(const pipeline::Runner &Runner,
+                           pipeline::Runner::InvalidationMap &Invalidations);
```
aleclearmind, line 28: This class has also a registry, but the `register*` methods do not register stuff into the registry.
What about `computeInvalidations`?

tools/pipeline/Main.cpp
```diff
@@ -73,6 +80,16 @@ static alias SilenceAlias1("s",
...
 
+static alias SilenceAlias2("quiet",
+                           desc("Alias for --silent"),
+                           aliasopt(Silence),
+                           cat(PipelineCategory));
+
+static alias SilenceAlias3("q",
+                           desc("Alias for --silent"),
+                           aliasopt(Silence),
+                           cat(PipelineCategory));
```
aleclearmind, line 101: Why?

tools/pipeline/Main.cpp
```diff
@@ -59,6 +59,13 @@ static opt<bool> ProduceAllPossibleTargets("produce-all",
                                            cat(PipelineCategory),
                                            init(false));
 
+static opt<bool> InvalidateAll("invalidate-all",
+                               desc("Try invalidating all possible "
+                                    "targets after producing them. Used for "
+                                    "debug porpuses"),
+                               cat(PipelineCategory),
+                               init(false));
```
aleclearmind, line 67: What's the purpose of this?

include/revng/PipelineC/PipelineC.h
```diff
@@ -149,6 +156,14 @@ void rp_manager_destroy(rp_manager *manager);
  */
 uint64_t rp_manager_containers_count(rp_manager *manager);
 
+/**
+ * Applies the diff to the model and triggers a ModelInvalidationEvent
+ *
+ * \param manager cannot be null.
+ * \param diff cannot be null.
```
aleclearmind, line 163: Drop these. Nothing can ever be null by default.

include/revng/PipelineC/PipelineC.h
```diff
@@ -149,6 +156,14 @@ void rp_manager_destroy(rp_manager *manager);
...
+ */
+void rp_apply_model_diff(rp_manager *manager, const rp_model_diff *diff);
```
aleclearmind, line 165: This API doesn't make much sense without an API to extract the model and compute the `diff`.

include/revng/Pipeline/LLVMContainer.h
drblallo: done!
in this case it was that linkInModule returned true on failure for some reason.

tools/pipeline/Main.cpp
```diff
@@ -137,7 +147,7 @@ static void runPipeline(Runner &Pipeline) {
   for (const auto &Target : Targets)
     AbortOnError(parseTarget(ToProduce, Target, Registry));
 
-  auto *Stream = Silence ? nullptr : &dbgs();
+  auto *Stream = (Verbose or not Silence) ? &outs() : nullptr;
```
aleclearmind, line 150: * The way you're mixing up verbose and silence is confusing. The tool by default should be silent, and in case of verbose it should print extra stuff on `stderr`. I really think you should be using `Logger`, `revng-pipeline` is acting more and more as a special kid in this sense.
* Also, you should print on `stdout` only stuff that makes sense to save on a file and has a well-defined format.

include/revng/Pipeline/LLVMGlobalKindBase.h
```diff
@@ -54,6 +54,19 @@ class LLVMGlobalKindBase : public KindForContainer<LLVMContainer> {
     return false;
   }
 
+  llvm::Error
+  verify(const ContainerBase &Container, const Target &T) const override {
+    if (const auto *Casted = llvm::dyn_cast<LLVMContainer>(&Container);
+        Casted != nullptr)
```
aleclearmind, line 60: You can drop `Casted != nullptr`

include/revng/Pipeline/Container.h
```diff
@@ -55,13 +58,20 @@ class ContainerBase {
   /// not after(this)->enumerate().contains(Targets);
   virtual bool remove(const TargetsList &Targets) = 0;
 
+  virtual llvm::Error serialize(llvm::raw_ostream &OS) const = 0;
```
aleclearmind, line 61: Doc

lib/PipelineC/PipelineC.cpp
```diff
@@ -396,3 +397,33 @@ void rp_apply_model_diff(rp_manager *manager, const rp_model_diff *diff) {
...
+    llvm::consumeError(std::move(Error));
+    return nullptr;
+  }
+  Serialized.flush();
+  return copy_string(Out);
+}
+
+bool rp_manager_deserialize_global(rp_manager *manager,
+                                   const char *serialized,
+                                   const char *global_name) {
```
aleclearmind, line 416:
```c++
bool rp_manager_set_global(rp_manager *manager,
                                   const char *global_name.
                                   const char *serialized)
```

aleclearmind (changes requested):

cmake/TupleTreeGenerator.cmake
```diff
@@ -230,3 +230,63 @@ function(tuple_tree_generator_generate_jsonschema
...
+
+endfunction()
+
+function(
+  target_tuple_tree_generate
+  HEADER_DIRECTORY
+  NAMESPACE
+  SCHEMA_PATH
+  TARGET_ID
+  )
```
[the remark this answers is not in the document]
drblallo: it is used in a later commit, in can squash them but it does not make much sense.

yeah the name is not great, but it generate the tuple tree and the schema, not sure how to call it.

tuple_tree_and_schema_generate handles itself all the managment regarding installation, dependencies, naming of the target and stuff like that, which has no reason to exposed to the user.

tuple_tree_generate hides some fields that are optional, the result is the much clearer

```
target_tuple_tree_generate(
  Pipes
  revng::pipes
  "${CMAKE_CURRENT_BINARY_DIR}/schema.yml"
  revngPipes
  "${CMAKE_SOURCE_DIR}/include/revng/Pipes/CallGraphCache.h")
```

lib/Pipes/FileContainer.cpp
```diff
@@ -54,6 +54,16 @@ FileContainer &FileContainer::operator=(const FileContainer &Other) noexcept {
   return *this;
 }
 
+bool FileContainer::remove(const pipeline::TargetsList &Target) {
```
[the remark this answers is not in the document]
drblallo: it is not a failure, it should return false if the thing that were requested to be removed were not present from the start.

include/revng/Pipeline/InvalidationEvent.h
```diff
@@ -0,0 +1,19 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "revng/Pipeline/Runner.h"
+
+namespace pipeline {
+
+class InvalidationEvent {
+public:
+  llvm::Error run(Runner &Runner) const;
```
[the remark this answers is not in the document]
drblallo: yes, apply is better

include/revng/Pipeline/InvalidationEvent.h
drblallo: yeah computeInvalidations is better
yeah, to accumulate, calculating invalidations is usually expensive.
it needs the state of the pipeline and the context, which is pretty much a entire runner.

include/revng/Pipeline/Container.h
```diff
@@ -55,13 +58,20 @@ class ContainerBase {
...
+
+  virtual llvm::Error deserialize(const llvm::MemoryBuffer &Buffer) = 0;
+
+  /// Must reset the state of the container to the just built state
+  virtual void clear() = 0;
```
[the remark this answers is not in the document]
drblallo: because if we are loading the entire state of a folder and a file is missing this conceptually means that the container should be default constructed. But we don't know how to default construct it because it is a virtual class and thus we need to be able to reset it.

include/revng/Pipeline/Container.h
```diff
@@ -55,13 +58,20 @@ class ContainerBase {
...
+
   /// The implementation must ensure that there exists a file at the provided
   /// path that contains the serialized version of this object.
-  virtual llvm::Error storeToDisk(llvm::StringRef Path) const = 0;
+  virtual llvm::Error storeToDisk(llvm::StringRef Path) const;
 
   /// The implementation must esure that the content of this file will be
   /// loaded from the provided path.
-  virtual llvm::Error loadFromDisk(llvm::StringRef Path) = 0;
+  virtual llvm::Error loadFromDisk(llvm::StringRef Path);
```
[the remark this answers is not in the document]
drblallo: yeah, but i would prefer not too. this allows for stuff like temporaryfile to not require to load the file from disk by providing a custom load

tools/pipeline/Main.cpp
drblallo: testing and debugging

tools/pipeline/Main.cpp
drblallo: using revng translate --verbose yields a output that makes sense for translate but not for revng pipeline
this allows to do 
```
revng translate -- -q
revng translate -- --verbose
```
when one is debugging the pipeline itself

include/revng/Pipes/ModelInvalidationRule.h
drblallo: ah, no i just forgot to pass down the diff across the calls

include/revng/Pipes/ModelInvalidationRule.h
drblallo: i could, but i prefer this way since this is resilient to initialization order fiasco

include/revng/Pipes/ModelInvalidationRule.h
drblallo: why not just getInvalidations?

include/revng/PipelineC/PipelineC.h
drblallo: yeah, it is just missing the thing about reading diffs on which i was working, but that is way less trivial than expected

include/revng/Pipeline/LLVMGlobalKindBase.h
drblallo: why? Container is a argument that may not be a LLVMContainer

cmake/TupleTreeGenerator.cmake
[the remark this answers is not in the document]
aleclearmind: OK, then `tuple_tree_and_schema_generate` must become `tuple_tree_generator`, the one and only entrypoint.

If you want to have a version with less arguments, the right solution is to switch to `cmake_parse_arguments`, which would improve the API overall.

I really want one, single, well document, public entry point called `tuple_tree_generator`.

lib/Pipes/FileContainer.cpp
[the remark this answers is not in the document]
aleclearmind: OK, make sure there's some doc about this somewhere.

include/revng/Pipes/ModelInvalidationRule.h
aleclearmind: I still don't see how this connects to `Kind::invalidate(TupleTreeDiff)`.

lib/Pipeline/InvalidationEvent.cpp
aleclearmind: > this method is what one would do anyway if it had to manually remove the deduced invalidations

I'm not sure I understand.

include/revng/Pipes/ModelInvalidationRule.h
aleclearmind: Usually we solve it using `ManagedStatic`. I'd prefer to threat the same problem in the same way all over the codebase, it eases refactoring.

tools/pipeline/Main.cpp
aleclearmind: In an unrelated branch, I'm dropping `--`. I don't think anyone ever used it and if you're running multiple commands it's not really obvious who is gong to receive what comes after `--`.
If you want to debug things, you've better using `--verbose`, copy paste and customize invocation.

include/revng/PipelineC/PipelineC.h
aleclearmind: ACK, let's move this to the invalidation PR.

include/revng/Pipeline/LLVMGlobalKindBase.h
aleclearmind:
```c++
if (const auto *Casted = llvm::dyn_cast<LLVMContainer>(&Container);
        Casted != nullptr)
```
is the same as
```c++
if (const auto *Casted = llvm::dyn_cast<LLVMContainer>(&Container))
```
isn't it?

include/revng/Pipeline/Container.h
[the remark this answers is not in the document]
aleclearmind: Are you actually doing this? If not, drop `virtual`.
Anyway, seems like a bit of a footgun for subtly mismatched behavior.

lib/PipelineC/PipelineC.cpp
```diff
@@ -213,11 +221,19 @@ bool rp_manager_produce_targets(rp_manager *manager,
...
-  llvm::consumeError(std::move(Error));
-  return false;
+  std::string Out;
+  llvm::raw_string_ostream Serialized(Out);
+  const auto &ToFilter = Targets[container->second->name()];
+  const auto &Cloned = container->second->cloneFiltered(ToFilter);
+  llvm::cantFail(Cloned->serialize(Serialized));
+  Serialized.flush();
+
+  return copy_string(Out);
```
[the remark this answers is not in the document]
aleclearmind: OK, still, it could be done. This API might transfer biiiig chunks of data.
You could have an heap allocated buffer, have the stream write in there, and then steal the ownership of data and pass it to the caller.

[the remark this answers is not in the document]
drblallo: yeah, the correct solution would be to write a custom output stream that writes in a buffer. i would delay this until it becomes a performance bottleneck toh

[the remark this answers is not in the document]
aleclearmind: Works! My original comment was "add a TODO about this in the code".

include/revng/Pipeline/Container.h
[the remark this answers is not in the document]
drblallo: eh, yeah, it is a bit riskier and not raccomended, but we are doing it yeah

include/revng/Pipeline/LLVMGlobalKindBase.h
drblallo: ahh, right, it just look more explicit the current way, but ok, make sense

tools/pipeline/Main.cpp
drblallo: --verbose is unusable on translate to do debugging about the pipeline. it runs the pipeline when you actually just want to see how it is being invoked. 
that one time when antonio tried out the pipeline it would have found helpfull multiple time in a day to have --. 
i myself quite use the -- mechanism to debug it.

i am very against droping -- in translate.

lib/Pipeline/InvalidationEvent.cpp
drblallo: i mean. this method rappresents the entire action of triggering a invalidation event, figure out the dependencies and drop all  those affected. Somewhere this method must be located, and it must be exactly like this. I can move it to another file, but i don't see much a reason to do that. 

this gives you the syntax
```
SomeInvalidaitonEvent(SomeData).apply(Runner);
```
and everything is done as expected

lib/PipelineC/PipelineC.cpp
```diff
@@ -396,3 +397,33 @@ void rp_apply_model_diff(rp_manager *manager, const rp_model_diff *diff) {
   model::Binary &Model(*getWritableModelFromContext(manager->context()));
   diff->apply(Model);
 }
+
+const char *rp_manager_create_serialized_global(rp_manager *manager,
+                                                const char *global_name) {
```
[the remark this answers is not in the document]
drblallo: this has to be rp_manager_create_get_global because return stuff that must be pruned by the user

lib/Pipeline/InvalidationEvent.cpp
drblallo: will we really ever have multiple events? we will send a single diff or a user triggered invalidation 99% of the time.

if it becomes a issue, we will add the batching

lib/PipelineC/PipelineC.cpp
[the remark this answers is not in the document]
aleclearmind: `rp_manager_create_global_copy`?

lib/PipelineC/PipelineC.cpp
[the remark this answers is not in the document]
aleclearmind: `TODO`s are typically things that never get done. But once you identify the problem, you look at the code and you find a suggestion on how to fix a problem we foresaw. Imagine that the person finding the comment might not be you,

lib/Pipeline/InvalidationEvent.cpp
aleclearmind: > will we really ever have multiple events

Who knows? Maybe at a certain point we will have a queue of events.

`Runner.getInvalidation` is basically an internal thing of the runner, I don't see why it should be called by someone that should only mind about his own business: computing a set of invalidations.

include/revng/Pipes/ModelInvalidationRule.h
drblallo: kinds must implement 

```
  virtual void getInvalidations(const pipeline::ContainerBase &Container,
                                const ModelInvalidationEvent &Event) const = 0;
```
or
```
  virtual void getInvalidations(pipeline::TargetsList &ToRemove,
                                const ModelInvalidationEvent &Event) const = 0;
```
would that work?

tools/pipeline/Main.cpp
drblallo: fixed, toh, outs() supports color coded stuff, dbgs does not

cmake/TupleTreeGenerator.cmake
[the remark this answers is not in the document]
drblallo: i would really like to keep the simplicity introduced by the two wrappers, and i agree that cmake_parse_arguments is the correct solution. toh adopting everyhwere is a bit invasive. Can i write a wrapper that uses cmake_parse_arguments and then dispatch to the current implementation that does not use cmake_parse_arguments?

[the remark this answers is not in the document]
aleclearmind: Please take your time and fix the other uses, there's just two of them:

```
$ git grep 'tuple_tree_generator(' | cat
lib/Model/CMakeLists.txt:tuple_tree_generator(
tests/unit/UnitTests.cmake:tuple_tree_generator(
```

One of them would really benefit from optional arguments (`tests/unit/UnitTests.cmake`).

aleclearmind:
Converting to draft until #152 is in.

aleclearmind:
Can we rename this PR to "`revng-pipeline`: introduce validation support"?
Also, I'd like to see actual implementation of some invalidation logic before I can make meaningful comments on the design.
Can you write the invalidation logic for:
* `RootKind Root`: if there are changes in `model::Binary::ExtraCodeAddresses`, we need to invalidate it.
* `TaggedFunctionKind Isolated`: if the corresponding `model::Function` is erased or its type or its attributes change, we need to invalidate it.

Also: please pull out the `CallGraphCache` commit and put it on some branch we can recover later or open a new PR.
I want to get in the basic invalidation infrastructure for the two simple situations above first.

include/revng/TupleTree/TupleTreeDiff.h
```diff
@@ -38,15 +44,58 @@ void addToContainer(C &Container, const typename C::value_type &Value) {
...
+  using Types = std::variant<char>;
+};
+
+template<typename T>
+using TupleTreeEntriesT = typename TupleTreeEntries<T>::Types;
+
+template<>
+struct TupleTreeEntries<model::Binary> {
+  using Types = model::AllTypes;
+};
```
aleclearmind, line 58: You can't have this here, this file has to be model-agnostic.

include/revng/TupleTree/TupleTreeDiff.h
```diff
@@ -38,15 +44,58 @@ void addToContainer(C &Container, const typename C::value_type &Value) {
...
+                       TupleTreeEntriesT<T> New) {
+    return Change(std::move(Path), std::move(Old), std::move(New));
+  }
+
+  using EntryType = std::optional<TupleTreeEntriesT<T>>;
+
+  TupleTreePath Path;
+  EntryType Old = std::nullopt;
+  EntryType New = std::nullopt;
+};
```
aleclearmind, line 89: In general, and in future too, please try to put methods/fields in the conventional order we use:

```c++
class MyClass {
private:
  // private usings, if any

public:
  // public usings, if any

private: // Or public
  // Fields of the class

public:
  // Constructors

public:
  // static factory methods

public:
  // Accessor methods

public:
  // Non-accessor `const` methods

public:
  // Non-accessor const methods

private:
  // Helper private methods
};
```
This kind of conventions meets the readers expectations and make him feel comfy.

Also, at least for important data structures, make sure you have a `verify` (is having both fields `nullopt` valid?) and a `dump() debug_function` method.
Future developer debugging this will love you.

I've noted these [here (B.16, B.17)](https://pad.rev.ng/s/SJNUet_ku#).

include/revng/TupleTree/TupleTreeDiff.h
```diff
@@ -57,14 +106,20 @@ struct TupleTreeDiff {
     return Result;
   }
 
-  void add(const TupleTreePath &Path, void *What) {
-    Changes.push_back({ Path, nullptr, What });
+  template<typename ToAdd>
+  void add(const TupleTreePath &Path, ToAdd What) {
+    Changes.push_back(Change::addition(Path, What));
```
aleclearmind, line 111: `assert` that the type pointed by `Path` is `ToAdd`.
Below too.

include/revng/TupleTree/TupleTreeDiff.h
```diff
@@ -74,7 +129,113 @@ struct TupleTreeDiff {
...
+template<DiffSpecialization T>
+struct llvm::yaml::MappingTraits<T> {
+  static void mapping(IO &io, T &info) {
+    io.mapOptional("changes", info.Changes);
+  }
+};
+
+template<ChangeSpecialization T, typename X>
+struct llvm::yaml::SequenceElementTraits<T, X> {
+  static const bool flow = false;
```
aleclearmind, line 150: Wrong case, are you running `clang-tidy`?

include/revng/TupleTree/TupleTreeDiff.h
```diff
@@ -74,7 +129,113 @@ struct TupleTreeDiff {
...
+template<typename T>
+concept DiffSpecialization = is_specialization_v<T, TupleTreeDiff>;
+
+template<typename T>
+concept ChangeSpecialization = is_specialization_v<T, Change>;
+
+template<DiffSpecialization T>
+struct llvm::yaml::MappingTraits<T> {
+  static void mapping(IO &io, T &info) {
+    io.mapOptional("changes", info.Changes);
```
aleclearmind, line 144: In every `YAMLTraits` we use the same casing as the type.
We might wanna change, but let's be coherent.

scripts/tuple_tree_generator/tuple_tree_generator/templates/all_types_variant_decls.h.tpl
```diff
@@ -0,0 +1,25 @@
...
+
+// This file is autogenerated! Do not edit it directly
+
+
+// Forward declarations
+namespace /*= namespace =*/ {
+
+using AllTypes = std::variant<
+/** for name in all_types|sort **/
+/*= name =*//** if not loop.last **/,/** endif **/
```
aleclearmind, line 21: Gosh, where is `loop` defined? Some jinja magic?

tests/unit/Model.cpp
```diff
@@ -291,6 +291,42 @@ BOOST_AUTO_TEST_CASE(TestTupleTreeDiff) {
...
+
+  auto Diff2 = llvm::cantFail(deserialize<TupleTreeDiff<model::Binary>>(S));
+
+  std::string S2;
+  llvm::raw_string_ostream Stream2(S2);
+  serialize(Stream2, Diff2);
+  Stream2.flush();
+  llvm::errs() << S2;
+
+  BOOST_TEST(S == S2);
```
aleclearmind, line 327: Nice

include/revng/Pipes/CallGraphCache.h
```diff
@@ -0,0 +1,46 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include <string>
+
+#include "revng/Pipeline/YamlizableGlobal.h"
+#include "revng/Support/YAMLTraits.h"
+#include "revng/TupleTree/TupleLikeTraits.h"
+
+/* TUPLE-TREE-YAML
```
aleclearmind, line 13: This commit is not ready to be merged.

include/revng/Pipeline/InvalidationEvent.h
aleclearmind: * Please address all the comments above.
* `Runner::getInvalidations` is also misleading.
* Are you sure that accepting a `Runner` is the right thing? I like more `ModelInvalidationRule`'s approach which accepts `const pipeline::ContainerBase &`. Maybe the runner should have the responsibility of invoking  `getInvalidations` on all the containers?

include/revng/Pipes/ModelInvalidationRule.h
aleclearmind: Doc is still missing! Please double check before you mark as resolved.
I haven't reviewed users of `ModelInvalidationRule`, but I've this question: is this abstraction necessary?
Can't we collapse this in `Kind`? I fail to see the big picture.

lib/Pipes/ModelInvalidationRule.cpp
```diff
@@ -0,0 +1,33 @@
...
+    auto &StepInvalidations = Map[Step.getName()];
+    for (const auto &Cotainer : Step.containers()) {
+      if (not Cotainer.second)
+        continue;
+
+      auto &ContainerInvalidations = StepInvalidations[Cotainer.first()];
+      for (const auto *Rule : *Rules)
+        Rule->getInvalidations(*Cotainer.second, ContainerInvalidations, Event);
+    }
+  }
```
aleclearmind, line 32: I've already said this in another comment but please consider shifting this responsibility to `Runner`, which already knows about invalidations. It might make the dependency graph between components cleaner.

include/revng/Pipes/PipelineManager.h
```diff
@@ -122,7 +122,9 @@ class PipelineManager {
 
   /// A helper function used to produce all possible targets. It is used for
   /// debug purposes to see if any particular target crashes.
-  llvm::Error printAllPossibleTargets(llvm::raw_ostream &Stream);
+  llvm::Error produceAllPossibleTargets(llvm::raw_ostream &Stream);
+
+  llvm::Error invalidateAllPossibleTargets(llvm::raw_ostream &OS);
```
aleclearmind, line 127: Why "Possible"?

include/revng/PipelineC/PipelineC.h
aleclearmind: I see many comments of my previous review not being addressed.

tests/unit/Pipeline.cpp
```diff
@@ -1219,4 +1219,52 @@ BOOST_AUTO_TEST_CASE(InspectorKindTest) {
...
+  Error = Pipeline.invalidate(Invalidations);
+
+  BOOST_TEST(!Error);
+
+  BOOST_TEST(C1.get(T) == 0);
+  BOOST_TEST(C1End.get(T) == 0);
+
+  BOOST_TEST(C2End.get(ToProduce) == 0);
+}
+
```
aleclearmind, line 1269: This change seems to be in the wrong commit.

include/revng/Pipeline/Container.h
```diff
@@ -78,6 +78,9 @@ class ContainerBase {
   /// The implementation must esure that the content of this file will be
   /// loaded from the provided path.
   virtual llvm::Error loadFromDisk(llvm::StringRef Path);
+
+  /// Checks that the content of the this container is well formed.
```
aleclearmind, line 82: `s/well formed/valid/`
I'm trying to push a terminology similar to XML, where "well formed" means that it's parsable but might fail to respect a schema. If it validates a schema it's "valid". Here we're more in the latter situation.

include/revng/Pipes/RootKind.h
```diff
@@ -4,15 +4,20 @@
...
 #include "revng/Pipeline/LLVMContainer.h"
 #include "revng/Pipeline/LLVMGlobalKindBase.h"
+#include "revng/Pipes/ModelInvalidationRule.h"
+#include "revng/TupleTree/TupleTreeDiff.h"
+#include "revng/TupleTree/TupleTreePath.h"
 
 namespace revng::pipes {
 
-class RootKind : public pipeline::LLVMKind {
+class RootKind : public pipeline::LLVMKind, ModelInvalidationRule {
```
aleclearmind, line 20: I see that you want to compose things and make things optional, but all this flexibility makes it really really hard to explain what's going on.

When I had to explain to @pfez how to create a `Kind`, he wanted to dig under the surface (rightly so) and if you start from `TaggedFunctionKind`, it's quite a journey. We have 1) kinds that can or cannot enumerate containers, 2) we have a decoupling for LLVM, 3) kinds that are coupled to `llvm::Function`s through `FunctionTags` and I might still be missing something.
Clearly 3) is really needed, but the others to me keep striking as complexity with little to no return.

tl;dr: I'm happy to see that you opted for having `Kind::verify` to be mandatory, can we do the same for `getInvalidations` and avoid further increasing complexity?

lib/Pipes/FileContainer.cpp
```diff
@@ -97,8 +97,12 @@ void FileContainer::mergeBackImpl(FileContainer &&Container) {
 }
 
 llvm::Error FileContainer::storeToDisk(llvm::StringRef Path) const {
-  if (this->Path.empty())
+  if (this->Path.empty()) {
+    if (llvm::sys::fs::exists(Path))
+      llvm::sys::fs::remove(Path);
```
aleclearmind, line 102: What's the ratio for this? Seems surprising.

aleclearmind (changes requested):

lib/Pipes/RootKind.cpp
```diff
@@ -33,3 +33,23 @@ IsolatedRootKind::symbolToTarget(const llvm::Function &Symbol) const {
...
+
+using Diff = TupleTreeDiff<model::Binary>;
+void RootKind::getInvalidations(TargetsList &ToRemove,
+                                const Diff &Event) const {
+
+  using Reference = TupleTreeReference<SortedVector<MetaAddress>,
+                                       model::Binary>;
+  const TupleTreePath
+    ToCheck = Reference::fromString(static_cast<model::Binary *>(nullptr),
+                                    "/ExtraCodeAddresses")
```
aleclearmind, line 45: Make an overload for `Reference::fromString` that takes no binary.

lib/Pipes/RootKind.cpp
```diff
@@ -33,3 +33,23 @@ IsolatedRootKind::symbolToTarget(const llvm::Function &Symbol) const {
 
   return std::nullopt;
 }
+
+using Diff = TupleTreeDiff<model::Binary>;
+void RootKind::getInvalidations(TargetsList &ToRemove,
+                                const Diff &Event) const {
+
+  using Reference = TupleTreeReference<SortedVector<MetaAddress>,
+                                       model::Binary>;
```
aleclearmind, line 42: `decltype(Binary::ExtraCodeAddresses)`?

tests/tools/invalidate/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+      -i begin:Strings1:${OPT_INPUT} \
+      -l ${CMAKE_BINARY_DIR}/lib/librevngStringContainerLibrary.so \
+      -p ${CMAKE_BINARY_DIR}/${TEST_NAME} \
+      --step ${OPT_STEP} &&
+      ${CMAKE_BINARY_DIR}/libexec/revng/revng-invalidate \
+      -P ${OPT_PIPELINE} \
+      ${OPT_INVALIDATIONS} \
+      -l ${CMAKE_BINARY_DIR}/lib/librevngStringContainerLibrary.so \
+      -p ${CMAKE_BINARY_DIR}/${TEST_NAME} \
+      --step ${OPT_STEP}")
```
aleclearmind, line 25: I suspect `${CMAKE_BINARY_DIR}` is not needed.
I understand you want to be explicit, but I prefer to make command line looks approachable when debugging stuff.

tests/unit/DiffInvalidationEvent.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+  MetaAddress Address(0x1000, MetaAddressType::Code_aarch64);
+  New.ExtraCodeAddresses.insert(Address);
+
+  auto Diff = diff(Empty, New);
+
+  TargetsList ToRemove;
+  Root.getInvalidations(ToRemove, Diff);
+  BOOST_TEST(ToRemove.size() == 1);
+  BOOST_TEST(&ToRemove.front().getKind() == &Root);
```
aleclearmind, line 36: Oh gosh. This kinda nice, but it's also weird. So we produce the list of *all* the targets that need to be invalidated, no matter whether they are actually present?
It's elegant but I'm a bit concerned about performance, aren't we wasting computation for stuff that's not there in the first place?

Probably not the bottleneck, but give it a think.

tools/CMakeLists.txt
```diff
@@ -5,3 +5,4 @@
 add_subdirectory(model)
 add_subdirectory(pipeline)
 add_subdirectory(link-for-translation)
+add_subdirectory(invalidate)
```
aleclearmind, line 8: `pipeline/invalidate`?

aleclearmind (changes requested):

include/revng/Support/YAMLTraits.h
```diff
@@ -228,3 +229,26 @@ std::string serializeToString(const T &ToDump) {
...
+}
+
+template<NotTupleTreeCompatible T>
+llvm::Expected<T> deserializeFile(const llvm::StringRef &Path) {
+  auto MaybeBuffer = llvm::MemoryBuffer::getFile(Path);
+  if (not MaybeBuffer)
+    return llvm::errorCodeToError(MaybeBuffer.getError());
+
+  return deserialize<T>((*MaybeBuffer)->getBuffer());
+}
```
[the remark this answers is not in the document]
drblallo: we can't use the tuple tree one, that is the point. TupleTreeDiff has the yaml traits, but not the tuple tree intrastructure, but i still need to deserialize it.

that is why i added template<NotTupleTreeCompatible T>

include/revng/TupleTree/TupleTreeDiff.h
drblallo: ah, yeah, forgot to move it

include/revng/Pipes/PipelineManager.h
drblallo: because not all targets that one can build can be produced, not sure i understand the question

lib/Pipes/FileContainer.cpp
drblallo: the issue is that if the output binary has been produced and then it gets somehow invalidated, without this the file it refers too would not be erased. Thus at the second execution it would be reloaded. 

The issue in reading this may be that this->Path is not Path the argument of the function. this->Path is the tmp file, Path is the final destination of the file

tests/unit/DiffInvalidationEvent.cpp
drblallo: it was like you say before, but we decided to have it this way so that it can be tested without end to end testing infrastructure

include/revng/Support/YAMLTraits.h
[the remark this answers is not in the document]
aleclearmind: Ah so this is symmetric to `serialize` in the same file, I didn't realize this.
Then please see if you can implement `TupleTree::deserialize` through your new functions.
We really need to reduce code duplication for this kind of stuff.

include/revng/Pipes/PipelineManager.h
aleclearmind: Would dropping the "Possible" from the name induce anyone in thinking this does something it doesn't actually do?

lib/Pipes/FileContainer.cpp
aleclearmind: I noticed the difference between `this->Path` and `Path` (maybe rename the argument).

If *not* producing a file is an expected behavior of `storeToDisk`, I think it should be the same for all the `storeToDisk`, not just `FileContainer::storeToDisk`.
The idea should be 1) delete the file if exists, 2) output something if it needs the be produced.

tests/unit/DiffInvalidationEvent.cpp
aleclearmind: OK, hopefully this won't become a bottleneck.

tools/model/apply/Main.cpp
```diff
@@ -0,0 +1,57 @@
...
+                                      cl::value_desc("model"));
+
+static cl::opt<std::string> DiffPath(cl::Positional,
+                                     cl::cat(ThisToolCategory),
+                                     cl::desc("<model diff>"),
+                                     cl::value_desc("model"));
+
+static ModelOutputOptions<false> Options(ThisToolCategory);
+
+int main(int Argc, char *Argv[]) {
```
aleclearmind, line 37: I'd like `-R` (`man diff | grep -- -R`) and a roundtrip test to ensure that diffing and applying produces a result the same as the original one.

scripts/tuple_tree_generator/tuple_tree_generator/templates/all_types_variant_decls.h.tpl
drblallo: no idea, i did as cremo told me to, i guess so.

include/revng/Pipes/PipelineManager.h
drblallo: i would not know it is all possible or the current ones

lib/Pipes/FileContainer.cpp
drblallo: the difference is that other containers (llvm containers mostly) are not truly empty when their enumeration is empty. LLVMContainers have all the various helpers that stick around

tests/tools/invalidate/CMakeLists.txt
drblallo: tried but it does not work, i think that the execution directory here is ${CMAKE_CURRENT_BINARY_DIR}

tools/model/apply/Main.cpp
drblallo: how do you suggest to test the model without committing handwritten models for this purpose?

include/revng/Support/YAMLTraits.h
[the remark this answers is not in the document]
drblallo: i tried, but what happens is that the current implementation of TupleTree::deserialize is transformed from

```
  static llvm::ErrorOr<TupleTree> deserialize(llvm::StringRef YAMLString) {
    TupleTree Result;

    Result.Root = std::make_unique<T>();
    llvm::yaml::Input YAMLInput(YAMLString);
    YAMLInput >> *Result.Root;

    std::error_code EC = YAMLInput.error();
    if (EC)
      return EC;

    // Update references to root
    Result.initializeReferences();

    return Result;
  }

```
```
 static llvm::ErrorOr<TupleTree> deserialize(llvm::StringRef YAMLString) {
    TupleTree Result;

    Result.Root = std::make_unique<T>();
    auto Content = ::deserialize<TupleTree>(Input):
    if (not Content)
        return Content.takeError();

    *Result.Root = std::move(*Content); 
    
    // Update references to root
    Result.initializeReferences();

    return Result;
  }
```
which is actually longer and has 1 move more

scripts/tuple_tree_generator/tuple_tree_generator/templates/all_types_variant_decls.h.tpl
aleclearmind: You did not answer `:P`.
Ask cremo and answer.

include/revng/Pipes/PipelineManager.h
aleclearmind: What would be the point of producing all the produced targets?
Also, invalidating all produced targets or all targets, produces the same result.

lib/Pipes/FileContainer.cpp
aleclearmind: Mmmmmmh, I find this behavior surprising, but OK.

tools/model/apply/Main.cpp
aleclearmind: Uhm, you can pick the `x86` `calc` artifact, extract the model, `sed s/functionname/otherfunctionname/`, compute the diff, apply, apply the opposite, test they are equal.
If you don't like `sed`, you can commit a patch.

include/revng/Support/YAMLTraits.h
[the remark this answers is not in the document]
aleclearmind: Yes please, I'd like to have `llvm::yaml::Input::operator>>` used in a single place, eventually.
It scares me.

tools/CMakeLists.txt
drblallo: done that, but do you want to rename them with a particular convetion? eg revng pipeline invalidate? that requires manipulating the driver as well.

include/revng/Pipes/PipelineManager.h
drblallo: yeah, but one is used to see the current state of the system, the other is used see if the propagation of targets across contracts is correct.

tools/model/apply/Main.cpp
drblallo: instead of using sed, which would only test changes of names, i extract the model from two unrelated binaries so it tests additions, changes and removals.

include/revng/Pipes/PipelineManager.h
aleclearmind: Can you please clarify what you mean?

drblallo: i mean that when one is trying to understand what is going on with a pipeline  it has written, he usually cares about two things, all the targets currently in each container, and all targets that may be ever produced. The first is needed to understand if the pipeline had loaded what you think has loaded, and the second is used to understand if the contract are being evaluated as you think they should be evaluated.

It is similarly useful in the frontend, where one can have a list all possible targets, where those already produced are marked green.

drblallo: i am ok with replacing allPossible with something different like get allProducible  (even if this makes it looks like that it does not contain the already produced ones) or allAchievable

drblallo: what about `getCurrentTargets()` and `getAllTargets()`

include/revng/Model/BinaryDiff.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
```
aleclearmind, line 2: License

include/revng/Model/BinaryDiff.h
```diff
@@ -0,0 +1,10 @@
...
+#include "revng/Model/Binary.h"
+#include "revng/Model/Generated/AllTypesVariant.h"
+#include "revng/TupleTree/TupleTreeDiff.h"
+
+template<>
+struct TupleTreeEntries<model::Binary> {
+  using Types = model::AllTypes;
+};
```
aleclearmind, line 10: Uhm, we can generate this. In fact, instead of `AllTypes` we could just specialize the `TupleTreeEntries` trait.

include/revng/TupleTree/TupleTreeDiff.h
```diff
@@ -38,17 +43,106 @@ void addToContainer(C &Container, const typename C::value_type &Value) {
...
+    return Change(std::move(Path), std::nullopt, std::move(New));
+  }
+
+  static Change change(TupleTreePath Path,
+                       TupleTreeEntriesT<T> Old,
+                       TupleTreeEntriesT<T> New) {
+    revng_check(detail::checkTypeIsCorrect<T>(Path, New));
+    revng_check(detail::checkTypeIsCorrect<T>(Path, Old));
+    return Change(std::move(Path), std::move(Old), std::move(New));
+  }
```
aleclearmind, line 131: Can you prefix these methods with `create...`?

include/revng/TupleTree/TupleTreeDiff.h
```diff
@@ -57,24 +151,139 @@ struct TupleTreeDiff {
...
+      std::string SerializedPath;
+      IO.mapRequired("path", SerializedPath);
+      auto MaybePath = stringAsPath<Model>(SerializedPath);
+      revng_assert(MaybePath.has_value());
+      Info.Path = std::move(*MaybePath);
+    }
+
+    mapSingleEntry(IO, Info, "add", Info.New);
+    mapSingleEntry(IO, Info, "remove", Info.Old);
+  }
```
aleclearmind, line 286: Let's use capital letters for YAML document keys. I'm thinking to switch to lower case eventually, but I prefer coherence for now.

include/revng/TupleTree/TupleTreeDiff.h
```diff
@@ -38,17 +43,106 @@ void addToContainer(C &Container, const typename C::value_type &Value) {
...
+    revng_check(detail::checkTypeIsCorrect<T>(Path, Old));
+    return Change(std::move(Path), std::move(Old), std::move(New));
+  }
+};
+
+template<typename T>
+using ChangesVector = std::vector<Change<T>>;
+
 template<typename T>
 struct TupleTreeDiff {
```
aleclearmind, line 138: Pretty good job you did!

lib/Pipes/FileContainer.cpp
aleclearmind: Please add a comment about this.

aleclearmind (changes requested):

include/revng/Model/BinaryDiff.h
drblallo: i tried, but importing tupleTreeDiff.h not as last headers among the generated one broke everything and i have been unable to figure out what the issue was. I think it has something to do with template declaration order, i can put TupleTreeEntries in a stand alone header toh.

aleclearmind: Works. What matters is that we generate stuff and we don't have to do this for every user of `TupleTreeGenerator`.

lib/Pipes/PipelineManager.cpp
```diff
@@ -279,7 +279,42 @@ PipelineManager::store(llvm::ArrayRef<std::string> StoresOverrides) {
 }
 
 llvm::Error
-PipelineManager::printAllPossibleTargets(llvm::raw_ostream &Stream) {
+PipelineManager::invalidateAllPossibleTargets(llvm::raw_ostream &Stream) {
```
aleclearmind, line 282: Please use `Logger`. Do not use `llvm::dbgs()`.

include/revng/PipelineC/PipelineC.h
aleclearmind: Nice we're getting to this point!
However, as previously said, this API doesn't make much sense without an API to extract the model and compute the diff.
I'd expect something like:
```c++
 void rp_model_diff_apply(const rp_model_diff *diff, rp_manager *manager);
// Or...
 void rp_manager_apply_diff(rp_manager *manager, const rp_model_diff *diff);

// Model diff I/O to make the API usable from Python
const char *rp_model_diff_serialize(const rp_model_diff *diff);
rp_model_diff *rp_model_diff_create(const char *);
```

aleclearmind (changes requested):

aleclearmind:
Merged in 8c31b6d. Thanks!

PR #136, opened 2022-02-22
Fix: revng-model-inject on modeless modules

drblallo:
Inject is meant to inject a model into a module, but it is failing when operating on a module that had no model metadata at all.
This commit fixes the bug


aleclearmind:
Merged in b982b69. Thanks!

PR #137, opened 2022-02-23
EarlyFunctionAnalysis: leverage ABI-specific layer

antoniofrighetto:


include/revng/MFP/MFP.h
```diff
@@ -43,13 +43,13 @@ concept MonotoneFrameworkInstance = requires(const MFI &I,
...
+  std::same_as<typename MFI::Label,
+               typename llvm::GraphTraits<typename MFI::GraphType>::NodeRef>;
   // Disable clang-format, because it does not handle concepts very well yet
   // clang-format off
-  { I.combineValues(E1, E2) } ->same_as<typename MFI::LatticeElement>;
-  { I.isLessOrEqual(E1, E2) } ->same_as<bool>;
-  { I.applyTransferFunction(L, E2) } ->same_as<typename MFI::LatticeElement>;
+  { I.combineValues(E1, E2) } ->std::same_as<typename MFI::LatticeElement>;
+  { I.isLessOrEqual(E1, E2) } ->std::same_as<bool>;
+  { I.applyTransferFunction(L, E2) } ->std::same_as<typename MFI::LatticeElement>;
```
Cvelth, line 52: There's a minor whitespace issue here. It seems to have been there ever since type shrinking was imported. This looks a good chance to fix it.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -843,6 +875,105 @@ void FunctionEntrypointAnalyzer::interproceduralPropagation() {
...
+           RegisterStateMap &ABIRegisterMap) -> std::optional<RegisterState> {
+    llvm::StringRef Name = model::Register::getCSVName(Reg);
+    if (llvm::GlobalVariable *CSV = M.getGlobalVariable(Name, true)) {
+      auto It = ABIRegisterMap.find(CSV);
+      if (It != ABIRegisterMap.end())
+        if (isYesOrDead(It->second))
+          return It->second;
+    }
+
+    return {};
```
Cvelth, line 901: I feel like default-constructing an `std::optional` looks a bit misleading, `std::nullopt` is also a lot more explicit about what's actually going on. Don't you agree, @antoniofrighetto?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -843,6 +875,105 @@ void FunctionEntrypointAnalyzer::interproceduralPropagation() {
...
+  using Register = model::Register::Values;
+
+  auto GPRArguments = listGeneralPurposeArgumentRegisters(Binary->DefaultABI);
+  auto GPRRetVals = listGeneralPurposeReturnValueRegisters(Binary->DefaultABI);
+  auto VRArguments = listVectorArgumentRegisters(Binary->DefaultABI);
+  auto VRRetVals = listVectorReturnValueRegisters(Binary->DefaultABI);
+
+  auto TryGetRegisterState =
+    [this](Register Reg,
+           RegisterStateMap &ABIRegisterMap) -> std::optional<RegisterState> {
```
Cvelth, line 892: I feel like this a pretty sad way to format a function, but there's nothing much that can be done to improve it if we want to leave it as a lambda. Maybe it's worth to look into upgrading it into a free (or even a member) function?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -843,6 +875,105 @@ void FunctionEntrypointAnalyzer::interproceduralPropagation() {
...
+    if (auto MaybeState = TryGetRegisterState(VRArg,
+                                              Summary.ABIResults
+                                                .ArgumentsRegisters))
+      Map[VRArg].IsUsedForPassingArguments = *MaybeState;
+
+  for (auto &VRRetVal : VRRetVals)
+    if (auto MaybeState = TryGetRegisterState(VRRetVal,
+                                              Summary.ABIResults
+                                                .FinalReturnValuesRegisters))
+      Map[VRRetVal].IsUsedForReturningValues = *MaybeState;
```
Cvelth, line 926: These lines starting from `.` feel pretty bad outside of method chaining idiom. And I kind of feel responsible for changes to the convention enforcing file (I'm referring to some old mattermost discussion).

Maybe something can done to make them a tad bit prettier? Like an alias or shortening the names of both `MaybeState` and `TryGetRegisterState`.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -843,6 +875,105 @@ void FunctionEntrypointAnalyzer::interproceduralPropagation() {
...
+  using State = abi::RegisterState::Values;
+
+  if (!ABIEnforcement.getNumOccurrences() || ABIEnforcement == NoABIEnforcement)
+    return;
+
+  for (const model::Function &Function : Binary->Functions) {
+    auto &Summary = Oracle.at(Function.Entry);
+
+    abi::RegisterState::Map StateMap(Binary->Architecture);
+    initializeMapForDeductions(Summary, StateMap);
```
Cvelth, line 939: I might bit wrong, but wouldn't it be nicer to just return `StateMap` from the `initializeMapForDeductions`?

I feel opposed to taking something by reference just to initialize it. It feels way too `c`-like. Luckily copy elision also takes care of any possible overhead (in a sense, it's still gonna do the same thing (allocate memory and then pass it as a first argument to be filled in) BUT it'd be hidden from the plain view. I don't know about you but I prefer the evil I can't see over the one I can).

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -843,6 +875,105 @@ void FunctionEntrypointAnalyzer::interproceduralPropagation() {
...
+          if (Summary.ABIResults.ArgumentsRegisters.count(CSV) != 0)
+            Summary.ABIResults.ArgumentsRegisters[CSV] = State::No;
+
+        if (MaybeRV == State::Yes || MaybeRV == State::YesOrDead)
+          if (Summary.ABIResults.FinalReturnValuesRegisters.count(CSV) != 0)
+            Summary.ABIResults.FinalReturnValuesRegisters[CSV] = State::Yes;
+
+        if (MaybeRV == State::No || MaybeRV == State::NoOrDead)
+          if (Summary.ABIResults.FinalReturnValuesRegisters.count(CSV) != 0)
+            Summary.ABIResults.FinalReturnValuesRegisters[CSV] = State::No;
```
Cvelth, line 969: I feel like we're discarding any information about `Dead` registers. Is it done on purpose, if so why?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: If the function formatting is not among our upmost concerns, I'd would leave a lambda function, as it seems to be actually the case for it to exist here (it has captures, being used only within this function, not really needed to promote it to member function).

Cvelth: Yeah, that's what I meant by "member function", because to me personally a lambda that only captures `this` and is only used within the class is the same thing as a private member - except it looks somewhat worse.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Yeah, a hundred percent, it was still under experiment.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Moved it as a member function!

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Yeah, I definitely agree, it's way more explicit and quickly understandable.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Honestly, I don't have any strong feelings on this, maybe it kinda boils down to subjective feelings? It's true that it feels more `C`-like, but there are so many cases in which we modify the object by reference (even in revng), that I don't see it as that much big of a deal.

Cvelth: Yeah-yeah, it's by no means a strong opinion!

Feel free to see this comment like "In my opinion, X is slightly nicer than Y, but either way is fine."!

antoniofrighetto: Sure! Let's see if perhaps @aleclearmind has some particular concern about this, though I think it's OK.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -141,6 +143,34 @@ static opt<std::string> AAWriterPath("aa-writer",
...
+                                                    "for incompatibility, if "
+                                                    "unsure about the ABI."),
+                                         clEnumValN(FullABIEnforcement,
+                                                    "full",
+                                                    "Enforce ABI-specific "
+                                                    "information, and do not "
+                                                    "allow for any possible "
+                                                    "violations of the ABI "
+                                                    "found.")));
+
```
antoniofrighetto, line 173: After some testing, I feel like we may wish to default-initialize `ABIEnforcement` option to `FullABIEnforcement`, and fully leverage the ABI-specific layer. @aleclearmind, what do you think?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -424,6 +455,9 @@ class FunctionEntrypointAnalyzer {
                                                 const FunctionSummary &,
                                                 const efa::BasicBlock &);
   FunctionSummary importPrototype(model::FunctionType::Values, model::TypePath);
+  void initializeMapForDeductions(FunctionSummary &, abi::RegisterState::Map &);
+  std::optional<abi::RegisterState::Values>
+  tryGetRegState(model::Register::Values, const ABIAnalyses::RegisterStateMap &);
```
aleclearmind, line 460: Do not abbreviate

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -424,6 +455,9 @@ class FunctionEntrypointAnalyzer {
                                                 const FunctionSummary &,
                                                 const efa::BasicBlock &);
   FunctionSummary importPrototype(model::FunctionType::Values, model::TypePath);
+  void initializeMapForDeductions(FunctionSummary &, abi::RegisterState::Map &);
+  std::optional<abi::RegisterState::Values>
```
aleclearmind, line 459: Leave a blank line

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -843,6 +877,98 @@ void FunctionEntrypointAnalyzer::interproceduralPropagation() {
...
+    if (auto MaybeState = tryGetRegState(VRRetVal,
+                                         Results.FinalReturnValuesRegisters))
+      Map[VRRetVal].IsUsedForReturningValues = *MaybeState;
+}
+
+void FunctionEntrypointAnalyzer::applyABIDeductions() {
+  using namespace abi;
+  using State = RegisterState::Values;
+
+  if (!ABIEnforcement.getNumOccurrences() || ABIEnforcement == NoABIEnforcement)
```
aleclearmind, line 927: Don't use implicit bool conversion. `ABIEnforcement.getNumOccurrences() == 0`

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -843,6 +877,98 @@ void FunctionEntrypointAnalyzer::interproceduralPropagation() {
...
+          if (Summary.ABIResults.FinalReturnValuesRegisters.count(CSV) != 0)
+            Summary.ABIResults.FinalReturnValuesRegisters[CSV] = State::Yes;
+
+        if (MaybeRV == State::No || MaybeRV == State::NoOrDead)
+          if (Summary.ABIResults.FinalReturnValuesRegisters.count(CSV) != 0)
+            Summary.ABIResults.FinalReturnValuesRegisters[CSV] = State::No;
+      }
+    }
+
+    Summary.dump(EarlyFunctionAnalysisLog);
```
aleclearmind, line 968:
```c++
if (EarlyFunctionAnalysisLog.isEnabled()) {
  EarlyFunctionAnalysisLog << "Some header: \n";
  Summary.dump(EarlyFunctionAnalysisLog);
  EarlyFunctionAnalysisLog << DoLog;
}
```

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -843,6 +877,98 @@ void FunctionEntrypointAnalyzer::interproceduralPropagation() {
...
+  auto VRRetVals = listVectorReturnValueRegisters(Binary->DefaultABI);
+
+  const auto &Results = Summary.ABIResults;
+  for (auto &GPRArg : GPRArguments)
+    if (auto MaybeState = tryGetRegState(GPRArg, Results.ArgumentsRegisters))
+      Map[GPRArg].IsUsedForPassingArguments = *MaybeState;
+
+  for (auto &GPRRetVal : GPRRetVals)
+    if (auto MaybeState = tryGetRegState(GPRRetVal,
+                                         Results.FinalReturnValuesRegisters))
```
aleclearmind, line 910: Multiline `if`: add `{ ... }`

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -843,6 +877,98 @@ void FunctionEntrypointAnalyzer::interproceduralPropagation() {
...
+  using namespace abi;
+
+  auto GPRArguments = listGeneralPurposeArgumentRegisters(Binary->DefaultABI);
+  auto GPRRetVals = listGeneralPurposeReturnValueRegisters(Binary->DefaultABI);
+  auto VRArguments = listVectorArgumentRegisters(Binary->DefaultABI);
+  auto VRRetVals = listVectorReturnValueRegisters(Binary->DefaultABI);
+
+  const auto &Results = Summary.ABIResults;
+  for (auto &GPRArg : GPRArguments)
+    if (auto MaybeState = tryGetRegState(GPRArg, Results.ArgumentsRegisters))
```
aleclearmind, line 905: I'm confused:
* Why do you initialize them only if `isYesOrDead`?
* Why do you pass only `listGeneralPurposeArgumentRegisters` and not all of them? I don't think EFA should be aware of any detail of ABI deduction such as  `listGeneralPurposeArgumentRegisters`. It should just know about `applyRegisterStateDeductions`.

aleclearmind (changes requested):

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Needed to abbreviate here otherwise check-convention would mess around later:
```cpp
  for (auto &GPRRetVal : GPRRetVals)
    if (auto MaybeState = tryGetRegisterState(GPRRetVal,
                                              Results
                                                .FinalReturnValuesRegisters))
```

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: Solved by making everything inline :D

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -843,6 +878,98 @@ void FunctionEntrypointAnalyzer::interproceduralPropagation() {
...
+void FunctionEntrypointAnalyzer::applyABIDeductions() {
+  using namespace abi;
+  using State = RegisterState::Values;
+
+  if (ABIEnforcement.getNumOccurrences() == 0
+      || ABIEnforcement == NoABIEnforcement)
+    return;
+
+  for (const model::Function &Function : Binary->Functions) {
+    auto &Summary = Oracle.at(Function.Entry);
```
aleclearmind, line 924: * Did we already merge call-site and callee information at this stage?
* We should perform this kind of deductions for indirect function calls too.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -843,6 +878,98 @@ void FunctionEntrypointAnalyzer::interproceduralPropagation() {
...
+    RegisterState::Map StateMap(Binary->Architecture);
+    initializeMapForDeductions(Summary, StateMap);
+
+    bool EnforceABIConformance = ABIEnforcement == FullABIEnforcement;
+    std::optional<abi::RegisterState::Map>
+      ResultMap = EnforceABIConformance ?
+                    enforceRegisterStateDeductions(StateMap,
+                                                   Binary->DefaultABI) :
+                    tryApplyRegisterStateDeductions(StateMap,
+                                                    Binary->DefaultABI);
```
aleclearmind, line 935:
```c++
    std::optional<abi::RegisterState::Map> ResultMap;
    if (EnforceABIConformance) {
      ResultMap = enforceRegisterStateDeductions(StateMap,
                                                   Binary->DefaultABI) 
    } else {
      ResultMap = tryApplyRegisterStateDeductions(StateMap,
                                                    Binary->DefaultABI);
    }
```

aleclearmind (changes requested):

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
Cvelth: This is not possible. `ResultMap` has a `const` field. And @antoniofrighetto predicted that you're going to hate the ternary operator :D

Cvelth: A separate function like this might be a good compromise thought:
```cpp
std::optional<abi::RegisterState::Map> doEnforceSomethingOrOther(const abi::RegisterState::Map &StateMap, model::ABI::Values ABI) {
  if (EnforceABIConformance) {
    return enforceRegisterStateDeductions(StateMap,
                                               Binary->DefaultABI) 
  } else {
    return tryApplyRegisterStateDeductions(StateMap,
                                                Binary->DefaultABI);
  }
}
```

antoniofrighetto: Achieving this would require extra change to `abi::RegisterState::Map`: the copy assignment operator is implicitly deleted because that class has a private `const` field (expression not assignable). So, I eventually decided, for the sake of simplicity, to use the ternary operator. Does it sound OK to you?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -843,6 +878,98 @@ void FunctionEntrypointAnalyzer::interproceduralPropagation() {
...
+          if (Summary.ABIResults.ArgumentsRegisters.count(CSV) != 0)
+            Summary.ABIResults.ArgumentsRegisters[CSV] = State::No;
+
+        if (RegisterState::isYesOrDead(MaybeRV))
+          if (Summary.ABIResults.FinalReturnValuesRegisters.count(CSV) != 0)
+            Summary.ABIResults.FinalReturnValuesRegisters[CSV] = State::Yes;
+
+        if (MaybeRV == State::No || MaybeRV == State::NoOrDead)
+          if (Summary.ABIResults.FinalReturnValuesRegisters.count(CSV) != 0)
+            Summary.ABIResults.FinalReturnValuesRegisters[CSV] = State::No;
```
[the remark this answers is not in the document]
antoniofrighetto: If `MaybeArg == State::YesOrDead`, don't we want the final State to be promoted to `Yes`?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
aleclearmind: I think we should drop the `const` from the field.
`const` is hardly ever a good idea, I think. Usually you can achieve the same with data-hiding (i.e., making the field private and only provide a read-only accessor).

Data structures with private fields are needlessly cumbersome to use.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: - We did, `interproceduralPropagation` has already run by the time `applyABIDeductions` runs.
- Uhmm, we are actually only doing this per function, not per call-site (the results of the ABI analysis per call-site are left untouched). Should we do enforce the ABI also for each call-site?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
Cvelth: Given how `abi::RegisterState::Map` is structured - it's either `const` or a template. And the second option leads to its own problems.

The point is that it only keeps track of relevant registers (limited by a single architecture) - const field is there to denote which architecture it is (it's passed in the constructor).

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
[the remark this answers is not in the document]
aleclearmind: Why? This phase should just let the ABI deductions refine results. Why do some additional arbitrary refinement here?

What's the final decision (argument or not) it's a competence of the phase in which we commit to the model. I don't see why we should do this here.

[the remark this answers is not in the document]
antoniofrighetto: I'm just not sure why we cannot simply promote `YesOrDead` to `Yes` directly at this point while we are at it, but sure, that's fine.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
aleclearmind: Mmmh, why can't we simply drop `const`?

Cvelth: Because it would make it easy to achieve invalid state. The const parameter determines both how many fields the container has and their relative position. By having a way to change it at runtime, a broken overall state becomes achievable. And for what? So that it can be assigned?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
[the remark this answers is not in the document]
aleclearmind: Because there might be further refinement passes one day that decide differently.
I see no point in doing any extra processing w.r.t. to what ABI deductions decided...

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
aleclearmind: How can the user lead to an invalid state?

> So that it can be assigned to?

Well... Yes! `:)` It cannot even be moved, it's a pretty rigid data structure.

aleclearmind: If we drop `const`, *a user of the class* (which is a programmer `:)`), how can it bring it into an invalid state without changing it?
If the answer is "it can't" I think there's no benefit and we should drop `const` in favor of making it assignable, which it can always come in handy.

Cvelth: By "programmer" I meant "A person who might want to modify the class" over "a person who is going to use the class", (sorry for dodgy word selection 😉).

Technically, adding additional members (especially accessors) to a class feels less dangerous than removing a "const" from a field - so it's a lot more likely to happen.

aleclearmind: I think you're trying to defend against something you can't defend against. If I wanted to add a wrong accessor, I'd just drop `const` and go ahead and do it `:D`.

@drblallo suggests to add an assertion that `getArchitecture` doesn't return a non-`const` reference.
If you prefer we can store `FirstRegister` and `LastRegister`, but `const` fields are really annoying. In the model, we wanted to have them (for the key fields), but it made things so unergonomic that we decide to leave things public and postpone to when we had a generator to do data-hiding (which I really think is the proper way of handling the situation).

Cvelth: (Now I'm starting to feel weird about thinking to bring up to you the fact that I wanted more `const` support from tuple tree generator for #161 😅)

but back on topic. Well yeah, I can't protect it from you, but it might just help someone who isn't as quick to drop all the `const`s they see :wink:.

Either way, ok, I give up. I'll send the patch to Antonio tomorrow.

aleclearmind:
Merged in 427a3ca. Thanks!

PR #140, opened 2022-03-01
Check-conventions: handle dirs with 0 python files

AlviseDeFaveri:
Same fix as https://github.com/revng/revng/pull/139, but better according to @fcremo


aleclearmind:
Merged in 9e32668. Thanks!

PR #141, opened 2022-03-02
FilteredGraphs: add filter() method

pfez:
The method allows to access the filter function and is useful for templating.


aleclearmind:
Merged in 269d729. Thanks!

PR #142, opened 2022-03-02
EarlyFunctionAnalysis: fix end address of a jump target basic block

antoniofrighetto:
EarlyFunctionAnalysis was considering jump targets as part of the
current basic block, thus computing the final address erroneously.

### DIFF
--- a/lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
+++ b/lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
@@ -790,7 +790,10 @@ CFEPAnalyzer<FunctionOracle>::collectDirectCFG(OutlinedFunction *F) {
           } else {
             Instruction *I = &(*Succ->begin());
 
-            if (auto *Call = getCallTo(I, PreHookMarker.F)) {
+            if (isa<ReturnInst>(I)) {
+              // Did we meet the end of the cloned function? Do nothing
+              revng_assert(Succ->getInstList().size() == 1);
+            } else if (auto *Call = getCallTo(I, PreHookMarker.F)) {
               MetaAddress Destination;
               model::FunctionEdgeType::Values Type;
               auto *CalleePC = Call->getArgOperand(1);


lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -780,6 +780,12 @@ CFEPAnalyzer<FunctionOracle>::collectDirectCFG(OutlinedFunction *F) {
           } else {
             Instruction *I = &(*Succ->begin());
 
+            // Did we meet the end of the cloned basic block?
```
aleclearmind, line 783:
```suggestion
            // Did we meet the end of the cloned function?
```
?

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -780,6 +780,12 @@ CFEPAnalyzer<FunctionOracle>::collectDirectCFG(OutlinedFunction *F) {
...
+            if (isa<ReturnInst>(I)) {
+              revng_assert(Succ->getInstList().size() == 1);
+              continue;
+            }
+
             if (auto *Call = getCallTo(I, PreHookMarker.F)) {
```
aleclearmind, line 789:
```suggestion
              // Do nothing
            } else if (auto *Call = getCallTo(I, PreHookMarker.F)) {
```

aleclearmind (changes requested):

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
antoniofrighetto: IMHO `continue` here makes the code more readable (if it's OK with you).

aleclearmind: Nah. `continue` introduces (in this case unnecessary) non-obvious control-flow. What you're handling is just another case of one of the situations handled below.

antoniofrighetto: Sure.

PR #143, opened 2022-03-04
Typescript generator

mrjackv:
Add typescript generator for tuple tree
Note: this PR depends on:
* ~~https://github.com/revng/llvm-project/pull/10~~
* ~~#150~~
* https://github.com/revng/orchestra/pull/79

### DIFF
--- a/cmake/TupleTreeGenerator.cmake
+++ b/cmake/TupleTreeGenerator.cmake
@@ -13,8 +13,11 @@ function(tuple_tree_generator_impl)
       INCLUDE_PATH_PREFIX
       JSONSCHEMA_PATH
       ROOT_TYPE
-      PYTHON_PATH)
-  set(multiValueArgs HEADERS STRING_TYPES SEPARATE_STRING_TYPES)
+      GLOBAL_NAME
+      PYTHON_PATH
+      TYPESCRIPT_PATH)
+  set(multiValueArgs HEADERS TYPESCRIPT_INCLUDE STRING_TYPES
+                     SEPARATE_STRING_TYPES)
   cmake_parse_arguments(GENERATOR "" "${oneValueArgs}" "${multiValueArgs}"
                         "${ARGN}")
   if(NOT DEFINED GENERATOR_JSONSCHEMA_PATH)
@@ -22,6 +25,9 @@ function(tuple_tree_generator_impl)
   endif()
   if(NOT DEFINED GENERATOR_ROOT_TYPE)
     set(GENERATOR_ROOT_TYPE "")
+  endif()
+  if(NOT DEFINED GENERATOR_GLOBAL_NAME)
+    set(GENERATOR_GLOBAL_NAME "")
   endif()
   if(NOT DEFINED GENERATOR_STRING_TYPES)
     set(GENERATOR_STRING_TYPES "")
@@ -31,6 +37,9 @@ function(tuple_tree_generator_impl)
   endif()
   if(NOT DEFINED GENERATOR_PYTHON_PATH)
     set(GENERATOR_PYTHON_PATH "")
+  endif()
+  if(NOT DEFINED GENERATOR_TYPESCRIPT_PATH)
+    set(GENERATOR_TYPESCRIPT_PATH "")
   endif()
 
   #
@@ -85,11 +94,27 @@ function(tuple_tree_generator_impl)
     list(APPEND EXTRA_TARGETS ${GENERATOR_PYTHON_PATH})
   endif()
 
+  #
+  # Produce TypeScript code, if requested
+  #
+  if(NOT "${GENERATOR_TYPESCRIPT_PATH}" STREQUAL "")
+    tuple_tree_generator_generate_typescript(
+      "${GENERATOR_SCHEMA_PATH}"
+      "${GENERATOR_NAMESPACE}"
+      "${GENERATOR_ROOT_TYPE}"
+      "${GENERATOR_GLOBAL_NAME}"
+      "${GENERATOR_TYPESCRIPT_INCLUDE}"
+      "${GENERATOR_STRING_TYPES}"
+      "${GENERATOR_SEPARATE_STRING_TYPES}"
+      "${GENERATOR_TYPESCRIPT_PATH}")
+    list(APPEND EXTRA_TARGETS ${GENERATOR_TYPESCRIPT_PATH})
+  endif()
+
   add_custom_target(
     "${GENERATOR_TARGET_NAME}"
     DEPENDS "${GENERATOR_SCHEMA_PATH}" ${LOCAL_GENERATED_HEADERS}
             ${LOCAL_GENERATED_IMPLS} ${GENERATOR_JSONSCHEMA_PATH}
-            ${EXTRA_TARGETS})
+            ${EXTRA_TARGETS} generate-node_modules)
 endfunction()
 
 # Extracts tuple_tree_generator YAML definitions from the given header files
@@ -147,6 +172,8 @@ set(CPP_TEMPLATES
 
 set(PYTHON_TEMPLATES "${TEMPLATES_DIR}/tuple_tree_gen.py.tpl")
 
+set(TYPESCRIPT_TEMPLATES "${TEMPLATES_DIR}/tuple_tree_gen.ts.tpl")
+
 set(SCRIPTS_ROOT_DIR "${CMAKE_SOURCE_DIR}/scripts/tuple_tree_generator")
 # The list of Python scripts is build as follows:
 #
@@ -159,15 +186,17 @@ set(TUPLE_TREE_GENERATOR_SOURCES
     "${SCRIPTS_ROOT_DIR}/tuple-tree-generate-cpp.py"
     "${SCRIPTS_ROOT_DIR}/tuple-tree-generate-jsonschema.py"
     "${SCRIPTS_ROOT_DIR}/tuple-tree-generate-python.py"
-    "${SCRIPTS_ROOT_DIR}/tuple_tree_generator/__init__.py"
-    "${SCRIPTS_ROOT_DIR}/tuple_tree_generator/generators/__init__.py"
+    "${SCRIPTS_ROOT_DIR}/tuple-tree-generate-typescript.py"
     "${SCRIPTS_ROOT_DIR}/tuple_tree_generator/generators/cppheaders.py"
+    "${SCRIPTS_ROOT_DIR}/tuple_tree_generator/generators/__init__.py"
     "${SCRIPTS_ROOT_DIR}/tuple_tree_generator/generators/jinja_utils.py"
     "${SCRIPTS_ROOT_DIR}/tuple_tree_generator/generators/jsonschema.py"
     "${SCRIPTS_ROOT_DIR}/tuple_tree_generator/generators/python.py"
-    "${SCRIPTS_ROOT_DIR}/tuple_tree_generator/schema/__init__.py"
+    "${SCRIPTS_ROOT_DIR}/tuple_tree_generator/generators/typescript.py"
+    "${SCRIPTS_ROOT_DIR}/tuple_tree_generator/__init__.py"
     "${SCRIPTS_ROOT_DIR}/tuple_tree_generator/schema/definition.py"
     "${SCRIPTS_ROOT_DIR}/tuple_tree_generator/schema/enum.py"
+    "${SCRIPTS_ROOT_DIR}/tuple_tree_generator/schema/__init__.py"
     "${SCRIPTS_ROOT_DIR}/tuple_tree_generator/schema/schema.py"
     "${SCRIPTS_ROOT_DIR}/tuple_tree_generator/schema/struct.py")
 
@@ -227,6 +256,55 @@ function(
       ${STRING_TYPE_ARGS} ${SEPARATE_STRING_TYPE_ARGS} "${YAML_DEFINITIONS}"
     OUTPUT "${OUTPUT_PATH}"
     DEPENDS "${YAML_DEFINITIONS}" ${TUPLE_TREE_GENERATOR_SOURCES})
+endfunction()
+
+# Generates typescript files
+function(
+  tuple_tree_generator_generate_typescript
+  # Path to the yaml definitions
+  YAML_DEFINITIONS
+  # Base namespace of the generated classes (e.g. model)
+  NAMESPACE
+  # Type to use as the root of the schema
+  ROOT_TYPE
+  # Name of the global type, e.g. Model
+  GLOBAL_NAME
+  # Files to be included in the prouduced output
+  INCLUDE_FILES
+  # Types equivalent to plain strings
+  STRING_TYPES
+  # Types equivalent to plain strings that get a separate type definition
+  EXTERNAL_TYPES
+  # Output path
+  OUTPUT_PATH)
+
+  set(INCLUDE_FILES_ARGS)
+  foreach(IF ${INCLUDE_FILES})
+    list(APPEND INCLUDE_FILE_ARGS --external-file "${IF}")
+  endforeach()
+
+  set(STRING_TYPE_ARGS)
+  foreach(ST ${STRING_TYPES})
+    list(APPEND STRING_TYPE_ARGS --string-type "${ST}")
+  endforeach()
+
+  set(EXTERNAL_TYPE_ARGS)
+  foreach(ET ${EXTERNAL_TYPES})
+    list(APPEND EXTERNAL_TYPE_ARGS --external-type "${ET}")
+  endforeach()
+
+  add_custom_command(
+    COMMAND
+      "${SCRIPTS_ROOT_DIR}/tuple-tree-generate-typescript.py" --namespace
+      "${NAMESPACE}" --root-type "${ROOT_TYPE}" --output "${OUTPUT_PATH}"
+      --global-name "${GLOBAL_NAME}" --prettier
+      "${CMAKE_BINARY_DIR}/node_build/node_modules/.bin/prettier"
+      ${INCLUDE_FILE_ARGS} ${STRING_TYPE_ARGS} ${EXTERNAL_TYPE_ARGS}
+      "${YAML_DEFINITIONS}"
+    OUTPUT "${OUTPUT_PATH}"
+    DEPENDS "${YAML_DEFINITIONS}" ${TYPESCRIPT_TEMPLATES}
+            "${CMAKE_SOURCE_DIR}/typescript/model.ts"
+            ${TUPLE_TREE_GENERATOR_SOURCES})
 endfunction()
 
 # Generates python files
@@ -301,10 +379,13 @@ function(target_tuple_tree_generator TARGET_ID)
       SCHEMA_PATH
       JSONSCHEMA_PATH
       ROOT_TYPE
+      GLOBAL_NAME
       INCLUDE_PATH_PREFIX
       PYTHON_PATH
+      TYPESCRIPT_PATH
       HEADERS_PATH)
-  set(multiValueArgs HEADERS STRING_TYPES SEPARATE_STRING_TYPES)
+  set(multiValueArgs HEADERS TYPESCRIPT_INCLUDE STRING_TYPES
+                     SEPARATE_STRING_TYPES)
   cmake_parse_arguments(GEN "${options}" "${oneValueArgs}" "${multiValueArgs}"
                         "${ARGN}")
 
@@ -341,12 +422,18 @@ function(target_tuple_tree_generator TARGET_ID)
     "${GEN_JSONSCHEMA_PATH}"
     ROOT_TYPE
     ${GEN_ROOT_TYPE}
+    GLOBAL_NAME
+    ${GEN_GLOBAL_NAME}
     STRING_TYPES
     "${GEN_STRING_TYPES}"
     SEPARATE_STRING_TYPES
     "${GEN_SEPARATE_STRING_TYPES}"
     PYTHON_PATH
-    ${GEN_PYTHON_PATH})
+    ${GEN_PYTHON_PATH}
+    TYPESCRIPT_PATH
+    ${GEN_TYPESCRIPT_PATH}
+    TYPESCRIPT_INCLUDE
+    ${GEN_TYPESCRIPT_INCLUDE})
   if(GEN_INSTALL)
     install(DIRECTORY ${GEN_HEADERS_PATH}
             DESTINATION include/revng/${GEN_HEADER_DIRECTORY})
--- a/scripts/tuple_tree_generator/tuple-tree-generate-typescript.py
+++ b/scripts/tuple_tree_generator/tuple-tree-generate-typescript.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python3
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+import argparse
+import os
+from pathlib import Path
+from subprocess import DEVNULL, run
+from tempfile import mkstemp
+
+from tuple_tree_generator import Schema, generate_typescript
+
+import yaml
+
+argparser = argparse.ArgumentParser()
+argparser.add_argument("schema", help="YAML schema")
+argparser.add_argument("--output", "-o", help="Output to this file")
+argparser.add_argument("--namespace", required=True, help="Base namespace for generated types")
+argparser.add_argument("--root-type", required=True, help="Schema root type")
+argparser.add_argument("--global-name", required=True, help="Name of the top-level object")
+argparser.add_argument("--prettier", help="Path to the prettier binary for formatting")
+argparser.add_argument("--external-file", action="append", help="Additional ts file to include")
+argparser.add_argument("--string-type", action="append", help="Treat this type as a string")
+argparser.add_argument(
+    "--external-type", action="append", help="Assume these types are externally defined. "
+)
+
+
+def main(args):
+    with open(args.schema) as f:
+        raw_schema = yaml.safe_load(f)
+
+    schema = Schema(raw_schema, args.namespace)
+    source = generate_typescript(
+        schema,
+        args.root_type,
+        args.global_name,
+        args.external_file,
+        args.string_type,
+        args.external_type,
+    )
+
+    if args.prettier is not None:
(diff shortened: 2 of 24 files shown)


scripts/tuple_tree_generator/tuple_tree_generator/generators/typescript.py
```diff
@@ -0,0 +1,191 @@
...
+    StructField,
+    SimpleStructField,
+    SequenceStructField,
+    ReferenceStructField,
+)
+
+int_re = re.compile(r"(u)?int(8|16|32|64)_t")
+CPP_TYPE = re.compile(r"^((?P<container>[^<>]+<(?P<contained>.*)>)|(?P<simple>[^<>]*))$")
+
+BUILTINS = ["string", "boolean", "bigint"]
```
aleclearmind, line 25: Uhm, `bigint` or switch to `std::string`?

scripts/tuple_tree_generator/tuple_tree_generator/generators/typescript.py
```diff
@@ -0,0 +1,191 @@
...
+CLS_BUILTINS = ["MetaAddress"]
+ACLS_BUILTINS = ["Reference", "Identifier"]
```
aleclearmind, line 27: `MetaAddress` and `Identifier` cannot be hardcoded.
Use `STRING_TYPES` and `SEPARATE_STRING_TYPES` as we do for Python.

scripts/tuple_tree_generator/tuple_tree_generator/templates/model_prelude.ts
```diff
@@ -0,0 +1,231 @@
...
+
+function randomBigint(start: bigint, end: bigint): bigint {
+    const diff = end - start;
+    for (;;) {
+        const randomLength = diff.toString().length * Math.random();
+        const value = BigInt(`0x${randomBytes(randomLength)}`);
+        if (value < diff) {
+            return value + start;
+        }
+    }
```
aleclearmind, line 133: This seem like a `do { ... } while` loop

scripts/tuple_tree_generator/tuple_tree_generator/templates/model_prelude.ts
```diff
@@ -0,0 +1,231 @@
...
+}
+
+export function genGuid(): string {
+    return randomBigint(2n ** 10n + 1n, 2n ** 64n - 1n).toString();
+}
+
+export function genPrimitiveTypeGuid(rawObject: IType): string {
+    const realObject = rawObject as IPrimitiveType;
+    const index = PrimitiveTypeKindValues.findIndex((e) => e === realObject.PrimitiveKind);
```
aleclearmind, line 142: Everything under `tuple_tree_generator` should be agnostic from the model.
This means that this code should be usable also for non-model purposes. We already have other users of `tuple_tree_generator`.

scripts/tuple_tree_generator/tuple_tree_generator/templates/model_prelude.ts
```diff
@@ -0,0 +1,231 @@
...
+
+        const doc = new yaml.Document(model, parseOptions);
+        return doc.toString(options);
+    }
+} catch(err) {
+    var _parseIModel = (undefined as (file: string) => IModel);
+    var _parseModel = (undefined as (file: string) => Model);
+    var _dumpModel = (undefined as (model: Model | IModel) => string);
+    console.log("revng-model: package yaml not found, not creating parse wrappers")
+}
```
aleclearmind, line 189: This `try ... catch` seems too big.

scripts/tuple_tree_generator/tuple_tree_generator/templates/model_prelude.ts
```diff
@@ -0,0 +1,231 @@
...
+
+export var parseIModel: (file: string) => IModel = _parseIModel;
+export var parseModel: (file: string) => Model = _parseModel;
+export var dumpModel: (model: Model | IModel) => string = _dumpModel;
+
+function isArray(obj: any): obj is Array<unknown> {
+  return Object.prototype.toString.call(obj) === "[object Array]";
```
aleclearmind, line 196: Oh jeez, for real?

scripts/tuple_tree_generator/tuple_tree_generator/templates/model_prelude.ts
```diff
@@ -0,0 +1,231 @@
...
+}
+
+function getElementByPathArr<T>(path: string[], obj: any): T {
```
aleclearmind, line 199: * Do not abbreviate
* What is this for?

scripts/tuple_tree_generator/tuple_tree_generator/generators/typescript.py
```diff
@@ -0,0 +1,191 @@
...
+                contained = match.groupdict()["contained"]
+                parts = contained.split(",")
+                return cls._remove_cpp_type(parts[-1])
+            else:
+                raise ValueError()
+        else:
+            raise ValueError()
+
+    @classmethod
+    def get_guid_field(cls, clazz: StructDefinition) -> Optional[StructField]:
```
aleclearmind, line 136: `clazz`? `:/`

scripts/tuple_tree_generator/tuple_tree_generator/generators/typescript.py
```diff
@@ -0,0 +1,191 @@
...
+from ..schema import (
+    StructDefinition,
+    StructField,
+    SimpleStructField,
+    SequenceStructField,
+    ReferenceStructField,
+)
+
+int_re = re.compile(r"(u)?int(8|16|32|64)_t")
+CPP_TYPE = re.compile(r"^((?P<container>[^<>]+<(?P<contained>.*)>)|(?P<simple>[^<>]*))$")
```
aleclearmind, line 23: What's this?

scripts/tuple_tree_generator/tuple_tree_generator/templates/model.ts.j2
```diff
@@ -0,0 +1,89 @@
+/**
```
aleclearmind, line 1: You'll need to guide me through this.

scripts/tuple_tree_generator/tuple-tree-generate-typescript.py
```diff
@@ -33,6 +39,18 @@ def main(args):
         print(source)
 
 
+def prettify_typescript(source: str) -> str:
+    tmp_dir = Path(mkdtemp())
+    source_file = tmp_dir / "source.ts"
+    source_file.write_text(source)
+    run_opts = {"cwd": tmp_dir, "check": True, "stdout": DEVNULL, "stderr": DEVNULL}
+    run(["npm", "install", "prettier"], **run_opts)
+    run(["./node_modules/.bin/prettier", "--write", "source.ts"], **run_opts)
```
aleclearmind, line 48: Uhm, we need to install `npm` to do this. We need to think how we want to this.
Squash this in the previous commit, no need to fixup new files.

cmake/TupleTreeGenerator.cmake
```diff
@@ -259,6 +259,8 @@ function(
       "${SCRIPTS_ROOT_DIR}/tuple-tree-generate-typescript.py" --namespace
       "${NAMESPACE}" --root-type "${ROOT_TYPE}" --output "${OUTPUT_PATH}"
       "${YAML_DEFINITIONS}"
+    COMMAND "${CMAKE_BINARY_DIR}/ts-ttg-test/prep-ts-serdes-test.sh"
```
aleclearmind, line 262: * `prep`? No abbreviations.
* What's serdes?

aleclearmind (changes requested):

cmake/TupleTreeGenerator.cmake
mrjackv: `prep` is for preparation, it compiles the `model.ts` file to javascript
serdes stands for `ser`ialization-`des`erialization, I'll change name to make it clearer

scripts/tuple_tree_generator/tuple_tree_generator/generators/typescript.py
mrjackv: there are other fields that have a `uint64_t` so bigint needs to stay

scripts/tuple_tree_generator/tuple_tree_generator/generators/typescript.py
aleclearmind: >  a field might contain UpcastablePointer<model::Type> but I need just Type

Where does this happen? We should have got rid of those.

scripts/tuple_tree_generator/tuple_tree_generator/generators/typescript.py
aleclearmind: Mmmh, at this point let's stick with `uint64_t` and wait for people to complain that they are interpreted as float.

scripts/tuple_tree_generator/tuple_tree_generator/templates/model_prelude.ts
mrjackv: `type = getElementByPath<PrimitiveType>('/Types/PrimitiveType-8', model)`

scripts/tuple_tree_generator/tuple_tree_generator/generators/typescript.py
mrjackv: must have missed them, removed

scripts/tuple_tree_generator/tuple_tree_generator/generators/typescript.py
mrjackv: I've refactored the `schema` package a bit, the regex is gone

scripts/tuple_tree_generator/tuple_tree_generator/templates/model_prelude.ts
mrjackv: removed the `try ... catch`, now the generated file is a bit more opinionated.

scripts/tuple_tree_generator/tuple_tree_generator/generators/typescript.py
mrjackv: I don't remember the exact reason for the switch (I think it had something to do with `bigint` or a wrapper being deserialized incorrectly), however this problem is not present anymore, so `ID`s are now `bigint`

scripts/tuple_tree_generator/tuple_tree_generator/templates/model_prelude.ts
mrjackv: Fixed

scripts/tuple_tree_generator/tuple-tree-generate-typescript.py
mrjackv: Done

scripts/tuple_tree_generator/tuple_tree_generator/generators/typescript.py
mrjackv: Fixed

cmake/TupleTreeGenerator.cmake
mrjackv: Renamed `serdes` to `roundtrip` to be consistent w.r.t. python

scripts/tuple_tree_generator/tuple_tree_generator/templates/model_prelude.ts
mrjackv: Fixed

scripts/tuple_tree_generator/tuple_tree_generator/schema/struct.py
```diff
@@ -136,19 +143,18 @@ def __init__(self, *, name, pointee_type, root_type, doc=None, optional=False, c
...
+            pointee_type = scalar_converter(self.pointee_type)
 
         if self.resolved_root_type:
             root_type = self.resolved_root_type.name
         else:
-            root_type = _scalar_python_type(self.root_type)
+            root_type = scalar_converter(self.root_type)
 
-        return f"Reference[{pointee_type}, {root_type}]"
+        return TypeInfo(pointee_type, False, root_type)
```
aleclearmind, line 157: All of this is quite confusing and calls for some re-engineering, but let's go with it for now.

typescript/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+add_custom_command(
+  OUTPUT "${CMAKE_BINARY_DIR}/node_modules"
```
aleclearmind, line 6: Using directories as target seems a bit fragile. Does `ninja clean` drops them?

typescript/generate_node_modules.sh
```diff
@@ -0,0 +1,11 @@
...
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+set -euo pipefail
+
+SCRIPT_DIR="$( dirname -- "${BASH_SOURCE[0]}" )"
+
+cp "$SCRIPT_DIR/package.json" .
+yarn install --yarnrc "$YARNRC"
```
aleclearmind, line 11: 1. Mmmh, it seems you're assuming some set `YARNRC`. Why? Isn't there a variable that `yarn` reads so we can make this `orchestra`-agnostic?
2. I guess the idea of `YARNRC` is that makes sure you don't fetch anything from the internet but it's fetched from the local mirror, is that right?
3. You're copying stuff from the source dir, but CMake does not about this. I think you should use `configure_file` in CMake and drop the script entirely in favor of just invoking `yarn install` from CMake.

scripts/tuple_tree_generator/tuple_tree_generator/__init__.py
```diff
@@ -1,6 +1,11 @@
 # This file is distributed under the MIT License. See LICENSE.md for details.
 
-from .generators import CppHeadersGenerator, JSONSchemaGenerator, PythonGenerator
+from .generators import (
+    CppHeadersGenerator,
+    JSONSchemaGenerator,
+    PythonGenerator,
+    TypescriptGenerator,
```
aleclearmind, line 7: It's called `TypeScript`, fix everywhere.

scripts/tuple_tree_generator/tuple_tree_generator/generators/typescript.py
```diff
@@ -0,0 +1,235 @@
...
+        string_types: Optional[List[str]] = None,
+        external_types: Optional[List[str]] = None,
+    ):
+        self.schema = schema
+        self.metadata = {"root": root_type, "namespace": schema.base_namespace}
+        self.string_types = string_types if string_types is not None else []
+        self.external_types = external_types if external_types is not None else []
+        self.external_files = external_files if external_files is not None else []
+        self.global_name = global_name
+        self.jenv = Environment(loader=loader)
```
aleclearmind, line 37: `jenv` is a bad name

scripts/tuple_tree_generator/tuple_tree_generator/templates/tuple_tree_gen.prelude.ts
```diff
@@ -0,0 +1,512 @@
...
+  only1: Set<T>;
+  only2: Set<T>;
+  common: Set<T>;
+}
+
+function isNull(obj: any): obj is undefined | null {
+  return obj === undefined || obj === null;
+}
+
+function objComm(obj1: any, obj2: any): Comm<string> {
```
aleclearmind, line 198: `Comm`? What's that?

scripts/tuple_tree_generator/tuple_tree_generator/templates/tuple_tree_gen.ts.tpl
```diff
@@ -0,0 +1,203 @@
+/**
+ * This file is distributed under the MIT License. See LICENSE.md for details.
+ */
+/* eslint-disable @typescript-eslint/naming-convention */
+
+import * as yaml from "yaml"
+
+type TupleTreeRoot = {{ metadata.root }}
+
+{{ load_file('tuple_tree_gen.prelude.ts') }}
```
aleclearmind, line 10: I'm not really a fan of this, isn't there a cleaner way to do this?

cmake/TupleTreeGenerator.cmake
```diff
@@ -229,6 +259,54 @@ function(
...
+  foreach(ST ${STRING_TYPES})
+    list(APPEND STRING_TYPE_ARGS --string-type "${ST}")
+  endforeach()
+
+  set(EXTERNAL_TYPE_ARGS)
+  foreach(ET ${EXTERNAL_TYPES})
+    list(APPEND EXTERNAL_TYPE_ARGS --external-type "${ET}")
+  endforeach()
+
+  add_custom_command(
```
aleclearmind, line 297: I see an `add_custom_command` without an `add_custom_target`.
How does the Python generator do this?
Who injects the dependency on the output of this command?

tests/analysis/AnalysisTests.cmake
```diff
@@ -98,6 +98,24 @@ macro(artifact_handler CATEGORY INPUT_FILE CONFIGURATION OUTPUT TARGET_NAME)
...
     )
+
+    #
+    # Do round-trip model -> ts -> model to check proper roundtrip
+    #
+    set(ROUNDTRIP_SERIALIZATION_TEST_NAME
+        test-tuple-tree-generator-typescript-model-serialization-roundtrip-${CATEGORY}-${TARGET_NAME}
+    )
+    add_test(NAME "${ROUNDTRIP_SERIALIZATION_TEST_NAME}"
+             COMMAND "${CMAKE_BINARY_DIR}/ts-ttg-test/run-ts-roundtrip-test.sh"
```
aleclearmind, line 109: Never use abbreviations.

tests/tuple-tree-generator/typescript/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+add_custom_command(
+  OUTPUT "${CMAKE_BINARY_DIR}/ts-ttg-test/out/model.js"
+  COMMAND "${CMAKE_BINARY_DIR}/ts-ttg-test/prepare-ts-roundtrip-test.sh"
+          "${CMAKE_BINARY_DIR}/lib/typescript/model.ts"
+  DEPENDS generate-revngModel-tuple-tree-code)
+
+add_custom_target(generate-model_js ALL
+                  DEPENDS "${CMAKE_BINARY_DIR}/ts-ttg-test/out/model.js")
+
+macro(artifact_handler CATEGORY INPUT_FILE CONFIGURATION OUTPUT TARGET_NAME)
```
aleclearmind, line 31: Defining `artifact_handler` without a call to `register_derived_artifacts` doesn't make sense.
What were you trying to do?

tests/tuple-tree-generator/typescript/run-ts-roundtrip-test.sh
```diff
@@ -0,0 +1,23 @@
...
+FILE="$1"
+
+revng model opt -Y --verify < "$FILE" > "$MODEL_TMP_FILE"
+node out/test.js "$MODEL_TMP_FILE" "$OUT_MODEL_TMP_FILE"
+if ! revng model diff "$MODEL_TMP_FILE" "$OUT_MODEL_TMP_FILE"; then
+  echo "Model of $FILE did not reserialize as original"
+  exit 1;
+fi
+
+rm "$MODEL_TMP_FILE" "$OUT_MODEL_TMP_FILE"
```
aleclearmind, line 23: Cleanup not happening in case of failure

typescript/CMakeLists.txt
```diff
@@ -9,3 +9,25 @@ add_custom_command(
...
+add_custom_command(
+  OUTPUT "${CMAKE_BINARY_DIR}/model.ts.tgz"
+  COMMAND "${CMAKE_BINARY_DIR}/model.ts-package/build.sh"
+  WORKING_DIRECTORY "${CMAKE_BINARY_DIR}/model.ts-package"
+  DEPENDS generate-revngModel-tuple-tree-code generate-node_modules)
+
+add_custom_target(generate-model.ts.tgz ALL
+                  DEPENDS "${CMAKE_BINARY_DIR}/model.ts.tgz")
+
+install(FILES "${CMAKE_BINARY_DIR}/model.ts.tgz"
```
aleclearmind, line 32: `.tar.gz`?

typescript/tsconfig-model.json
```diff
@@ -0,0 +1,9 @@
+{
+  "__license": "See LICENSE.md for details.",
+  "compilerOptions": {
+    "module": "commonjs",
+    "target": "es2020",
+    "declaration": true,
+    "outDir": "dist"
+  },
+}
```
aleclearmind, line 9: No newline at end of file. Are you using `revng check-conventions`?

typescript/build-model.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+set -euo pipefail
+
+cp ../lib/typescript/model.ts .
```
aleclearmind, line 8: I think this script should run from `${CMAKE_BINARY_DIR}`, `rm -rf model.ts-package`, recreate `model.ts-package`, enter, do its stuff, leave the archive in `${CMAKE_BINARY_DIR}`.

aleclearmind (changes requested):

typescript/generate_node_modules.sh
mrjackv: 1. Rechecked, there is in fact a variable. Updated orchestra PR
2. Yes, basically it sets the correct path to the local cache and tells `yarn install` to use `--offline`
3. Done

scripts/tuple_tree_generator/tuple_tree_generator/templates/tuple_tree_gen.prelude.ts
mrjackv: https://man7.org/linux/man-pages/man1/comm.1.html
It's a linux tool that given two files it splits it in 3 columns (only 1) (only 2) (common)
The behavior here is similar where I need to know keys that are only in `obj1`, `obj2` and in common

tests/analysis/AnalysisTests.cmake
mrjackv: Fixed

tests/tuple-tree-generator/typescript/run-ts-roundtrip-test.sh
mrjackv: Fixed

tests/tuple-tree-generator/typescript/CMakeLists.txt
mrjackv: I want to run round-trip tests on all bitcode files, the script does:
1. run `revng model opt` on the bitcode to generate the model
2. run the typescript wrapper to deserialize the yaml and reserialize it
3. check with `revng model diff` that nothing changed, so we know that (2) worked correctly

typescript/CMakeLists.txt
mrjackv: Node packages have the `.tgz` extension (e.g. look in `node_cache`), so I'm sticking with that extension

typescript/build-model.sh
mrjackv: Done

typescript/tsconfig-model.json
mrjackv: There is a bug, adding fix to rcc PR

scripts/tuple_tree_generator/tuple_tree_generator/templates/tuple_tree_gen.prelude.ts
aleclearmind: Uhm, OK, add documentation about that. It's not intuitive.

typescript/tsconfig-model.json
aleclearmind: The PR went in before that.

mrjackv: https://github.com/revng/revng/pull/172

cmake/TupleTreeGenerator.cmake
mrjackv: https://github.com/revng/revng/blob/develop/cmake/TupleTreeGenerator.cmake#L85
The output of the command (e.g. `model.ts`) is added as a file-level dependency to `EXTRA_TARGETS` and then all of them are added as targets

aleclearmind: Uhm OK

scripts/tuple_tree_generator/tuple_tree_generator/templates/tuple_tree_gen.prelude.ts
mrjackv: Added docstrings

scripts/tuple_tree_generator/tuple_tree_generator/templates/tuple_tree_gen.ts.tpl
mrjackv: Split out the prelude, now it's a standalone `tuple_tree.ts`

scripts/tuple_tree_generator/tuple_tree_generator/schema/struct.py
```diff
@@ -4,13 +4,19 @@
...
 
 from .definition import Definition
 
 # TODO: should we return the fields in a stable order (in all_fields, all_optional_fields, etc)
 
 
+class TypeInfo(NamedTuple):
+    type: str  # noqa: A003
+    is_array: bool
+    root_type: str
```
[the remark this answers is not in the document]
mrjackv: 1. It's needed for Typescript to properly generate the `Reference<T, M>` type
2. Fixed

scripts/tuple_tree_generator/tuple_tree_generator/generators/typescript.py
mrjackv: Renamed to `jinja_environment`

typescript/CMakeLists.txt
mrjackv: Fixed

aleclearmind:
Merged in 581b468. Thanks!

PR #144, opened 2022-03-04
Generated python model v2

fcremo:
**Note**: this PR depends on https://github.com/revng/llvm-project/pull/10

With this PR we won't rely on datamodel-codegen for generating python classes from tuple-tree-generator schemas, which instead now supports emitting python classes directly.


scripts/tuple_tree_generator/tuple_tree_generator/templates/tuple_tree_gen.py.tpl
```diff
@@ -0,0 +1,254 @@
+###- This template file is distributed under the MIT License. See LICENSE.md for details. -###
```
aleclearmind, line 1: Outline everything that does not strictly need to be generated.

python/CMakeLists.txt
```diff
@@ -59,37 +59,17 @@ foreach(SCRIPT ${SCRIPTS})
...
-  revng/model/_common/monkeypatches.py
+  revng/model/_util.py
+  revng/model/metaaddress.py
   revng/model/v1/__init__.py
-  revng/model/v1/base.py
-  revng/model/v1/metaaddress.py
-  revng/model/v1/reference.py
+  revng/model/v1/external.py
 )
+set(PYTHON_GENERATED_MODEL_PATH revng/model/v1/_generated.py)
```
aleclearmind, line 72: Dedup with `GENERATED_PYTHON_MODEL_PATH` in `lib/Model/CMakeLists.txt`

python/revng/model/_util.py
```diff
@@ -0,0 +1,52 @@
...
+
+            Size = kwargs.get("Size")
+            if Size is None:
+                raise ValueError("Must provide Size as kwarg to PrimitiveType constructor")
+
+            primitive_kind_value = enum_value_to_index(PrimitiveKind)
+            kwargs["ID"] = primitive_kind_value << 8 | Size
+        original_init(self, *args, **kwargs)
+
+    base_class.__init__ = init_with_computed_id
```
aleclearmind, line 31: I'd really like to get rid entirely of monkey patching. Can we export a `PrimitiveType` that inherits the generated `PrimitiveType` but overrides `__init__` in a simple and explicit way?

python/revng/model/metaaddress.py
```diff
@@ -0,0 +1,162 @@
...
+
+    @classmethod
+    def yaml_representer(cls, dumper: yaml.dumper.Dumper, instance: Enum):
+        return dumper.represent_str(instance.name)
+
+
+@dataclass
+class MetaAddress:
+    Address: int
+    Type: MetaAddressType
```
aleclearmind, line 30:
```suggestion
    Type: MetaAddressType = field(default=Invalid)
```
?

python/revng/model/metaaddress.py
```diff
@@ -0,0 +1,162 @@
...
+    Epoch: int = field(default=0)
+    AddressSpace: int = field(default=0)
+
+    def __init__(self, *args, **kwargs):
```
aleclearmind, line 34: This is a quite obscure constructor and lacks docs.
Can't we have a simple constructor and maybe one or more `static` methods with explicit names and simple parameter list? For instance `MetaAddress.fromString` or the like.

python/revng/model/metaaddress.py
```diff
@@ -0,0 +1,162 @@
...
+        return dumper.represent_str(repr(instance))
+
+    @staticmethod
+    def __reject_additional_kwargs(kwargs: dict):
+        additional_kwargs = kwargs.keys() - {"Address", "Type", "Epoch", "AddressSpace"}
+        if additional_kwargs:
+            raise ValueError(f"Unexpected arguments supplied to MetaAddress constructor: {additional_kwargs}")
+
+    @staticmethod
+    def __get_arg(args, kwargs, idx, name, mandatory=True, default_if_not_mandatory=None):
```
aleclearmind, line 121: * Why `__`?
* This all seems quite complicated. What's the goal? Is it strictly necessary?

scripts/parse_revng_model.py
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env python3
+import sys
+import yaml
+
+from revng import model
+
+m = yaml.load(sys.stdin, Loader=model.YamlLoader)
+
+print(yaml.dump(m, Dumper=model.YamlDumper))
```
aleclearmind, line 9: * `def main`
* What's the purpose of this tool? Is this is for testing only? Add docs or consider inlining it in `scripts/run-serialization-roundtrip-test.sh` with `python -c`.

scripts/run-serialization-roundtrip-test.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
aleclearmind, line 1: Testing scripts should stay somewhere in `tests/`

scripts/run-serialization-roundtrip-test.sh
```diff
@@ -0,0 +1,27 @@
...
+
+set -o errexit
+set -o pipefail
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+export PYTHONPATH="$BUILD_DIR/lib/python"
```
aleclearmind, line 8: You're assuming `$BUILD_DIR` is set. Basically you're assuming orchestra.
I would assume instead that `$PWD` is the one set by `add_test`.

scripts/run-serialization-roundtrip-test.sh
```diff
@@ -0,0 +1,27 @@
...
+REVNG="$BUILD_DIR/bin/revng"
```
aleclearmind, line 9: See `tests/abi/test.sh` and `tests/abi/CMakeLists.txt`. Set `PATH` through `add_test`.

scripts/run-serialization-roundtrip-test.sh
```diff
@@ -0,0 +1,27 @@
...
+
+FILE="$1"
+
+echo "Testing $FILE"
+if ! "$REVNG" model opt --verify < "$FILE" > /dev/null; then
+  echo "ERROR: Model $FILE is invalid from the start"
+  exit 1;
+fi
+
+if ! "$REVNG" model opt -Y < "$FILE" | "$SCRIPT_DIR/parse_revng_model.py" | "$REVNG" model opt --verify > /dev/null; then
```
aleclearmind, line 24:
```suggestion
if ! "$REVNG" model opt --verify -Y < "$FILE" | "$SCRIPT_DIR/parse_revng_model.py" | "$REVNG" model opt --verify > /dev/null; then
```

scripts/tuple_tree_generator/tuple-tree-generate-python.py
```diff
@@ -0,0 +1,50 @@
...
+        external_types=args.external_type,
+    )
+
+    if args.output:
+        output_path = Path(args.output)
+        # Ensure the containing directory is created
+        output_path.parent.mkdir(exist_ok=True)
+        output_path.write_text(source)
+    else:
+        print(source)
```
aleclearmind, line 45: `print` adds a `\n`, this is different from `write_text`.

aleclearmind (changes requested):

scripts/tuple_tree_generator/tuple_tree_generator/templates/tuple_tree_gen.py.tpl
fcremo: That piece is not included in the generated file (`### marks a template comment ###`), do you want to remove it anyway?

python/revng/model/_util.py
fcremo: I didn't do it that way because I thought that require code duplication when we'll have multiple versions of the model, but we can do something like this:

`common.py`
```python
def create_primitive_type_class(base_class):
    class PrimitiveType(base_class):
        def __init__(...):
            ...
    return PrimitiveType
```
    

`v1/__init__.py`
```python
from common import create_primitive_type_class
from _generated import PrimitiveType as GeneratedPrimitiveType

# Create a derived PrimitiveType
PrimitiveType = create_primitive_type_class(GeneratedPrimitiveType)
```

`v2/__init__.py` would be similar

what do you think?

python/revng/model/metaaddress.py
fcremo: Yeah for sure the constructor needs documentation. I took the decision to implement a polymorphic constructor because it's what I would like to have as a python user, given that MetaAddress is pervasive I would like to have the convenience of being able to do all of the following without hassle
```python
MetaAddress("0x0:Invalid") # construction from serialized form
MetaAddress(0x0, "Invalid")
MetaAddress(0x0, MetaAddressType.Invalid)
MetaAddress(0x0, Type=MetaAddressType.Invalid)
MetaAddress(0x0, MetaAddressType.Invalid, 123, 456) # 123 is the epoch, 456 the address space
MetaAddress(0x0, MetaAddressType.Invalid, Epoch=123, AddressSpace=456)
MetaAddress(Address=0x0, Type=MetaAddressType.Invalid, Epoch=123, AddressSpace=456)
kwargs = {"Address": 0, Type="Invalid"} # this could come from a JSON
MetaAddress(**kwargs)
```
Having factory methods is pretty inconvenient IMO as long as the constructor is documented and predictable, but I recognize it's an opinionated opinion :)

There's another reason why a polymorphic constructor is "required", namely to not special case it in the YAML deserialization process, as the YAML constructor function invokes the regular constructor for all types that are not part of the schema.

python/revng/model/metaaddress.py
fcremo: `__` because AFAIK it ensures that I'm not overriding something from the dataclasses base class.
Re: 2nd question, it's a helper to get an argument that can be passed either as positional or as keyword, see the thread above

scripts/parse_revng_model.py
fcremo: Just for testing, I'll inline it

scripts/tuple_tree_generator/tuple_tree_generator/templates/tuple_tree_gen.py.tpl
aleclearmind: No, I mean, all the classes that have nothing generated. Outline them.

python/revng/model/_util.py
aleclearmind: Mmmh, what about
```python
class PrimitiveType(GeneratedPrimitiveType, MyOtherTypeOverridingInit):
  pass
```
? Would this work?
I feel like we should strive for avoiding having functions create new types or monkeypatching. Autocompleters will become very unhappy.

python/revng/model/metaaddress.py
aleclearmind: This is quite complicated to implement and document. Having `MetaAddress.fromString` is obvious and self-documenting.
I'd have a constructor that accepts `kwargs` and a `MetaAddress.fromString` `@classmethod` and be done with it.
KISS.

lib/Model/CMakeLists.txt
```diff
@@ -58,6 +58,7 @@ tuple_tree_generator(
   GENERATED_HEADERS
   GENERATED_IMPLS
   "${JSONSCHEMA_PATH}"
+  "${CMAKE_BINARY_DIR}/lib/python/${PYTHON_GENERATED_MODEL_PATH}"
```
aleclearmind, line 61: Wouldn't it be better to push `${CMAKE_BINARY_DIR}/lib/python/` within `PYTHON_GENERATED_MODEL_PATH`?
Otherwise it has a strange value.

python/revng/model/_util.py
```diff
@@ -0,0 +1,27 @@
+from enum import Enum
+
+
+def enum_value_to_index(enum_value: Enum):
+    """Converts an enum value to its index"""
+    return list(enum_value.__class__.__members__).index(enum_value.value)
+
+
+def force_constructor_kwarg(base_class: type, kwarg_name, kwarg_value):
```
aleclearmind, line 9:
```suggestion
# TODO: once tuple_tree_generator becomes kind-aware, this will no longer be necessary
def force_constructor_kwarg(base_class: type, kwarg_name, kwarg_value):
```

aleclearmind:
Did you test with `vscode`/`pycharm`? Do we get a nice autocomplete?

scripts/tuple_tree_generator/tuple_tree_generator/templates/tuple_tree_gen.py.tpl
```diff
@@ -0,0 +1,108 @@
...
+from .external import 't'
+## endfor ##
+
+
+class YamlLoader(_ExternalYamlLoader):
+    pass
+
+
+class YamlDumper(_ExternalYamlDumper):
+    pass
```
aleclearmind, line 24: Did you understand if the registry is reset every time you inherit or what?

python/revng/model/_yaml.py
```diff
@@ -0,0 +1,28 @@
+import yaml
+
+
+class YamlLoader(yaml.Loader):
+    pass
```
aleclearmind, line 5: I guess we can drop this

python/revng/model/base.py
```diff
@@ -0,0 +1,170 @@
+from dataclasses import dataclass, fields
```
aleclearmind, line 1: It seems to me that the `python/revng/model/` directory contains non-model specific stuff.
It's probably a good time to split model-specific stuff (e.g., `MetaAddress`) from `tuple_tree_generator`'s stuff.

python/revng/model/base.py
```diff
@@ -0,0 +1,170 @@
...
+from enum import Enum
+from typing import Generic, Type, TypeVar, get_args, get_origin, get_type_hints
+
+import yaml
+
+
+no_default = object()
+
+
+@dataclass
```
aleclearmind, line 11: The other day I was using `@dataclass`, but I find it to have a subtle and undesirable feature: you can set any field you want, even if it was not defined. Is there a way to disable this behavior?

python/revng/model/base.py
```diff
@@ -0,0 +1,170 @@
...
+                raise ValueError(f"Field {field_name} is not allowed for type {cls.__name__}")
+
+            try:
+                # isinstance raises if field_spec.type is an instance of `typing._SpecialForm`
+                # such as `Optional[something]`
+                if isinstance(field_value, field_spec.type):
+                    constructor_kwargs[field_name] = field_value
+                    continue
+            except TypeError:
+                pass
```
aleclearmind, line 29: You sure `try` is the only solution here?

python/revng/model/base.py
```diff
@@ -0,0 +1,170 @@
...
+                constructor_kwargs[field_name] = instances
+
+            elif origin is Reference:
+                if isinstance(field_value, str):
+                    constructor_kwargs[field_name] = Reference(field_value)
+                else:
+                    constructor_kwargs[field_name] = Reference.from_type(field_value)
+
+            elif isinstance(field_value, field_spec_type):
+                constructor_kwargs[field_name] = field_value
```
aleclearmind, line 70: This seems redundant w.r.t. the code above using `try`.

python/revng/model/base.py
```diff
@@ -0,0 +1,170 @@
...
+
+                    instances.append(v_inst)
+
+                constructor_kwargs[field_name] = instances
+
+            elif origin is Reference:
+                if isinstance(field_value, str):
+                    constructor_kwargs[field_name] = Reference(field_value)
+                else:
+                    constructor_kwargs[field_name] = Reference.from_type(field_value)
```
aleclearmind, line 67: What's the situation here?

python/revng/model/base.py
```diff
@@ -0,0 +1,170 @@
...
+                    if isinstance(v, underlying_type):
+                        v_inst = v
+                    elif isinstance(v, str) and issubclass(underlying_type, Enum):
+                        v_inst = underlying_type(v)
+                    elif isinstance(v, dict) and issubclass(underlying_type, StructBase):
+                        v_inst = underlying_type.from_dict(**v)
+                    elif isinstance(v, str) and hasattr(underlying_type, "from_string"):
+                        v_inst = underlying_type.from_string(v)
+                    else:
+                        raise ValueError(f"Unhandled value of type {type(v)}: {v}")
```
aleclearmind, line 57: It feels like this should be its own function: it's deserialization logic even more generic than `StructBase.from_dict`.

python/revng/model/base.py
```diff
@@ -0,0 +1,170 @@
...
+import yaml
+
+
+no_default = object()
+
+
+@dataclass
+class StructBase:
+    @classmethod
+    def from_dict(cls, **kwargs):
```
aleclearmind, line 14: This function needs to be thoroughly commented.

python/revng/model/base.py
```diff
@@ -0,0 +1,170 @@
...
+    @classmethod
+    def yaml_representer(cls, dumper: yaml.dumper.Dumper, instance: Enum):
+        return dumper.represent_str(instance.name)
+
+
+_PointedType = TypeVar("_PointedType")
+_RootType = TypeVar("_RootType")
+
+
+class Reference(Generic[_PointedType, _RootType]):
```
aleclearmind, line 132: This is not really generic as of now. It basically only captures `TypePath`, not all `TupleTreeReference`s.
We'll need to get back to this, but in the meantime, put a bigass note about this.

python/revng/model/base.py
```diff
@@ -0,0 +1,170 @@
...
+
+    @classmethod
+    def yaml_representer(cls, dumper: yaml.dumper.Dumper, instance: "Reference"):
+        return dumper.represent_str(repr(instance))
+
+    def __repr__(self):
+        return self._ref_str
+
+
+def init_reference_yaml_machinery(loader: Type[yaml.Loader], dumper: Type[yaml.Dumper]):
```
aleclearmind, line 162: Not a great name

python/revng/model/metaaddress.py
```diff
@@ -0,0 +1,110 @@
...
+        if not self.is_default_address_space():
+            components.append(str(self.AddressSpace))
+        return ":".join(components)
+
+    @classmethod
+    def yaml_representer(cls, dumper: yaml.dumper.Dumper, instance: "MetaAddress"):
+        return dumper.represent_str(repr(instance))
+
+    @staticmethod
+    def __parse_int(i):
```
aleclearmind, line 86: Why `__`?

aleclearmind (changes requested):

python/revng/model/base.py
fcremo: That's standard python behavior, not dataclasses. But yes, it should be possible to override `__setattr__` in the base class and deny setting unexpected members

python/revng/model/base.py
fcremo: `Reference` can be constructed from a string `/Types/WhateverType-ID` or directly from a type

fcremo:
VSCode autocompletion works to a basic extent, however it does not suggest constructor arguments unfortunately, it's not as smart as pycharm and does not have special support for dataclasses. It should be possible to autogenerate type hints however.
image

scripts/tuple_tree_generator/tuple_tree_generator/templates/tuple_tree_gen.py.tpl
fcremo: Yes, I don't know exactly how, but running this
```python
import yaml

class YamlDumper1(yaml.Dumper):
    pass

class YamlDumper2(yaml.Dumper):
    pass

class MyClass:
    @classmethod
    def representer(cls, dumper: yaml.dumper.Dumper, instance):
        return dumper.represent_str("asdasd")


YamlDumper1.add_representer(MyClass, MyClass.representer)

instance = MyClass()

print(yaml.dump(instance, Dumper=YamlDumper1))
print(yaml.dump(instance, Dumper=YamlDumper2))
```
prints
```
asdasd
...

!!python/object:__main__.MyClass {}
```
because `YamlDumper2` does not have an associated representer for `MyClass`

python/revng/model/_yaml.py
fcremo: We could, I left it there so that it's possible to globally customize the YamlLoader as well as YamlDumper by applying modifications to the base class, without touching generated files. As of now there's no such usage, that's true, so your call if you want me to remove it.

python/revng/model/base.py
fcremo: I can't think of a better place than the `revng.model` module where `MetaAddress` could be stored. Apart maybe from pushing it up to a `revng.metaaddress` module? IMO it belongs to the model

python/revng/model/base.py
fcremo: This simple trick seems to do the job :)
```
def __setattr__(self, key, value):
    # Prevent setting undefined attributes
    if self.__dataclass_fields__.get(key) is None:
        raise AttributeError(f"Cannot set attribute {key} for class {type(self).__name__}")
    super().__setattr__(key, value)
```

python/revng/model/base.py
fcremo: Nope, I found a way more elegant solution

scripts/tuple_tree_generator/tuple_tree_generator/templates/tuple_tree_gen.py.tpl
aleclearmind: Gotcha

lib/Model/CMakeLists.txt
aleclearmind: Why not?
Maybe he is the special kid and you want to compute the relative path there?
Otherwise `PYTHON_GENERATED_MODEL_PATH` is quite misleading.

python/revng/model/base.py
aleclearmind: I expect that during YAML-deserialization only one will be called. Which one? Are we sure we want to keep both?

python/revng/model/base.py
aleclearmind: I think there was a misunderstanding. `MetaAddress` belongs to the model. But many others in this file are *not* model-specific but generic to tupletree: `StructBase`, `EnumBase`, `Reference` (in the future).

fcremo:
Attributes on an instance do get autocompleted, there's no autocompletion on the arguments of the constructor.

python/revng/model/base.py
fcremo: During deserialization only the string one will be called, I'll remove the other case

lib/Model/CMakeLists.txt
fcremo: `python_module` assumes that the paths it is passed have to be installed in `${CMAKE_BINARY_DIR}/lib/python`, I tried to refactor it and anything I came up with is worse than this

python/revng/tupletree.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def _field_is_optional(field):
+    field_metadata = field.metadata
+    if field_metadata is None:
+        return False
+    return field_metadata.get("optional", False)
+
+
+class YamlLoader(yaml.Loader):
```
aleclearmind, line 173:
```
$ wc -l asd.yml 
21307 asd.yml
$ cat asd.yml | time -v python lel.py  |& grep -i wall
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:02.27
```
```suggestion
class YamlLoader(yaml.CLoader):
```
```
$ cat asd.yml | time -v python lel.py  |& grep -i wall
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.59
```

[They suggest](https://pyyaml.org/wiki/PyYAMLDocumentation) to `try` loading it, might not always be available.
```python
try:
    from yaml import CLoader as Loader, CDumper as Dumper
except ImportError:
    from yaml import Loader, Dumper
```
Maybe we can emit a warning about this?

python/revng/tupletree.py
```diff
@@ -0,0 +1,199 @@
+from dataclasses import dataclass, fields
+from enum import Enum
+from typing import Generic, Type, TypeVar, get_args, get_origin, get_type_hints
+
+import yaml
+try:
+    from yaml import CLoader as Loader, CDumper as Dumper
+except ImportError:
+    from yaml import Loader, Dumper
```
aleclearmind, line 9: Warning?
It's not super nice but it could bring a huge performance improvement to users and they might want to know about it.

python/revng/model/metaaddress.py
```diff
@@ -0,0 +1,110 @@
...
+        # Optional epoch
+        rf"""(:(?P<Epoch>\d+))?"""
+        # Optional address space
+        rf"""(:(?P<AddressSpace>\d+))?"""
+        # End of the string
+        r"\Z"
+    )
+
+
+def init_metaaddress_yaml_classes(loader: Type[yaml.Loader], dumper: Type[yaml.Dumper]):
```
aleclearmind, line 107: Are we fine with this?
There's another occurrence of `yaml.Loader`.

fcremo: There's no common base class between `Loader`/`CLoader` and `Dumper`/`CDumper` unfortunately

aleclearmind: Can't you import the right one use `Loader`?

python/revng/tupletree.py
fcremo: Added a warning

aleclearmind:
Merged in 70a460c. Thanks!

PR #148, opened 2022-03-15
Add stack argument support for abi utilities

Cvelth:


lib/ABI/FunctionType.cpp
```diff
@@ -740,28 +805,60 @@ Layout::Layout(const model::CABIFunctionType &Function) :
...
   for (const model::TypedRegister &Register : Function.ReturnValues)
     ReturnValue.Registers.emplace_back(Register.Location);
 
-  // Fill callee saved registers.
-  CalleeSavedRegisters.resize(Function.PreservedRegisters.size());
-  llvm::copy(Function.PreservedRegisters, CalleeSavedRegisters.begin());
+  fillLayoutFromRawFunctionTypeHelper(Function, *this);
+}
+
+Layout::Layout(const model::RawFunctionType &Function, model::ABI::Values ABI) {
```
aleclearmind, line 841: Why do you need `ABI`? `RawFunctionType` is ABI-agnostic!

Cvelth: This is one of overloads - it sorts argument registers in the "right" order based on the passed ABI.

lib/ABI/FunctionType.cpp
```diff
@@ -154,6 +155,11 @@ class ConversionHelper {
     for (auto &Argument : *ArgumentList)
       Result.Arguments.insert(Argument);
 
+    auto StackArgumentList = convertStackArguments(Function.StackArgumentsType,
+                                                   Result.Arguments.size());
```
aleclearmind, line 159: Why is the number of arguments relevant?

lib/ABI/FunctionType.cpp
```diff
@@ -197,12 +210,40 @@ class ConversionHelper {
         revng_assert(ArgumentIterator != Function.Arguments.end());
         const model::Argument &Argument = *ArgumentIterator;
 
-        /// TODO: handle stack arguments properly.
-        /// \note: different ABIs could use different stack types.
-        /// \sa: `clrcall` ABI.
+        model::StructField Field;
+        Field.Offset = CombinedStackArgumentSize;
+        Field.CustomName = Argument.CustomName;
+        Field.OriginalName = Argument.OriginalName;
```
aleclearmind, line 216:
```suggestion
```
Ibidem

lib/ABI/FunctionType.cpp
```diff
@@ -197,12 +210,40 @@ class ConversionHelper {
...
+        Field.Type = Argument.Type;
+        StackArguments.Fields.insert(std::move(Field));
+
+        auto MaybeSize = Argument.Type.size();
+        revng_assert(MaybeSize.has_value() && MaybeSize.value() != 0);
+
+        // Take stack alignment into consideration.
```
aleclearmind, line 223:
```suggestion
        // TODO: take stack alignment into consideration.
```
?

lib/ABI/FunctionType.cpp
```diff
@@ -338,13 +377,14 @@ class ConversionHelper {
...
-              Second.CustomName = "unnamed";
-            First.CustomName.append(("+" + Second.CustomName).str());
-          }
+
+          if (First.name().empty())
+            First.CustomName = "unnamed";
+          if (Second.name().empty())
+            Second.CustomName = "unnamed";
+          First.CustomName = ("combination_of_" + First.name()).str();
+          First.CustomName.append(("_and_" + Second.name()).str());
```
aleclearmind, line 386: We never generate any name.

lib/ABI/FunctionType.cpp
```diff
@@ -370,6 +411,29 @@ class ConversionHelper {
...
 
+  static llvm::SmallVector<model::Argument, 8>
+  convertStackArguments(model::TypePath StackArgumentTypes,
+                        size_t IndexOffset) {
+    if (StackArgumentTypes.get() == nullptr)
+      return {};
+
+    auto *Pointer = llvm::dyn_cast<model::StructType>(StackArgumentTypes.get());
+    revng_assert(Pointer != nullptr,
+                 "`RFT::StackArgumentsType` must be a struct");
```
aleclearmind, line 422:
```suggestion
                 "`RawFunctionType::StackArgumentsType` must be a struct");
```

lib/ABI/FunctionType.cpp
```diff
@@ -559,6 +623,14 @@ class ConversionHelper {
...
     }
 
+    // if (DA.Registers.size() > 1 && !AT::ArgumentsArePositionBased) {
+    //   // Register based data can be further affected by the endianness
+    //   // TODO: investigate the cases where this little "reverse" is not
+    //   enough. constexpr auto Architecture = model::ABI::getArchitecture(ABI);
+    //   if (model::Architecture::isLittleEndian(Architecture))
+    //     std::reverse(DA.Registers.begin(), DA.Registers.end());
+    // }
+
```
aleclearmind, line 633: We do not commit comment code

lib/ABI/FunctionType.cpp
```diff
@@ -740,28 +805,60 @@ Layout::Layout(const model::CABIFunctionType &Function) :
...
     revng_assert(StackStruct,
                  "`RawFunctionType::StackArgumentsType` must be a struct.");
     typename Layout::Argument::StackSpan StackSpan{ 0, StackStruct->Size };
-    Arguments.emplace_back().Stack = std::move(StackSpan);
+    Output.Arguments.emplace_back().Stack = std::move(StackSpan);
   }
 
+  // Fill callee saved registers.
+  Output.CalleeSavedRegisters.resize(Function.PreservedRegisters.size());
+  llvm::copy(Function.PreservedRegisters, Output.CalleeSavedRegisters.begin());
```
aleclearmind, line 823: So this *overwrites* existing entries? `:O`

aleclearmind (changes requested):

lib/ABI/FunctionType.cpp
```diff
@@ -169,24 +175,31 @@ class ConversionHelper {
 
     model::RawFunctionType Result;
     Result.CustomName = Function.CustomName;
```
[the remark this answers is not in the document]
Cvelth: As a matter of a fact, we do not. But neither are we adding this function to the model `;)`. It's up to the user what to do with the returned `model::RawFunctionType` object.

lib/ABI/FunctionType.cpp
```diff
@@ -169,24 +175,31 @@ class ConversionHelper {
...
                                              TheBinary);
-          Argument.CustomName = FinalName;
+
+          std::string TemporaryName = ArgumentName.str().str();
+          if (ArgumentStorage.Registers.size() > 1 && !TemporaryName.empty())
+            TemporaryName = "part_of_" + std::move(TemporaryName) + "_"
+                            + std::to_string(++Index) + "_out_of_"
+                            + std::to_string(ArgumentStorage.Registers.size());
+          if (llvm::StringRef{ TemporaryName }.take_front(8) != "unnamed_")
+            Argument.CustomName = TemporaryName;
```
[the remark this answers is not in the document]
Cvelth: > What's going on here? 

When an argument has to be split, I prefer to "mark" it in some what to clearly see where it comes from.

> We never generate `CustomNames`.

Should I just write it into `OriginalName` then?

lib/ABI/FunctionType.cpp
aleclearmind: Mmmh, I understand how it could be useful, but I really want `RawFunctionType` (and all views on it) to be ABI-independent.
We're fine if it looks bad in the decompiled code, one more reason to make sure it never shows up!

lib/ABI/FunctionType.cpp
Cvelth: Nope, it's closer to "this next section is here to `take stack alignment into consideration.`".

lib/ABI/FunctionType.cpp
Cvelth: `First.OriginalName` doesn't work either?

lib/ABI/FunctionType.cpp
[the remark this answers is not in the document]
aleclearmind: Mmmmh, we're not adding it to the model, but we're adding other `StructType`s! Either we add everything or we don't add anything.
There should be only one subject responsible for adding all the things to the model. If it can't be `toRaw`'s caller, it has to be `toRaw` itself.

lib/ABI/FunctionType.cpp
Cvelth: This is just a helper to avoid code duplication. There's never anything in there.

lib/ABI/FunctionType.cpp
aleclearmind: I don't understand your answer. I think you think I was implying something with my question that I was not implying `:)`

lib/ABI/FunctionType.cpp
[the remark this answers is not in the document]
Cvelth: But if we're adding resulting function to the model, I'd push for removal of the original one and replacement of all the references. But at that point it feels like this function is doing way too much "damage" to the original model.

lib/ABI/FunctionType.cpp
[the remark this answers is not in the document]
aleclearmind: No, as a policy, we never commit to the model any generated name. They are tied to a specific version of revng and if they look bad we can't change them because they're in the model.

If you want to have nice names, we need a function that computes them dynamically reasoning on the type system.
But we never commit generated names to the model.

lib/ABI/FunctionType.cpp
aleclearmind: Right, this is not *always* a generated name, only under certain conditions.

lib/ABI/FunctionType.cpp
aleclearmind: My bad, I was in doubt since I would have expected the word align to pop up somewhere below.

lib/ABI/FunctionType.cpp
aleclearmind: No OK, but you can use `llvm::copy` to overwrite entries in a container. I've *always* and only used it with `std::back_inserter`. I just got a new insight into output iterators.

lib/ABI/FunctionType.cpp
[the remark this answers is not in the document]
Cvelth: I'm never relying on this name being there or anything. It's there for human readers. If, say, we have a function
```cpp
void function(my_cool_struct this_name_is_important_to_me);
```
but we have no choice but to split the struct into a couple of `int`s, I'd rather it become
```cpp
void function(uint64_t part_of_this_name_is_important_to_me_1_out_of_2,
              uint64_t part_of_this_name_is_important_to_me_2_out_of_2);
```
than something nameless.

lib/ABI/FunctionType.cpp
aleclearmind: What would be the use case for that?
Consider that this way, we might have two distinct `llvm::Module`s containing functions with the same `RawFunctionType` but with a different `llvm::FunctionType` due to the fact that a different ABI was used to generate the layout. And this fact is not encoded anywhere in the model.

Cvelth: The case I'm using it currently for is when I'm checking whether function semantics changed during a conversion (it also works as the `Layout` test). I just thought somebody else might find this useful as well.

If you're opposed against it that deeply, I'm Ok with moving it closer to the user.

lib/ABI/FunctionType.cpp
aleclearmind: Interesting. It'd be nice to have an `append` function that does this in `STLExtras.h`!

lib/ABI/FunctionType.cpp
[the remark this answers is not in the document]
Cvelth: None than I can think of. In all honesty, that's what bulk version (the `revng-model-opt` pass) is doing already.

I have no practical opposition to moving it all higher up. I do feel like it's wrong, but whatever.

lib/ABI/FunctionType.cpp
aleclearmind: I suggest:
1. Consider if you can change tests not to rely on this behavior
2. If it's too much effort, `toRaw` must also add to the model the converted function. Then you can expose another function (used internally by `toRaw`) which has a name that suggest that you're not going to add it to the model *but* that other stuff is going to be added to the model nevertheless and put some good Doxygen.

aleclearmind: Ah, the problem is that you might get two different orderings?
Well, then the solution is that `Layout` should provide a predictable, ABI-independent ordering!
It can be trivial (register order) or you can define a custom one you like in `namespace Architecture { }` in `Model/Register.h`.

Cvelth: The only difference is sorting - it's easy to do. So it's not the question of behaviour, it's the question of whether someone else might like the access to this functionality (I think, for example, @antoniofrighetto can find this useful at some point).

lib/ABI/FunctionType.cpp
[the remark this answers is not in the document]
Cvelth: What matters to me is that these "parts" can still be traceable - nothing more. As such "OriginalName` works.

lib/ABI/FunctionType.cpp
aleclearmind: Ah without changing the `Layout` itself? I thought you meant `orderArgumentRegisters` was to change the layout.

I honestly don't see a big use case for that, but if it the `Layout` of a `RawFunctionType` remains ABI-indepedent, I'm not against it.

lib/ABI/FunctionType.cpp
Cvelth: Here it's a bit more tricky that the other way - here we have two different arguments - there's no way to determine which one of the two to discard.

lib/ABI/FunctionType.cpp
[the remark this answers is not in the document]
aleclearmind: I think we want `toRaw` inject the new prototype in the model.
Then, we could have another function that exposes an internal functionality, but you would need to call it `convertToRawButDontAddToModelButAddSomeOtherStuffToTheModel`.

lib/ABI/FunctionType.cpp
Cvelth: Yeah, initially I intended for it to change the layout. But if you're opposed extracting also works.

lib/ABI/FunctionType.cpp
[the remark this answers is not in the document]
Cvelth: Sounds good to me.

lib/ABI/FunctionType.cpp
aleclearmind: If we think this is very important, we can turn `OriginalName` into a vector one day. Feel free to put a TODO, and if we see a surging demand for something like this, we can address it.

lib/ABI/FunctionType.cpp
Cvelth: Done: https://github.com/revng/revng/pull/148/commits/b109b7f99c65f914c8c102b26455aafbd500a730.

lib/ABI/FunctionType.cpp
Cvelth: Done: https://github.com/revng/revng/pull/148/commits/5e467c713d5d3465942fd48d20b0e0f5b9255c13.

lib/ABI/FunctionType.cpp
Cvelth: Is this what you had in mind?
https://github.com/revng/revng/pull/148/commits/093426913074536c0fb649aedcade3be3b18e8d8
https://github.com/revng/revng/pull/148/commits/037706b30087a8df7206bb3ef956f5513f1ef6b5

aleclearmind: Yes! Maybe add a simple `concept` to catch the presence of `size` and everything else can be easily checked with a concept.

lib/ABI/FunctionType.cpp
aleclearmind: Uhm, I'm more talking about the output, we're going from `RawFunctionType` to `CABIFunctionType`.

Let me try to explain myself better.

1. We're going from a `struct` representing all stack arguments (`RawFunctionType::StackArgumentsType`) to multiple arguments that will end up on the stack (`CABIFunctionType`). Right?
2. One option would be to pass a single `struct` representing all of the stack arguments, while you seem to be taking the more aggressive (and prettier!) route to add multiple arguments that will end up on the stack with the *exact same* layout as `RawFunctionType::StackArgumentsType`. Right?

What I'm saying is: if we can do this safely (i.e., no mismatches in layout), let's do it!

Cvelth: Sounds good! I just asked because maybe you knew about some irregular alignment patterns under some architectures I haven't tested under just yet. If there's nothing obvious, it's worth a shot - worst case we can always go back to "preserving" the struct in those corner cases.

lib/ABI/FunctionType.cpp
Cvelth: Like this?
https://github.com/revng/revng/pull/148/commits/5c87f65c1bb2301931b3425e57eacc7619065eca
https://github.com/revng/revng/pull/148/commits/1b83961229d3e0e2d5a5db3475bc0538935cdaa4

lib/ABI/FunctionType.cpp
Cvelth: What do you think?
https://github.com/revng/revng/pull/148/commits/4abfd3becf925c1f99a2dfed350ac7991781893d
https://github.com/revng/revng/pull/148/commits/ed4589cf2961ce2c3538d776ead4ff686f85e432

lib/ABI/FunctionType.cpp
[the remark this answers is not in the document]
Cvelth: Done: https://github.com/revng/revng/pull/148/commits/bdda96b74c961763ebbab381626c4eed68cd353e.

lib/ABI/FunctionType.cpp
aleclearmind: Don't the belong to `model::ABI` at this point, more than `Layout`?
I haven't looked into testing yet, once I do, I will be able to comment on their usage there.

lib/ABI/FunctionType.cpp
aleclearmind: That's great to hear! Be bold and test your boldness! `:)`

lib/ABI/FunctionType.cpp
aleclearmind: Yes, except for `will_be_std` `:D`

lib/ABI/FunctionType.cpp
Cvelth: Well, it does rely on ABI trait to sort them - so it "technically" has `abi` dependencies and `model` shouldn't depend on that.

lib/ABI/FunctionType.cpp
Cvelth: Any suggestions on how to name that? `:D`

lib/ABI/FunctionType.cpp
aleclearmind: OK, sorry, I meant `abi`, but not `Layout`.

lib/ABI/FunctionType.cpp
aleclearmind: I would just put a TODO there.
Several things in `Concepts.h` and `STLExtras.h` are expected to be superseded by standard stuff.
I'm pretty sure we'll review those files once upgrade the compiler or the C++ standard.

So I suggest no namespace, the fact itself that they in one of those two files is a guarantee we'll look into them.

lib/ABI/FunctionType.cpp
Cvelth: Sure, I guess. It's just if we already have a free function like `SomeType doSomething(MyWellKnownObject, SomeFlag)`, it might as well just be `SomeType MyWellKnownObject::doSomething(SomeFlag)`. Don't you agree?

lib/ABI/FunctionType.cpp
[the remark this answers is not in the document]
Cvelth: Done: https://github.com/revng/revng/pull/148/commits/91b7d815be9d8bbf0a526577804567ddb991e229.

lib/ABI/FunctionType.cpp
aleclearmind: Ah it's not a `static` method. You might drop the requirement for `Layout` and pass in a vector of registers.

lib/ABI/FunctionType.cpp
Cvelth: Done: https://github.com/revng/revng/pull/148/commits/91b7d815be9d8bbf0a526577804567ddb991e229.

lib/ABI/FunctionType.cpp
Cvelth: https://github.com/revng/revng/pull/148/commits/9acfd377760665ff6fd9163cf2deeaf028c06d3a
https://github.com/revng/revng/pull/148/commits/bc0d874f0ec8f97333080a979a903494e557d179

lib/ABI/FunctionType.cpp
Cvelth: How about now?
https://github.com/revng/revng/pull/148/commits/a4b4a95015d243e55d79f82962238b0cff026404
https://github.com/revng/revng/pull/148/commits/980dd4d6ce8dc13814c1373924ff8357ef143211

aleclearmind: Works! Why not a `const Container &Registers` instead of `Container &&Registers`?

Cvelth: It's a universal reference. Since container is template anyway, it'll deduce `const` if such a container is passed and so on. I guess there's no advantage of it now, since `SortedVector::BatchInserter::insert()` is pretty basic, but maybe one day this will allow avoiding a copy or something.

Cvelth: What I mean is that using `std::forward` it's possible to propagate a r-value status down to the elements in the `for` loop - and then those elements can be moved when they are used. I didn't do that here though, since `insert` takes a `const T&` anyway.

aleclearmind: Uhm, OK, but `std::mov`ing `enum` values does not make much sense.
Universal references are not a widely known concept so I'd limit it to where it makes sense/infrastructure or very general code.

Cvelth: Yeah, sure it's not that useful on an enum - but this function is not limited to those. It accepts a `sized_range`-like of any objects that are implicitly convertible to the said enum.

Cvelth: I guess even just seeing something like
```cpp
template<typename T> void function(const T &);
```
feels so wrong to me because it's information lost for no reason. Even if that information was not used, it still doesn't feel good to throw it away.

Either way, here you go: https://github.com/revng/revng/pull/148/commits/0eefb2cc8ad17b662d51d26e282054c9ac917c9e

aleclearmind: I understand but, while you might be very familiar with universal references, their usage and implications, for the average dev (including me!) they are delicate and not super-straightforward and trigger some further thinking.

So, one of my key goals during review is: does this change has a good return on complexity coefficient?
That is, is this piece of code as simple as it could be without compromising on some aspect (functionality, performance)?
If the answer is no, I reject the extra complexity!

tests/abi/tools/revng-ensure-rft-equivalence/Main.cpp
```diff
@@ -0,0 +1,119 @@
+/// \file Main.cpp
+/// \brief
```
aleclearmind, line 2: Can you describe *in detail* what this tool does?
> This tool takes two models, looks for all the `RawFunctionType`s with the same ID...
> Using `renvg model diff` would not work since...

aleclearmind, line 2: Can you describe *in detail* what this tool does?
> This tool takes two models, looks for all the `RawFunctionType`s with the same ID...
> Using `renvg model diff` would not work since...

tests/abi/tools/revng-ensure-rft-equivalence/Main.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+      RightModel->Model->Types.erase(LeftStackArguments->key());
+      auto *Struct = llvm::dyn_cast<model::StructType>(RightStackArguments);
+      revng_assert(Struct != nullptr);
+      auto Copy = model::UpcastableType::make<model::StructType>(*Struct);
+      Copy->ID = LeftStackArguments->ID;
+      auto ToPath = RightModel->Model->recordNewType(std::move(Copy));
+
+      Replacements.emplace(FromPath, ToPath);
+    }
```
aleclearmind, line 103: So you're overriding `StackArgumentsType` on the A model with the one from `B` model, purge duplicates and then `diff`?
Please comment.

aleclearmind, line 103: So you're overriding `StackArgumentsType` on the A model with the one from `B` model, purge duplicates and then `diff`?
Please comment.

tests/abi/tools/revng-abi-verify/Verify.cpp
```diff
@@ -7,374 +7,228 @@
...
+    if (CurrentArtifact == ParsedArtifact.end()) {
+      // Ignore types with no artifact provided.
+      continue;
+    }
+
+    if (auto *CABI = llvm::dyn_cast<model::CABIFunctionType>(Type.get())) {
+      if (auto Error = verifyImpl(abi::FunctionType::Layout(*CABI),
+                                  *CurrentArtifact,
+                                  ABI))
+        return Error;
```
aleclearmind, line 223: Why not simply...
```suggestion
      return verifyImpl(abi::FunctionType::Layout(*CABI),
                                  *CurrentArtifact,
                                  ABI);
```
?
Alsom you could keep in the `if` only the computation of the layout and have a single invocation to `verifyImpl`. No biggie though.

tests/abi/tools/revng-abi-verify/Verify.h
```diff
@@ -9,17 +9,21 @@
...
+  OnlyContinuousStackArgumentsAreSupported = 5,
+  UnknownRegister = 6,
+  UnknownStackOffset = 7,
+  ArgumentCouldNotBeLocated = 8,
+  FoundUnexpectedReturnValue = 9,
+  ExpectedReturnValueNotFound = 10,
+  UnknownReturnValueRegister = 11,
+  ReturnValueCouldNotBeLocated = 12,
+  CombinedStackArgumentsSizeIsWrong = 13
 };
```
aleclearmind, line 25: Please consider dropping this in favor of `llvm::createStringError`. In terms of error reporting, that's direction we want to go and I prefer to increase its adoption w.r.t. custom error handling.

tests/abi/tools/revng-abi-verify/Verify.cpp
```diff
@@ -7,374 +7,228 @@
...
+      if (OffsetIndex >= Iteration.Stack.size())
+        return std::nullopt;
+
+      auto &Result = Iteration.Stack[OffsetIndex];
+      revng_assert(Result.Offset >= 0);
+      if (static_cast<uint64_t>(Result.Offset) != OffsetIndex * RegisterSize)
+        return std::nullopt;
+      else
+        return Result.Value;
+    };
```
aleclearmind, line 122: This situation typically calls for a class.

aleclearmind (changes requested):

tests/abi/test.sh
```diff
@@ -15,53 +15,71 @@ test -n "$ABI_NAME"
 test -n "$RUNTIME_ABI_ANALYSIS_RESULT"
 test -n "$BINARY"
 
-mkdir -p "model"
+TEMPORARY_FILES="$(pwd)/model"
```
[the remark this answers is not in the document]
Cvelth: I figured, adding `set -e` counteracts this problem. But yeah, there's no such thing as too careful.

tests/abi/tools/revng-abi-verify/ABIArtifactParser.cpp
```diff
@@ -0,0 +1,188 @@
...
+  : GenericSmallVectorYamlTrait<State::Registers> {};
+template<>
+struct llvm::yaml::SequenceTraits<State::Stack>
+  : GenericSmallVectorYamlTrait<State::Stack> {};
+template<>
+struct llvm::yaml::SequenceTraits<State::Arguments>
+  : GenericSmallVectorYamlTrait<State::Arguments> {};
+template<>
+struct llvm::yaml::SequenceTraits<State::Deserialized>
+  : GenericSmallVectorYamlTrait<State::Deserialized> {};
```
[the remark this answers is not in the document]
Cvelth: I had no idea those were defined already. I guess it's safe to drop them.

tests/abi/tools/revng-abi-verify/ABIArtifactParser.cpp
```diff
@@ -0,0 +1,188 @@
...
+// `llvm::yaml` traits for the internal state.
+//
+
+template<>
+struct llvm::yaml::MappingTraits<State::Register> {
+  static void mapping(IO &IO, State::Register &N) {
+    IO.mapRequired("Name", N.Name);
+    IO.mapRequired("Value", N.Value);
+  }
+};
```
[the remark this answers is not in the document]
Cvelth: Mostly because I never actually touched it. Writing a straightforward mapping seemed faster to me at a time, especially considering that this yaml is pretty simple.

Either way, I have nothing against moving to `TupleTreeGenerator`.

tests/abi/tools/revng-abi-verify/CMakeLists.txt
```diff
@@ -3,10 +3,13 @@
 #
 
 revng_add_executable(revng-abi-verify
+  ABIArtifactParser.cpp
   Verify.cpp
   Main.cpp)
 
+llvm_map_components_to_libnames(LLVM_LIBRARIES Support Core)
```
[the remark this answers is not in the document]
Cvelth: I didn't find anything specific. But I prefer to see where the variable comes from. Especially now that I realized it's not expected to be set to something.

tests/abi/tools/revng-ensure-rft-equivalence/Main.cpp
Cvelth: Whoops. I'm sorry I forgot to fill this in.

tests/abi/tools/revng-ensure-rft-equivalence/Main.cpp
Cvelth: Nope. I'm overriding the ID and only the ID. The struct itself stays the same (as it should, otherwise test should fail).

Comment? 👌🏻

tests/abi/tools/revng-abi-verify/Verify.cpp
Cvelth: It's inside a loop. I want to drop down if there's an error but keep going if not.

tests/abi/tools/revng-abi-verify/Verify.h
Cvelth: Sure, I'll take a look.

tests/abi/tools/revng-abi-verify/Verify.cpp
Cvelth: It sure does 😅

tests/abi/tools/revng-abi-verify/ABIArtifactParser.cpp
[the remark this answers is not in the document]
aleclearmind: Let's keep it as it is for now.
Things are moving in other PRs on the CMake side of `TupleTreeGenerator`.
Please leave a TODO about this.

tests/abi/tools/revng-abi-verify/Verify.cpp
aleclearmind: Aahh right

tests/abi/tools/revng-abi-verify/ABIArtifactParser.cpp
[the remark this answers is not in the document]
Cvelth: Sure.

tests/abi/tools/revng-abi-verify/ABIArtifactParser.cpp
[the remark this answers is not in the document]
Cvelth: On the other hand, I feel like "give me the default `SmallVector` sequence traits" is more explicit than defining `SequenceElementTraits` for the same type. And since a three-line specialization is required in either case, does it really matter if I just choose the more explicit one?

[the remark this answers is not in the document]
aleclearmind: If the body ever changes on the LLVM side to some internal changes, we'll have to update it.
Let's stick with default.

tests/abi/test.sh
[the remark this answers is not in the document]
Cvelth: Done: https://github.com/revng/revng/pull/148/commits/5edb58fd833e6a950eaa0f02344c7e9e7afe8e0b.

tests/abi/tools/revng-abi-verify/ABIArtifactParser.cpp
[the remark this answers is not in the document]
Cvelth: Done: https://github.com/revng/revng/pull/148/commits/cd1439dfa5a530eedd9c0fbaa674f2406750e83a.

tests/abi/tools/revng-abi-verify/ABIArtifactParser.cpp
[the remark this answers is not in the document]
Cvelth: Done: https://github.com/revng/revng/pull/148/commits/cd1439dfa5a530eedd9c0fbaa674f2406750e83a#diff-1ff4591728310ca98928087735ded8f7d2f661d82ccb9668a4e41fb01e4ee09fR33.

tests/abi/tools/revng-ensure-rft-equivalence/Main.cpp
Cvelth: Done: https://github.com/revng/revng/pull/148/commits/e02a96539f4bdb67ba41bb038122743ebaa07e12.

tests/abi/tools/revng-ensure-rft-equivalence/Main.cpp
Cvelth: What do you think?
https://github.com/revng/revng/pull/148/commits/e02a96539f4bdb67ba41bb038122743ebaa07e12

tests/abi/tools/revng-abi-verify/Verify.h
Cvelth: What about something like this?
https://github.com/revng/revng/pull/148/commits/90621ae1874c78344be2e4cf838b72aaf15485c7

tests/abi/tools/revng-abi-verify/Verify.cpp
Cvelth: Done: https://github.com/revng/revng/pull/148/commits/709a1641dcd68bc5373d98c9d1f95d42426d98c1.

tests/abi/tools/revng-abi-verify/Verify.h
aleclearmind: I don't think there's much benefit in using `std::error_code`. Either it's something you can act on (and then you should probably use a custom error, not a string error) or let's just go with error string + `llvm::inconvertibleErrorCode`, which is what we do in other parts of the codebase.

Cvelth: The whole point is that I want the code to be carried out of the test - it's easier for me to find the problem having "`revng-abi-verify` failed with code 11" rather than some log entry about what happened. It's probably better for other people to have the message - that's why both seems like a solid middle ground.

PR #149, opened 2022-03-16
Miscellaneous fixes and improvements for `model::Qualifier` and `model::QualifiedType`

pfez:
This PR does the following.

1. Fixes the order of `model::Qualifier`s in `model::QualifiedType`s generated when importing by DWARF. Before, the order was broken because it assumed `Qualifier`s should be emitted in the same order as in C, while they actually need to be emitted left-to-right. This is particularly important because in C nested array qualifiers have opposite order w.r.t. to the other qualifiers, e.g. `int * const x` means "a constant pointer to int" (right-to-left), but `int *x[3][4]` means "an array with 3 elements, each of which is an array of 4 pointers to `int` (left-to-right, but only for the arrays, because the pointer is considered at the end). This weirdness of nested arrays means the C does not have a single natural order of qualifiers, so we cannot assume the ordering of `model::Qualifier`s to be like C (technically we could, but that would be cumbersome to handle, and unintuitive for users to specify in the model in YAML), so to second most natural choice is left-to-right.

2. Provides a set of useful helpers for inspecting properties of `model::Qualifier`s and `model::QualifiedType`.

### DIFF
--- a/include/revng/Model/QualifiedType.h
+++ b/include/revng/Model/QualifiedType.h
@@ -37,10 +37,18 @@ public:
   std::optional<uint64_t> size() const debug_function;
   RecursiveCoroutine<std::optional<uint64_t>> size(VerifyHelper &VH) const;
 
+  /// Checks if is a scalar type, unwrapping typedefs
   bool isScalar() const;
+  /// Checks if is a primitive type, unwrapping typedefs
   bool isPrimitive(model::PrimitiveTypeKind::Values V) const;
-  bool isVoid() const { return isPrimitive(model::PrimitiveTypeKind::Void); }
+  /// Checks if is float, unwrapping typedefs
   bool isFloat() const { return isPrimitive(model::PrimitiveTypeKind::Float); }
+  /// Checks if is void, unwrapping typedefs
+  bool isVoid() const { return isPrimitive(model::PrimitiveTypeKind::Void); }
+  /// Checks if is an array type, unwrapping typedefs
+  bool isArray() const;
+  /// Checks if is a pointer type, withouth unwrapping typedefs
+  bool isPointer() const;
 
 public:
   bool verify() const debug_function;
--- a/lib/Model/Type.cpp
+++ b/lib/Model/Type.cpp
@@ -545,7 +545,7 @@ static bool isOnlyConstQualified(const QualifiedType &QT) {
   if (QT.Qualifiers.empty() or QT.Qualifiers.size() > 1)
     return false;
 
-  return QT.Qualifiers[0].isConstQualifier();
+  return Qualifier::isConst(QT.Qualifiers[0]);
 }
 
 struct VoidConstResult {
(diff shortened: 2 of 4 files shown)


lib/Model/Type.cpp
```diff
@@ -648,28 +649,91 @@ QualifiedType::size(VerifyHelper &VH) const {
...
+}
+
+bool QualifiedType::isEventuallyArray() const {
+  return isEventuallyArrayImpl<true>(*this);
+}
+
+template<bool TraverseTypedefs = true>
+static RecursiveCoroutine<bool>
+isEventuallyPointerImpl(const model::QualifiedType &QT) {
+  const auto &IsNonConst = std::not_fn(Qualifier::isConst);
```
aleclearmind, line 685: Wouldn't `NotIsConst` be more appropriate?

include/revng/Model/QualifiedType.h
```diff
@@ -37,10 +37,34 @@ class model::QualifiedType : public model::generated::QualifiedType {
   std::optional<uint64_t> size() const debug_function;
   RecursiveCoroutine<std::optional<uint64_t>> size(VerifyHelper &VH) const;
 
+  /// \brief Checks if is a scalar type, without unwrapping typedefs
   bool isScalar() const;
+  /// \brief Checks if is a scalar type, unwrapping typedefs
+  bool isEventuallyScalar() const;
```
aleclearmind, line 43: * [ ] `\brief` is deprecated
* [ ] Honestly, the non-`Eventually` version sounds like a footgun. Let's be bold and not keep around unused, dangerous functions. This should also shrink this PR significantly.

aleclearmind (changes requested):

lib/Model/Type.cpp
pfez: Will do

include/revng/Model/QualifiedType.h
pfez: The non-`Eventually` versions are footguns unless they are exactly what you need. There's a couple of places in `revng-c` where they are exactly what we need. But well, we can probably hand-roll our step-one versions there. What do you think?

aleclearmind: Can you describe the `revng-c` use case?

pfez: About `\brief`, what should I do? Switch to non-doxygen comments, or just remove `\brief` but keep the `///` comment?

lib/Model/Type.cpp
pfez: Done!

include/revng/Model/QualifiedType.h
aleclearmind: Just drop `\brief`, it's a [stupid feature](https://lists.llvm.org/pipermail/llvm-dev/2015-May/085152.html) that weshould stop using.
We still use doxygen though! `:)`

pfez: The `revng-c` use case is when we're building pointer types in DLA, but now that I look more closely it's just assertions that a given type we've just built is a pointer at step one. Might as well use the `Eventually` version.

pfez: The `\brief` thingy is gone. I'm happy to see it disappear. It was indeed stupid.

PR #150, opened 2022-03-22
Remove tags from model YAML

mrjackv:
This PR removes tags from the model, allowing interoperability with JSON.
It also renames the type kinds to the relative class name


include/revng/Model/TypeKind.h
```diff
@@ -1,22 +0,0 @@
...
-  - name: Enum
-  - name: Typedef
-  - name: Struct
-  - name: Union
-  - name: CABIFunctionType
-  - name: RawFunctionType
-TUPLE-TREE-YAML */
-
-#include "revng/Model/Generated/Early/TypeKind.h"
-#include "revng/Model/Generated/Late/TypeKind.h"
```
aleclearmind, line 22: Nice!

python/revng/model/v1/__init__.py
```diff
@@ -5,13 +5,15 @@
...
+    (TypedefType, TypeKind.TypedefType),
+    (StructType, TypeKind.StructType),
+    (UnionType, TypeKind.UnionType),
     (CABIFunctionType, TypeKind.CABIFunctionType),
     (RawFunctionType, TypeKind.RawFunctionType),
+    (FunctionEdge, FunctionEdgeBaseKind.FunctionEdge),
+    (CallEdge, FunctionEdgeBaseKind.CallEdge),
 ]
 for t, k in _types_to_kinds:
     force_constructor_kwarg(t, "Kind", k)
```
aleclearmind, line 19: All of this can be reworked to be done directly in generated code now.
Take a look and if you have doubts ask @fcremo.

include/revng/Model/Type.h
```diff
@@ -24,8 +24,6 @@ name: Type
 doc: Base class of model types used for LLVM-style RTTI
 type: struct
 fields:
-  - name: Kind
-    type: model::TypeKind::Values
```
aleclearmind, line 28: Some observations:
* It is nice that we can drop this, but it makes things a bit magic for someone passing by.
However, even if we were to force the user to put by hand the `Kind` field... what type would it have?
* Also, below `Kind` is listed as a part of the key, we're really forcing the user to be aware that if we have `abstract: true`, there's going to be a magic field `Kind`.
* Other question: does it make sense to have a `struct` where `Kind` is not part of the key?
  We have an invariant that says that inheritors must have the same key as the base class.
  I guess it's probably sane to have it, but not strictly necessary.

Overall, I think that if `abstract: true` we should ask the user to have a field named kind whose type is `${STRUCT_NAME}Kind`. If it's not there, we should print a nice error message that instructs him on what to do.
`${STRUCT_NAME}Kind` would be autogenerated.

What's your take?

include/revng/Model/FunctionEdgeBase.h
```diff
@@ -35,6 +35,7 @@ type: struct
     doc: Type of the CFG edge
     type: FunctionEdgeType::Values
```
aleclearmind, line 36: Before, we used to identify the class based by looking at the value of `Type` (i.e., certain values were mapped to `FunctionEdge`, certain others to `CallEdge`). I'm not a fan of this newly introduced redundancy, but the CFG (including `FunctionEdgeBase` and friends) are going out of the model, so they won't be user-facing API.
So no biggie.

aleclearmind (changes requested):

scripts/tuple_tree_generator/tuple_tree_generator/schema/schema.py
```diff
@@ -90,6 +91,40 @@ def _resolve_references(self):
...
+                parent = self.get_definition_for(value._inherits)
+                if parent.fullname in children:
+                    children[parent.fullname][1].append(value)
+                else:
+                    children[parent.fullname] = (
+                        parent,
+                        [
+                            value,
+                        ],
+                    )
```
fcremo, line 107: Does the dict need to be `parent_name: str -> parent: Definition`? I don't see parent_name used below and IIRC `get_definition_for` returns always the same instance for a given type name.
I'm not sure `Definition` is hashable at the moment, but in case it's not implementing `__hash__` as `return id(self)` should be good enough.
Also, I would suggest using a `defaultdict(list)` so you can remove the `if parent.fullname in children` test and just do `children[parent].append(value)`.

include/revng/Model/CallEdge.h
```diff
@@ -52,12 +52,18 @@ class model::CallEdge : public model::generated::CallEdge {
   CallEdge() : model::generated::CallEdge() { Type = AssociatedType; }
 
   CallEdge(MetaAddress Destination, FunctionEdgeType::Values Type) :
-    model::generated::CallEdge(Destination, Type) {}
+    model::generated::CallEdge(model::FunctionEdgeBaseKind::CallEdge,
```
fcremo, line 55: I would expect the Kind to be automatically provided by the generated constructor base class, can't we do that?

include/revng/Model/Type.h
aleclearmind: Yeah, documentation is always good, but I think that letting the user make a small mistake and give a very informative error on how to fix it, is also effective.

python/revng/tupletree.py
```diff
@@ -128,6 +128,18 @@ def __setattr__(self, key, value):
         super().__setattr__(key, value)
 
 
+class UpcastableStructBase(StructBase):
+    _children: List[Type] = []
+
+    @classmethod
+    def from_dict(cls, **kwargs):
+        if UpcastableStructBase in cls.__bases__ and "Kind" in kwargs:
```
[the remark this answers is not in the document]
mrjackv: That wouldn't work
e.g. `FunctionEdgeBase(UpcastableStructBase)` -> `FunctionEdge(FunctionEdgeBase, StructBase)`
so `FunctionEdge` is a subclass of `UpcastableStructBase` indirectly

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
```diff
@@ -65,7 +65,8 @@ struct /*= struct.fullname =*/
   /*= struct.name =*/(
     /*#- Inherited fields #*/
     /**- for field in struct.inherits.fields **/
-    /*=- field.type =*/ /*= field.name =*/,
+    /*=- field.type =*/ /*= field.name =*/
+    /** if (struct.fields | length > 0) or (not loop.last) **/, /** endif **/
```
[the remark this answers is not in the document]
mrjackv: yes, `FunctionEdgeBase` -> `FunctionEdge`

python/revng/tupletree.py
```diff
@@ -128,6 +128,18 @@ def __setattr__(self, key, value):
...
+            child_cls = next(c for c in cls._children if c.__name__ == kwargs["Kind"])
```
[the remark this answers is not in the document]
mrjackv: Agree, changed accordingly

scripts/tuple_tree_generator/tuple_tree_generator/schema/schema.py
mrjackv: Changed, this is way more elegant

python/revng/tupletree.py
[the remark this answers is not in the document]
fcremo: I'm not sure why that would be a problem, actually, I'm not sure why we need the test at all.
Say we have `Base` and `Derived(Base)`. `cls` is `Base` when the function is called like `Base.from_dict(...)`, as is the case during deserialization. Even if someone called `Derived.from_dict` it would still work (*), as Derived also inherits the `_children` from `Base`:
```
class A:
  pass
class B:
  pass
A.x = 1
assert B.x == 1
```

Granted, this would mean that `Derived.from_dict` could create an instance of something else which admittedly might be a bit unexpected, is this what you mean?

Because of this reasoning I argue we actually might want to raise an exception if `Kind` is not in kwargs, I don't see how that could not be an error, but I guess that the base class `from_dict` fill eventually do that anyway.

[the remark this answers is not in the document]
mrjackv: The problem is that without `__bases__` there will be a loop.
Example:

1. `from_dict` arrives at a `FunctionEdgeBase` and calls its `from_dict`
2. in `FunctionEdgeBase.from_dict` we have `UpcastableStructBase` in `__bases__` and a `Kind`, we call the relative class, let's say `FunctionEdge`
3. now we are in `FunctionEdge.from_dict` and we want to bail back in `StructBase.from_dict`, if we check `issubclass` it will return `True` since we have the chain `FunctionEdge -> FunctionEdgeBase -> UpcastableStructBase` and we go back to (2) whereas `__bases__` will be `[ FunctionEdgeBase, StructBase ]` and so the check passes

include/revng/Model/Type.h
mrjackv: Ok, I see your point.

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
```diff
@@ -118,7 +118,7 @@ struct /*= struct.fullname =*/
 
   bool localCompare(const /*= struct.user_fullname =*/ &Other) const;
 
-  static constexpr const char *Tag = "!/*= struct.name =*/";
+  inline static const std::string ClassName = "/*= struct.name =*/";
```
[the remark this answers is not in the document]
mrjackv: should I keep the `std::string` or convert to `char *`?

include/revng/ADT/UpcastablePointer/YAMLTraits.h
```diff
@@ -14,7 +14,9 @@ void initializeOwningPointer(llvm::yaml::IO &TheIO, O &Obj) {
 
   if constexpr (I < std::tuple_size_v<concrete_types>) {
     using type = typename std::tuple_element_t<I, concrete_types>;
-    if (TheIO.mapTag(type::Tag)) {
+    std::string kind;
```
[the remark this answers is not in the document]
mrjackv: Done

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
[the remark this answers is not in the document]
mrjackv: Done

aleclearmind:
I'd like this MR to go in after #125, which makes significant changes to the model which make some things in here more palatable (since they will no longer be exposed to the user).

include/revng/Model/TypeKind.h
```diff
@@ -1,22 +0,0 @@
...
-  - name: EnumType
-  - name: TypedefType
-  - name: StructType
-  - name: UnionType
-  - name: CABIFunctionType
-  - name: RawFunctionType
-TUPLE-TREE-YAML */
-
-#include "revng/Model/Generated/Early/TypeKind.h"
-#include "revng/Model/Generated/Late/TypeKind.h"
```
aleclearmind, line 22: Yeeee

scripts/tuple_tree_generator/tuple_tree_generator/schema/schema.py
```diff
@@ -90,6 +92,38 @@ def _resolve_references(self):
...
+                raise ValueError("Kind field must be present in abstract classes")
+            if kind_field[0].type != f"{parent.user_namespace}::{parent.name}Kind::Values":
+                raise ValueError(
+                    f"""
+                    Kind field in {parent.name} must have the type
+                    '{parent.user_namespace}::{parent.name}Kind::Values'.
+                    This enum will be autogenerated with the child names
+                    of this class.
+                """
+                )
```
aleclearmind, line 114: I think this message will be printed with indentation.

scripts/tuple_tree_generator/tuple_tree_generator/schema/schema.py
```diff
@@ -90,6 +92,38 @@ def _resolve_references(self):
...
+
+            kind_enum_namespace = f"{parent.user_namespace}::{parent.name}Kind"
+            kind_enum = EnumDefinition(
+                namespace=kind_enum_namespace,
+                user_namespace=kind_enum_namespace,
+                name=f"{parent.name}Kind",
+                members=[EnumMember(name=c.name) for c in child_list],
+            )
+            kind_enum.autogenerated = True
+            self.definitions[f"{kind_enum_namespace}::Values"] = kind_enum
```
aleclearmind, line 125: I expected to see this in jinja, but this maybe even better.

scripts/tuple_tree_generator/tuple_tree_generator/schema/struct.py
```diff
@@ -208,7 +209,10 @@ def resolve_references(self, schema):
         for dep in self.dependencies:
             dep_definition = schema.get_definition_for(dep)
             if dep_definition:
-                self.includes.add(dep_definition.filename)
+                if not dep_definition.autogenerated:
+                    self.includes.add(dep_definition.filename)
+                else:
+                    self.includes.add(f"Generated/Early/{dep_definition.filename}")
```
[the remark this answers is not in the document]
aleclearmind: @mrjackv Why did you mark this as solved? Please do not mark as solved comments that are still relevant and for which you did not give any answer.

The situation for this is bad and is getting worse, we need to simplify this stuff.

python/revng/tupletree.py
```diff
@@ -128,6 +128,24 @@ def __setattr__(self, key, value):
...
+            child_cls = cls._children.get(kwargs["Kind"])
+            if child_cls:
+                if cls != child_cls:
+                    return child_cls.from_dict(**kwargs)
+                else:
+                    return super().from_dict(**kwargs)
+            else:
+                raise ValueError(f"No class found to deserialize {kwargs['Kind']}")
+        else:
+            raise ValueError("Upcastable types must have a Kind field")
```
aleclearmind, line 154: Do not nest unnecessarily: go from
```python
if a:
  ...
else:
  raise
```
to:
```python
if not a:
  raise
...
```

python/revng/tupletree.py
```diff
@@ -128,6 +128,24 @@ def __setattr__(self, key, value):
         super().__setattr__(key, value)
 
 
+class UpcastableStructBase(StructBase):
```
aleclearmind, line 139: More like `AbstractStructBase`?

scripts/tuple_tree_generator/tuple_tree_generator/templates/tuple_tree_gen.py.tpl
```diff
@@ -95,6 +101,20 @@ class 'struct.name'(
 
 ## endfor ##
 
+## for struct in structs ##
+## if struct.abstract -##
+'struct.name'._children = {
+        ##- for child in struct.children -##
+        "'-child.name'": '-child.name' ##- if not loop.last -## , ##- endif -##
+        ##- endfor -##
+    }
```
aleclearmind, line 110: Mmmh, this out-of-line definition doesn't sound right. Can we push this in the body of the `class` definition?

scripts/tuple_tree_generator/tuple_tree_generator/templates/tuple_tree_gen.py.tpl
```diff
@@ -95,6 +101,20 @@ class 'struct.name'(
...
+
+## for child in struct.children -##
+force_constructor_kwarg('child.name', "Kind", 'struct.name'Kind.'child.name')
+## endfor ##
```
aleclearmind, line 114: Comment, I've no idea what this does.

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
```diff
@@ -56,7 +64,11 @@ struct /*= struct.fullname =*/
     /**- for field in struct.key_fields **/
     /*=- field.name =*/(/*= field.name =*/)/** if not loop.last **/, /** endif **/
     /**- endfor **/
-    /**- endif **/ {}
+    /**- endif **/ {
+      /**- if struct.inherits and not 'Kind' in struct.key_fields | map(attribute='name') -**/
+      Kind = AssociatedKind;
```
aleclearmind, line 69: Why don't you do this in the initialization list?

python/revng/tupletree.py
```diff
@@ -28,6 +29,13 @@ def _create_instance(field_value, field_type):
         raise TypeError(f"Invalid type {type(field_value)}, was expecting {field_type}")
 
 
+# Hot function, hence the lru_cache
+# Called once per field for each object instantiation
+@lru_cache(maxsize=1024, typed=True)
+def get_type_hint(clazz, name):
+    return get_type_hints(clazz)[name]
```
aleclearmind, line 36: * `clazz`?
* Call this `get_type_hints_cached` or something like this

aleclearmind (changes requested):

scripts/tuple_tree_generator/tuple_tree_generator/schema/schema.py
mrjackv: This is transparent to the generator, so if tomorrow someone writes e.g. a `ruby` model generator it automatically gets the computed kind enum

scripts/tuple_tree_generator/tuple_tree_generator/templates/tuple_tree_gen.py.tpl
mrjackv: No, because there is a cyclic dependency:
* The base class needs to be defined for the children to subclass it
* The children would need to be before the base class to have a child list in the base class

This is (AFAIK) the most straightforward way of solving this.

scripts/tuple_tree_generator/tuple_tree_generator/templates/tuple_tree_gen.py.tpl
mrjackv: It's a function that patches the `__init__` of the child class, it does the following:
* Automatically setting the `Kind` field if not specified.
* If `Kind` is present, assert that it is the correct one.
See https://github.com/mrjackv/revng/blob/feature/remove-model-tag/python/revng/model/_util.py#L10

This is the autogenerated version of [this](https://github.com/revng/revng/blob/de033b50f20019a2ca1d996e0b845e3962688cbe/python/revng/model/v1/__init__.py#L17)

scripts/tuple_tree_generator/tuple_tree_generator/schema/struct.py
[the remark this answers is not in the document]
mrjackv: We discussed the situation on the internal chat, the solution for now was to include a comment [here](https://github.com/mrjackv/revng/blob/38c13fa6224a4a4651e3d2eefe7477b582a5fffe/scripts/tuple_tree_generator/tuple_tree_generator/schema/struct.py#L208) and [here](https://github.com/mrjackv/revng/blob/38c13fa6224a4a4651e3d2eefe7477b582a5fffe/cmake/TupleTreeGenerator.cmake#L308)

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
mrjackv: 2 reasons:
* Kind belongs to the base class, it cannot be initialized in the initialization list
* Putting the kind in the parent's initialization (e.g. `Type(AssociatedKind)`) would mean foregoing the default constructor.

Also, this replaces the same [manual implementation](https://github.com/revng/revng/blob/de033b50f20019a2ca1d996e0b845e3962688cbe/include/revng/Model/StructType.h#L38)

scripts/tuple_tree_generator/tuple_tree_generator/templates/tuple_tree_gen.py.tpl
aleclearmind: I mean: add a comment.

mrjackv: Ok 👍

python/revng/tupletree.py
mrjackv: `class` is a reserved keyword, so there are alternative names like `class_` or `clazz`
I checked [PEP-8](https://peps.python.org/pep-0008/#descriptive-naming-styles) and the preferred name is `class_`, so I'll be switching to that

python/revng/tupletree.py
mrjackv: Fixed

python/revng/tupletree.py
mrjackv: Yes, that is more appropriate

scripts/tuple_tree_generator/tuple_tree_generator/schema/schema.py
mrjackv: Indeed, fixed

mrjackv:
> Some commits do not respect the 50/72 rule.

Fixed

scripts/tuple_tree_generator/tuple_tree_generator/templates/tuple_tree_gen.py.tpl
aleclearmind: Sad, Python really needs forward declarations `:P`

scripts/tuple_tree_generator/tuple_tree_generator/schema/struct.py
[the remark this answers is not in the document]
aleclearmind: I'm not really happy with this, we'll need to go through this again.

aleclearmind:
Merged in 364be5e. Thanks!

PR #152, opened 2022-03-22
Feature/pipeline call graph cache preliminaries

drblallo:


include/revng/Pipeline/GenericLLVMPipe.h
```diff
@@ -133,6 +141,13 @@ class LLVMPassWrapper : public LLVMPassWrapperBase {
   std::unique_ptr<LLVMPassWrapperBase> clone() const override {
     return std::make_unique<LLVMPassWrapper>(*this);
   }
+
+  void print(llvm::raw_ostream &OS) const override {
+    if constexpr (LLVMPrintablePass<T>)
+      PipePass.print(OS);
+    else
+      OS << "-" << T::Name;
```
aleclearmind, line 149: First case is for custom LLVM pipes that might run multiple passes, while the second is for contractless passes.
Correct?

lib/Pipes/LinkForTranslationPipe.cpp
```diff
@@ -214,4 +213,19 @@ void LinkForTranslationPipe::run(const Context &Ctx,
...
+  linkForTranslationImpl(Ctx,
+                         InputBinary.name(),
+                         *InputBinary.path(),
+                         *ObjectFile.path(),
+                         OutputBinary.getOrCreatePath());
+}
+
+void LinkForTranslationPipe::print(const Context &Ctx,
+                                   llvm::raw_ostream &OS) const {
+}
```
aleclearmind, line 229: Why empty?

lib/Pipes/PipelineManager.cpp
```diff
@@ -47,6 +47,8 @@ class LoadModelPipePass {
   void registerPasses(llvm::legacy::PassManager &Manager) {
     Manager.add(new LoadModelWrapperPass(Wrapper));
   }
+
+  void print(llvm::raw_ostream &OS) const { OS << ""; }
```
aleclearmind, line 51: Why empty?

lib/PipelineC/PipelineC.cpp
```diff
@@ -29,7 +29,10 @@ void rp_set_custom_abort_hook(AbortHook Hook) {
   setAbortHook(Hook);
 }
 
-static char *copy_string(llvm::StringRef str) {
+/// Used when we want to return a stack allocated string.
+/// Copies the string onto the heap and gives ownership of
+/// to the caller
```
aleclearmind, line 34: This is not 80-cols aligned. Please make sure your editor is correctly set up, otherwise I always have to fix these.

tests/tools/pipeline/StringContainerLibrary.cpp
```diff
@@ -117,6 +119,6 @@ static llvm::RegisterPass<ExamplePass> X2("example-pass", "ExamplePass");
 
 char StringContainer::ID;
 
-static RegisterPipe<CopyPipe<StringContainer>> E1(Root);
+static RegisterPipe<CopyPipe<StringContainer>> E1(StringKind);
 static RegisterDefaultConstructibleContainer<StringContainer> C("StringContaine"
                                                                 "r");
```
aleclearmind, line 124: Very sad

lib/Pipes/CallGraphCache.cpp
```diff
@@ -0,0 +1,56 @@
...
+template<>
+const char pipeline::YamlizableGlobal<revng::pipes::CallGraphCache>::ID = '0';
+
+class CreateCallGraphCachePass : public llvm::ModulePass {
+public:
+  static char ID;
+  bool runOnModule(llvm::Module &Module) override {
+    ToWrite.Name = "ASD";
+    return true;
+  }
```
aleclearmind, line 23: This is not ready to be merged, please move this commit to the other PR.

aleclearmind (changes requested):

include/revng/Pipeline/GenericLLVMPipe.h
drblallo: yes and the first one is for those that have custom arguments to be set as well.

include/revng/TupleTree/TupleTreeDiff.h
```diff
@@ -271,9 +271,29 @@ inline void TupleTreeDiff<T>::dump(llvm::raw_ostream &OutputStream) const {
...
+};
+
 template<typename T>
-concept IterableAndNotStdString
-  = Iterable<T> and not std::is_same_v<std::string, T>;
+concept IterableAndNotStdString = Iterable<T> and
+                                  not std::is_same_v<std::string, T> and
+                                  not StringRefConvertible<T> and
+                                  not TwineConvertible<T> and
+                                  not StringLike<T>;
```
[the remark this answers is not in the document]
drblallo: it is used to exclude classes that extend std::string such as identifier if i remember correctly

tests/tools/pipeline/StringContainerLibrary.cpp
drblallo: eh, i mean, i can use "using", but beside that here is not a single character i can drop.

aleclearmind: `using` in `.cpp` files solves many issues.
Otherwise you can declare the string on another line.

lib/Pipes/PipelineManager.cpp
drblallo: this one is actually legit, the load model pass is added by revng opt by default, so we don't need to add it ourself

drblallo: this one is actually legit, the load model pass is added by revng opt by default, so we don't need to add it ourself

drblallo: i guess i can leave a comment

aleclearmind:
Merged in 7439b87. Thanks!

PR #153, opened 2022-03-25
Improve initialization of type-based `OpaqueFunctionsPool`

pfez:
This PR does the following things.

- preliminary work: move some useful concepts from `UpcastablePointer.h` to a new header `include/ADT/TypeTraits.h` so that they can be reused elsewhere
- new functionality: add two method for OpaqueFunctionsPool that use a pointer to `llvm::Type` (or a class derived from `llvm::Type`) as key. Both methods are used to initialize the pool with pre-existing functions inside an `llvm::Module` that match a specific `FunctionTags::Tag`:
    - `initializeFromReturnType` uses the return `llvm::Type` of the  function as key in the pool
    - `initializeFromNthArgType` takes an additional argument which is an  index N, and uses the return `llvm::Type` of the N-th argument as a  key in the pool

  These methods are useful for avoiding unnecessary proliferation of opaque functions in an `llvm::Module`, re-using the existing ones instead if they are already present.
- uses the newly created methods where possible


include/revng/ADT/TypeTraits.h
```diff
@@ -0,0 +1,13 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include <type_traits>
+
+template<typename T>
+concept NotVoid = not std::is_void_v<T>;
+
+template<class Derived, class Base>
+concept DerivesFrom = std::is_base_of_v<Base, Derived>;
```
aleclearmind, line 13: This seems to belong to `Concepts.h`, see if we can avoid a new header.

include/revng/Support/OpaqueFunctionsPool.h
```diff
@@ -94,4 +95,29 @@ class OpaqueFunctionsPool {
 
     return get(Key, FunctionType::get(ReturnType, Arguments, false), Name);
   }
+
+  /// Initialize the pool with all the functions in M that match the tag TheTag,
+  /// using the return type as key.
+  void initializeFromReturnType(const FunctionTags::Tag &TheTag) requires
+    DerivesFrom<std::remove_pointer_t<KeyT>, llvm::Type> {
```
aleclearmind, line 101: Can you make a local `concept` for this, given that you're using it twice?

include/revng/Support/OpaqueFunctionsPool.h
```diff
@@ -94,4 +95,29 @@ class OpaqueFunctionsPool {
 
     return get(Key, FunctionType::get(ReturnType, Arguments, false), Name);
   }
+
+  /// Initialize the pool with all the functions in M that match the tag TheTag,
+  /// using the return type as key.
+  void initializeFromReturnType(const FunctionTags::Tag &TheTag) requires
```
aleclearmind, line 100: The function name does not reflect the fact that it works with tags only.
Consider accepting a `Function *` range instead or having a more descriptive name.
No biggie though, it can also stay like this.

aleclearmind (changes requested):

include/revng/Support/OpaqueFunctionsPool.h
pfez: >  Consider accepting a Function * range instead or having a more descriptive name.
I don't understand what you mean.

aleclearmind: Something like:
```c++
   void initializeFromReturnType(auto X) requires ... {
     for (llvm::Function &F : X) {
     }
   }
```
Call as:
```c++
initializeFromReturnType(TheTag.functions(M));
```
Maybe instead of `auto` above you can use `range` from `Concepts.h`.

include/revng/ADT/TypeTraits.h
pfez: Done. Also moved `Concepts.h` to `ADT` since it belongs there to be reusable.

include/revng/Support/OpaqueFunctionsPool.h
pfez: So you want the API to ignore the `Tag` and just get a list of `Function *`?

pfez: Honestly, not taking a list of `Function *` was the whole point of this API: relieving the user from the task of manually collecting the function list manually. That would be a totally different API that we can still add in another moment, if someone wants it.

I'd go for renaming this API to incorporate tag:
- `initializeFromTaggedWithReturnTypeKey`
- `initializeFromTaggedWithNthArgTypeKey`

The naming is very verbose though. Do you have better suggestions? Can't we just be happy with the fact that there's documentation and the only argument is a `Tag`? What else should the API use?

aleclearmind: OK, no biggie.

pfez: I'm confused now. Do you want me to change the name or just leave it like it is?

aleclearmind: I marked the issue as solved. No further changes are required.

PR #157, opened 2022-03-30
RegisterLLVMPass.h: add missing include

pfez:
Missing `#include "revng/Pipeline/Loader.h"` causes compilation error when calling `Loader.registerLLVMPass<LLVMPass>(Name);`.

This is a problem since whoever includes `RegisterLLVMPass.h` should include `Loader.h` **before** `RegisterLLVMPass.h` to compile without errors. This is subject to breakage if `Loader.h` ever changes name and `clang-format` starts reordering the include **after** `RegisterLLVMPass.h`.

This commit inlcudes `Loader.h` directly in `RegisterLLVMPass.h` to fix the root cause.

### DIFF
--- a/include/revng/Pipeline/RegisterLLVMPass.h
+++ b/include/revng/Pipeline/RegisterLLVMPass.h
@@ -4,6 +4,7 @@
 // This file is distributed under the MIT License. See LICENSE.md for details.
 //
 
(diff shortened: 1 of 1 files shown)


aleclearmind:
Merged in de033b5. Thanks!

PR #160, opened 2022-04-12
Preliminary work for migrating `revng-c` to `revng-pipeline`

pfez:


lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -49,15 +49,19 @@ TaggedFunctionKind::compactTargets(const pipeline::Context &Ctx,
...
+  const auto IsNotFake = [](const auto &F) {
+    return F.Type != model::FunctionType::Fake;
+  };
+  const auto IsInTargetSet = [&TargetsSet](const model::Function &F) {
+    return TargetsSet.contains(F.Entry.toString());
+  };
+  if (llvm::all_of(llvm::make_filter_range(Model.Functions, IsNotFake),
+                   IsInTargetSet)) {
     return pipeline::TargetsList({ AllFunctions });
   }
```
aleclearmind, line 66: Why do you need this change?
Was it crashing?

lib/Pipes/CMakeLists.txt
```diff
@@ -19,10 +19,29 @@ revng_add_library_internal(
...
+  Linker
   MC
+  Object
+  Passes
+  Remarks
+  ScalarOpts
   Support
+  Target
+  TransformUtils
+  Vectorize
```
aleclearmind, line 45: Shouldn't we only import libraries for which we call `llvm::initialize$LIB`?
I see `Object`, `Linker`, `IRReader` which seem a bit suspicious.

python/revng/cli/lift.py
```diff
@@ -52,6 +52,20 @@ def run(self, options: Options):
...
+                [
+                    "model",
+                    "opt",
+                    "--add-primitive-types",
+                    model.name,
+                    "-o",
+                    model.name,
+                ]
+                + arg_or_empty(args, "base")
+                + arg_or_empty(args, "entry"),
```
aleclearmind, line 65: Drop

aleclearmind (changes requested):

python/revng/cli/lift.py
pfez: :+1:

lib/Pipes/TaggedFunctionKind.cpp
pfez: I don't remember exactly if it was crashing, but it wasn't working, either crashing or reporting an error.

In particular, whenever I had even a single `Fake` function in the model, the `*` target wasn't working properly. The fake functions would be found in the model but they wouldn't have a counterpart in LLVM IR, because they were not tagged as `Isolated` (which IMHO makes sense, given that you cannot really isolate a fake function). So the `AllFunctions` target would be deemed impossible to produce, yielding a failure.

So, discussing with @drblallo we reached the conclusion that the sane thing to do was not consider `Fake` functions. Do you have a better idea?

python/revng/cli/lift.py
pfez: Done!

lib/Pipes/TaggedFunctionKind.cpp
aleclearmind: One day fake functions will disappear. They will just be regular functions with an `inline` attribute.
They will be inlined in callers but we could decompile them individually nevertheless.
For now this makes sense, but it'll change.

lib/Pipes/CMakeLists.txt
pfez: Done

lib/Pipes/TaggedFunctionKind.cpp
pfez: If we're gonna have fake functions in the LLVM IR as well, that's great! I'll be more than happy to revert this! It also means we're gonna be able to show C code for fake functions, which is not currently possible.

However, for now this is necessary for properly handling binaries with fake functions. Downstream tests in `revng-c` don't pass without this at the moment.

PR #161, opened 2022-04-14
Make ABI definition YAML-serializable

Cvelth:


Cvelth:
The relevant changes with go in as part of the #232, which eclipses this PR.

PR #166, opened 2022-04-21
Fix: model diffs with upcastable pointers

drblallo:
Model diffs that contained a upcastable pointer were broken.

When parsing a path, fields on upcasted classes would be ignored, even thought the key of the accessed object did contained all information needed to access the correct class.


aleclearmind:
Merged in 016a946. Thanks!

PR #167, opened 2022-04-25
sanitize coverage pass for revng

pwnslinger:


aleclearmind:
It seems like you're trying to import an LLVM pass into rev.ng. There's no need to do that. LLVM passes are automatically available and you can just schedule them by name in the pipeline. For instance, you can add `sroa` and it will be executed.

aleclearmind:
Closing due to inactivity.

PR #169, opened 2022-04-26
Feature/pipeline analysis

drblallo:
Add support for pipeline analysis and moves efa to be one.


tests/tools/pipeline/StringContainerLibrary.cpp
```diff
@@ -88,6 +88,12 @@ class StringContainer : public Container<StringContainer> {
     return llvm::Error::success();
   }
 
+  llvm::Error extractOne(llvm::raw_ostream &OS,
+                         const pipeline::Target &Target) const override {
+    revng_check(false);
+    return llvm::Error::success();
```
aleclearmind, line 94: Wat?
```
revng_abort();
```

include/revng/Pipeline/Analysis.h
```diff
@@ -0,0 +1,25 @@
...
+//
+
+#include "revng/Pipeline/Pipe.h"
+
+namespace pipeline {
+
+class Analysis {
+private:
+  PipeWrapper Pipe;
+  std::string Name;
```
aleclearmind, line 14: * Mmmh pipes already have a name. We're adding a layer around pipes, this leads to think that the underlying pipe can be use on its own too, while we absolutely do not want pipes to change the model.
  I expected an analysis to be similar to a pipe, more than a wrapper of a pipe.
* I'd expect an analysis to produce a model diff or something.

include/revng/Pipeline/Loader.h
```diff
@@ -182,6 +186,7 @@ struct llvm::yaml::MappingTraits<pipeline::PipeInvocation> {
     TheIO.mapRequired("UsedContainers", Info.UsedContainers);
     TheIO.mapOptional("Passes", Info.Passes);
     TheIO.mapOptional("EnabledWhen", Info.EnabledWhen);
+    TheIO.mapOptional("Name", Info.AliasName);
```
aleclearmind, line 189: * What is `AliasName`?
* Please use exactly the same name in the serialized version and in the `struct`.

include/revng/Pipeline/Loader.h
```diff
@@ -194,6 +199,7 @@ struct llvm::yaml::MappingTraits<pipeline::StepDeclaration> {
     TheIO.mapOptional("Pipes", Info.Pipes);
     TheIO.mapOptional("EnabledWhen", Info.EnabledWhen);
     TheIO.mapOptional("Artifacts", Info.Artifacts);
+    TheIO.mapOptional("Analysis", Info.Analysis);
```
aleclearmind, line 202: Why do we allow a single analysis per step? I think we want to have more than one.

include/revng/Pipeline/Runner.h
```diff
@@ -116,6 +125,10 @@ class Runner {
                   const ContainerToTargetsMap &Targets,
                   llvm::raw_ostream *DiagnosticLog = nullptr);
 
+  llvm::Error runAnalyis(llvm::StringRef AnalysisName,
+                         const ContainerToTargetsMap &Targets,
+                         llvm::raw_ostream *DiagnosticLog = nullptr);
```
aleclearmind, line 130: * `runAnalysis`
* `DiagnosticLog` should disappear, either return the information in the `llvm:Error` or in a `Logger`. If you want to do this later, please make sure you do not add any new uses of this.

include/revng/PipelineC/Prototypes.h
```diff
@@ -184,6 +182,19 @@ const char *rp_manager_produce_targets(rp_manager *manager,
...
+/**
+ * Request to run the required analysis
+ *
+ * \param tagets_count must be equal to the size of targets.
+ *
+ * \return 0 if an error was encountered, the serialized model diff otherwise
+ */
+const char *rp_manager_run_analysis(rp_manager *manager,
+                                    uint64_t targets_count,
+                                    rp_target *targets[],
```
aleclearmind, line 194: So this has to become a list of lists right?
Please put this PR in draft if it's not ready.

include/revng/PipelineC/Prototypes.h
```diff
@@ -316,8 +333,6 @@ void rp_target_destroy(rp_target *target);
 rp_kind *rp_target_get_kind(rp_target *target);
 
 /**
- * Serializes target into a string, target cannot be NULL.
- *
```
aleclearmind, line 320: Why?

lib/Pipeline/Runner.cpp
```diff
@@ -190,6 +190,27 @@ Error Runner::loadFromDisk(llvm::StringRef DirPath) {
...
+                         llvm::raw_ostream *DiagnosticLog) {
+
+  auto MaybeStep = llvm::find_if(Steps, [AnalysisName](const auto &Step) {
+    return Step.second.getAnalysis().getName() == AnalysisName;
+  });
+
+  if (MaybeStep == Steps.end())
+    return createStringError(inconvertibleErrorCode(),
+                             "Could not find a step with a anlisys named %s\n",
+                             AnalysisName.str().c_str());
```
aleclearmind, line 204: * If you have multi-line statements, please use `{ }`
* Is `c_str()` necessary?

python/revng/cli/translate.py
```diff
@@ -47,9 +47,13 @@ def run(self, options: Options):
 
         command = [
             "pipeline",
-            "output::Translated",
-            "--step",
-            step_name,
+        ]
+
+        # if args.isolate:
```
aleclearmind, line 52: What's this?

tests/analysis/AnalysisTests.cmake
```diff
@@ -64,8 +64,19 @@ macro(artifact_handler CATEGORY INPUT_FILE CONFIGURATION OUTPUT TARGET_NAME)
         "${INPUT_FILE}"
         "${OUTPUT}"
         Lift
+        Lift
+        --analysis
```
aleclearmind, line 68: Didn't we say we want two separate invocations?

aleclearmind (changes requested):

include/revng/Pipeline/Analysis.h
drblallo: we need two names. One is the name (it is the type in the yml) used to refer to a particular pipe type among the registered ones, the other is the name assigned to the analysis used to disambiguate which analysis the user wants to run. We can't use the type of the pipe because multiple analysis may have the name LLVMPipe

lib/Pipeline/Runner.cpp
drblallo: yes, when used to do string formatting, a c string is necessary as far as i could tell.

tests/analysis/AnalysisTests.cmake
drblallo: not sure what you mean, it is divided into two steps, there is the && in the middle

we must provide both the step and the analysis name, because otherwise we don't know from which step we want to extract the output. I guess i can make it so that analysis name are aliases for their step, but i don't like much to be honest

include/revng/PipelineC/Prototypes.h
drblallo: erased the entire line by accident, meant to only erase about the NULL thing

tests/analysis/AnalysisTests.cmake
aleclearmind: Ah, I didn't notice, I'm dumb.

tests/tools/pipeline/StringContainerLibrary.cpp
aleclearmind: Not resolved.

include/revng/Pipeline/AnyDiff.h
```diff
@@ -0,0 +1,54 @@
...
+  static void serializeImpl(llvm::raw_ostream &OS, const AnyDiff &Diff) {
+    auto &Casted = Diff.getAs<T>();
+
+    ::serialize(OS, Casted);
+  }
+
+private:
+  std::unique_ptr<InvalidationEventBase> Event;
+  std::any Diff;
+  SerializationMethod Serialize;
```
aleclearmind, line 29: This is weird. `std::any` is type erasing the destructor and then you have `Serialize`.
Wouldn't it be better not to use `std::any` and just have a `virtual` class with one inheritor for each class to type erase (i.e., classical type-erasure)?

include/revng/Pipeline/Context.h
```diff
@@ -20,73 +20,68 @@
...
+
+  template<typename T, typename... ArgsT>
+  void addGlobal(llvm::StringRef Name, ArgsT &&...Args) {
+    Globals.emplace<T>(Name, std::forward<T>(Args)...);
+  }
+
+  template<typename T>
+  void addLesserGlobal(llvm::StringRef Name, T &ToAdd) {
+    LesserGlobals.try_emplace(Name, std::make_any<T *>(&ToAdd));
+  }
```
aleclearmind, line 72: What about:
* `Global` -> `MutableGlobal`
* `LesserGlobal` -> `ImmutableGlobal` or `ConstGlobal`?

*Lesser* doesn't tell me much.

include/revng/Pipeline/SavableObject.h
```diff
@@ -24,24 +31,178 @@ class SavableObjectBase {
...
   virtual llvm::Error loadFromDisk(llvm::StringRef Path);
 };
 
-/// Crtp class to describe a object that can be saved and load from disk
-template<typename MostDerived>
-class SavableObject : public SavableObjectBase {
-public:
-  SavableObject() : SavableObjectBase(&MostDerived::ID) {}
+template<TupleTreeCompatible Object>
+class TupleTreeGlobal : public SavableObjectBase {
```
aleclearmind, line 46: `MutableGlobal`? Let's try and stay coherent with terminology.

include/revng/PipelineC/Prototypes.h
```diff
@@ -458,6 +470,21 @@ rp_container_extract_one(rp_container *container, rp_target *target);
 
 /** \} */
 
+/**
+ *
+ * \defgroup rp_diff_map rp_diff_map methods
+ * \{
+ */
+void rp_diff_map_destroy(rp_diff_map *to_free);
```
aleclearmind, line 478: Doc

include/revng/PipelineC/Prototypes.h
```diff
@@ -152,6 +152,17 @@ bool rp_manager_set_global(rp_manager *manager,
...
+/**
+ * \returns the number of serializable global objects
+ */
+int rp_manager_get_globals_count(rp_manager *manager);
+
+/**
+ * \returns the owning pointer to the name of serializable global object
+ * with the provided index, nullptr if the index was out of bound.
+ */
+const char *rp_manager_create_global_name(rp_manager *manager, int index);
```
aleclearmind, line 164: * Why the `index`?
* What is this function for?

include/revng/PipelineC/Prototypes.h
```diff
@@ -458,6 +470,21 @@ rp_container_extract_one(rp_container *container, rp_target *target);
...
+ * \defgroup rp_diff_map rp_diff_map methods
+ * \{
+ */
+void rp_diff_map_destroy(rp_diff_map *to_free);
+
+/**
+ * \returns nullptr if global_name did not named a global variable in the
+ * diff_map else return the serialized diff of the indicated global
+ */
+const char *rp_diff_map_create_diff(rp_diff_map *map, const char *global_name);
```
aleclearmind, line 484: I'm confused. Is this WIP?

aleclearmind (changes requested):

include/revng/Pipeline/Context.h
drblallo: yeah it was a tmp name i forgot to change

include/revng/PipelineC/Prototypes.h
drblallo: now we can have multiple model like globals, we need to fetch a key for them, in this case the name

include/revng/PipelineC/Prototypes.h
drblallo: why? it follows the convetions that create means a owning pointer. it returns the serialized content of the map

include/revng/PipelineC/Prototypes.h
aleclearmind: Aaaah, so this is `rp_manger_get_global_name`, but you put `_create` due to the fact it returns an owning pointer.
Why does it have to be owning? Can't we state that its lifetime is associated to `rp_manager`'s?
Isn't this the case for other strings too?

include/revng/PipelineC/Prototypes.h
aleclearmind: Ahh, so this is `rp_diff_map_serialize`. The idea of always having `create_` in the name is not working very well.
It makes certain names very counter-intuitive and we're not being consistent (`rp_manager_run_analysis` returns an owning pointer, but it does not contain `create`).

Shall we try to change approach?
What about:
```c++
#define OWNING(x) x

/// \note The returned pointer is owning and needs to be freed by the caller
OWNING(const char *) rp_diff_map_serialize(...);
```

drblallo: yeah i guess it is more reasonable

include/revng/PipelineC/Prototypes.h
drblallo: it was just because the underlying structure is a string map and there was no guarantee of being zero terminated

include/revng/Pipeline/Context.h
drblallo: i kept global for global and i simply renamed add and get lesser into add and get, since you are pretty much simply adding a pointer and retrieving it.

lib/Pipes/StringMapContainer.cpp
```diff
@@ -35,6 +35,22 @@ StringMapContainer::cloneFiltered(const TargetsList &Targets) const {
...
+llvm::Error
+StringMapContainer::extractOne(llvm::raw_ostream &OS,
+                               const pipeline::Target &Target) const {
+  revng_check(Target.getPathComponents().back().isSingle());
+  revng_check(&Target.getKind() == TheKind);
+
+  std::string MetaAddrStr = Target.getPathComponents().back().getName();
+
+  auto It = Map.find(MetaAddress::fromString(MetaAddrStr));
+  revng_check(It != Map.end());
```
aleclearmind, line 47: Why `revng_check`?

tests/unit/Pipeline.cpp
```diff
@@ -146,6 +147,12 @@ class MapContainer : public Container<MapContainer> {
     return true;
   }
 
+  llvm::Error
+  extractOne(llvm::raw_ostream &OS, const Target &Target) const override {
+    revng_check(false);
```
aleclearmind: `revng_abort`. Not the first time, please `git grep revng_check(false)` and make sure to replace with `revng_abort`.

aleclearmind (changes requested):

include/revng/Pipeline/Analysis.h
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  void run(Context &Ctx, ContainerSet &Containers) override {
+    Invokable.run(Ctx, Containers);
+  }
+
+  std::vector<std::string> getRunningContainersNames() const override {
+    return Invokable.getRunningContainersNames();
+  }
+  std::string getName() const override { return Invokable.getName(); }
```
aleclearmind, line 69: What's the difference between `getName` and `getCustomName`?
Is this naming reflected in the pipeline?

include/revng/Pipeline/AnyDiff.h
```diff
@@ -0,0 +1,76 @@
...
+#include "revng/Pipeline/TupleTreeInvalidationEvent.h"
+#include "revng/Support/YAMLTraits.h"
+#include "revng/TupleTree/TupleTreeDiff.h"
+
+namespace pipeline {
+class AnyDiffBase {
+public:
+  virtual void serialize(llvm::raw_ostream &OS) const = 0;
+  virtual std::unique_ptr<InvalidationEventBase>
+  getInvalidationEvent() const = 0;
```
aleclearmind, line 18: Does a diff *needs* to be associated with an existing invalidation event? Or is this *creating* a new invalidation event?
In the latter case, `createInvalidationEvent` might be a more proper name.

include/revng/Pipeline/AnyDiff.h
```diff
@@ -0,0 +1,76 @@
...
+
+#include <any>
+
+#include "revng/Pipeline/InvalidationEvent.h"
+#include "revng/Pipeline/TupleTreeInvalidationEvent.h"
+#include "revng/Support/YAMLTraits.h"
+#include "revng/TupleTree/TupleTreeDiff.h"
+
+namespace pipeline {
+class AnyDiffBase {
```
aleclearmind, line 14: What about `GenericTupleTreeDiff`? `AnyDiff` doesn't tell me much.

include/revng/Pipeline/Context.h
```diff
@@ -15,78 +15,73 @@
...
 /// A class that contains every object that has a lifetime longer than a
 /// pipeline.
 ///
 /// This includes the kinds and the NamedGlobals that will be available in that
 /// context.
 class Context {
 private:
-  llvm::StringMap<SavableObjectBase *> Globals;
+  GlobalsMap Globals;
+  llvm::StringMap<std::any> Anys;
```
aleclearmind, line 31: `Anys` is quite obscure and leaks an implementation detail.

Proposal: we have `Globals` that are mutable and must be `TupleTree`s, then we have `ImmutableGlobals` (was: `Anys` that can be of any type.

What's your take?

Let's doc this.

include/revng/Pipeline/Global.h
```diff
@@ -0,0 +1,211 @@
+#pragma once
```
aleclearmind, line 1: Consider splitting this file in 3 files, one per class.

include/revng/Pipeline/Invokable.h
```diff
@@ -0,0 +1,278 @@
+#pragma once
```
aleclearmind, line 1: I assume this file is mostly a refactoring of existing stuff. Please point me to some point you'd like to get feedback on, if any.

include/revng/Pipeline/Loader.h
```diff
@@ -50,6 +57,7 @@ struct StepDeclaration {
   std::vector<PipeInvocation> Pipes;
   std::vector<std::string> EnabledWhen = {};
   ArtifactsDeclaration Artifacts = {};
+  std::vector<AnalysisDeclaration> Analysis = {};
```
aleclearmind, line 60: The plural is `Analyses`, please check in the whole patchset.

include/revng/Pipeline/TupleTreeInvalidationEvent.h
```diff
@@ -0,0 +1,33 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "revng/Pipeline/InvalidationEvent.h"
+#include "revng/TupleTree/TupleTreeDiff.h"
+
+namespace pipeline {
+
+template<TupleTreeCompatible T>
+class TupleTreeInvalidationEvent
```
aleclearmind, line 13: Given the current design, will we ever have anything different from a `TupleTreeInvalidationEvent`?
Maybe there's a simplification opportunity here?

include/revng/PipelineC/Prototypes.h
```diff
@@ -275,6 +301,47 @@ rp_kind *rp_step_get_artifacts_kind(rp_step *step);
  */
 rp_container *rp_step_get_artifacts_container(rp_step *step);
 
+/**
+ * \return the number of analysis present in this step
+ */
+int rp_step_get_analysis_count(rp_step *step);
```
aleclearmind, line 307: `analyses`, double check everywhere.

include/revng/PipelineC/Prototypes.h
```diff
@@ -275,6 +301,47 @@ rp_kind *rp_step_get_artifacts_kind(rp_step *step);
...
+                                                    int argument_index);
+
+/**
+ * \return the pointer to a acceptable kind for the container with index
+ * argument_index within a analysis, nullptr if kind_index is >= than
+ * rp_analysis_argument_acceptable_kinds_count(analysis, argument_index)
+ */
+const rp_kind *rp_analysis_get_argument_acceptable_kind(rp_analysis *analysis,
+                                                        int argument_index,
+                                                        int kind_index);
```
aleclearmind, line 349: Would dropping `_acceptable` take away any valuable information?

include/revng/PipelineC/Prototypes.h
```diff
@@ -316,12 +383,12 @@ void rp_target_destroy(rp_target *target);
...
 
 /**
- * Serializes target into a string, target cannot be NULL.
+ * Serializes target into a string.
  *
  * \note The return string is owned by the caller. Destroy with
  *       rp_string_destroy.
  */
-char *rp_target_create_serialized_string(rp_target *target);
+char * /*owning*/ rp_target_create_serialized_string(rp_target *target);
```
aleclearmind, line 397: Is this passing `check-conventions`? I'd add a space between the comment delimiters.

include/revng/PipelineC/Prototypes.h
```diff
@@ -408,6 +475,36 @@ bool rp_container_store(rp_container *container, const char *path);
...
+ * frees the provided map
+ */
+void rp_diff_map_destroy(rp_diff_map *to_free);
+
+/**
+ * \returns nullptr if global_name did not named a global variable in the
+ * diff_map else return the serialized diff of the indicated global
+ */
+const char * /*owning*/
+rp_diff_map_get_diff(rp_diff_map *map, const char *global_name);
```
aleclearmind, line 512: Argh, I suspect multi-line declaration might break `grep owning` `:(`
What can we do about it?

include/revng/Pipes/AllFunctions.h
```diff
@@ -11,7 +11,7 @@
 namespace revng::pipes {
 
 pipeline::TargetsList
-compactFunctionTargets(const model::Binary &Model,
+compactFunctionTargets(const TupleTree<model::Binary> &Model,
```
aleclearmind, line 14: It seems to me that you squashed changes that could have been a standalone commit in this one, please try to keep them isolated next time.

lib/Pipeline/Step.cpp
```diff
@@ -106,6 +101,23 @@ Step::cloneAndRun(Context &Ctx, ContainerSet &&Input, llvm::raw_ostream *OS) {
   return Containers.cloneFiltered(InputEnumeration);
 }
 
+void Step::runAnalysis(llvm::StringRef AnalysisName,
+                       Context &Ctx,
+                       const ContainerToTargetsMap &Targets,
+                       llvm::raw_ostream *OS) {
+  ContainerToTargetsMap Map = Containers.enumerate();
+  revng_check(Map.contains(Targets),
+              "An analysis was requested, but not all targets are aviable");
```
aleclearmind, line 110: Let's use `revng_assert`, I'd keep `revng_check` only for stuff we deem to be security/safety-sensitive.
If we don't have a clear policy future refactorings will make us cry.

On the other hand, if you think this might be due to a user mistake, we need to return an error, not abort.

python/revng/api/_capi.py
```diff
@@ -3,13 +3,62 @@
...
 
 from cffi import FFI
 from cffi.backend_ctypes import CTypesBackend
 
+
+class ApiWrapper:
+    function_matcher = re.compile(
+        r"(?P<return_type>[\w_]+)\s*\*\s*\/\*\s*owning\s*\*\/\s*(?P<function_name>[\w_]+)",
+        re.M | re.S,
+    )
```
aleclearmind, line 21: @mrjackv does this work with multiline prototype declarations too?

aleclearmind:
* `Add missing annotations to PipelineC` is a fixup commit w.r.t. introducing `owning`, squash it.
* [Fix revng daemon debug page](https://github.com/revng/revng/pull/169/commits/bb9c879ef4cc9911bb916b080e4c6720f3eeadf0): this should be the first commit.
* [python: handle multithreading](https://github.com/revng/revng/pull/169/commits/3ae48b236917fb954ac7b240f4823a6f8c48ed0a): this should be the second commit.

Usually we put preliminary commits that fix problems that are beneficial also out of the scope of the PR at the beginning of the branch.

aleclearmind (changes requested):

include/revng/Pipeline/AnyDiff.h
drblallo: well, i mean, it returns a unique pointer, it must be a new one every time

include/revng/PipelineC/Prototypes.h
drblallo: @mrjackv which format do you want for the string `/*onwing*/` ?

include/revng/PipelineC/Prototypes.h
drblallo: jack said he handles it

include/revng/Pipes/AllFunctions.h
drblallo: yeah, the patch set got convoluted due to rebasing on pervasive changes and at one point it was easier to just sqush

include/revng/Pipeline/Analysis.h
drblallo: "name" is the "type" in the yaml, "customName" is the "name" in the yaml

include/revng/Pipeline/Context.h
drblallo: the issue is that they are not immutable, they are non savable and non serializable. 

why does it leak implementation? it is privite and there is just a getter and a setter.

include/revng/Pipeline/AnyDiff.h
drblallo: i mean, generic tuple tree diff does not convey much more either. `TypeErasedDiff`?

include/revng/Pipeline/Global.h
drblallo: the first two classes are closely related and do not make sense to split, i put the last one in a stand alone file

include/revng/Pipeline/Invokable.h
drblallo: the only weird thing is that there is `InvokableImpl` that implements `InvokableBase`. `InvokableImpl` is used as a memeber of `PipeImpl` and then `PipeImpl` reimplements `InvokableBase` by dispatching the call to the field.

lib/Pipeline/Step.cpp
drblallo: yeah, the issue here is that we need to do that pervasive change were we allow to return errors in run and analyze

include/revng/PipelineC/Prototypes.h
drblallo: it can always be put in the comment, but i feelt hat `get argument kind` is saying that the argument has a kind, `acceptable` makes it more clear it is a filter

include/revng/Pipeline/AnyDiff.h
aleclearmind: Yeah, you're right, but I think `create` gives a more specific hint to the fact that we're not getting something that already existed but we're creating it from scratch.

include/revng/Pipeline/Analysis.h
aleclearmind: OK, let's make these coherent. No need to deviate.

include/revng/Pipeline/Context.h
aleclearmind: In order to find a good name we need to identify the core of the difference between the two: one can be modified by the user and generates invalidation events, the other is just some global context for the benefit of pipes that the user is not interested in.

Proposals:
* `Globals` -> `MutableGlobals` and `Anys` -> `ImmutableGlobals`
* `Anys` -> `ImmutableGlobals`
* `Anys` -> `PrivateGlobals`
* `Anys` -> `Contexts`
* `Anys` -> `NamedContexts`
* `Globals` -> `Configurations` and `Anys` -> `Contexts`

include/revng/Pipeline/AnyDiff.h
aleclearmind: Mh, `TypeErased` is the technique through which you achieve the goal, more than the actual role of the class.
I can't come up with anything better than `Generic`. `Opaque`?

include/revng/PipelineC/Prototypes.h
mrjackv: > Is this passing check-conventions?
Yes
> which format do you want for the string /*onwing*/ ?
doesn't matter whitespace-wise

include/revng/PipelineC/Prototypes.h
mrjackv: Yes, multiline is not a problem

lib/Pipeline/Step.cpp
aleclearmind: Hopefully `Pipe::run` should not fail, we discussed about checking prerequisites, but that's a different story.
Here, either we return a graceful error or we should run a pipeline to ensure those targets are produced, which is how I thought this would work, or am I missing some reason for which this not viable?

python/revng/api/_capi.py
mrjackv: Yes, `re.M` is for multiline

include/revng/PipelineC/Prototypes.h
aleclearmind: The comment solution is not so exciting after all, one day we could 1) introduce a custom attribute/annotation, 2) write a clang-based tool that generated a metadata file where it prints all the functions that return an owning pointer.

But at this stage, it seems overkill, so let's go with what we have.

include/revng/Pipeline/Analysis.h
drblallo: the issue is that the name is as well the name in the class declaration, like in pipes

aleclearmind: Yes, but pipe names cannot be customized.
I'd go for `getName`+`getTypeName`.

include/revng/Pipeline/AnyDiff.h
drblallo: yeah but this time around we do care about the technique, we had a bunch of templates Diff<T> and we are making a Diff without the T

include/revng/Pipeline/Context.h
drblallo: Uh, Contexts is not a bad pick.

* Anys -> Contexts (addExternalContext, getExternalContext)

with this globals, can stay named globals

include/revng/Pipeline/AnyDiff.h
aleclearmind: I don't know, I'm not a fan of having `TypeErased` in the name, we should put it in many other places. The property of being type-erased is already in the `Base` suffix. We can switch to an `Abstract` prefix if you prefer (and love Java).

We're a bit overdiscussing this: go with what you like but make sure we don't have `Any` and we have `TupleTreeDiff` in the name. This is my key point, the rest I'll leave it to you.

lib/Pipeline/Step.cpp
drblallo: ah right yeah, if this check is triggered it is a pipe implementation issue, can be a assert

include/revng/Pipeline/TupleTreeInvalidationEvent.h
drblallo: not really, TupleTreeInvalidationEvent is a template, we would still need a base class, that is IvalidationEvent

include/revng/Pipeline/AnyDiff.h
drblallo: GlobalDiff? In the end nobody else uses it

include/revng/Pipeline/TupleTreeInvalidationEvent.h
aleclearmind: I see.
Still, under the assumption that we just want `TupleTreeDiff`s as invalidation events, we already have a type-erased `TupleTreeDiff`, so we could discard the type erasure for invalidation event and adopt the type-erased `TupleTreeDiff`.
It'd simplify things a bit.

aleclearmind: I'm saying: all invalidation events are `TupleTreeInvalidationEvent`. So the first upcast is not necessary.

drblallo: but what is the point of that then? it just reduces flexibility, produces the same number of checks in client code except things have longer names. I would rather scrap events entirely and move around diffs, rather than have events that are just a wrapper.

aleclearmind: Well, my friend, that's not a bad idea at all!

PR #170, opened 2022-04-27
Use /self/proc/maps to determine root

mrjackv:


lib/Support/PathList.cpp
```diff
@@ -24,6 +27,32 @@ std::string getCurrentExecutableFullPath() {
   return FullPath.str().str();
 }
 
+std::string getCurrentRoot() {
+  using llvm::sys::path::parent_path;
+
+  std::ifstream Maps("/proc/self/maps");
```
aleclearmind, line 33: Let's avoid `ifstream`, let's use `MemoryBuffer::getFile` and then `StringRef.split("\n")`.

lib/Support/PathList.cpp
```diff
@@ -24,6 +27,32 @@ std::string getCurrentExecutableFullPath() {
   return FullPath.str().str();
 }
 
+std::string getCurrentRoot() {
```
aleclearmind, line 30: This function sometimes returns `""`, but callers do not handle such situation.
Let's abort in those case so we can guarantee that if the function returns, returns sensible data..

aleclearmind (changes requested):

lib/Support/PathList.cpp
mrjackv: Done

lib/Support/PathList.cpp
mrjackv: Fixed

lib/Support/PathList.cpp
```diff
@@ -2,28 +2,56 @@
...
+
+  auto MaybeBuffer = llvm::MemoryBuffer::getFileAsStream("/proc/self/maps");
+  if (!MaybeBuffer)
+    revng_abort("Unable to open /proc/self/maps");
+
+  auto Maps = (*MaybeBuffer)->getBuffer();
+  llvm::SmallVector<llvm::StringRef, 128> MapsLines;
+  Maps.split(MapsLines, "\n");
+
+  for (auto Line = MapsLines.begin(); Line < MapsLines.end(); Line++) {
```
aleclearmind, line 37: Can you?
```suggestion
  for (const auto &Line : MapsLines) {
```

aleclearmind:
Merged in 13bf5a7. Thanks!

Please remember to put the dot the end of lines in commits `:)`

PR #172, opened 2022-04-28
revng-check-conventions: refactor run_revng_checks

mrjackv:
This PR splits `run_revng_checks`:
* `run_revng_cpp_checks` are C/CPP-specific checks, and now uses the variables `C_FILES` and `CXX_FILES` to clearly indicate the nature of the file list
* `run_revng_checks` now only has general checks that apply to all files (this fixes a bug where the absence of C-related bailed the checks early and skipped the generic file checks)

Additionally, change the isort configuration so that the sorting is stable w.r.t. the config file location

### DIFF
--- a/scripts/isort.cfg
+++ b/scripts/isort.cfg
@@ -6,8 +6,9 @@
 profile=black
 quiet=true
 
-known_revng=revng
-known_flask=flask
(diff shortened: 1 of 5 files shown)


aleclearmind:
Merged in b3535d8. Thanks!

PR #173, opened 2022-04-28
Add a disassembly pipeline

Cvelth:


lib/Lift/LoadBinaryPass.cpp
```diff
@@ -40,3 +40,16 @@ bool LoadBinaryWrapperPass::runOnModule(llvm::Module &M) {
...
   return false;
 }
+
+std::optional<std::pair<RawBinaryView, std::unique_ptr<llvm::MemoryBuffer>>>
+loadBinary(const model::Binary &Model, llvm::StringRef BinaryPath) {
+  revng_assert(Model.verify());
+
+  auto FileContentsBuffer = llvm::MemoryBuffer::getFileOrSTDIN(BinaryPath);
+  if (FileContentsBuffer.getError())
+    return std::nullopt;
```
aleclearmind, line 50: No reason to drop the error, return an `llvm::Expected<...>`

include/revng/Support/FunctionTags.h
```diff
@@ -121,6 +121,30 @@ class Tag : public DynamicHierarchy<Tag> {
...
+    using namespace llvm;
+    auto Filter = [this](const GlobalVariable &G) { return isTagOf(&G); };
+    return make_filter_range(M->globals(), Filter);
+  }
+
+  auto exactGlobals(const llvm::Module *M) const {
+    using namespace llvm;
+    auto Filter = [this](const GlobalVariable &G) { return isExactTagOf(&G); };
+    return make_filter_range(M->globals(), Filter);
+  }
```
aleclearmind, line 147: Can you please interleave these with the non-`const` counterparts? I feel like it reduces the probability of them going out of sync.

include/revng/Yield/Assembly/Assembly.h
```diff
@@ -0,0 +1,105 @@
...
+
+#include "llvm/ADT/SmallVector.h"
+
+#include "revng/ADT/KeyedObjectTraits.h"
+#include "revng/ADT/SortedVector.h"
+#include "revng/Support/MetaAddress.h"
+
+namespace assembly {
+
+struct Instruction {
```
aleclearmind, line 18: Did you consider using tuple tree generator? It'd enable us to ship to clients a YAML document with this information (as opposed to HTML) with no effort.
I understand it might seem overkill, but I'd also appreciate your reasoning why you think it's excessive, because we might want to reduce friction to the point this is no longer a problem.

include/revng/Yield/Assembly/DisassemblerInstance.h
```diff
@@ -0,0 +1,38 @@
...
+#include "llvm/MC/MCInstPrinter.h"
+#include "llvm/MC/MCInstrInfo.h"
+#include "llvm/MC/MCObjectFileInfo.h"
+#include "llvm/MC/MCRegisterInfo.h"
+#include "llvm/MC/MCSubtargetInfo.h"
+
+#include "revng/Support/MetaAddress.h"
+#include "revng/Yield/Assembly/Assembly.h"
+
+class DisassemblerInstance {
```
aleclearmind, line 21: Maybe `LLVMDisassembler`?

lib/Yield/Assembly/DisassemblerInstance.cpp
```diff
@@ -0,0 +1,338 @@
...
+               "Impossible to create a disassembler for a non-code section");
+  auto Architecture = llvm::Triple::getArchTypeName(*LLVMArchitecture);
+
+  // Workaround for ARM
+  if (*LLVMArchitecture == llvm::Triple::ArchType::arm)
+    Architecture = "armv7";
+
+  std::string CPUFeatures = "";
+  if (AddressType == MetaAddressType::Code_arm_thumb)
+    CPUFeatures += "+thumb-mode";
```
aleclearmind, line 59: Can you push this logic close to `MetaAddressType` (`MetaAddressType::getLLVMCPUFeatures(MetaAddressType::Values)`)?

lib/Yield/Assembly/DisassemblerInstance.cpp
```diff
@@ -0,0 +1,338 @@
...
+
+  auto &SI = *SubtargetInformation;
+  Disassembler.reset(LLVMTarget->createMCDisassembler(SI, *Context));
+  revng_assert(Disassembler != nullptr, "Disassembler object creation failed.");
+
+  InstructionInformation.reset(LLVMTarget->createMCInstrInfo());
+
+  unsigned AssemblyDialect = 0;
+  if (*LLVMArchitecture == llvm::Triple::ArchType::x86
+      || *LLVMArchitecture == llvm::Triple::ArchType::x86_64)
```
aleclearmind, line 101: If you have multi-line statements, I suggest `{ ... }`

lib/Yield/Assembly/DisassemblerInstance.cpp
```diff
@@ -0,0 +1,338 @@
...
+// static std::optional<size_t>
+// findLastOpenTag(const assembly::Instruction::Fragment &Fragment) {
+//   for (size_t Index = Fragment.Tags.size() - 1;
+//        Index != std::numeric_limits<size_t>::max();
+//        --Index)
+//     if (Fragment.Tags[Index].ToIndex == std::numeric_limits<size_t>::max())
+//       return Index;
+//
+//   return std::nullopt;
+// }
```
aleclearmind, line 156: Why the commented code? We don't keep commented code.

lib/Yield/Assembly/DisassemblerInstance.cpp
```diff
@@ -0,0 +1,338 @@
...
+  if (ImmediateStyle == ImmediateStyles::CHexadecimal)
+    Printer->setPrintHexStyle(llvm::HexStyle::C);
+  else if (ImmediateStyle == ImmediateStyles::AsmHexadecimal)
+    Printer->setPrintHexStyle(llvm::HexStyle::Asm);
+
+  Printer->setSymbolizeOperands(ShouldSymbolizeOperands);
+  Printer->setUseMarkup(true);
+}
+
+static std::pair<std::optional<llvm::MCInst>, uint64_t>
```
aleclearmind, line 129: I'd put the `std::optional` outside.

lib/Yield/Assembly/DisassemblerInstance.cpp
```diff
@@ -0,0 +1,338 @@
...
+getInstruction(const MetaAddress &Address,
```
aleclearmind, line 130: Consider making this (and friends) methods of `DisassemblerInstance`.

lib/Yield/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+revng_add_library_internal(
+  revngYield
+  SHARED
+  Assembly/DisassemblerInstance.cpp)
+
+set(DISASSEMBLER_ARCHITECTURE_SUPPORT "")
+foreach(CURRENT_ARCHITECTURE AArch64 AMDGPU ARM Mips SystemZ X86)
```
aleclearmind, line 13: We don't really support `AMDGPU` but OK.

include/revng/Yield/Assembly/DisassemblyHelper.h
```diff
@@ -0,0 +1,38 @@
...
+}
+namespace efa {
+class FunctionMetadata;
+}
+class DisassemblerInstance;
+class RawBinaryView;
+
+class DissassemblyHelper {
+private:
+  detail::HiddenDisassemberStorage *Internal;
```
aleclearmind, line 26: Why pimpl? Caliban was doing this a lot, but I'm not really a fan of it.
Consider dropping the pimpl pattern.

include/revng/Yield/Assembly/DisassemblyHelper.h
```diff
@@ -0,0 +1,38 @@
...
+public:
+  explicit DissassemblyHelper();
+  ~DissassemblyHelper();
+
+  assembly::Function disassemble(const model::Function &Function,
+                                 const efa::FunctionMetadata &Metadata,
+                                 const RawBinaryView &BinaryView);
+
+private:
+  DisassemblerInstance &access(MetaAddressType::Values AddressType);
```
aleclearmind, line 37: `access` -> `getInstance`?

lib/Yield/Assembly/DisassemblyHelper.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include "revng/Model/Function.h"
+#include "revng/Model/RawBinaryView.h"
+#include "revng/Support/Debug.h"
+#include "revng/Yield/Assembly/DisassemblerInstance.h"
+#include "revng/Yield/Assembly/DisassemblyHelper.h"
+
+namespace detail {
+
+class HiddenDisassemberStorage
+  : public std::map<MetaAddressType::Values, DisassemblerInstance> {};
```
aleclearmind, line 19: You sure you want inheritance as opposed to composition?

include/revng/Pipes/Yield/Assembly.h
```diff
@@ -0,0 +1,37 @@
...
+#include "revng/Pipes/StringMapContainer.h"
+
+namespace revng::pipes {
+
+class YieldAssemblyPipe {
+public:
+  static constexpr const auto Name = "yield-assembly";
+
+public:
+  std::array<pipeline::ContractGroup, 1> getContract() const;
```
aleclearmind, line 24: @drblallo can you review the contract?

aleclearmind (changes requested):

aleclearmind:
This PR reads very very smoothly. I think you did a very good job. Congratulations.

include/revng/Yield/Assembly/Assembly.h
Cvelth: There are some non-trivial changes happening when this stuff gets converted into html (like function/basic_block resolution and the line), so I don't think this is suitable for the client.
 
On the other hand, I want to have access to this stuff when building graphs though. So switching to the tuple tree generator might be a good idea either way.

include/revng/Yield/Assembly/DisassemblerInstance.h
Cvelth: `LLVMDisassemblerInterface` then?

lib/Yield/Assembly/DisassemblerInstance.cpp
Cvelth: Whoops, this is obsolete. I just forgot to remove it.

lib/Yield/Assembly/DisassemblerInstance.cpp
Cvelth: I was worried about disclosing to much interface so I hid it for the time being - but either way works.

lib/Yield/CMakeLists.txt
Cvelth: This snipped is straight up copied from caliban (Maybe I renamed a variable or something, I don't remember), so we have always supported AMDGPU ;) As far as disassembly concerned that is.

include/revng/Yield/Assembly/DisassemblyHelper.h
Cvelth: I hate the idea of having to include all that ~~sketchy~~ internal LLVM stuff.

I'd prefer that as few of my translation units include `MCDisassembler/MCDisassembler.h` and friends. Even if it means pimpl.

lib/Yield/Assembly/DisassemblyHelper.cpp
Cvelth: Yep, the storage IS the map.

The whole point of even having a separate name is to hide the template parameters of the map from the header.

Cvelth:
> This PR reads very very smoothly

Then it wasn't in-the-making for half a year for nothing then :D Glad to hear you like it.

cc: @aleclearmind

lib/Yield/Assembly/DisassemblerInstance.cpp
aleclearmind: There's more commented code below, take that out too.

Cvelth: Yeah, sorry.

include/revng/Pipes/Yield/Assembly.h
drblallo: the contract provided states that the input is both the binary and any quantity of isolated functions, and transforms each of them into disassembled functions. As far as understand that should be correct

include/revng/Yield/Assembly/Assembly.h
Cvelth: If you think the client would benefit from getting yaml over html, I'm all for it - but that yaml shouldn't look much like this data-structure.

lib/Lift/LoadBinaryPass.cpp
Cvelth: Done: https://github.com/revng/revng/pull/173/commits/5b6f6182f7ecdbabf16eea4ebe6d3bd5121507d2 and https://github.com/revng/revng/pull/173/commits/2e0f17ec67ad1a9a5fb1ec3a0a5db2e740f31e73.

lib/Yield/Assembly/DisassemblerInstance.cpp
aleclearmind: I'd say that as long as adding `private` methods doesn't pull in new headers, it's fine.

include/revng/Pipes/Yield/Assembly.h
aleclearmind: Enterprise.

include/revng/Yield/Assembly/DisassemblyHelper.h
Cvelth: I tried everything I could to make either one of those work. But I kept getting `sizeof` problem for some reason - in the end I just gave up and went with the ugly but simple approach.

Cvelth: If I recall correctly, it boiled down to the fact that every single translation unit including the problem header also had to include the original llvm stuff - otherwise it wouldn't compile properly. Which kind of defeats the purpose of forward declaring this stuff.

aleclearmind: On godbolt I was having `sizeof` issues if I left the inline constructor, adding an out-of-line constructor fixed the problem for me.

Can you get to the bottom of this? It's good to know for the future too.

include/revng/Yield/Assembly/Assembly.h
Cvelth: In practice it wouldn't broaden it that much - the target yaml contains pretty much all the same things html contains now but ordered slightly differently. But yeah, some other day.

include/revng/Yield/Assembly/DisassemblyHelper.h
Cvelth: Wait, this actually be relevant. I'll give it a try.

include/revng/Yield/Assembly/DisassemblerInstance.h
Cvelth: Done: https://github.com/revng/revng/pull/173/commits/b6a22bf923ffda4d532db35491cfe67995392441 and https://github.com/revng/revng/pull/173/commits/d4ebf7ce88899ab242332c365e978077773f8e0b.

lib/Pipes/Yield/Assembly.cpp
```diff
@@ -0,0 +1,91 @@
...
+    auto HTML = html::yield(Disassembled, *Metadata, *Model);
+    Output.insert_or_assign(Function.Entry, std::move(HTML));
+  }
+}
+
+void YieldAssemblyPipe::print(const pipeline::Context &,
+                              llvm::raw_ostream &OS,
+                              llvm::ArrayRef<std::string>) const {
+  OS << *revng::ResourceFinder.findFile("bin/revng");
+  OS << " ma-a-a-agic ^_^\n";
```
Cvelth, line 58: @aleclearmind, BTW, this still needs to be fixed. Do you have any suggestions about how the command should look like?

lib/Yield/Assembly/DisassemblerInstance.cpp
Cvelth: Done: https://github.com/revng/revng/pull/173/commits/99d94861b2bae194d3aafeb42592e101eb6d8902.

lib/Yield/Assembly/DisassemblerInstance.cpp
Cvelth: Done: https://github.com/revng/revng/pull/173/commits/99d94861b2bae194d3aafeb42592e101eb6d8902.

lib/Pipes/Yield/Assembly.cpp
Cvelth: Wait, it's as simple as asking some other part of the pipeline for the path?

I was under the impression that I'd need to actually create a separate tool that mirrors the functionality of this pipe and then give the hint as to how it should be invoked here.

include/revng/Yield/Assembly/DisassemblyHelper.h
Cvelth: How about something like this https://github.com/revng/revng/pull/173/commits/c0975737d3057076daadb9c08ed0c007a4dfc479?

include/revng/Yield/Assembly/DisassemblyHelper.h
Cvelth: Seems to work: https://github.com/revng/revng/pull/173/commits/c0975737d3057076daadb9c08ed0c007a4dfc479.

Also I went with pimpl after all. Now that we disclosed internals of the `interface` as private methods - it does rely on those headers after all.

But, tbh, I like it this way a bit more - let the `interface` be a private-ish header that should be only touched when it's needed while the `helper` is the official interface.

lib/Yield/CMakeLists.txt
Cvelth: Done.

include/revng/Yield/Assembly/DisassemblyHelper.h
aleclearmind: OK.
Maybe rename `detail::HiddenDisassemberStorage` `DissassemblyHelperImpl`.

include/revng/Yield/Assembly/DisassemblyHelper.h
aleclearmind: `Specific` seems redundant, but OK.

Cvelth: Mostly a hint at "thumb one is gonna be different" as in "it heavily depends on the type you pass in".

lib/Yield/Assembly/DisassemblerInstance.cpp
Cvelth: Actually, I feel like we should go back on this - this can let us properly handle `MCDisassembler::SoftFail`s: the current instruction might be unrecoverable, but we can still proceed recovering the next ones!

What I mean is, the instruction size part of the output (`uint64_t`) can still have meaning even if the disassembly failed.

Any objections?

include/revng/Yield/Assembly/DisassemblyHelper.h
Cvelth: Done: https://github.com/revng/revng/pull/173/commits/b2f9dbc3c7141222fa5b7c16eab9c996004fd000.

lib/Pipes/Yield/Assembly.cpp
Cvelth: What do you think?

lib/Yield/Assembly/DisassemblerInstance.cpp
aleclearmind: Works!

include/revng/Yield/Assembly/DisassemblyHelper.h
aleclearmind: There's no way to fetch a *non-specific* disassembler, no need to specify this.

Cvelth: Why would anyone ever need a __non-specific__ disassembler though? You always want the one that can deal with the code before you.

I'll rename it though, it's no big deal.

tools/yield/Common.cpp
```diff
@@ -0,0 +1,91 @@
...
+  if (!MaybeModel) {
+    // Explicit model path is not usable, use the model from the ir.
+    return *Module.getModel();
+  } else {
+    // A valid path is provided - use the model from there.
+    return *MaybeModel->getModel();
+  }
+}
+
+static revng::pipes::StringMapContainer makeTheMap() {
```
aleclearmind, line 60:
```suggestion
static revng::pipes::StringMapContainer createMap() {
```

lib/Yield/Assembly/LLVMDisassemblerInterface.cpp
```diff
@@ -0,0 +1,388 @@
...
+size_t getBackwardsConsecutiveCount(llvm::StringRef String,
+                                    LambdaType Lambda,
+                                    size_t StartFrom) {
+  for (size_t Index = StartFrom - 1; Index != size_t(-1); --Index)
+    if (!Lambda(String[Index]))
+      return StartFrom - Index - 1;
+
+  return StartFrom;
+}
+
```
aleclearmind, line 187: Obscure. Docs please.

lib/Yield/Assembly/HTML.cpp
```diff
@@ -0,0 +1,334 @@
...
+static constexpr auto InstructionTargets = "instruction-targets";
+
+static constexpr auto ExternalWhitespace = "external-whitespace";
+static constexpr auto InternalWhitespace = "internal-whitespace";
+static constexpr auto Untagged = "untagged";
+} // namespace tags
+
+namespace templates {
+static constexpr auto Span = "<span class='{0}'>{1}</span>";
+static constexpr auto Div = "<div class='{0}'>{1}</div>\n";
```
aleclearmind, line 47: * I'd prefer `"` as delimiters. `'` seem quite 90s.
* About the following:
  ```html
  <span class='instruction-address'>0x40055c:Code_x86_64</span>
  ```
  I'd like this to become
  ```html
  <div class='instruction' id="block_0x40055c:Code_x86_64">
  ```
  If we make the link become `<a href="#block_0x40055c:Code_x86_64">`, we get intra-function navigation for free!
* Let's not print `<a class='unknown-link' href='revng://unknown'>(unknown)</a>`, no point.

share/revng/pipelines/yield/assembly.yml
```diff
@@ -0,0 +1,13 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+From:            Isolate
+Containers:
+ - Name:            assembly.html-body.yml
```
aleclearmind, line 7:
```suggestion
 - Name:            assembly.html.yml
```

lib/Yield/Assembly/HTML.cpp
aleclearmind: I'd also like to see the list of targets after a `;` (even if probably the comment character varies on an asm-by-asm basis). Can you update the CSS?

Cvelth: Sure.

Cvelth: I looked into that, it shouldn't be a problem to request what the comment character is for a given architecture from llvm, so that's going to be fine as well.

aleclearmind: Enterprise

include/revng/Yield/Assembly/DisassemblyHelper.h
Cvelth: Done: https://github.com/revng/revng/pull/173/commits/68130aaa102516083aa1ca776d7cf60685c3ae8c.

tools/yield/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+    revngEarlyFunctionAnalysis
+    revngLift
+    revngModel
+    revngPipes
+    revngSupport
+    revngYield
+    ${LLVM_LIBRARIES})
+endmacro()
+
+# add_tool(revng-yield Main.cpp)
```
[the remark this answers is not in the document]
Cvelth: Gone: https://github.com/revng/revng/pull/173/commits/6e274c9aa6211c5c6599269e980c1363294ed971.

tools/yield/Main.cpp
```diff
@@ -0,0 +1,56 @@
...
+  llvm::cl::HideUnrelatedOptions({ &options::ThisToolCategory });
+  llvm::cl::ParseCommandLineOptions(Argc, Argv);
+
+  if (llvm::find(AllowedTypes, options::YieldType) == AllowedTypes.end())
+    llvm::outs() << "Unknown yield type: " << options::YieldType << '\n';
+
+  if (llvm::find(AllowedFormats, options::YieldFormat) == AllowedFormats.end())
+    llvm::outs() << "Unknown yield format: " << options::YieldFormat << '\n';
+
+  llvm::outs() << "Add `--help` for more information.\n\n";
```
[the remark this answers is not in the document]
Cvelth: Gone: https://github.com/revng/revng/pull/173/commits/0ade97f33e880b5f469dce39cefd15d85338cd2c.

tools/yield/Common.cpp
Cvelth: Done: https://github.com/revng/revng/pull/173/commits/a767c545a8024adfa827a018f6fe6f3198371f6d.

lib/Yield/Assembly/LLVMDisassemblerInterface.cpp
Cvelth: Something like this?
https://github.com/revng/revng/pull/173/commits/805448a63b1ea6d0dbd45eccf1181b187ab4fb3e.

share/revng/pipelines/yield/assembly.yml
Cvelth: Done: https://github.com/revng/revng/pull/173/commits/f169cc5244b07ecc15e525061a3899834d29af8f and https://github.com/revng/revng/pull/173/commits/9895ae429cd543d4e21eb482db80f4aebc64358f.

lib/Pipes/Yield/Assembly.cpp
Cvelth: Here's the link to what I'm referring to: https://github.com/revng/revng/pull/173/files#diff-8b3d47dd3d3992b5f8d455b306ebe9d3f715dd0c8e6d94b45acdfd4017502640R57-R58.

tools/yield/Common.cpp
```diff
@@ -0,0 +1,91 @@
...
+                                  desc("<input model>"),
+                                  value_desc("model"),
+                                  init("-"),
+                                  cat(ThisToolCategory));
+
+} // namespace options
+
+static const model::Binary &
+findTheModel(const ModelInModule &Module,
+             llvm::Optional<ModelInModule> &MaybeModel) {
```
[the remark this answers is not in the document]
Cvelth: Now that I think about it, maybe there's a better alternative out there?

The point is that I have a `llvm::Expected` that I need to preserve, but working with it is really clunky, so I just went with `llvm::expectedToOptional` and use that instead. What would you normally done in a case like this?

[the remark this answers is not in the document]
aleclearmind: https://github.com/revng/revng/blob/develop/tools/pipeline/Main.cpp#L130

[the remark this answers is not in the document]
Cvelth: Sorry for bad explanation. That's not it. There is no problem if this `llvm::Expected` is an `llvm::Error`- I just need to know about that and be able to propagate that fact (like using an optional).

tools/yield/Common.cpp
```diff
@@ -0,0 +1,91 @@
...
+  ObjectLifetimeController Result;
+
+  // Load the llvm module
+  Result.IR = ExitOnError(ModelInModule::loadModule(options::LLVMModulePath));
+  revng_assert(Result.IR.hasModule());
+
+  // Locate and load the model
+  using namespace llvm;
+  Result.Model = expectedToOptional(ModelInModule::load(options::ModelPath));
+  const model::Binary &Model = findTheModel(Result.IR, Result.Model);
```
aleclearmind, line 79: `ModelInModule` is useful when you don't really care about the LLVM module, in this case it's probably better to 1) just manually parse the IR, 2) `if (hasModule(M)) Model = loadModel(M); else Model = deserialize(options::ModelPath)`.

lib/Yield/Assembly/HTML.cpp
Cvelth: 1 is also done: https://github.com/revng/revng/pull/173/commits/8350ef4214d5b1cae6cad28ffb0166b95d56ab81.

python/revng/cli/pipeline.py
```diff
@@ -14,7 +14,9 @@ def register_arguments(self, parser):
         pass
 
     def run(self, options: Options):
-        pipelines = collect_files(options.search_prefixes, ["share", "revng", "pipelines"], "*.yml")
+        pipelines = collect_files(
+            options.search_prefixes, ["share", "revng", "pipelines"], "**/*.yml"
```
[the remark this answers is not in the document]
Cvelth: Dropped the commit :(

tools/yield/Common.cpp
Cvelth: What about something like this then: https://github.com/revng/revng/pull/173/commits/6ba47650c5d095ef1c875456ca1e62a040e18277?

aleclearmind: I was suggesting not to use `ModelInModule` entirely, since it's designed to serve a different use case.

But this is not too bad.

Cvelth: When I see "just manually parse X", I'd rather go any other way I can possibly go with :D

At the very least this simplifies the support this tool ~~I never even wanted to create~~ will possibly need since it relies on a well-supported part of the project instead of hand-rolling a parser for a changing format.

Cvelth: I have no idea why, but something like `ModelInModule` gives more of a "stable-API-that-is-definitely-not-going-anywhere" feeling to it. I guess I shouldn't make decisions like this based on hunches, but yeah, either way works!

aleclearmind: Either we extend `ModelInModule` to handle the situation without having to call `ModelInModule::load*` twice, or let's do things by hand as I suggested.

Cvelth: Gotcha.

Cvelth: Something like this?

https://github.com/revng/revng/pull/173/commits/5a4b366e115afac428ba2de58dd13194d1e7ad6b

aleclearmind:
Merged in 0d447f0. Thanks!

PR #174, opened 2022-04-28
New graphql server

mrjackv:


python/revng/cli/daemon.py
```diff
@@ -13,37 +14,56 @@ class DaemonCommand(Command):
...
 
-    def register_arguments(self, parser):
-        parser.add_argument("-p", "--port", type=str)
+    def register_arguments(self, parser: ArgumentParser):
+        parser.add_argument("-p", "--port", type=str, default="8000", help="Port to use")
+        parser.add_argument(
+            "--export",
+            action="store_true",
+            help="Don't start the server, only print the variables that would be used to start it",
+        )
```
aleclearmind, line 23: Do we still need this now that we have the production server in-house?

aleclearmind:
Please justify in the commit message the reasons for which we're switching away from flask.

python/revng/cli/daemon.py
```diff
@@ -13,37 +14,56 @@ class DaemonCommand(Command):
...
+        parser.add_argument("--prod", action="store_true", help="Production mode")
```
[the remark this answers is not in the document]
mrjackv: 1. Fixed
2. Reworded the help message to clarify

python/revng/cli/daemon.py
mrjackv: I dropped `--export`
However I added an option `--hypercorn-args` to allow some flexibility while running the server (e.g. enable hot reload), which seems the real reason behind the need for `--export`

mrjackv:
> Please justify in the commit message the reasons for which we're switching away from flask.

Reworded the main commit, should be more explicative.

aleclearmind:
Please comment on why we're switching to hypercorn in the commit. Then, I think we're good to go.

aleclearmind:
Merged in f42fd7f. Thanks!

PR #176, opened 2022-05-10
Improve `StringMapContainer` to make it more `revng-pipeline` friendly

pfez:
This MR is blocking for porting `revng-c` to work with `revng-pipeline`

### DIFF
--- a/include/revng/Pipes/StringMapContainer.h
+++ b/include/revng/Pipes/StringMapContainer.h
@@ -83,6 +83,11 @@ public:
     return Map.insert_or_assign(Key, std::move(Value));
   };
 
+  bool contains(MetaAddress Key) const { return Map.contains(Key); }
+
+  Iterator find(MetaAddress Key) { return Map.find(Key); }
+  ConstIterator find(MetaAddress Key) const { return Map.find(Key); }
+
   Iterator begin() { return Map.begin(); }
   Iterator end() { return Map.end(); }
 
--- a/include/revng/Pipes/StringMapContainer.h
+++ b/include/revng/Pipes/StringMapContainer.h
@@ -14,7 +14,14 @@ namespace revng::pipes {
 
 class StringMapContainer : public pipeline::Container<StringMapContainer> {
 public:
-  using MapType = std::map<MetaAddress, std::string>;
+  /// Wrapper for std::string that allows YAML-serialization as multiline string
+  struct String {
+    std::string TheString;
+    operator std::string() const { return TheString; }
+  };
+
+public:
+  using MapType = std::map<MetaAddress, String>;
   using ValueType = MapType::value_type;
   using Iterator = MapType::iterator;
   using ConstIterator = MapType::const_iterator;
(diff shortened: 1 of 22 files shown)


include/revng/Pipes/StringMapContainer.h
```diff
@@ -14,7 +14,14 @@ namespace revng::pipes {
 
 class StringMapContainer : public pipeline::Container<StringMapContainer> {
 public:
-  using MapType = std::map<MetaAddress, std::string>;
+  /// Wrapper for std::string that allows YAML-serialization as multiline string
+  struct String {
+    std::string Str;
```
aleclearmind: Do not abbreviate

lib/Pipes/StringMapContainer.cpp
```diff
@@ -101,7 +116,7 @@ namespace revng::pipes {
 
 llvm::Error StringMapContainer::serialize(llvm::raw_ostream &OS) const {
   llvm::yaml::Output YAMLOutput(OS);
-  YAMLOutput << const_cast<std::map<MetaAddress, std::string> &>(Map);
+  YAMLOutput << const_cast<std::map<MetaAddress, String> &>(Map);
   return llvm::Error::success();
```
aleclearmind, line 121: Can you switch to?
```
include/revng/Support/YAMLTraits.h-void serialize(S &Stream, const T &Element) {
```
I'd like to concentrate `const_cast` in a single place.

include/revng/Pipes/StringMapContainer.h
```diff
@@ -29,17 +33,22 @@ class StringMapContainer : public pipeline::Container<StringMapContainer> {
...
 
 public:
   static char ID;
 
 public:
   StringMapContainer(llvm::StringRef Name,
                      llvm::StringRef MIMEType,
-                     const pipeline::Kind &K) :
+                     const pipeline::Kind &K,
+                     const model::Binary *Model) :
```
aleclearmind: Can this be `nullptr`? If no, use references.

lib/Pipes/AllFunctions.cpp
```diff
@@ -0,0 +1,48 @@
+//
```
aleclearmind, line 1: Missing proper header for `.cpp` files.

aleclearmind (changes requested):

include/revng/Pipes/StringMapContainer.h
pfez: It can be if you don't use `enumerate`. But for now we don't have any place where we need to pass a `nullptr` to it, nor I have in mind cases in the near future where we might wan to do that. I'll switch to references for now.

include/revng/Pipes/StringMapContainer.h
pfez: Done!

lib/Pipes/StringMapContainer.cpp
pfez: Done!

include/revng/Pipes/StringMapContainer.h
pfez: Done!

lib/Pipes/AllFunctions.cpp
pfez: Done!

Just so you know, in `revng-c` we're not using either of those on top of `.cpp` files. I think we never did, or if we did we got rid of them a while back. You might want to take that into consideration when merging the two repositories in the future.

PR #179, opened 2022-05-17
Emit CFG in SVG

Cvelth:


include/revng/Yield/Internal/Graph.h
```diff
@@ -0,0 +1,67 @@
+#pragma once
```
aleclearmind, line 1: This commit says `Outline yield::Graph` but I only see new code. Why is that?

lib/Yield/Support/SugiyamaStyleGraphLayout.cpp
```diff
@@ -0,0 +1,11 @@
...
+/// \brief
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "revng/Yield/Support/SugiyamaStyleGraphLayout.h"
+
+bool yield::sugiyama::layout(Graph &Graph, const Configuration &Configuration) {
+}
```
aleclearmind, line 2326: Uh?

aleclearmind (changes requested):
Sugiyama is gigantic, directory + multiple files? You seem to already have split things nicely into commits.

lib/Yield/Internal/Verify.cpp
```diff
@@ -0,0 +1,54 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "revng/Yield/Internal/BasicBlock.h"
+#include "revng/Yield/Internal/Function.h"
+#include "revng/Yield/Internal/Instruction.h"
+#include "revng/Yield/Internal/Tag.h"
+
+bool yield::Tag::verify(model::VerifyHelper &VH) const {
+  VH.maybeFail(Type != TagType::Invalid);
```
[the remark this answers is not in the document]
Cvelth: 1. Yeah, I had no idea how that part of the model worked so I just copied what I saw. I'll fix this :D

2. Love the idea!

lib/Pipes/Yield/AssemblyPipes.cpp
```diff
@@ -44,20 +45,19 @@ void YieldAssemblyPipe::run(pipeline::Context &Context,
...
 }
 
-void YieldAssemblyPipe::print(const pipeline::Context &,
-                              llvm::raw_ostream &OS,
-                              llvm::ArrayRef<std::string> Files) const {
+void ProcessAssemblyPipe::print(const pipeline::Context &,
+                                llvm::raw_ostream &OS,
+                                llvm::ArrayRef<std::string> Files) const {
   OS << *revng::ResourceFinder.findFile("bin/revng") << " yield assembly html "
      << Files[0] << " " << Files[1] << '\n';
```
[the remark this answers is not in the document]
Cvelth: The tools weren't changed, so it is as correct as it is in the current develop. But the tools still need a little bit of love. I think these shouldn't be in the `yield` "~~name~~-tool-space" after all - they are way too internal-like. I can't really imagine user wanting to do much with these. I imagine the python script I've been using for simple binary-to-a-directory-with-htmls have should become the foundation on `revng yield` instead.

include/revng/Yield/ControlFlow/Configuration.h
```diff
@@ -0,0 +1,55 @@
...
+
+  Graph::Dimension InstructionFontSize;
+  Graph::Dimension InstructionAddressFontSize;
+  Graph::Dimension InstructionBytesFontSize;
+  Graph::Dimension CommentFontSize;
+  Graph::Dimension LabelFontSize;
+
+  size_t NodeCornerRoundingFactor;
+
+  bool UseOrtogonalBends;
```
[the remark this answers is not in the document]
Cvelth: Whoops :D

include/revng/Yield/Internal/Graph.h
Cvelth: It only OUTLINEs an interface. The actual usage of the new interface is introduced in the next commit.

lib/Yield/Support/SugiyamaStyleGraphLayout.cpp
Cvelth: This commit (similarly to the previous thread) only describes the interface. The actual functionality comes in the next commits.

lib/Yield/Assembly/HTML.cpp
```diff
@@ -527,62 +637,120 @@ static std::string instruction(const assembly::Instruction &Instruction,
...
+    return llvm::formatv(templates::SimpleDiv,
+                         tags::BasicBlock,
+                         std::move(Result));
+  else
+    return llvm::formatv(templates::BlockDiv,
+                         tags::BasicBlock,
+                         basicBlockID(BasicBlock.Address),
+                         std::move(Result));
+}
+#include <iostream>
```
aleclearmind, line 693: Wat

Cvelth: 🤦🏻, an indication of some painful debugging at one point I forgot to remove. There's little as embarrassing as for someone to see your dirty includes 😓

lib/Yield/Internal/Verify.cpp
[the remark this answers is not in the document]
Cvelth: 1. Done.
2. I'm closing this thread, but hopefully this idea doesn't get lost.

include/revng/Yield/ControlFlow/Configuration.h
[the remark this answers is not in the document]
Cvelth: Fixed.

lib/Yield/Assembly/HTML.cpp
Cvelth: As if it was never there.

lib/Pipes/Yield/AssemblyPipes.cpp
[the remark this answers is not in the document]
Cvelth: Dropped the tools: https://github.com/revng/revng/pull/179/commits/742dbf8e03c4201f48bb5bce3798cde51b090e86.

include/revng/Yield/Internal/Graph.h
aleclearmind: Uhm OK, as long as it compiles.

lib/Yield/Support/SugiyamaStyleGraphLayout.cpp
aleclearmind: At least put a `revng_abort`. Doesn't this trigger warnings?

include/revng/Yield/Internal/Graph.h
Cvelth: It does, that's why the empty bodies and stuff are there.

lib/Yield/Support/SugiyamaStyleGraphLayout.cpp
Cvelth: `orc install revng@debug` didn't complain, so I guess not.

`revng_abort` works.

include/revng/Yield/Internal/Graph.h
Cvelth: Closing this one, keeping the second one open for now.

lib/Yield/Support/SugiyamaStyleGraphLayout.cpp
Cvelth: Done.

include/revng/Yield/Internal/Function.h
```diff
@@ -0,0 +1,58 @@
...
+
+/* TUPLE-TREE-YAML
+
+name: Function
+type: struct
+fields:
+  - name: Entry
+    type: MetaAddress
+  - name: Name
+    type: std::string
```
aleclearmind, line 24: If possible, we should avoid copying names in intermediate artifacts, since it needlessly increase the situations under which we need to recompute this.

tl;dr Can you fetch this from the model later down the pipeline?

include/revng/Yield/Internal/Function.h
```diff
@@ -0,0 +1,58 @@
...
+
+  - name: BasicBlocks
```
aleclearmind, line 26: Would it make sense to
```suggestion
  - name: ControlFlowGraph
```
For coherence with EFA?

include/revng/Yield/Internal/BasicBlock.h
```diff
@@ -0,0 +1,79 @@
...
+      upcastable: true
+      elementType: efa::FunctionEdgeBase
+
+  - name: Instructions
+    sequence:
+      type: SortedVector
+      elementType: yield::Instruction
+
+  - name: IsLabelAlwaysRequired
+    type: bool
```
aleclearmind, line 45: Doc

include/revng/Yield/Internal/BasicBlock.h
```diff
@@ -0,0 +1,79 @@
...
+
+  - name: CommentIndicator
+    type: std::string
+  - name: LabelIndicator
+    type: std::string
```
aleclearmind, line 50: * [ ] What are these used for? Can you add doc?
* [ ] Drop `Indicator` suffix

include/revng/Yield/Internal/Instruction.h
```diff
@@ -0,0 +1,79 @@
...
+using ByteContainer = llvm::SmallVector<uint8_t, 16>;
+}
+
+/* TUPLE-TREE-YAML
+
+name: Instruction
+type: struct
+fields:
+  - name: Address
+    type: MetaAddress
```
aleclearmind, line 26: Can we have `Start`/`End`?

include/revng/Yield/Internal/Instruction.h
```diff
@@ -0,0 +1,79 @@
...
+  - name: Bytes
+    type: yield::ByteContainer
+  - name: Raw
+    type: std::string
```
aleclearmind, line 30:
```suggestion
  - name: RawBytes
    type: yield::ByteContainer
  - name: Disassembly
    type: std::string
```

include/revng/Yield/Internal/Instruction.h
```diff
@@ -0,0 +1,79 @@
...
+
+  - name: Tags
+    sequence:
+      type: std::vector
```
aleclearmind, line 34: * Consider using `SortedVector` and make `yield::Tag` with a key spanning all the fields of the data structure? It can work as a set, avoiding issues with duplicate tags.
* I'm afraid `mnemonic` +  tag set will inflate in a more and more sophisticated data structure and I really wonder if we should avoid representing the structure here entirely and just switch to XML at this stage, i.e., drop mnemonic and tags and stuff XML into the `Raw`/`Disassembly` field. We're going from 1) pseudo-XML from LLVM disassembler, 2) to out-of-line markup here, 3) back to XML when we'll ship this to the client and finally 4) in VSCode we'll translate it back to out-of-line markup again. Maybe we can avoid 2) and go from 1) to 3)?

include/revng/Yield/Internal/Instruction.h
```diff
@@ -0,0 +1,79 @@
...
+      elementType: yield::Tag
+    optional: true
+
+  - name: Opcode
```
aleclearmind, line 38: * Uhm, I feel like *opcode* identifies the number identifying the instruction, while here probably the most appropriate term is *mnemonic*, I think.
* Why do we need this?

include/revng/Yield/Internal/Instruction.h
```diff
@@ -0,0 +1,79 @@
...
+    type: std::string
+    optional: true
+  - name: Comment
+    type: std::string
+    optional: true
+  - name: Error
+    type: std::string
+    optional: true
```
aleclearmind, line 46: What is this for? Doc.

include/revng/Yield/Internal/Instruction.h
```diff
@@ -0,0 +1,79 @@
...
+
+  - name: HasDelayedSlot
```
aleclearmind, line 48: What is the semantic of this field? Is this `yield::Instruction` representing both the jump instruction and the instruction in the delay slot?

```suggestion
  - name: HasDelaySlot
```

include/revng/Yield/Internal/Tag.h
```diff
@@ -0,0 +1,53 @@
...
+
+name: Tag
+type: struct
+fields:
+  - name: Type
+    type: yield::TagType::Values
+  - name: FromPosition
+    type: size_t
+  - name: ToPosition
+    type: size_t
```
aleclearmind, line 23: Please use portable types in serializable data structures.
```suggestion
  - name: From
    type: uint64_t
  - name: To
    type: uint64_t
```

include/revng/Yield/Internal/BasicBlock.h
```diff
@@ -0,0 +1,79 @@
+#pragma once
```
aleclearmind, line 1: Can you find a more appropriate name than `Internal`? This seems like a relic from caliban.

include/revng/EarlyFunctionAnalysis/ControlFlowGraph.h
```diff
@@ -0,0 +1,124 @@
...
+template<IsGenericGraph GraphType,
+         IsBasicBlock BasicBlockType,
+         typename ...OtherTs,
+         template<typename...> typename Container>
+  requires std::is_constructible_v<typename GraphType::Node,
+                                   const MetaAddress &>
+std::pair<GraphType, std::map<MetaAddress, typename GraphType::Node *>>
+buildControlFlowGraph(const Container<BasicBlockType, OtherTs...> &BasicBlocks,
+                      const MetaAddress &EntryAddress,
+                      const model::Binary &Binary) {
```
aleclearmind, line 81: See if you can make this more accessible and add some doc.

lib/Yield/Internal/Verify.cpp
```diff
@@ -0,0 +1,72 @@
...
+bool yield::Instruction::verify(model::VerifyHelper &VH) const {
+  if (Address.isInvalid())
+    return VH.fail("An instruction has to have a valid address.");
+  if (Raw.empty())
+    return VH.fail("A raw view of an instruction cannot be empty.");
+  if (Bytes.empty())
+    return VH.fail("An instruction has to be at least one byte big.");
+
+  // TODO: Tags might need special verification as well.
+  // Their order, the fact that they don't overlap in undesirable ways, etc.
```
aleclearmind, line 35: If you switch to a `SortedVector` of elements sorted by `Start, End, Type` you have no problems with ordering by construction!

aleclearmind (changes requested):

include/revng/Pipes/FunctionStringMap.h
```diff
@@ -23,6 +25,7 @@ class FunctionStringMap : public pipeline::Container<FunctionStringMap> {
   struct String {
     std::string TheString;
     operator std::string() const { return TheString; }
+    operator llvm::StringRef() const { return TheString; }
```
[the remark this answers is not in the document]
Cvelth: This causes problems, sadly, I already have an idea on how to solve it, once it's a bit more develop, I'll ping `@fez` and discuss it with him.

include/revng/EarlyFunctionAnalysis/ControlFlowGraph.h
Cvelth: Always a good idea! :)

include/revng/Yield/Internal/Tag.h
Cvelth: Ah, right, sorry.

include/revng/Yield/Internal/Instruction.h
Cvelth: This field is set to true for an instructions that contains a delayed slot. The actual delayed instruction is an instruction that follows the one with this field set.

I went with this approach because that's how MCDissassembler stores these. This allows generate instructions independently from each other - no need to carry over any information between neighbouring ones.

include/revng/Yield/Internal/Instruction.h
Cvelth: Sure, sorry.

include/revng/Yield/Internal/Instruction.h
Cvelth: I feel like this way to store it gives the most flexibility possible. If at some point we decide to change the approach, there'll be a lot less work to do with having tags and the raw text separated.

include/revng/Yield/Internal/Instruction.h
Cvelth: Sure.

include/revng/Yield/Internal/Instruction.h
Cvelth: Sure.

Cvelth: Do we really need `End` though? Consider that instructions are always stored in a vector and are considered packed. As such the `End` of one instructions is just the `Start` of the next one. The `End` of the last instruction in a basic block is the `End` of the corresponding block.

include/revng/Yield/Internal/BasicBlock.h
Cvelth: These are what indicates `Comment` and `Label` correspondingly in the official way to print assembly. For example, for most platforms `LabelIndicator` is `:`.

`CommentIndicator` varies wildly, like `@` for arm, `//` for aarch64 and so on.

2. Do you have better name ideas then? I feel like just calling them `Comment` and `Label` is going to get really confusing.

include/revng/Yield/Internal/Function.h
Cvelth: The only issue with that is that it'd need to depend on the model in the later stages as well (as of now they are self-contained, as in they don't need anything but this intermediate structure).

But, now that you say it, yeah that would make a lot of sense! There's no reason to invalidate the entirety of this intermediate structure just because user renamed a function.

What's the safest way to store a reference to a model function? `TupleReference`? Are there problems if I were to serialize those outside the model? As in, model could change unexpectedly between serializing and deserializing those.

include/revng/Yield/Internal/BasicBlock.h
aleclearmind: To me it makes sense to have the file that defines `yield::BasicBlock` in `include/revng/Yield/BasicBlock.h`.
Those are data structures as valid as EFA's ones.
All our data structures are always WIP, I wouldn't worry too much.

lib/Yield/Internal/Verify.cpp
Cvelth: I actually not sure whether we need that or not. I can't think of a case where there are two tags beginning and ending at the same points.

include/revng/Yield/Internal/Instruction.h
aleclearmind: OK, document this. Unless you saw *delayed* somewhere let's go with *delay*.

Also, in the verify assert that this is the semi-last instruction of the list.

include/revng/Yield/Internal/BasicBlock.h
Cvelth: Yeah, thinking about it, I'm not opposed to dropping the `Internal` "namespace".

include/revng/Yield/Internal/Instruction.h
aleclearmind: So this is the name *LLVM* gives to the opcode? Then consider calling it `LLVMOpcode`.
Also, consider that this data structure could be reused with different disassemblers one day.

include/revng/Yield/Internal/Instruction.h
aleclearmind: Works, but then state this somewhere in the doc and check this in verify.

include/revng/Yield/Internal/Instruction.h
Cvelth: I looked it up, it's my bad. For some reason I mixed up `delayed load` and `delay slot` when naming this. Will fix.

2. Do we care where it is really? Sure, we should probably check that it's not the last, but other than that let the instruction be wherever it is, imo.

include/revng/Yield/Internal/Instruction.h
Cvelth: 1. yeah, it's llvm notation, sorry for not making that clear.

2. That's actually the reason why I kept it as just an `opcode`, every disassembler should theoretically have something similar after all - hence LLVM in the name could be redundant.

include/revng/Yield/Internal/Function.h
aleclearmind: I feel like not depending on the model is a bit of a misguided goal.
The safest thing is to use the `Entry` address to make a lookup on the model and get the Right Name™.

lib/Yield/Internal/Verify.cpp
aleclearmind: ACK, the  consider implementing a custom `operator<`.

include/revng/Yield/Internal/Instruction.h
aleclearmind: 1. :+1:
2. Mmh, I like our data structures to offer strong, useful, predictable guarantees. If in our mind we expect something, why not enforce it?

include/revng/Yield/Internal/BasicBlock.h
Cvelth: Yeah, that was my main worry as well, but keeping track of those manually didn't felt right. This allows as to just LLVM for these, but it also means we store a whole lot of duplicates.

I thought about moving them higher up, but the highest we can possible have them in the current datastructure state is at a function level which isn't much better than the basic block level.

Do you prefer we store these in the headers somewhere then?

include/revng/Yield/Internal/Function.h
Cvelth: Feels like that should work.

lib/Yield/Internal/Verify.cpp
Cvelth: What a weird way to write "the spaceship operator" /s

include/revng/Yield/Internal/Instruction.h
Cvelth: 2. Mostly, in this case, if we try to enforce something this rigid, I see no point of having this flag at an instruction level. If the delay slot can only be located at the very end of a block, it should be a block-level flag!

Now that I think about it, maybe that's actually the way to go. What do you think? Should `EndsInDelaySlot` be a basic-block field after all?

include/revng/Yield/Internal/BasicBlock.h
aleclearmind: In [`Architecture.h` we already have similar stuff](https://github.com/revng/revng/blob/develop/include/revng/Model/Architecture.h#L258).

include/revng/Yield/Internal/Instruction.h
aleclearmind: Yes!

include/revng/Yield/Internal/Instruction.h
aleclearmind: `OpcodeIdentifier`? We can decouple this from LLVM and just provide the guarantee that the identifier uniquely identifies the instruction in this architecture.
If you prefer, not to change the name, make sure the doc explains something as the thing above.

include/revng/Yield/Internal/Instruction.h
aleclearmind: OK, let's stick with this.

I'm closing this discussion in favor of the other one where you mentioned `<=>`.

include/revng/Yield/Internal/Instruction.h
Cvelth: Sure, I don't really have any attachment to the name.

include/revng/Yield/Internal/BasicBlock.h
Cvelth: [Done](https://github.com/Cvelth/revng/blob/snapshot/wip_cfg_in_svg/include/revng/Yield/BasicBlock.h#L44).

include/revng/Yield/Internal/Function.h
Cvelth: [Done](https://github.com/Cvelth/revng/blob/snapshot/wip_cfg_in_svg/include/revng/Yield/Function.h#L24).

include/revng/Yield/Internal/BasicBlock.h
Cvelth: Done: https://github.com/Cvelth/revng/blob/snapshot/wip_cfg_in_svg/include/revng/Model/Architecture.h#L366.

include/revng/Yield/Internal/Instruction.h
Cvelth: [Done](https://github.com/Cvelth/revng/blob/snapshot/wip_cfg_in_svg/lib/Yield/Verify.cpp#L53).

include/revng/Yield/Internal/Instruction.h
Cvelth: [Done](https://github.com/Cvelth/revng/blob/snapshot/wip_cfg_in_svg/include/revng/Yield/Instruction.h#L29).

include/revng/Yield/Internal/Instruction.h
Cvelth: [Done](https://github.com/Cvelth/revng/blob/snapshot/wip_cfg_in_svg/include/revng/Yield/Instruction.h#L40).

include/revng/Yield/Internal/Instruction.h
Cvelth: [Done](https://github.com/Cvelth/revng/blob/snapshot/wip_cfg_in_svg/include/revng/Yield/Instruction.h#L50).

include/revng/Yield/Internal/Tag.h
Cvelth: [Done](https://github.com/Cvelth/revng/blob/snapshot/wip_cfg_in_svg/include/revng/Yield/Tag.h#L20).

include/revng/Yield/Internal/BasicBlock.h
Cvelth: [Done](https://github.com/Cvelth/revng/tree/snapshot/wip_cfg_in_svg/include/revng/Yield).

include/revng/Pipes/FunctionStringMap.h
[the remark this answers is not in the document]
Cvelth: I went like this instead: https://github.com/Cvelth/revng/commit/b75fbf4fcb4ce8250d46443ecf5b41ccfb26ae94

What do you think?

include/revng/Yield/Internal/Instruction.h
Cvelth: What do you think?
https://github.com/revng/revng/pull/179/commits/c514425b189fe2fcbf337ff546726575e8c3c0f6

include/revng/EarlyFunctionAnalysis/ControlFlowGraph.h
Cvelth: Something like this? https://github.com/revng/revng/pull/179/commits/17c6a939a0f08d8e81eb151378dbadbb6ef878d9

lib/Yield/Internal/Verify.cpp
Cvelth: What do you think?
https://github.com/revng/revng/pull/179/commits/a9a153ef846bdce867a44a0baa2ced49f31ce148 and https://github.com/revng/revng/pull/179/commits/d03f75b9360e0862c498787af8908a1d304c2de1

include/revng/Yield/Internal/Instruction.h
aleclearmind: Looks good!

* I'd go for `HasDelaySlot`, there's doc for better specifying the meaning.
* `InstructionReturnValueTuple`: you often do this, can you find a more semantic name instead of simply describing the role? I find that providing more dignity to data structure is always beneficial.

include/revng/EarlyFunctionAnalysis/ControlFlowGraph.h
aleclearmind: Yeah!

include/revng/Yield/Internal/Instruction.h
Cvelth: 1. Sure.
2. I do it for the data structures I wouldn't have added if I could get away with it. For example, in this case it's literally `std::tuple<yield::Instruction, bool, size_t>` except that's too long and somewhat confusing (What does this `bool` mean? What's the int for?), hence the struct with named fields instead. If you have any name suggestions - I'm not opposed, BUT, it's only ever used as the return value of this specific function and is expected to be unrolled using structured bindings at the call site.

Cvelth: 2. It wouldn't provide it much dignity to call it its real name: `InstructionAndTheFieldsThatWereRemovedFromIt` :0

aleclearmind: I'd call it "LLVMDisassemblerInterface::Instruction" and be done with it. Its scoped, no one will use it for other purpose, it makes sense as a name. Occam's razor!

Cvelth: I guess that's one way to do it. It is somewhat confusing to have `Instruction` class that has an `Instruction` member with type `Instruction` but, either way, this is not much worse the the current solution, so I'm fine with that.

PR #182, opened 2022-06-02
Add `revng-model-graph` tool

AlviseDeFaveri:
This MR introduces a `ModelGraphPrinter` class whose purpose is to generate a `.dot` representation of a given `model::Type`.

Features:
- `Type.dumpDotOnFile("filename")` to be used e.g. inside `gdb`
- `revng model graph <model or LLVM>`

Example of the output (`.dot` file, produced from `printf/x86-64/canonicalized.bc`
[model.txt](https://github.com/revng/revng/files/8826284/model.txt)

### DIFF
--- a/include/revng/Model/Type.h
+++ b/include/revng/Model/Type.h
@@ -85,6 +85,7 @@ public:
   bool verify(bool Assert) const debug_function;
   RecursiveCoroutine<bool> verify(VerifyHelper &VH) const;
   void dump() const debug_function;
+  void dumpTypeGraph(const char *Path) const debug_function;
 };
 
 namespace model {
--- a/include/revng/Model/Binary.h
+++ b/include/revng/Model/Binary.h
@@ -141,6 +141,7 @@ public:
   bool verify(VerifyHelper &VH) const;
   void verify(ErrorList &EL) const;
   void dump() const debug_function;
+  void dumpTypeGraph(const char *Path) const debug_function;
   std::string toString() const debug_function;
 };
 
--- a/include/revng/Model/Function.h
+++ b/include/revng/Model/Function.h
@@ -77,6 +77,7 @@ public:
   bool verify(bool Assert) const debug_function;
   bool verify(VerifyHelper &VH) const;
   void dump() const debug_function;
+  void dumpTypeGraph(const char *Path) const debug_function;
 };
 
 #include "revng/Model/Generated/Late/Function.h"
--- a/include/revng/Model/TypeSystemPrinter.h
+++ b/include/revng/Model/TypeSystemPrinter.h
@@ -1,0 +1,63 @@
+#pragma once
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "revng/Model/QualifiedType.h"
+
+/// Use this class to print a graph representing one or more model types
+/// and their subtypes.
+class TypeSystemPrinter {
+private:
+  /// Where to print the `.dot` graph
+  llvm::raw_ostream &Out;
+  /// Visited == all the edges have already been emitted
+  llvm::SmallPtrSet<const model::Type *, 16> Visited;
+  /// Map each Type to the ID of the corresponding emitted `.dot` node
+  std::map<const model::Type *, uint64_t> NodesMap;
+
+  /// Next free ID to assign to a node
+  uint64_t NextID = 0;
+
+public:
+  TypeSystemPrinter(llvm::raw_ostream &Out, bool OrthoEdges = false);
+  ~TypeSystemPrinter();
+
+private:
+  /// \brief Print the part of a node's label that represents a single field.
+  void dumpField(const model::QualifiedType &FieldQT,
+                 const model::Type *Parent,
+                 uint64_t Offset,
+                 uint64_t FieldIdx);
+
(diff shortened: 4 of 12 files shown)


lib/Model/ModelGraphPrinter.cpp
```diff
@@ -0,0 +1,235 @@
...
+    for (auto &Field : T->edges())
+      Fields.push_back({ Field, 0 });
+  }
+
+  return Fields;
+}
+
+ModelGraphPrinter::ModelGraphPrinter(llvm::raw_ostream &Out,
+                                     bool OrthoEdges /*= false*/,
+                                     bool DisplayUnionsVertically /*= false*/) :
```
aleclearmind, line 42:
```suggestion
                                     bool OrthoEdges,
                                     bool DisplayUnionsVertically) :
```

lib/Model/ModelGraphPrinter.cpp
```diff
@@ -0,0 +1,235 @@
...
+    case model::QualifierKind::Pointer:
+      Out << "Pointer (size " << Qual.Size << ")";
+      break;
+
+    default:
+      break;
+    }
+
+    Out << ",\\n";
+  }
```
aleclearmind, line 163: There's a spurious comma at the end.
Something like:
```suggestion
  const char *Prefix = "";
  for (auto &Qual : QT.Qualifiers) {
    Out << Prefix;
    Prefix = ",\\n";
    switch (Qual.Kind) {
    case model::QualifierKind::Array:
      Out << "Array[" << Qual.Size << "]";
      break;

    case model::QualifierKind::Pointer:
      Out << "Pointer (size " << Qual.Size << ")";
      break;

    default:
      break;
    }
  }
```

include/revng/Model/ModelGraphPrinter.h
```diff
@@ -0,0 +1,55 @@
+#pragma once
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "revng/Model/QualifiedType.h"
+
+/// Use this class to print a graph representing one or more model types
+/// and their subtypes.
```
aleclearmind, line 10: I'm impressed by your discipline in documentation!

aleclearmind:
Pretty cool!

The code prints the graph of the types, not of the model in general. Can you:
1. Rename the tool to `revng-model-export-type-graph`. We have `revng model import ...`, I'd also like `revng model export ...`.
2. `s/ModelGraphPrinter/TypeSystemPrinter/`

tools/model/graph/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+revng_add_executable(revng-model-graph Main.cpp)
+
+llvm_map_components_to_libnames(LLVM_LIBRARIES IRReader BitWriter)
```
aleclearmind, line 7: Why `BitWriter`?

aleclearmind (changes requested):

tools/model/graph/CMakeLists.txt
AlviseDeFaveri: Leftover from copypasta, sorry

include/revng/Model/ModelGraphPrinter.h
AlviseDeFaveri: thanks!

AlviseDeFaveri:
Latest version looks like this
Screenshot from 2022-06-20 19-24-53

include/revng/Model/Type.h
```diff
@@ -85,6 +85,7 @@ class model::Type : public model::generated::Type {
   bool verify(bool Assert) const debug_function;
   RecursiveCoroutine<bool> verify(VerifyHelper &VH) const;
   void dump() const debug_function;
+  void dumpDotOnFile(const char *FileName) const debug_function;
```
aleclearmind, line 88: * @pfez was saddened by the fact that the graph doesn't allow to easily get to a function prototype.
What do you think of moving this to `model::Binary` and include a node for each `model::Function` and an edge for the prototype and another for the stack frame.
We could also do this for `model::Segment`s, once antonio's stuff is in.
* Rename the method to `dumpTypeGraph`, call the argument "Path" and use `StringRef` instead of `const char *`.

pfez: Small aside, I believe using `const char *` is a trick for enabling to call the function from `gdb`. If that's the case we could have 2 overloads, one with `const char *` and one with `StringRef`. Does this make sense?

aleclearmind: @pfez right! Let's leave it like it is.

AlviseDeFaveri: 1. yes good idea thanks for the suggestion. I also realized typedefs are missing, so I'll add those.
2. fez is right

AlviseDeFaveri:
- Print Typedefs
- Print `model::Function`s
- Renamed `Type::dumpDotOnFile` into `Type::dumpTypeGraph`
- Added `Function::dumpTypeGraph`
- Added `Binary::dumpTypeGraph` (prints all functions and type)

Looks like this

Screenshot from 2022-07-06 19-53-23

PR #192, opened 2022-06-29
Add `QualifiedType` field to `model::Segment` for decompilation

antoniofrighetto:
`model::Segment` now includes a `Type` of struct kind. This is intended to
emit the segment as a struct later in decompilation.

### DIFF
--- a/lib/Model/Type.cpp
+++ b/lib/Model/Type.cpp
@@ -262,6 +262,11 @@ const std::set<llvm::StringRef> ReservedKeywords = {
   "asm",
 };
 
+const std::set<llvm::StringRef> ReservedPrefixes = { "unnnamed_",
+                                                     "function_",
+                                                     "dynamic_function_",
+                                                     "segment_" };
+
 static llvm::cl::opt<uint64_t> ModelTypeIDSeed("model-type-id-seed",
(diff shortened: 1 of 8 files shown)


lib/Model/Type.cpp
```diff
@@ -523,7 +527,9 @@ PrimitiveType::PrimitiveType(uint64_t ID) :
 }
 
 static bool beginsWithReservedPrefix(llvm::StringRef Name) {
-  return Name.startswith("unnamed_");
+  for (const auto &Prefix : ReservedPrefixes)
+    return Name.startswith(Prefix);
```
aleclearmind, line 531: This is incorrect, you're returning at the first iteration

aleclearmind (changes requested):

pfez:
@antoniofrighetto `model::Binary::verify()` should use `CheckCustomName` to make sure the names of the segments are not clashing with names of other global stuff in the model, such as `Function`s and the like. If you look into `lib/Model/Binary.cpp` somewhere between lines 110 and 140 you'll see that this already happens for `Function`s, `DynamicFunction`s, `Type`s. You just have to do the same for segments.

PR #194, opened 2022-07-04
Introduce revng-artifact

drblallo:


include/revng/Pipes/PipelineManager.h
```diff
@@ -126,7 +130,7 @@ class PipelineManager {
 
   /// A helper function used to produce all possible targets. It is used for
   /// debug purposes to see if any particular target crashes.
-  llvm::Error produceAllPossibleTargets();
+  llvm::Error produceAllPossibleTargets(bool ExpandTargets = true);
```
aleclearmind, line 130: Mmmh, I'm not a fan of `bool` args, in particular if they have a single user, in particular if they have a default value, in particular if the default value is different from the one of the single user.

Do we really need this?
Can we make a method with a different, more eloquent name?

include/revng/Pipes/PipelineManager.h
```diff
@@ -83,6 +83,9 @@ class PipelineManager {
   /// Path
   llvm::Error overrideModel(llvm::StringRef Path);
 
+  /// Entirelly replaces the model with the one provided
```
aleclearmind, line 83: * Why do we still have model-specific methods in `PipelineManager`? I was hoping it could decay to be a super-regular global variable after the recent changes.
* The other method should be called `overrideModelFromFile`
* Please, pretty please, install a vim plugin spell checking comments and possibly also commit messages.
Pretty pretty please.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
```diff
@@ -2087,6 +2087,9 @@ void FunctionEntrypointAnalyzer::runInterproceduralAnalysis() {
 
 template<bool ShouldAnalyzeABI>
 bool EarlyFunctionAnalysis<ShouldAnalyzeABI>::runOnModule(Module &M) {
+  if (not M.getFunction("root") or M.getFunction("root")->isDeclaration())
+    return false;
```
aleclearmind, line 2091: * Are you sure we're not supposed to `assert` here?
* You seem to be using the `not getFunction(x) or M.getFunction(x)->isDeclaration()` quite a bit. Make an helper function in `IRHelpers.h`.

lib/Pipeline/Runner.cpp
```diff
@@ -75,9 +77,9 @@ static void explainPipeline(const ContainerToTargetsMap &Targets,
   ExplanationLogger << Requirements.back().ToExecute->getName() << ":\n";
   prettyPrintStatus(Targets, ExplanationLogger, 2);
 
-  for (size_t I = Requirements.size(); I != 1; I--) {
-    StringRef StepName = Requirements[I - 2].ToExecute->getName();
-    const auto &TargetsNeeded = Requirements[I - 1].Objectives;
+  for (size_t I = Requirements.size(); I != 0; I--) {
+    StringRef StepName = Requirements[I - 1].ToExecute->getName();
+    const auto &TargetsNeeded = Requirements[I - 1].Input;
```
aleclearmind, line 82: I see magic numbers changing here, which is usually scary.
Can you briefly summarize what's going on?

lib/Pipeline/Runner.cpp
```diff
@@ -291,32 +293,33 @@ llvm::Expected<DiffMap> Runner::runAllAnalyses() {
...
   }
 
   ExplanationLogger << "PRODUCED \n";
   indent(ExplanationLogger, 1);
   ExplanationLogger << EndingStepName << ":\n";
-  CurrentContainer.enumerate().dump(ExplanationLogger, 2, false);
+  ToExec.back().ToExecute->containers().enumerate().dump(ExplanationLogger,
+                                                         2,
+                                                         false);
   ExplanationLogger << DoLog;
```
aleclearmind, line 323: Wrap in `if (ExplanationLogger.isEnabled())`. Clearly isolating logging logic makes it easier to read the code.

lib/Pipes/FileContainer.cpp
```diff
@@ -123,6 +130,12 @@ llvm::Error FileContainer::storeToDisk(llvm::StringRef Path) const {
     return llvm::Error::success();
   }
 
+  if (Path == "-") {
+    auto Buffer = unfailableGetFileAsStream(this->Path);
+
+    llvm::outs() << Buffer->getBuffer();
+  }
```
aleclearmind, line 137: Can we adopt `ToolOutputFile`? It handles `-` and does not leave around stuff in case of failure.
We're already using it [elsewhere](https://github.com/revng/revng/blob/develop/include/revng/Model/ToolHelpers.h#L235).

aleclearmind (changes requested):

include/revng/Pipes/PipelineManager.h
drblallo: it was mostly to implement the -m option in tools, i can move it to main if you prefer

lib/Pipeline/Runner.cpp
drblallo: we were operating on outputs of the previous step to deduce inputs of the current one. The induction variable started from end and walked backward, so I -1 is the current guy, I-2 is the guy which outputs feed into the current guy.

This has been changed to operate in the space of inputs, rather than outputs that are transformed into inputs for the next step, so there is no need any longer to inspect the I-2 guy

include/revng/Pipes/PipelineManager.h
aleclearmind: Yes, makes sense.

lib/EarlyFunctionAnalysis/EarlyFunctionAnalysis.cpp
aleclearmind: What scenario could trigger this?
If one asks for the binary at the end of step that includes `collect-cfg` and therefore we run the step but `cloneFilter`ing only for things we're interested in, which means, only the binary and therefore we have no `root`?

drblallo: yes

lib/Pipes/FileContainer.cpp
drblallo: i could but to do so it would require to load into memory the file in the general case

```
  auto Error = errorCodeToError(llvm::sys::fs::copy_file(this->Path, Path));
  auto Perm = cantFail(errorOrToExpected(fs::getPermissions(this->Path)));
  fs::setPermissions(Path, Perm);
```

aleclearmind: How does that work with `-`?

include/revng/Pipes/PipelineManager.h
drblallo: toh multiple tools need to use it to implement -m so that is a better place to keep it.

lib/Pipes/FileContainer.cpp
drblallo: i load it and dump it to stdout

aleclearmind: I mean, `copy_file`, if an argument is `-` it will fail/create a file named `-`.

drblallo: yeah, that is why we have a if before everything else that check if the output path is -. In that case it executes the code in the top snippet

include/revng/Pipes/PipelineManager.h
drblallo: i can do that, but this is only used to implement --produce-all and --produce-all-single, and would introduce a copy of that function just to have a argument less. 

i would prefer less code duplication and keep it this way this time around

aleclearmind: You can factor the implementation through an `...Impl` method.
I'd suggest to have `--produce-all-individually` and `produceAllPossibleTargetsIndividually`.
The fact that `ExpandTargets` means to produce everything individually is not very intuitive and the fact that is defaulted to `true` is quite strange too.

What's more informative for the reader?

```c++
produceAllPossibleTargets();
produceAllPossibleTargetsIndividually();
```

or

```c++
produceAllPossibleTargets(false);
produceAllPossibleTargets();
```

lib/Pipes/Pipes.cpp
```diff
@@ -118,15 +119,55 @@ class DetectABIAnalysis {
...
+  };
+};
+
+class DetectABIAnalysis {
+private:
+  LLVMAnalysisImplementation<CollectFunctionsFromCalleesWrapperPass,
+                             EarlyFunctionAnalysis::EarlyFunctionAnalysis<true>,
+                             CollectFunctionsFromUnusedAddressesWrapperPass,
+                             EarlyFunctionAnalysis::EarlyFunctionAnalysis<true>>
+    Implementation;
```
aleclearmind, line 153: `:O`

include/revng/Pipes/PipelineManager.h
aleclearmind: So you think there's a demand for an handy wrapper, gotcha.
Still he's a bit of special kid. Can we make this method non `extern` and make it a wrapper around an `overrideGlobal` method? Something like:
```c++
llvm::Error overrideModel(TupleTree<model::Binary> Binary) {
  return overrideGlobal<ModelGlobal>(ModelGlobalName, ...);
}
```

include/revng/Pipeline/LLVMContainer.h
```diff
@@ -46,6 +46,9 @@
 
 namespace pipeline {
 
+void makeGlobalObjectsArray(llvm::Module &Module,
+                            llvm::StringRef GlobalArrayName);
```
aleclearmind, line 51: What's this for? Doc

tests/pipeline/CMakeLists.txt
```diff
@@ -172,7 +172,7 @@ function(add_invalidation_test TEST_NAME)
       ${CMAKE_BINARY_DIR}/libexec/revng/revng-pipeline \
       -P ${OPT_PIPELINE} \
       ${OPT_TARGETS} \
-      -i begin:Strings1:${OPT_INPUT} \
+      -i ${OPT_INPUT}:begin/Strings1 \
```
aleclearmind, line 176: I think this change does not belong to this commit

tests/pipeline/StringContainerLibrary.cpp
```diff
@@ -53,7 +53,7 @@ class StringContainer : public Container<StringContainer> {
   }
 
   bool remove(const Target &Target) {
-    if (contains(Target))
+    if (not contains(Target))
```
aleclearmind, line 56: What was going on here?
Maybe this deserves its own commit.

tools/pipeline/artifact/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+revng_add_executable(revng-artifact Main.cpp)
+
+target_link_libraries(revng-artifact revngPipeline revngPipes revngRecompile
+                      revngEarlyFunctionAnalysis)
```
aleclearmind, line 8: Why `revngRecompile` and `revngEarlyFunctionAnalysis`? I guess we could just have `revngPipeline` here.

include/revng/Pipes/PipelineManager.h
```diff
@@ -127,7 +127,12 @@ class PipelineManager {
 
   /// A helper function used to produce all possible targets. It is used for
   /// debug purposes to see if any particular target crashes.
-  llvm::Error produceAllPossibleTargets(bool ExpandTargets = true);
+  llvm::Error produceAllPossibleTargets() {
+    return produceAllPossibleTargets(false);
+  }
+  llvm::Error produceAllPossibleSingleTargets() {
+    return produceAllPossibleTargets(true);
+  }
```
aleclearmind, line 135: This change doesn't seem to belong here. Can you squash it in the commit introducing `bool ExpandTargets`?

lib/Pipeline/Contract.cpp
```diff
@@ -34,6 +34,21 @@ void Contract::deduceResults(ContainerToTargetsMap &StepStatus,
...
+  if (TargetKind) {
+    for (size_t I = TargetKind->depth(); I < In.getKind().depth(); I++) {
+      if (In.getPathComponents()[I].isSingle()) {
+        return false;
+      }
+    }
+  }
+
+  return true;
+}
```
aleclearmind, line 50: This is a bit obscure. Will this disappear if we get rid of `*` in targets?

If not, can you describe from an high level the problem?

python/revng/cli/analyze.py
```diff
@@ -0,0 +1,25 @@
...
+    def register_arguments(self, parser):
+        pass
+
+    def run(self, options: Options):
+        pipelines = collect_files(options.search_prefixes, ["share", "revng", "pipelines"], "*.yml")
+        pipelines_args = interleave(pipelines, "-P")
+        command = build_command_with_loads(
+            "revng-analyze", pipelines_args + options.remaining_args, options
+        )
+        return run(command, options)
```
aleclearmind, line 22: I guess this might be a good time to factor this class  out.

include/revng/Pipes/ToolCLOptions.h
```diff
@@ -0,0 +1,61 @@
...
+#include <string>
+
+#include "llvm/Support/CommandLine.h"
+#include "llvm/Support/PluginLoader.h"
+
+#include "revng/Pipes/PipelineManager.h"
+
+/*
+ * This file is distributed under the MIT License. See LICENSE.md for details.
+ */
```
aleclearmind, line 12: Move right after `#pragma once`

aleclearmind (changes requested):

tests/pipeline/StringContainerLibrary.cpp
drblallo: that file is just a test file, it is like fixing a failing test. i can push it in a commit of it's own but there is no much point to it.

include/revng/Pipes/PipelineManager.h
drblallo: i will push it in the new CLToolOption class

tests/pipeline/StringContainerLibrary.cpp
aleclearmind: Bugfixes usually deserve their own commit, unless it's a "Rework heavily this part of the code" or "Fix many bugs" commit.
But I'll leave that to you.

drblallo: it was a wrong stub implementation of a method that never got tested and now it does

python/revng/cli/analyze.py
drblallo: since this affects the original pipeline tool tool i will add this change as a extra commit instead of modifying the various commits

include/revng/Pipeline/LLVMContainer.h
drblallo: it creates the array with a pointer to each global object so that they can't be erased by the linker.

lib/Pipeline/Contract.cpp
drblallo: unfortunately removing * will not help, if anything this will become even more obscure because it will have to check if all elements are present.

this is just checking that if we are going from function to root then all functions must be present.

drblallo: i opted for the following comment
```
/**
 * returns false iff the output kind has less depth than the input and the input
 * is not *
 */

```

aleclearmind: OK (without C-style comment `:)`).

Can we turn this into `bool Target::isCompatibleWith(Kind *Other)`? Maybe we can have a different name, but this looks like a `Targets`' method.

drblallo: what do you mean c style comment? it is intended to be documentation.

it is a private method not exposed.

we can't have a compatibility thing, because it depends if it forward or backward and a couple of other things too

include/revng/Pipes/PipelineManager.h
drblallo: isolated in a comit of its own

lib/Pipeline/Contract.cpp
aleclearmind: Doxygen documentation in C++ uses `///`.

> we can't have a compatibility thing, because it depends if it forward or backward and a couple of other things too

Mmmh, you sure you can't make it work with `isCompatibleWith`?
`satisfiedDeeperGranularityRequest` is a very obscure name, we dropped the *granularity* term. Maybe `Contract::targetCanBePromotedTo`/`Contract::targetCanBeDemotedTo`?

drblallo: targetCanBePromotedToShallowerRank?

aleclearmind: `Shallower` is a big judgemental but works.

drblallo: yeah but promote demote is unclear it it means to lower rank or higher, this is clear that means lower

tests/pipeline/StringContainerLibrary.cpp
aleclearmind: Up to you, please comment whether you take action or not and mark as solved.

PR #195, opened 2022-07-05
python: Fix deprecated stdlib import

Depau:
Importing abstract classes from `collections` is deprecated since 3.3 and `collections.abc` has to be used instead. In 3.10 they are no longer available.

```
Python 3.8.10 (default, Mar 15 2022, 12:22:08) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from collections import Iterable
<stdin>:1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working
```

```
Python 3.10.5 (main, Jun  6 2022, 18:49:26) [GCC 12.1.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from collections import Iterable
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'Iterable' from 'collections' (/usr/lib/python3.10/collections/__init__.py)
>>> 
```

This MR fixes the issue but I'd also like to raise the question: how did nobody on <3.10 see the deprecation warning? (as in: is there an issue that causes warnings to not be printed?)

### DIFF
--- a/python/revng/support/__init__.py
+++ b/python/revng/support/__init__.py
@@ -2,7 +2,7 @@
 # This file is distributed under the MIT License. See LICENSE.md for details.
 #
 
-from collections import Iterable as CIterable
+from collections.abc import Iterable as CIterable
 from pathlib import Path
 from typing import Iterable, TypeVar, Union
 


mrjackv:
> how did nobody on <3.10 see the deprecation warning?

If you import the module it doesn't emit warnings, e.g.
```
echo 'from collections import Iterable` > foo.py
python -c 'import foo'
```

Depau:
> Also see [here](https://docs.python.org/3/library/exceptions.html#DeprecationWarning)

Interesting.

Maybe we should consider forcing the "dev mode" when running Python tests to prevent this type of issues, i.e. `python -X dev -m pytest`

https://docs.python.org/3/library/devmode.html#devmode

aleclearmind:
If there an environment variable we can use in tests (and tests only), I'm in favor!

Depau:
> If there an environment variable we can use in tests (and tests only), I'm in favor!

Yes, `PYTHONDEVMODE=1`

PR #197, opened 2022-07-05
Introduce cross-references artifact (call graph)

Cvelth:


Cvelth:
The WIP implementation (based on a trait-defined relation tree) was replaced and can be found here:
https://github.com/Cvelth/revng/tree/feature/revng-binary-location-trait-trees

(new implementation integrates with the pipeline registries to provide a pipeline-like easy downstream extensibility)

include/revng/Model/Binary.h
```diff
@@ -108,6 +108,14 @@ class model::Binary : public model::generated::Binary {
                                 "/Types/" + getNameFromYAMLScalar(T->key()));
   }
 
+  model::TypePath getTypePath(const model::Type::Key &Key) {
+    return TypePath::fromString(this, "/Types/" + getNameFromYAMLScalar(Key));
+  }
+
+  model::TypePath getTypePath(const model::Type::Key &Key) const {
+    return TypePath::fromString(this, "/Types/" + getNameFromYAMLScalar(Key));
+  }
```
aleclearmind, line 117: The other two overloads should call these two methods.

lib/Yield/Support/SugiyamaStyleGraphLayout/GraphPreparation.cpp
```diff
@@ -229,8 +229,9 @@ partitionLongEdges(InternalGraph &Graph, NodeClassifier<Strategy> &Classifier) {
   revng_assert(Graph.getEntryNode() == nullptr);
   if (EntryNodes.size() > 1) {
     auto EntryPoint = Graph.addNode(nullptr);
-    for (auto &Node : EntryNodes)
-      EntryPoint->addSuccessor(Node, nullptr);
+    for (auto *Node : Graph.nodes())
+      if (!Node->hasPredecessors() && Node->Index != EntryPoint->Index)
```
aleclearmind, line 233: Can you compare pointers instead of `Index`?

lib/Yield/Support/SugiyamaStyleGraphLayout/NodeRanking.cpp
```diff
@@ -22,7 +22,7 @@ RankContainer rankNodes<BFS>(InternalGraph &Graph) {
   RankContainer Ranks;
 
   for (auto *Node : llvm::breadth_first(Graph.getEntryNode())) {
-    auto CurrentRank = Ranks.try_emplace(Node, Rank(-1)).first->second;
+    auto CurrentRank = Ranks.try_emplace(Node, Rank(4)).first->second;
```
aleclearmind, line 25: * I guess this is a workaround. If so, please add a `// TODO` about it.
* Can you describe an example that triggers this issue?
* Make a constant out of `4`

include/revng/ADT/StringTemplateParameter.h
```diff
@@ -0,0 +1,34 @@
...
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include <array>
+#include <string_view>
+
+template<size_t N>
+struct StringTemplateParameter {
```
aleclearmind, line 11: Can you put a couple of `static_assert`s to illustrate goals and usage?

aleclearmind:
Please reorder commit so that trivial commits are the top (e.g., "Fix typo").
It's typical for us to have a first commit "Whitespace and other minor changes".

aleclearmind (changes requested):

lib/Yield/Support/SugiyamaStyleGraphLayout/NodeRanking.cpp
Cvelth: 1. It is a workaround, but consider that ranks are, in the essence, relative values. And they are only ever used internally within the layouter.

2. The issue arrives when backedges cause backedges that span before the root node (as in, go into the negative rank). This workaround doesn't fix the issue, but offsets it by 3 or 4 layers (depending on what was the initial value of the constant). Consider that in real examples I never saw more than one layer, but theoretically more is possible, so I decided to push the constant further back.

3. Sure.

include/revng/ADT/StringTemplateParameter.h
Cvelth: What do you have in mind?

As far as I see it, the only (non copy/move) construction accepting an array is a good enough guarantee that the object makes sense if it is constructed. Sure, it's possible to make runtime copies (for some reason), but then it's just a boring wrapper over the array with limited interface.

Cvelth: And if you meant like example for other users to see, then it's probably better for them to just take a look where this thing is used, like at those scary string-accepting `std::get`s - definitely better than flat artificial examples.

lib/Yield/Support/SugiyamaStyleGraphLayout/GraphPreparation.cpp
aleclearmind: OK, I don't have enough context, but I trust you.
Consider maybe if it makes sense to implement `operator==` at this point. If not, feel free to mark this as solved.

lib/Yield/Support/SugiyamaStyleGraphLayout/NodeRanking.cpp
aleclearmind: It seems like we can arbitrarily increase this value with no side effects.
If this is the case, can't we have a post-processing phase that offset all the ranks so that the lowest is 0?

include/revng/ADT/StringTemplateParameter.h
aleclearmind: Yeah, but with no doc and no examples looking at it, it's not obvious how to use it.
An artificial example speaks more than a thousand words and saves you from `git grep`.

lib/Yield/Support/SugiyamaStyleGraphLayout/GraphPreparation.cpp
Cvelth: This is one of the core concepts behind how the layouter works. Adding a comparison operator (or a similar alternative) would cause a bit of the rework for everything that touches "internal" (and especially "fake") nodes - and I feel like I'm already somewhat behind the schedule with this PR.

lib/Yield/Support/SugiyamaStyleGraphLayout/NodeRanking.cpp
Cvelth: 1. There is a bit of a side effect - at one point, the `layer` structure ([this one](https://github.com/revng/revng/blob/develop/lib/Yield/Support/SugiyamaStyleGraphLayout/InternalGraph.h#L194)) is created based on the ranks. And since we have 3 (or 4) extra ranks - there are 3 (or 4) extra elements of the "outside" vector created there. And that's it - no other side-effects.  
P. S. If not for this I'd have set the value to like `1 << 16` or something and would call it a day.

2. This is, in fact, what happens when layers are finalized. To remove any "holes" (as in, layers we removed when processing the graph further), the ranks of all the nodes are shifted to ensure better [logical position](https://github.com/revng/revng/blob/develop/lib/Yield/Support/SugiyamaStyleGraphLayout/InternalGraph.h#L201) relations (e.g., if we see two nodes with neighboring ranks, we can be sure they are going to end up very close vertically in the final graph).

lib/Yield/Support/SugiyamaStyleGraphLayout/GraphPreparation.cpp
aleclearmind: Makes sense.

lib/Yield/Support/SugiyamaStyleGraphLayout/NodeRanking.cpp
aleclearmind: If we have an indexing mechanism whose "zero" is not predictable at the start, there might an opportunity to reorganize the code in terms of relative position among ranks, instead of working with absolute numbers.
Basically, drop the indexing system in favor of pointers to object representing ranks, which are part of a linked list or something similar where you can add elements in front or on the back (and maybe in the middle?).

But I don't have enough context, so please just drop a TODO of what you think it might a good long term solution to avoid the hack and let's move forward.

Cvelth: I think I'm confusing you - the `Rank`s and the `Index`es have nothing to do with each other - they are separate systems.

aleclearmind: Yeah, sorry, I meant rank system. Ranks are "indexed", in the sense that they are associated to an absolute value.
Does this make more sense?

Cvelth: `Index`es are unique node identifiers. They never change and assigned when a note is created.

`Rank`s are our way of sorting nodes, e.g. to store whether a specific node needs to be near the front of the graph, or near the back.

Ranks of all the nodes in the same horizontal layer are the same by definition. That's what we're looking for when assigning them. We're looking for sets of nodes that can be laid out in parallel to each other (on the same horizontal layer). But the relations between different ranks are purely relative: the higher the rank - the closer the node is going to be placed to the start of the graph - and that's it.

The problem with a "good long term solution" is that it would require quite a bit of a rework, and I'm not sure it's worth it - if we were to pour more time into the layouter, there are more pressing problems with it.

Cvelth: Indexing of the ranks is, in its essence, an optimization - it lets us use vectors for representing the layers. And since this data-structure does impact the bottle-neck, we don't have much wiggle room about how we want to structure this.

aleclearmind: I'm just suggesting to leave a TODO on how to improve the situation, not to actually do it.
But yeah, let's move on.

Cvelth: Yeah, I mean that if someone at some point sees this TODO and decides to "fix" this, their effort would be more useful in other parts of the layouter.

But sure, I'll leave one after all.

include/revng/Support/ArtifactLocation.h
```diff
@@ -0,0 +1,162 @@
+#pragma once
```
aleclearmind, line 1: Let's move this to `include/revng/Pipes` if we want this to be specific to revng, or `include/revng/Pipeline` if it's more general.
In `librevngSupport` we don't have any concept such as "artifact".

include/revng/Support/ArtifactLocation.h
```diff
@@ -0,0 +1,162 @@
...
+    std::string Result = '/' + serializeToString(std::get<Layer>(*this));
+    if constexpr (Layer < sizeof...(Types) - 1)
+      Result += toStringImpl<Layer + 1>();
+    return Result;
+  }
+
+  // clang-format off
+  template<size_t Layer = 0>
+  ArtifactLocation *fromStringImpl(llvm::ArrayRef<llvm::StringRef> Input) {
+    // clang-format on
```
aleclearmind, line 78: What problem do you have with `clang-format` here? Seems legit code.

include/revng/Support/ArtifactLocation.h
```diff
@@ -0,0 +1,162 @@
...
+    return Result;
+  }
+
+private:
+  template<size_t Layer = 0>
+  std::string toStringImpl() const {
+    std::string Result = '/' + serializeToString(std::get<Layer>(*this));
+    if constexpr (Layer < sizeof...(Types) - 1)
+      Result += toStringImpl<Layer + 1>();
+    return Result;
```
aleclearmind, line 72: Give a shot to `+ ...`.

include/revng/Support/ArtifactLocation.h
```diff
@@ -0,0 +1,162 @@
...
+struct ArtifactLocation : public std::tuple<Types...> {
+  using std::tuple<Types...>::tuple;
+  using NameHelper = Names;
+  friend pipeline::ArtifactLocationKind<Names, Types...>;
+
+  template<size_t Index>
+  using ElementType = std::tuple_element_t<Index, std::tuple<Types...>>;
+
+private:
+  std::string_view Kind;
```
aleclearmind, line 56: * `Kind` is quite an overloaded term in `revng`. What if we say that a location has a *name* and a set of *steps* or *arguments*?
* `std::string_view` is non-owning. Shall we go for `std::string` or `SmallString`?

include/revng/Support/ArtifactLocation.h
```diff
@@ -0,0 +1,162 @@
...
+
+namespace pipeline {
+
+template<typename Names = ArtifactLocationNameHelper<>, typename... Types>
+class ArtifactLocationKind;
+
+} // namespace pipeline
+
+template<typename Names, typename... Types>
+struct ArtifactLocation : public std::tuple<Types...> {
```
aleclearmind, line 47: Docs. What is this?

include/revng/Support/ArtifactLocation.h
```diff
@@ -0,0 +1,162 @@
...
+using L = ArtifactLocation<Names, Types...>;
+
+} // namespace detail
+
+namespace std {
+
+// clang-format off
+
+template<StringTemplateParameter Name, typename Helper, typename... Types>
+  requires(Helper::template find<Name>().has_value())
```
aleclearmind, line 108: You then to use `requires` a lot. I find that's 99% of the times better to define a new concept with a nice, intuitive name.
Can you give it a shot?

It'll also reduce the size of `clang-format off`.

include/revng/Support/ArtifactLocation.h
```diff
@@ -0,0 +1,162 @@
...
+namespace detail {
+
+template<typename Names, typename... Types>
+using L = ArtifactLocation<Names, Types...>;
+
+} // namespace detail
+
+namespace std {
+
+// clang-format off
```
aleclearmind, line 105: The following part of code is very very hard to read.
Please document what's going to happen, so the reader can at least have some expectation on what's going on.

tests/unit/Location.cpp
```diff
@@ -0,0 +1,128 @@
...
+  return Model.recordNewType(std::move(Ptr));
+}
+
+BOOST_AUTO_TEST_CASE(TypeIDAsTheKey) {
+  TupleTree<model::Binary> NewModel;
+  auto FunctionPath = makeFunction(*NewModel);
+
+  auto Field = revng::pipes::location::TypeField.make();
+  std::get<"type-key">(Field) = FunctionPath.get()->key();
+  std::get<"field-index">(Field) = 2; // And arbitrary field index. WIP
```
aleclearmind, line 52: WIP?

aleclearmind (changes requested):

include/revng/Support/ArtifactLocation.h
Cvelth: Whoops, just some leftovers from the previous optimization 😅

include/revng/Support/ArtifactLocation.h
Cvelth: I guess it's just a personal bias. If I can fit the condition into a single line, I'd rather place it right then and there than try to find what does this specific concept even means.

As a bonus, I don't need to come up with an extra name 😉

include/revng/Support/ArtifactLocation.h
Cvelth: 2. The non-owning nature is on-purpose. This value is only expected to ever be instantiated from a compile time literal - hence it points into a global lifetime literal. They are never changed or combiner or anything.

Moreover, the only reason this is not `const` is that it would break copy and move assignment.

tests/unit/Location.cpp
Cvelth: Something akin to "I'm not sure this should be an integer-index but I can't think of anything better right now, should look into this later".

Cvelth:
The WIP implementation (based on a trait-defined relation tree, working name: "attempt-6") was reworked from ground up and can be found here:
https://github.com/Cvelth/revng/tree/feature/revng-binary-location-tuples

include/revng/Model/Binary.h
Cvelth: Done: https://github.com/revng/revng/pull/197/commits/b31bc794623792d25b85584abd2a301e2657a550#diff-612e49a68d142703a9b058750b7936a1e80e98624734600fd4b22692aab9338bR110

include/revng/Support/ArtifactLocation.h
Cvelth: It lives in `Pipeline` now.

include/revng/Support/ArtifactLocation.h
Cvelth: This is gone.

include/revng/Support/ArtifactLocation.h
Cvelth: This is gone.

include/revng/Support/ArtifactLocation.h
Cvelth: This is gone: this value is not `constexpr` and is called `RankName`.

include/revng/Support/ArtifactLocation.h
Cvelth: This is gone.

include/revng/Support/ArtifactLocation.h
Cvelth: This is gone.

include/revng/Support/ArtifactLocation.h
Cvelth: This is gone.

tests/unit/Location.cpp
Cvelth: This is also gone.

lib/Yield/Support/SugiyamaStyleGraphLayout/NodeRanking.cpp
Cvelth: Done: https://github.com/revng/revng/pull/197/commits/1b87566a41eb9a426fb56b3652a6b1ba058aee9b.

include/revng/ADT/StringTemplateParameter.h
Cvelth: Is this about what you had in mind: https://github.com/revng/revng/pull/197/commits/38409697f5862d86f20f2625aac04158092db5a8?

include/revng/Pipes/Ranks.h
```diff
@@ -6,10 +6,10 @@
 
 #include "revng/Pipeline/Rank.h"
 
-namespace revng::pipes {
+namespace revng::pipes::rank {
```
aleclearmind, line 11: `s/rank/ranks/`

include/revng/Pipeline/Rank.h
```diff
@@ -17,30 +18,96 @@ class Rank : public DynamicHierarchy<Rank> {
...
+
+template<StringTemplateParameter Str, typename LocationType, typename ParentRnk>
+class RankDefinition : public Rank {
 public:
   static constexpr std::string_view RankName = Str;
   using Type = LocationType;
   using Parent = ParentRnk;
 
   using Definition = RankDefinition<Str, Type, Parent>;
+  using RankTag = Definition;
```
aleclearmind, line 60: I'm a bit confused by this intermediate layer? Can we prune some of them?

include/revng/Pipeline/Location.h
```diff
@@ -0,0 +1,181 @@
...
+template<typename Type>
+concept Rank = std::is_same_v<typename Type::RankTag, Type>;
+
+template<typename Type>
+concept RootRank = Rank<Type> && Type::Depth == 0;
+
+namespace detail {
+
+template<typename From, typename To>
+struct RechabilityHelper {
```
aleclearmind, line 33: Documentation: reachability of what?

include/revng/Pipeline/Location.h
```diff
@@ -0,0 +1,181 @@
...
+    return Result;
+  }
+
+public:
+  std::string toString() const {
+    std::string Result;
+
+    (Result += Separator) += RankDefinition::RankName;
+    constexprRepeat<Size>([&Result, this]<auto Index> {
+      (Result += Separator) += serializeToString(std::get<Index>(tuple()));
```
aleclearmind, line 127: Split `Result += Separator` on their own line.

include/revng/Pipeline/Location.h
```diff
@@ -0,0 +1,181 @@
...
+
+    return Result;
+  }
+
+  // clang-format off
+  template<typename AnotherRank>
+    requires (concepts::ConvertibleRank<RankDefinition, AnotherRank>
+              || concepts::ConvertibleRank<AnotherRank, RankDefinition>)
+  static constexpr Location<RankDefinition>
+  convert(const Location<AnotherRank> &Another) {
```
aleclearmind, line 109: Why do we need the `Location<AnotherRank> &&Another` version?

include/revng/Pipeline/Rank.h
```diff
@@ -17,30 +18,96 @@ class Rank : public DynamicHierarchy<Rank> {
...
+  return std::tuple<InitialTupleTypes..., TypesToAppend...>();
+}
+
+template<typename Tuple, typename... TypesToAppend>
+struct AppendToTupleHelper {
+  using type = decltype(appendImpl<TypesToAppend...>(std::declval<Tuple>()));
+};
+
+template<typename First, typename Second>
+struct CH {
```
aleclearmind, line 34: Bad name

include/revng/Pipeline/Rank.h
```diff
@@ -17,30 +18,96 @@ class Rank : public DynamicHierarchy<Rank> {
...
   explicit RankDefinition() : Rank(RankName) {}
+
+  constexpr static Location<Definition> makeLocation() {
+    return Location<Definition>{};
+  }
+
+  constexpr static std::optional<Location<Definition>>
+  locationFromString(std::string_view String) {
+    return Location<Definition>::fromString(String);
+  }
```
aleclearmind, line 110: I think ranks would benefit from being agnostic of `Location`.

lib/Yield/CrossRelations.cpp
```diff
@@ -0,0 +1,168 @@
...
+            }
+          }
+        }
+      }
+    }
+  }
+}
+
+template<typename AddNodeCallable, typename AddEdgeCallable>
+static void convertionHelper(const yield::CrossRelations &Input,
```
aleclearmind, line 55: `s/convertion/conversion/`

lib/Yield/CrossRelations.cpp
```diff
@@ -0,0 +1,168 @@
...
+    LookupHelper.at(From)->addSuccessor(LookupHelper.at(To), EL{ Kind });
+  };
+  convertionHelper(*this, AddNode, AddEdge);
+
+  return Result;
+}
+
+namespace concepts = pipeline::concepts;
+
+// WIP
```
aleclearmind, line 101: ?

aleclearmind (changes requested):

include/revng/ADT/GenericGraph.h
```diff
@@ -1224,9 +1224,28 @@ class GenericGraph
...
+  }
+  template<typename ...Ts>
+    requires (std::is_constructible_v<Node, Ts...> && EqualityComparable<Node>)
+  const_nodes_iterator findNode(Ts &&...Values) const {
+    Node Expected(Values...);
+    auto Comparator = [&Expected](auto &N) { return *N == Expected; };
+    auto InternalIt = std::find_if(Nodes.begin(), Nodes.end(), Comparator);
+    return const_nodes_iterator(InternalIt, getConstNode);
+  }
+  // clang-format on
```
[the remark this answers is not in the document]
Cvelth: Ah, wait! I got rid of the only user of this in the end, but forgot to remove the methods themselves. My bad.

include/revng/Pipeline/Rank.h
```diff
@@ -16,4 +17,38 @@ class Rank : public DynamicHierarchy<Rank> {
   Rank(llvm::StringRef Name, Rank &Parent) : DynamicHierarchy(Name, Parent) {}
 };
 
+  template<StringTemplateParameter Str, typename LocationType, typename ParentRnk>
+  class RankDefinition : public Rank {
```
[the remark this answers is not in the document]
Cvelth: 1. What do you mean? I don't think `check-conventions` complained.

2. Yeah, I don't like it either BUT I'm 98%-ish sure that we want to keep the `Rank` as `Rank` since that name is used all over the pipeline. As for the `Definition`, we can choose whatever - these objects should always be constructed using a helper-function anyway.

include/revng/Pipeline/Rank.h
```diff
@@ -16,4 +17,38 @@ class Rank : public DynamicHierarchy<Rank> {
   Rank(llvm::StringRef Name, Rank &Parent) : DynamicHierarchy(Name, Parent) {}
 };
 
+  template<StringTemplateParameter Str, typename LocationType, typename ParentRnk>
```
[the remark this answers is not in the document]
Cvelth: 1-2. There's not enough space there 😅 
3. Yeah, I forgot about this!!! Inheriting is insufficient though. `ParentRank` must be a definition. Will make a check though.
4. I hesitated on this one. IMO either way is fine. If it's not `YamlSerializeable` - you'll just get an error when you try to yaml-serialize it. But it's your call, we can restrict it.
5. Yeah, this is a lot better!!!

include/revng/Pipeline/Rank.h
```diff
@@ -16,4 +17,38 @@ class Rank : public DynamicHierarchy<Rank> {
...
+  class RankDefinition : public Rank {
+public:
+  static constexpr std::string_view RankName = Str;
+  using Type = LocationType;
+  using Parent = ParentRnk;
+
+  using Definition = RankDefinition<Str, Type, Parent>;
+
+public:
+  explicit RankDefinition(Parent &ParentObj) : Rank(RankName, ParentObj) {}
```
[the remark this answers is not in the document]
Cvelth: Pipeline does. If we get rid of this, we'll need to rework the original `Rank` quite a bit.

Also, it's way too problematic to make the original `Rank` a template (believe me, I tried)

include/revng/Pipeline/Rank.h
Cvelth: Yeah, I hate relying on `Tag`s as well. It feels so old-c++-like. But also, this is a simplest way to actually ensure that the object we're getting is indeed an object we're expecting.

What's your call? Should I make external concepts for those instead (the downside I don't like is that if someone is going to modify one of those, they would need to modify both the template and its concept). Or maybe you know a better way to "fingerprint" a template?

include/revng/Pipeline/Rank.h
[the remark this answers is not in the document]
aleclearmind: We usually go for `X` + `XBase` or `XImpl` + `X`.
If it's not supposed to be used explicitly by the user `XImpl` works.

include/revng/Pipeline/Location.h
```diff
@@ -0,0 +1,181 @@
...
+
+namespace pipeline {
+
+namespace concepts {
+
+template<typename Type>
+concept Rank = std::is_same_v<typename Type::RankTag, Type>;
+
+template<typename Type>
+concept RootRank = Rank<Type> && Type::Depth == 0;
```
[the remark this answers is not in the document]
Cvelth: They are in namespace `concepts`. And `concept`s are by definition `bool`.

Or maybe you don't like using namespaces for this?

include/revng/Pipeline/Location.h
Cvelth: Somehow I knew you're gonna say that 😅 Will do.

include/revng/Pipeline/Location.h
Cvelth: What if there are ever locations with expensive to copy keys? Or, maybe, keys that cannot be copied at all?

include/revng/Pipeline/Rank.h
Cvelth: 80-character long lines are brutal 😅 

Will rename the struct and make an alias for it though. Sorry.

include/revng/Pipeline/Rank.h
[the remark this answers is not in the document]
aleclearmind: 1. `StringTemplateParameter` -> `StringTemplate` or `ConstExprString` or `ConstString`. How does rcc reformats it if it gets too long?
2. Ibidem
3. :+1:
4. Let's be strict, I'm not a fan of APIs that you can only use partially depending on a template argument. Makes sense for super-duper-generic libraries but in regular code it's better to stop the user ASAP. The only point of a Location is to get serialized really.
5. :+1:

include/revng/Pipeline/Rank.h
Cvelth: How do you suggest we construct locations then? Free helper-functions or something?

lib/Yield/CrossRelations.cpp
Cvelth: Sorry, read this like `note to self: following is bunch of templates I'm still unsure about`. I used `WIP` so that `check-conventions` complains about it so that I don't forget to clean the following up.

include/revng/Pipeline/Rank.h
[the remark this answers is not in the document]
Cvelth: Anything that doesn't change the `Rank` name is fine by me

include/revng/Pipeline/Rank.h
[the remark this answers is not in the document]
Cvelth: Sounds good.

include/revng/Pipeline/Location.h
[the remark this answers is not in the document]
aleclearmind: I don't know, I don't a definitive opinion on naming of concepts, but the closest thing we had to concepts where `std::is_...`. In `requires` the `Is` prefix makes sense, while in place of `typename` it doesn't.

Can you look around for a best-practice so we adopt one once and for all?

include/revng/Pipeline/Rank.h
aleclearmind: Ah, now I understand. If you just want a marker maybe use a `constexpr bool IsWhatever = true`, assigning it to `Definition` seems to carry some other semantic meaning, which is a bit confusing.

In general, it depends what guarantees you want to offer? Maybe inheritance from `Rank`? Or the existence of certain members?

include/revng/Pipeline/Rank.h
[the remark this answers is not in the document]
Cvelth: 1-2. I don't like either `StringTemplate` because it can be misunderstood as well as `Const.*` because those are not what this is for (`std::string_view` is perfectly capable of being `constexpr` on its own after all). This trickery is only ever needed if it's used to define a template parameter - hence I wouldn't want to make a `ADT`-namespace name less explicit for the sake of its user.

include/revng/Pipeline/Location.h
aleclearmind: In most cases they will be intergers or `MetaAddress` (designed to be cheaply copyable).
I see more liability in duplicating the code (you end  updating only one of the two implementations) than an actual benefit in terms of efficiency.

include/revng/Pipeline/Rank.h
aleclearmind: Uhm, `Location<MyNiceRank>(...)`? Or maybe you can move this method as a `static` method in `Location`? Would that be possible?

include/revng/Pipeline/Rank.h
Cvelth: Members, actually. In essence, it is supposed to guarantee that all the following `typename Parent::.*` are valid. It's easily reproducible with a dedicated concept BUT it does feel like all the relevant names are mentioned twice then.

include/revng/Pipeline/Location.h
Cvelth: Once we move to c++23 these can be merged together using the new `this type deduction` trickery. Until then, I'm fine with either keeping a move-only copy or removing it outright.

include/revng/Pipeline/Rank.h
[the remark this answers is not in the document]
aleclearmind: `string_view` is a different concept than `ConstString`. `string_view` doesn't own the data, so there's no overlap in functionality.

include/revng/Pipeline/Rank.h
Cvelth: I guess that would work, I don't see why not. My main worry is that I don't really want a user having any control of the `Location`'s template argument. IMO it's much better to have
```
auto CoolLocation = revng::pipes::ranks::MyCoolRankDefinedInAHeader.makeLocation(...);
```
than
```
pipeline::Location CoolLocation(revng::pipes::ranks::MyCoolRankDefinedInAHeader, ...);
```
because then there's no ability explicitly pass anything as the `Location`'s template parameter.

Cvelth: And by suggesting a free function, I meant something like
```cpp
auto CoolLocation = pipeline::makeLocation(revng::pipes::ranks::MyCoolRankDefinedInAHeader, ...);
```

Doesn't work with a static member though, because then there's again an opportunity to explicitly specify the template parameter.

include/revng/Pipeline/Rank.h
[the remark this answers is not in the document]
Cvelth: Actually, if the string is `constexpr`, there's no need to worry about ownership - the compiler guarantees that the lifetime of the literals is as long as the lifetime of the program itself - hence if the string never ever changes - just a view is good enough - hence, in that context, it's the same thing.

include/revng/Pipeline/Rank.h
aleclearmind: Eh, but if you see `RankDefinition` as an implementation detail and the `concept` as the real thing, it makes sense.

Up to you, the tag is OK too, just make a `bool` maybe.

include/revng/Pipeline/Location.h
aleclearmind: Drop then. Can you point me to some this type deduction resource?
Does clang 14 supports it? `:P`

include/revng/Pipeline/Rank.h
aleclearmind: I'd prefer a `static` method if it's just an opportunity, but I'm fine with a free function too.
Let's avoid circular dependencies unless really needed.

include/revng/Pipeline/Location.h
Cvelth: As of now, only msvc 19.32 (vs 17.2) and EDG's compiler support that, sadly. But it's to be expected, I guess, the c++23 is not even finalized yet.

Here's the relevant proposal paper: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p0847r7.html

include/revng/Pipeline/Rank.h
Cvelth: 1. Well, it's an opportunity for a constructor (since those can deduce template parameter), for a static member it you'd have to specify those which is kind of painful.

2. There's no circular dependencies per se. Location relies on a `Rank-like` concept. And then the `RankDefinition` relies on the `Location`.

It's only now that I realized that I already have the system we've talked about in the `tag-vs-concept` thread in place - even if limited.

So, considering there's no circular dependencies, do you still prefer free-function-construction  over a static rank member?

include/revng/Pipeline/Rank.h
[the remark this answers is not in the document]
Cvelth: 1. I guess so. It still doesn't sit right with me.

2. I wish we could. There's an extra limitation on the template parameters:

> all base classes and non-static data members are public and non-mutable

for which `std::string_view` doesn't qualify.

include/revng/Pipeline/Rank.h
aleclearmind: 1. True, we need a free function.
2. Even if they are decoupled in the code, `Rank`-like is not really decoupled from `Rank`, it's pretty close to being a circular dependency. Conceptually, a `Location` needs a `Rank`, a `Rank` makes perfect sense in absence of `Location`s.

Cvelth: Ok, yeah, you convinced me. Let's go with a free helper function.

include/revng/Pipeline/Rank.h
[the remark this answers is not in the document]
aleclearmind: Types should have their own dignity as standalone creatures.
Therefore type names should represent what the class is, in abstract. It should not represent an exhaustive list of places where it's used.
Sometimes you give names such as `ReturnTypeOfX` for the return type of function `X`, and that's a no-no for me `:)`.

The fact that there's overlap with some other existing tool (`string_view`, `llvm::StringLiteral`...) should not be expressed in the name, maybe in the doc, but not the name.

include/revng/Pipes/Ranks.h
Cvelth: Done.

include/revng/ADT/StringTemplateParameter.h
```diff
@@ -0,0 +1,49 @@
...
+#include <string_view>
+
+/// Allows passing strings as template parameters
+///
+/// Objects of this type should only ever be constructed from a string literal.
+/// Internally, it stores the string (minus the last '\0' character as an array)
+template<size_t N>
+struct StringTemplateParameter {
+  /// \note: it has to be public in order to be used as a template parameter :{
+  std::array<char, N - 1> Internal;
```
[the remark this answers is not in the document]
Cvelth: Done.

include/revng/ADT/GenericGraph.h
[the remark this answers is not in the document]
Cvelth: Dropped the commit.

include/revng/Pipeline/Rank.h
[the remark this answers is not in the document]
Cvelth: 1. Oh, wait, this seems to be the exact same issue we observed when seeing the namespaces being badly formatted after concepts.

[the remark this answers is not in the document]
Cvelth: 1 is done.
2 is still pending since I'm unsure about how we should name things 😓

include/revng/Pipeline/Rank.h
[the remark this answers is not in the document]
Cvelth: Ended up going with this:
```cpp
template<ConstexprString String,
         detail::YamlizableOrVoid KeyType,
         RankSpecialization ParentRank>
class RankDefinition : public Rank {
// . . .
};
```

What do you think?

include/revng/Pipeline/Rank.h
```diff
@@ -16,4 +17,38 @@ class Rank : public DynamicHierarchy<Rank> {
...
+  using Type = void;
+  using Parent = void;
+
+public:
+  explicit RankDefinition() : Rank(RankName) {}
+};
+template<StringTemplateParameter Str>
+using RootRankDefinition = RankDefinition<Str, void, void>;
+
+// Rank template parameter deduction helper.
```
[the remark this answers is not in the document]
Cvelth: What do you think?
https://github.com/revng/revng/pull/197/commits/314170f11e01f18323f7391f3115d7a7d7ef17e7

include/revng/Pipeline/Rank.h
Cvelth: The tag is bool now. Also there's a concept looking for said tag.

include/revng/Pipeline/Location.h
Cvelth: Done: https://github.com/revng/revng/pull/197/files#diff-5e12934676b4472008086bc5d11b23bee4f9c170fd7a3594ba9d94ad73509958R119

include/revng/Pipeline/Location.h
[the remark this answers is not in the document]
Cvelth: Closing this thread for now since we had an extensive discussion on the mattermost. If there are any active points to still be discussed here, it'd be better to have separate threads for those.

include/revng/Pipeline/Location.h
Cvelth: Done: https://github.com/revng/revng/pull/197/commits/c7773a697b4c9eb46e98dbb9b28f1d3f7afaf40d

include/revng/Pipeline/Location.h
Cvelth: Gone.

include/revng/Pipeline/Rank.h
Cvelth: Gone.

include/revng/Pipeline/Rank.h
Cvelth: They are a set of free functions now.

lib/Yield/CrossRelations.cpp
Cvelth: I cleaned up this stuff somewhat and moved it to the `Location.h` header. If you have some suggestions about this, please open a new thread there.

lib/Yield/CrossRelations.cpp
Cvelth: https://github.com/revng/revng/pull/197/commits/337ff34e5d97fb0394f7b7c615f9178865b5add9

and also 

https://github.com/revng/revng/pull/197/commits/9f16bed463452f897f87e7d4ad268fefefec8b18

include/revng/Pipeline/Rank.h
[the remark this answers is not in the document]
Cvelth: Actually, no, the keys shouldn't be `void` anywhere but the root. And I did decide to fully separate the root rank from normal ranks for safety sake. As such, we can get away with using `Yamlizable` here: https://github.com/revng/revng/pull/197/commits/a3084ca759dae88f3a295c777f297298a8e04029

include/revng/Pipeline/Rank.h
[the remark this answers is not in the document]
aleclearmind: So what are the candidates as of now? Let's take a decision.

include/revng/ADT/STLExtras.h
```diff
@@ -267,3 +267,47 @@ auto append(FromType &&From, ToType &To) {
...
+constexpr bool constexprAnd(CallableType &&Callable) {
+  return detail::constexprAndImpl(std::make_index_sequence<IterationCount>(),
+                                  std::forward<CallableType>(Callable));
+}
+
+template<std::size_t IterationCount, typename CallableType>
+constexpr bool constexprOr(CallableType &&Callable) {
+  return detail::constexprOrImpl(std::make_index_sequence<IterationCount>(),
+                                 std::forward<CallableType>(Callable));
+}
```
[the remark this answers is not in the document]
Cvelth: Is this what you had in mind?
https://github.com/revng/revng/pull/197/commits/54ebf44c34abe30d2127bdfc4c94311c651aa345

include/revng/Pipeline/Rank.h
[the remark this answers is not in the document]
Cvelth: Technically, there's no limit to how we might want to name them.

Seriously though, there are three different names we need:
1. The runtime rank facing the pipeline internals (I prefer this to stay as `Rank` after all, since this is the most "stable" interface)
2. The rank definition - inherits the `Rank` and extends it with a number of template parameters letting it be aware of its relations with other `Rank`s.
3. The `RankSpecialization` (name pending) concept - used in interfaces to make sure that a valid rank definition is passed to, for example `Location::at` function and the like. This is the most outwards facing interface a user is most likely to encounter.

What's your stance? Do you still prefer `RankImpl` for 2 or something?

include/revng/ADT/Concepts.h
```diff
@@ -4,44 +4,178 @@
...
+/// A more strict version of the `SpecializationOf` concept. This one only
+/// allows direct specializations (and aliases) - no inheritance.
+///
+/// It's useful in the cases when template parameter deduction is important,
+/// e.g. when instanciating traits.
+template<typename Test, template<typename...> class Ref>
+concept StrictSpecializationOf = revng::detail::StrictSpecialization<Test, Ref>;
+
+template<typename Test, template<typename...> class Ref>
+concept NotStrictSpecializationOf = !StrictSpecializationOf<Test, Ref>;
```
aleclearmind, line 162: This kind of concepts are misleading. Does the `Not` applies to `Strict` or to the whole concept?
Consider:
* Using `requires` locally with `!`
* `Not` -> `Negate`
* Pushing the concept close to its usage so that is immediately evident what's going on

include/revng/ADT/Concepts.h
```diff
@@ -4,44 +4,178 @@
...
+// };
+//
+// template<typename Type, template<typename...> class Ref>
+// concept NotSpecializationOf = !SpecializationOf<Type, Ref>;
+//
+// static_assert(SpecializationOf<std::pair<int, long>, std::pair>);
+// static_assert(SpecializationOf<const std::pair<int, long>, std::pair>);
+// static_assert(SpecializationOf<std::string, std::basic_string>);
+// static_assert(SpecializationOf<std::ostream, std::basic_ios>);
+// static_assert(not SpecializationOf<std::string, std::basic_string_view>);
```
aleclearmind, line 129: `SpecializationOf` or `StrictSpecializationOf`?

include/revng/ADT/Concepts.h
```diff
@@ -4,44 +4,178 @@
...
+
+} // namespace revng::detail
+
+/// A more strict version of the `SpecializationOf` concept. This one only
+/// allows direct specializations (and aliases) - no inheritance.
+///
+/// It's useful in the cases when template parameter deduction is important,
+/// e.g. when instanciating traits.
+template<typename Test, template<typename...> class Ref>
+concept StrictSpecializationOf = revng::detail::StrictSpecialization<Test, Ref>;
```
aleclearmind, line 159: The `Strict` prefix was to distinguish it from places where we use were willing to accept derived classes to, but here it seems you don't even have a *non-*Strict version of the concept, so I think we can strip the prefix and be happy.

include/revng/ADT/Concepts.h
```diff
@@ -4,44 +4,178 @@
...
+// TODO: remove these after updating the libc++ version.
+//
 
+// clang-format off
 template<class Derived, class Base>
-concept DerivesFrom = std::is_base_of_v<Base, Derived>;
+concept derived_from = std::is_base_of_v<Base, Derived>
+                       && std::is_convertible_v<const volatile Derived *,
+                                                const volatile Base *>;
+// clang-format on
```
aleclearmind, line 24: I'm starting to think that this whole file should be wrapped in `// clang-format off`

share/revng/pipelines/yield-call-graph.yml
```diff
@@ -0,0 +1,15 @@
...
+Containers:
+  - Name: cross-relations.yml
+    Type: BinaryCrossRelations
+    Role: CrossRelations
+Branches:
+  - From: Isolate
+    Steps:
+      - Name: ProcessCallGraph
+        Pipes:
+          - Type: process-call-graph
```
aleclearmind, line 14: Pipe names are `CamelCase`
```suggestion
          - Type: ProcessCallGraph
```

include/revng/ADT/KeyedObjectContainer.h
```diff
@@ -9,46 +9,61 @@
...
-#include "revng/ADT/KeyedObjectTraits.h"
 #include "revng/ADT/STLExtras.h"
 #include "revng/ADT/UpcastablePointer.h"
 #include "revng/Support/Assert.h"
 
 template<typename T>
-using KOTKey = decltype(KeyedObjectTraits<T>::key(std::declval<T>()));
+struct KeyedObjectTraits {
+  // static * key(const T &);
+  // static T fromKey(* Key);
```
aleclearmind, line 19: Drop, we have a concept for this.

aleclearmind (changes requested):

include/revng/ADT/Concepts.h
Cvelth: Actually, I prefer `requires` with `!` hands down. The only reason I even went with this temporarily is because all of those `clang-format off` are starting to get out of hand and I'm afraid they might be causing some problems of their own.

include/revng/ADT/Concepts.h
Cvelth: `SpecializationOf` requires `clang-13 :(` ([this paper](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0315r4.pdf)), so for now I used `StrictSpecializationOf` everywhere (it seems to do the trick) and left the comments in a couple places to make it less strict after the update.

include/revng/ADT/Concepts.h
Cvelth: Actually, I'd prefer to keep it and reintroduce a "non-strict" version as soon as we are able (updating to clang13+ will do the trick)

include/revng/ADT/Concepts.h
Cvelth: Yeah, I thought so as well, but there are also a couple of non-concept things in here - and there can never be too many consistency checks.

include/revng/Pipeline/Location.h
```diff
@@ -0,0 +1,181 @@
...
+template<typename First, typename Second>
+concept ConvertibleRank = Rank<First>
+                          && Rank<Second>
+                          && detail::RechabilityHelper<Second, First>::value;
+// clang-format on
+
+} // namespace concepts
+
+template<concepts::Rank RankDefinition>
+class Location : public RankDefinition::Tuple {
```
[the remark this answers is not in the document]
aleclearmind: Mmmh, that sounds like a justification on its existence, not docs `:D`

What about
```
A location represents an entity which has a name and it's identified by a series of parameters.
A location is always associated to a `Rank`.
It is basically a name plus a tuple whose components are deduced by the key associated to its `Rank` and its ancestors.
```

include/revng/ADT/Concepts.h
aleclearmind: ACK, still.

Cvelth: Do you want to use this chance and pull the trigger on getting rid of all the `NotXXX` concepts then? I considered doing it, but didn't go through with it.

share/revng/pipelines/yield-call-graph.yml
Cvelth: Done: https://github.com/revng/revng/pull/197/commits/ea09d706e8e2291973e69882ba0704d94d33796d.

include/revng/ADT/KeyedObjectContainer.h
Cvelth: We still need to have at least a definition of this struct to be able to specialize it (which we do all over the place).

So I changed this declaration into a definition: https://github.com/revng/revng/pull/197/commits/8f6ce06ba692c3d7faa84361892396181d799560

include/revng/Pipeline/Location.h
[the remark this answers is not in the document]
Cvelth: Something like this?

https://github.com/revng/revng/pull/197/commits/55aae933fe6afdab126c0a4cf42fde579ce75cc0

include/revng/ADT/Concepts.h
Cvelth: For reference, this is what it would look like: https://github.com/revng/revng/pull/197/commits/e2d9a9bcf7eadcc49dfa5f5b018ce7077a1ed82d

Should I squash this commit? Or drop it?

include/revng/Pipeline/Rank.h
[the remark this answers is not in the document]
aleclearmind: * `TypedRank` expresses the fact that it's not much more than a regular `Rank` but adds type safety, I like it.
* `RankDefinition` OK but it describes more what it is used for than what actually the thing *is*.
* `RankImpl` sad but OK

Pick one.

include/revng/Pipeline/Location.h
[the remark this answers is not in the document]
aleclearmind: Perfect

include/revng/ADT/Concepts.h
aleclearmind: LGTM

include/revng/Pipeline/Rank.h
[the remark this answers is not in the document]
Cvelth: `TypedRank` it is then.

include/revng/ADT/Concepts.h
Cvelth: Squashed.

include/revng/Pipeline/Rank.h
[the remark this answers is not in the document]
Cvelth: Done: https://github.com/revng/revng/pull/197/commits/bf0bab006f5a1fe707fc90ddc42f929b41dc61ee.

lib/Yield/Assembly/PTML.cpp
```diff
@@ -92,22 +94,27 @@ static std::string targetPath(const MetaAddress &Target,
   if (auto Iterator = Binary.Functions.find(Target);
       Iterator != Binary.Functions.end()) {
     // The target is a function
-    return "/function/" + str(Iterator->key());
+    return pipeline::serializedLocation(revng::pipes::ranks::Function,
```
aleclearmind, line 97: Put a `using namespace` to avoid `pipeline::` or make an alias just for this function.
We'll use this a lot, let's try to set a standard where things do not get super-verbose.

Cvelth: Sure.

Cvelth: Does this work? https://github.com/revng/revng/pull/197/commits/5451c5e3396623ebbeb488e9cfeb66d5291be88e.

PR #199, opened 2022-07-18
model: Import idb

dwarfyresearch:


python/revng/cli/_commands/idb_converter.py
```diff
@@ -0,0 +1,655 @@
...
+            return m.PrimitiveTypeKind.Signed
+        elif idb_type.is_unsigned():
+            return m.PrimitiveTypeKind.Unsigned
+        else:
+            return m.PrimitiveTypeKind.Number
+    elif idb_type.is_decl_ptr():
+        # TODO: can we ever reach this point?
+        return m.PrimitiveTypeKind.PointerOrNumber
+    elif idb_type.is_decl_bitfield():
+        return m.PrimitiveTypeKind.Generic
```
aleclearmind, line 652: IIRC we don't support bitfields, we should ignore/mark the types containing them as invalid.

python/revng/cli/_commands/idb_converter.py
```diff
@@ -0,0 +1,655 @@
...
+                if not isinstance(argument_unqualified_type, m.PrimitiveType):
+                    continue
+
+                if argument_unqualified_type.PrimitiveKind != m.PrimitiveTypeKind.Void:
+                    continue
+
+                # TODO: Pick appropriate size.
+                argument.Type = self._get_primitive_type(m.PrimitiveTypeKind.Generic, 8)
+
+    def _fixup_structs(self):
```
aleclearmind, line 248: In this function there's *a lot* of logic for bitfields. We could ignore them, but then, since IIRC we compute the size of a IDB's `struct` as the sum of the size of all of its members, we'd lose the entirety of the `struct`.
Please review this logic and test it.
I don't feel super confident about it.

python/revng/cli/_commands/idb_converter.py
```diff
@@ -0,0 +1,655 @@
...
+                return True
+
+            mod_name = self.api.ida_nalt.get_import_module_name(mod_index)
+            # NOTE: IDA 7 has a bug when reporting imported library names - prints `.dynsym` for
+            # each import.
+            if mod_name and mod_name != ".dynsym":
+                self.imported_libraries.append(mod_name)
+            self.api.ida_nalt.enum_import_names(mod_index, cb)
+
+    def _fixup_void_cabifunctiontype_arguments(self):
```
aleclearmind, line 221: Why? When does this happen?
Very mysterious.

python/revng/cli/_commands/idb_converter.py
```diff
@@ -0,0 +1,655 @@
...
+        """Imports initial types from the IDB. The types will be incomplete and need to be fixed"""
+        til = self.idb.til
+        for type_definition in til.types.defs:
+            assert isinstance(type_definition, idb.typeinf.TILTypeInfo)
+            the_type = type_definition.type
+            assert isinstance(the_type, idb.typeinf.TInfo)
+            self._convert_idb_type_to_revng_type(the_type, ordinal=type_definition.ordinal)
+
+    def _import_functions(self):
+        metaaddr_type = revng_arch_to_metaaddr_code_type[self.arch]
```
aleclearmind, line 107: We should distinguish regular ARM from Thumb here.

python/revng/cli/_commands/idb_converter.py
```diff
@@ -0,0 +1,655 @@
...
+
+        for function_start_addr in self.api.idautils.Functions():
+            function = idb.analysis.Function(self.idb, function_start_addr)
+            function_name = function.get_name()
+
+            # NOTE: python-idb raises an exception when deserializing type
+            # info from some DBs (it tries to read a byte from an empty buffer).
+            try:
+                idb_function_type = function.get_signature()
+            except Exception as exception:
```
aleclearmind, line 117: * Did we look into why this happens?
* Is this the most-specific exception you can get? Isn't there an `IDBException` or something?

python/revng/cli/_commands/idb_converter.py
```diff
@@ -0,0 +1,655 @@
...
+                abi = revng_arch_to_abiname[self.arch]
+
+                qualified_return_type = self._get_primitive_type(m.PrimitiveTypeKind.Void, 0)
+                revng_function_type = m.CABIFunctionType(
+                    OriginalName=function_name,
+                    ABI=abi,
+                    ReturnType=qualified_return_type,
+                    Arguments=[],
+                )
+                self.revng_types_by_id[revng_function_type.ID] = revng_function_type
```
aleclearmind, line 137: What's going on here? When does this happen?
Let's at least emit a warning.

aleclearmind (changes requested):

python/revng/cli/_commands/idb_converter.py
```diff
@@ -0,0 +1,655 @@
...
+                unsigned = bool(match["unsigned"])
+                size = int(match["width"]) // 8
+                primitive_kind = (
+                    m.PrimitiveTypeKind.Unsigned if unsigned else m.PrimitiveTypeKind.Signed
+                )
+                qualified_type = self._get_primitive_type(primitive_kind, size)
+            elif type_name == "size_t":
+                primitive_kind = m.PrimitiveTypeKind.Unsigned
+                size = 8 if self.is64bit else 4
+                qualified_type = self._get_primitive_type(primitive_kind, size)
```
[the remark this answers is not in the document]
dwarfyresearch: Ouch... it would generate:

```
Types:
- ID: 1025
  Kind: PrimitiveType
  PrimitiveKind: Number
  Size: 1
- ID: 1284
  Kind: PrimitiveType
  PrimitiveKind: Unsigned
  Size: 4
- ID: 6940297126226104124
  Kind: TypedefType
  OriginalName: uint32_t
  UnderlyingType:
    Qualifiers: []
    UnqualifiedType: /Types/PrimitiveType-1284
- ID: 9514706589701521359
  Kind: TypedefType
  OriginalName: uint32_t
  UnderlyingType:
    Qualifiers: []
    UnqualifiedType: /Types/PrimitiveType-1284
- ABI: SystemV_x86_64
  Arguments:
  - Index: 0
    OriginalName: x
    Type:
      Qualifiers: []
      UnqualifiedType: /Types/TypedefType-9514706589701521359
  ID: 10813545522046549192
  Kind: CABIFunctionType
  OriginalName: fn
  ReturnType:
    Qualifiers: []
    UnqualifiedType: /Types/TypedefType-6940297126226104124
```

instead of:

```
Types:
- ID: 1025
  Kind: PrimitiveType
  PrimitiveKind: Number
  Size: 1
- ID: 2692631726386566055
  Kind: TypedefType
  OriginalName: uint32_t
  UnderlyingType:
    Qualifiers: []
    UnqualifiedType: /Types/PrimitiveType-1025
- ID: 10726857336666412012
  Kind: TypedefType
  OriginalName: uint32_t
  UnderlyingType:
    Qualifiers: []
    UnqualifiedType: /Types/PrimitiveType-1025
- ABI: SystemV_x86_64
  Arguments:
  - Index: 0
    OriginalName: x
    Type:
      Qualifiers: []
      UnqualifiedType: /Types/TypedefType-10726857336666412012
  ID: 352442133061364510
  Kind: CABIFunctionType
  OriginalName: fn
  ReturnType:
    Qualifiers: []
    UnqualifiedType: /Types/TypedefType-2692631726386566055
```
I will add a test for this.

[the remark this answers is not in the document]
dwarfyresearch: C producer:
```
typedef char uint32_t;

uint32_t fn(uint32_t x)
{
  return x;
}
```

python/revng/cli/_commands/idb_converter.py
```diff
@@ -0,0 +1,655 @@
...
+                    aliased_tiltypeinfo = type.til.types.find_by_name(aliased_type.name)
+                    if aliased_tiltypeinfo:
+                        aliased_type_ordinal = aliased_tiltypeinfo.ordinal
+
+                if aliased_type_ordinal is None:
+                    aliased_type = type.get_next_tinfo()
+
+                qualified_type = self._convert_idb_type_to_revng_type(
+                    aliased_type, ordinal=aliased_type_ordinal
+                )
```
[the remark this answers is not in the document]
dwarfyresearch: There is a line at the begging of the method that actually should solve that I guess.

python/revng/cli/_commands/idb_converter.py
```diff
@@ -0,0 +1,655 @@
...
+            )
+
+        elif type.is_decl_partial():
+            # Represents an unknown or void type with a known size.
+            revng_type = self.unwrap_qualified(
+                self._get_primitive_type(
+                    m.PrimitiveTypeKind.Generic,
+                    type.get_size(),
+                )
+            )
```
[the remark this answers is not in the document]
dwarfyresearch: Being primitive type acceptable means:
```
type.get_size() == 1 or type.get_size() == 2 or type.get_size() == 4 or \
                type.get_size() == 8 or type.get_size() == 10 or type.get_size() == 16
```
right?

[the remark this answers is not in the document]
dwarfyresearch: >We should check if this is compatible with being primitive type in terms of size. Another option would be to emit a char[] of the appropriate size.

Yeah, I agree.

python/revng/cli/_commands/idb_converter.py
```diff
@@ -0,0 +1,655 @@
...
+
+        else:
+            # IDA does not know anything about this type.
+            # TODO: In some cases we should emit a void type (when the type is always used as a
+            # pointer).
+            kind = m.PrimitiveTypeKind.PointerOrNumber
+            size = type.get_size()
+            if size == 0:
+                # Fallback.
+                size = 8 if self.is64bit else 4
```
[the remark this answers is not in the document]
dwarfyresearch: Yeah, you are right.

python/revng/cli/_commands/idb_converter.py
```diff
@@ -0,0 +1,655 @@
...
+                    f"Found invalid empty struct {type_name}, replacing with a void* typedef"
+                )
+                underlying_type = self._get_primitive_type(m.PrimitiveTypeKind.Void, 0)
+                underlying_type.Qualifiers.append(
+                    m.Qualifier(
+                        Kind=m.QualifierKind.Pointer,
+                        Size=8 if self.is64bit else 4,
+                    )
+                )
+                revng_type = m.TypedefType(OriginalName=type_name, UnderlyingType=underlying_type)
```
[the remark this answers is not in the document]
dwarfyresearch: Yeah. Agree.

python/revng/cli/_commands/idb_converter.py
dwarfyresearch: O yeah... ignoring the whole struct would be more appropriate way, right?

python/revng/cli/_commands/idb_converter.py
dwarfyresearch: Fixed.

python/revng/cli/_commands/idb_converter.py
dwarfyresearch: I thought that this is something that you discussed, but it makes no sense to me at all.

python/revng/cli/_commands/idb_converter.py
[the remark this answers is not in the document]
aleclearmind: ACK, then just make sure we have a test with a linked list, which is recursive.
We should already have a linked list `.c` in revng-qa IIRC.

python/revng/cli/_commands/idb_converter.py
aleclearmind: I don't know, I suggest you put `assert False` above `argument.Type = self._get_primitive_type(m.PrimitiveTypeKind.Generic, 8)` and run the tests, see what triggers it and decide what to do.
If nothing triggers it, drop.

dwarfyresearch: I made a script that creates an idb file out of each ```.compiled``` test, and nothing triggers this, I am dropping it.

python/revng/cli/_commands/idb_converter.py
dwarfyresearch: Yeah, warning is ok for sure. But I think that we should avoid creating prototype here, since IDA was not able to find the signature. For example, the case ```always-dead-return-value-revng-qa.compiled``` is a small example for this -- all 3 functions have no prototype info, and when we try ```import binary``` it looks as:
```
Functions:
  - Entry:           "0x401000:Code_x86_64"
    CustomName:      "prefix__start"
    OriginalName:    _start
    Type:            Regular
  - Entry:           "0x401006:Code_x86_64"
    CustomName:      "func"
    OriginalName:    func
    Type:            Regular
  - Entry:           "0x401013:Code_x86_64"
    CustomName:      "callee"
    OriginalName:    callee
    Type:            Regular
```
And I think that we should avoid production of the artificial type in the case of IDB as well.

aleclearmind: Having no prototype for a function is fine. It's a valid model.
No artificial type + warning makes sense I think.

python/revng/cli/_commands/idb_converter.py
dwarfyresearch: I've tried, but I have not found any case for this (in our test suite, or in the big cases you have shared with me that you used for ```defcons```).

>Is this the most-specific exception you can get? Isn't there an IDBException or something?

No, there is no specific exception in the package.

aleclearmind: If it doesn't happen drop it.

python/revng/cli/_commands/idb_converter.py
dwarfyresearch: Yeah, on the same page.

python/revng/cli/_commands/idb_converter.py
```diff
@@ -0,0 +1,655 @@
...
+                # Treat this case as `typedef void someothername`.
+                revng_void_type = self._get_primitive_type(primitive_kind, size)
+                revng_type = m.TypedefType(OriginalName=type_name, UnderlyingType=revng_void_type)
+            else:
+                revng_type = self.unwrap_qualified(self._get_primitive_type(primitive_kind, size))
+
+        elif type.is_decl_func():
+            # TODO: handle non C-ABI functions.
+            # We cannot handle stack arguments at the moment.
+            assert type.type_details.stkargs is None
```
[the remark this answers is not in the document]
dwarfyresearch: I was playing a bit with this, and my conclusion is that the ```python-idb``` does not populate/use this field at all. Also, I am seeing that (as far as I can tell), the IDA is not very good when recovering signatures of the functions with the stack arguments, so that may be the reason the field was not used at all in the package.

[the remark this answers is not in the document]
aleclearmind: Are you 100% sure? x86 32-bit only uses stack arguments, and IDA has been supporting that forever.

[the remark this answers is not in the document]
dwarfyresearch: Hmm, good example. I was playing with some very complex examples, but I made some very simple ones and IDA is able to recognize the signature. But anyhow, the ```type.type_details.stkargs``` are still ```None``` and all the arguments are being populated within ```type.type_details.args``` (for both stack and register arguments), so I guess we should just refactor the ```TODO``` by saying something like "This field is unused in the package....".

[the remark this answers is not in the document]
aleclearmind: OK. So we *do* support importing stack arguments, correct?
Up to you whether to drop the code entirely or leave a TODO.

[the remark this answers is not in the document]
dwarfyresearch: Right, we *do* support them.

python/revng/cli/_commands/idb_converter.py
aleclearmind: Wait, who is "it"?
What would be `size(my_struct)` in C?
What does IDA say from the UI?
What does our code say?

We either need to investigate more or drop. Up to you.

dwarfyresearch: Sorry for not being super clear.

The `sizeof`, as well as both `revng import binary` and IDA, are accurate - both report `8`.

Our code (the code from the original PR) implements the logic in the wrong way. I'd say that it is not trivial to implement it with the current information (I think we would need to improve/investigate the `python-idb` package to correctly calculate/pick the size of the struct from IDB files). Therefore I dropped this for now.

aleclearmind: Gotcha, leave a TODO and let's call it a day.

aleclearmind:
Nice! `:)`

PR #201, opened 2022-07-19
Pipeline: add write once containers

drblallo:


include/revng/Pipeline/Loader.h
```diff
@@ -211,10 +226,15 @@ class Loader {
...
 template<>
-struct llvm::yaml::MappingTraits<pipeline::ContainerDeclaration>
-  : public TupleLikeMappingTraits<pipeline::ContainerDeclaration> {};
+struct llvm::yaml::MappingTraits<pipeline::ContainerDeclaration> {
+  static void mapping(IO &TheIO, pipeline::ContainerDeclaration &Info) {
+    TheIO.mapRequired("Name", Info.Name);
+    TheIO.mapOptional("Type", Info.Type);
+    TheIO.mapOptional("Role", Info.Role);
+  }
+};
```
aleclearmind, line 237: Why?

include/revng/Pipeline/RegisterContainerFactory.h
```diff
@@ -56,4 +56,31 @@ class RegisterDefaultConstructibleContainer : Registry {
   void libraryInitialization() override {}
 };
 
+/// Instantiate a global object of this class
+/// for each default constructible container to be used as role you wish to
+/// register
+template<typename ContainerT>
+class RegisterDefaultConstructibleRole : Registry {
```
aleclearmind, line 63: Let's just have `RegisterRole`, we'll never have non default constructible roles.
It's just a list of strings, let's not make it more complex that it needs to be.

lib/Pipeline/Loader.cpp
```diff
@@ -123,14 +128,40 @@ Loader::parseInvocation(Step &Step, const PipeInvocation &Invocation) const {
...
   auto &Entry = It->second;
+  for (const auto &ContainerName : Invocation.UsedContainers) {
+
+    if (llvm::find(WriteOnceNames, ContainerName) == WriteOnceNames.end())
+      continue;
+
+    if (PipelineContext->hasRegisteredWriteOnceContainer(ContainerName))
+      return createStringError(inconvertibleErrorCode(),
+                               "Detected two uses of write once container %s\n",
+                               ContainerName.c_str());
```
aleclearmind, line 139:
```suggestion
    if (PipelineContext->hasRegisteredWriteOnceContainer(ContainerName)) {
      return createStringError(inconvertibleErrorCode(),
                               "Detected two uses of write once container %s\n",
                               ContainerName.c_str());
     }
```

include/revng/Pipeline/Context.h
```diff
@@ -96,6 +101,28 @@ class Context {
 
   const GlobalsMap &getGlobals() const { return Globals; }
 
+  void addWriteOnceContainer(llvm::StringRef Name,
```
aleclearmind, line 104: Can we come up with a name that's more intuitive than "WriteOnceContainer"?
Maybe "ConstantContainer" or "GlobalContainer" (sad).

aleclearmind (changes requested):

include/revng/Pipeline/Loader.h
drblallo: before type was mandatory, now it is either type or role

include/revng/Pipeline/RegisterContainerFactory.h
drblallo: mhh i think we will, in particular if a container must be configured with a type

aleclearmind: That can be a template argument. Why do you think we'll need a constructor?

drblallo: mistake, i meant a kind

include/revng/Pipeline/Loader.h
drblallo: i am doing the current pattern in all other mapping in that file, should i switch over to the suggested pattern in all of them?

include/revng/Pipeline/RegisterContainerFactory.h
aleclearmind: I'd expect a role to be a map from string to container, why do you envision the need for a kind?

include/revng/Pipeline/Loader.h
aleclearmind: Consider if it's worth it.
You could also consider switch to tuple trees.

PR #202, opened 2022-07-20
Introduce PTML output for Assembly

mrjackv:


lib/Yield/Support/PTML.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+  if (BufferEndsNewline)
+    OS << '\n';
+  TrailingNewline = BufferEndsNewline;
+}
+
+void PTMLIndentedOstream::writeIndent() {
+  if (IndentDepth > 0) {
+    OS << Tag(tags::Span, std::string(IndentSize * IndentDepth, ' '))
+            .add(attributes::scope, yield::ptml::scopes::Indentation)
```
aleclearmind, line 40: Why `scope`? Whitespace is much more a `token` then a `scope`.

include/revng/Yield/Support/PTML.h
```diff
@@ -0,0 +1,143 @@
...
+inline llvm::raw_ostream &operator<<(llvm::raw_ostream &OS, const Tag &TheTag) {
+  OS << TheTag.serialize();
+  return OS;
+}
+
+class PTMLIndentedOstream : public llvm::raw_ostream {
+private:
+  int IndentSize;
+  int IndentDepth;
+  bool TrailingNewline;
```
aleclearmind, line 125: What's this?

libexec/revng/revng-ptml
```diff
@@ -0,0 +1,267 @@
...
+if __name__ == "__main__":
+
+    def new_unraisablehook(arg):
+        if arg.exc_type != BrokenPipeError:
+            sys.__unraisablehook__(arg)
+
+    sys.unraisablehook = new_unraisablehook
+
+    with suppress(BrokenPipeError, KeyboardInterrupt):
+        main()
```
aleclearmind, line 267: Reasons to make this a standalone script as oppose to be a `revng.cli.commands_registry.Command`?

python/revng/daemon/schema.graphql.tpl
```diff
@@ -12,6 +12,8 @@ type Query {
     {%- for rank in structure.keys() %}
     {{ rank.name }}{{ rank | rank_param }}: {{ rank.name | capitalize }}!
     {%- endfor %}
+
+    debug: Debug!
```
aleclearmind, line 16: What's this?

aleclearmind (changes requested):

include/revng/ADT/Concepts.h
```diff
@@ -32,6 +32,11 @@ concept range = requires(T &R) {
   std::end(R);
 };
 
+template<class Container, typename ElementType>
+concept typed = range<Container> &&
+                std::is_convertible_v<typename Container::value_type,
+                                      ElementType>;
```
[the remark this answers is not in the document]
Cvelth: About this, the `ranges` namespace here was designed to mirror the c++20's ranges, since there's nothing like this one in the standard, I feel like it's better to move it outside the namespace - to avoid confusion when we update the compiler and get rid of the two other ones in favour of the standard ones.

lib/Yield/Support/PTML.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+#include "llvm/Support/ErrorHandling.h"
+
+#include "revng/Yield/Support/PTML.h"
+
+namespace yield::ptml {
+
+uint64_t PTMLIndentedOstream::current_pos() const {
+  return OS.tell();
+}
```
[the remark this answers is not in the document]
mrjackv: It's marked as abstract by the `llvm::raw_ostream` base class.

include/revng/ADT/Concepts.h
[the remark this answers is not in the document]
aleclearmind: (I didn't notice it was in a namespace)

libexec/revng/revng-ptml
mrjackv: I'd argue that since PTML will be "shipped" to the clients, it might be nice to have a standalone tool rather then have to drag the entirety of revng with it

lib/Yield/Support/PTML.cpp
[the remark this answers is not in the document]
aleclearmind: Then what about?
```suggestion
uint64_t PTMLIndentedOstream::current_pos() const {
  return OS.current_pos();
}
```

python/revng/api/exceptions.py
```diff
@@ -5,3 +5,7 @@
 
 class RevngException(Exception):
     pass
+
+
+class RevngCException(Exception):
+    pass
```
aleclearmind, line 11: What is this?

include/revng/Pipeline/Global.h
```diff
@@ -36,6 +36,8 @@ class Global {
   virtual llvm::Error applyDiff(const llvm::MemoryBuffer &Diff) = 0;
   virtual llvm::Error serialize(llvm::raw_ostream &OS) const = 0;
   virtual llvm::Error deserialize(const llvm::MemoryBuffer &Buffer) = 0;
+  virtual llvm::Error verify() const = 0;
+  virtual llvm::Error verifyOther(const llvm::MemoryBuffer &Buffer) const = 0;
```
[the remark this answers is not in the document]
mrjackv: Basically it's `verify` but instead of verifying itself it will verify if what's in the buffer, this is used to have a dry-run `set_global`

libexec/revng/revng-ptml
aleclearmind: Uhm, this calls for a more general question: what do we put in the `pip install`-able package?
In theory we could put a subset of those made available in `revng-distributable`.

include/revng/Pipeline/Global.h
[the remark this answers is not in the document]
aleclearmind: Mmmmh. Can you make it a static method?
If it doesn't use `this`, it doesn't make sense to make a method.

include/revng/Support/ErrorContainer.h
```diff
@@ -0,0 +1,82 @@
...
+#include "revng/Support/Assert.h"
+
+class ErrorContainer {
+private:
+  std::vector<llvm::Error> Errors;
+
+public:
+  ErrorContainer() = default;
+  ErrorContainer(ErrorContainer &&EC) = default;
+  ErrorContainer &operator=(ErrorContainer &&) = default;
```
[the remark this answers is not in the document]
aleclearmind: Yes, I know, I'm suggesting to
```suggestion
  ErrorContainer() = default;
  ErrorContainer(ErrorContainer &&EC) = default;
  ErrorContainer &operator=(ErrorContainer &&) = default;
  ErrorContainer(const ErrorContainer &EC) = delete;
  ErrorContainer &operator=(const ErrorContainer &) = delete;
```

include/revng/Pipeline/Context.h
```diff
@@ -62,6 +62,16 @@ class Context {
     return Globals.deserialize(GlobalName, Buffer);
   }
 
+  llvm::Expected<std::unique_ptr<Global>>
+  cloneGlobal(llvm::StringRef GlobalName) {
+    return Globals.clone(GlobalName);
+  }
+
+  llvm::Error applyDiffToGlobal(llvm::StringRef GlobalName,
+                                const llvm::MemoryBuffer &Buffer) {
```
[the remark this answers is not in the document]
mrjackv: Because the diff deserialization is a specialization of the global, so it cannot be done here
(I'm also not sure that it would work if we decided that diff = tupletreediff, since there might be some erasure)

include/revng/ADT/Concepts.h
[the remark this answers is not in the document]
mrjackv: Fixed

include/revng/Support/ErrorContainer.h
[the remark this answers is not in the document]
mrjackv: Ah ok, afaik they should be implicitly like that since `llvm::Error` has an `delete`d copy constructor

include/revng/Support/ErrorContainer.h
```diff
@@ -0,0 +1,82 @@
...
+      this->EC = ErrorContainer();
+      ECP = &(*this->EC);
+    } else {
+      ECP = EC;
+    }
+  }
+
+  ErrorContainer &operator*() { return *ECP; }
+  ErrorContainer *operator->() { return ECP; }
+};
```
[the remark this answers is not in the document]
mrjackv: It's a wrapper so that a `PipelineC` user can ignore errors with nullptr
e.g.
```cpp
bool rp_manager_apply_diff(..., rp_error_container *error_container) {
  // error_container might be nullptr
  ErrorContainerWrapper EC(error_container);
  // EC now either points to the error_container or to the class-variable EC
  revng_check(EC->empty());
```

lib/Yield/Support/PTML.cpp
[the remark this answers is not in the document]
mrjackv: `current_pos` is a private member, whereas `tell` isn't

python/revng/api/exceptions.py
mrjackv: It's a separate exception that is thrown when an `error_container` contains an error, this is to distinguish it from `RevngException` that's thrown if an error occurs on the python side

include/revng/Pipeline/Context.h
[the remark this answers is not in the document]
aleclearmind: Uhm, OK.

include/revng/Support/ErrorContainer.h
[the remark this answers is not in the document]
aleclearmind: Should or is? In any case in C++ we try to follow the [rule of five](https://en.wikipedia.org/wiki/Rule_of_three_(C%2B%2B_programming)). If you define certain guys, you define them all.

include/revng/Support/ErrorContainer.h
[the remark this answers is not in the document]
aleclearmind: OK, doc it.

lib/Yield/Support/PTML.cpp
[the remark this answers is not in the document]
aleclearmind: Ah but this comes from `mlir/IndentedOStream`. OK then.
Why can't you use the MLIR one again?

python/revng/api/exceptions.py
aleclearmind: `RevngC` has a quite specific meaning in rev.ng. Can you find a better name?
Maybe `RevngPipelineCException`?

include/revng/Pipeline/Global.h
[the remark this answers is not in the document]
mrjackv: It has to be `virtual` because each `Global` has its own specialized `verify` (and consequently `verifyOther`)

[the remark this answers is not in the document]
aleclearmind: This means that a `Global` represenst at the same time the type of a global and the data of that global, which is not super nice. They could be decoupled.

In order to avoid proliferation of `xOther` methods, I suggest we introduce some method such as `std::unique_ptr<Global> createNewOfSameType(MemoryBuffer)` and then you can call `verify` on that one.

include/revng/Pipeline/GlobalsMap.h
```diff
@@ -26,6 +26,18 @@ class GlobalsMap {
     return ToReturn;
   }
 
+  llvm::Error
+  applyDiff(llvm::StringRef GlobalName, const llvm::MemoryBuffer &Buffer) {
+    auto Iter = Map.find(GlobalName);
+    if (Iter == Map.end()) {
```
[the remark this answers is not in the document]
mrjackv: Fixed

include/revng/Pipeline/GlobalsMap.h
```diff
@@ -26,6 +26,18 @@ class GlobalsMap {
...
+      auto *Message = "pipeline loader context did not contain object %s";
```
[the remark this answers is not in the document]
mrjackv: Fixed

include/revng/Pipeline/GlobalsMap.h
```diff
@@ -26,6 +26,18 @@ class GlobalsMap {
...
+      return llvm::createStringError(llvm::inconvertibleErrorCode(),
+                                     Message,
+                                     GlobalName.str().c_str());
+    }
```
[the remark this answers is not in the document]
mrjackv: Done

lib/Yield/Support/PTML.cpp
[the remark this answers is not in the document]
mrjackv: Because It's not writing just spaces, it's outputting a PTML tag

include/revng/PipelineC/Prototypes.h
```diff
@@ -578,3 +594,36 @@ uint64_t rp_rank_get_depth(rp_rank *rank);
...
+
+/**
+ * \return if an error_container contains a valid error
+ */
+bool rp_error_container_is_empty(rp_error_container *error);
+
+/**
+ * \return number of errors present in the container
+ */
+uint64_t rp_error_container_error_count(rp_error_container *error);
```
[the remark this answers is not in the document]
mrjackv: Sounds good

include/revng/PipelineC/Prototypes.h
```diff
@@ -578,3 +594,36 @@ uint64_t rp_rank_get_depth(rp_rank *rank);
...
+ * \{
+ */
+
+/**
+ * \return a new error container
+ */
+rp_error_container * /*owning*/ rp_make_error_container();
+
+/**
+ * \return if an error_container contains a valid error
```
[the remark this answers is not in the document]
mrjackv: I think at one point it was something along the line of `rp_error_container_has_error`, but I probably changed my mind and forgot to update the docstring

python/revng/daemon/schema.graphql.tpl
mrjackv: I'm unsure where to include the `verify_{global,diff}` functions, so I put them in a separate section

libexec/revng/revng-ptml
mrjackv: Still, even in the case we have it in a pip package it would be a hassle to section off part of the `revng` command rather than `cp` the script

include/revng/PipelineC/Prototypes.h
```diff
@@ -154,11 +148,33 @@ rp_manager_create_global_copy(rp_manager *manager, const char *global_name);
...
+ * \param dry_run if true, the global is not set, but any error that would arise is still reported
+ * \param error_container container that will store the errors, if any
+ *      if nullptr is passed then errors are silently ignored
  * \return true on success
  */
 bool rp_manager_set_global(rp_manager *manager,
                            const char *serialized,
-                           const char *global_name);
+                           const char *global_name,
+                           bool dry_run,
```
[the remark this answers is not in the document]
mrjackv: Done

include/revng/PipelineC/Prototypes.h
```diff
@@ -154,11 +148,33 @@ rp_manager_create_global_copy(rp_manager *manager, const char *global_name);
...
+ * \param dry_run if true, the diff is not committed, but any error that would arise is still reported
+ * \param error_container container that will store the errors, if any
+ *      if nullptr is passed then errors are silently ignored
+ * \return true on success
+ */
+bool rp_manager_apply_diff(rp_manager *manager,
+                           const char *diff,
+                           const char *global_name,
+                           bool dry_run,
+                           rp_error_container *error_container);
```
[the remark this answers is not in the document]
mrjackv: Added a comment on top of `Prototypes.h` that explains the use of `rp_error_list` once, as to avoid duplication

include/revng/Support/ErrorContainer.h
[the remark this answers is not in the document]
mrjackv: Gotcha, fixed

include/revng/Support/ErrorContainer.h
```diff
@@ -0,0 +1,82 @@
...
+  bool empty() const { return Errors.empty(); }
+
+  void add(llvm::Error &&Error) {
+    if (Error)
+      Errors.push_back(std::move(Error));
+    else
+      llvm::consumeError(std::move(Error));
+  }
+
+  size_t count() const { return Errors.size(); }
```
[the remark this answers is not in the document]
mrjackv: Fixed

include/revng/Support/ErrorContainer.h
```diff
@@ -0,0 +1,82 @@
...
+
+  llvm::Error *get(size_t Index) { return &Errors[Index]; }
```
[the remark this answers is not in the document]
mrjackv: Fixed

include/revng/Support/ErrorContainer.h
```diff
@@ -0,0 +1,82 @@
...
+
+  ~ErrorContainer() {
+    for (auto &Error : Errors) {
+      llvm::consumeError(std::move(Error));
+    }
+  }
+
+  bool empty() const { return Errors.empty(); }
+
+  void add(llvm::Error &&Error) {
```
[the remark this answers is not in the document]
mrjackv: Fixed

include/revng/Support/ErrorContainer.h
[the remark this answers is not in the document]
mrjackv: Added a docstring atop the class, tell me if it's enough

lib/PipelineC/PipelineC.cpp
```diff
@@ -683,3 +716,34 @@ bool rp_diff_map_is_empty(rp_diff_map *map) {
...
+rp_error_container *rp_make_error_container() {
+  return new ErrorContainer();
+}
+
+bool rp_error_container_is_empty(rp_error_container *error_container) {
+  revng_check(error_container != nullptr);
+  return error_container->empty();
+}
+
+uint64_t rp_error_container_error_count(rp_error_container *error_container) {
```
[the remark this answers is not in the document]
mrjackv: Fixed

python/revng/api/exceptions.py
mrjackv: Renamed to `RevngPipelineCException`

python/revng/daemon/schema.graphql.tpl
aleclearmind: I'd put them on the side of `apply_diff`

libexec/revng/revng-ptml
aleclearmind: I think this problem will pop up with more tools and shipping them as standalone tools won't be a good long term solution.
I think in `revng.py` we should detect if `libexec/revng` exists and if not disable a bunch of commands.

python/revng/daemon/schema.graphql.tpl
mrjackv: technically they are not mutations, so putting them next to `apply_diff` would be misleading IMHO

aleclearmind: Next to where fetch a global then?

include/revng/Pipeline/Global.h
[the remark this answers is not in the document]
mrjackv: Done, now there's a `createNew` function
It also helped remove a couple of double deserializations, which is nice

include/revng/Pipeline/Global.h
```diff
@@ -87,6 +89,20 @@ class TupleTreeGlobal : public Global {
     return llvm::Error::success();
   }
 
+  llvm::Error verify() const override;
```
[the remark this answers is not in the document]
mrjackv: Added `TupleTreeCompatibleWithVerify`

python/revng/daemon/schema.graphql.tpl
mrjackv: Done

libexec/revng/revng-ptml
mrjackv: Fixed

lib/Yield/Support/PTML.cpp
mrjackv: Changed to token

include/revng/Yield/Support/PTML.h
mrjackv: Fixed

aleclearmind (changes requested):

lib/Yield/Assembly/HTML.cpp
```diff
@@ -5,65 +5,43 @@
 // This file is distributed under the MIT License. See LICENSE.md for details.
```
aleclearmind, line 5: Rename to `PTML`?

lib/Yield/Assembly/HTML.cpp
```diff
@@ -5,65 +5,43 @@
...
-static constexpr auto Untagged = "untagged";
+} // namespace tokenTypes
 
-} // namespace tags
+namespace scopes {
 
-namespace templates {
+static constexpr auto Function = "asm.function";
+static constexpr auto BasicBlock = "asm.basic-block";
+static constexpr auto Instruction = "asm.instruction";
```
aleclearmind, line 40: I'd like to see all this stuff documented in the [PTML pad](https://pad.rev.ng/emEtVjiLTz2dRCnf9avSng#).

include/revng/Yield/Support/PTML.h
```diff
@@ -0,0 +1,159 @@
...
+
+} // namespace tags
+
+namespace attributes {
+
+inline constexpr auto scope = "data-scope";
+inline constexpr auto token = "data-token";
+inline constexpr auto locationDefinition = "data-location-definition";
+inline constexpr auto locationReferences = "data-location-references";
+inline constexpr auto modelEditPath = "data-model-edit-path";
```
aleclearmind, line 32: This does not follow conventions for global variables casing

include/revng/Yield/Support/PTML.h
```diff
@@ -0,0 +1,158 @@
...
+
+} // namespace attributes
+
+namespace scopes {
+
+inline constexpr auto Annotation = "annotation";
+inline constexpr auto Comment = "comment";
+inline constexpr auto Error = "error";
+inline constexpr auto Link = "link";
```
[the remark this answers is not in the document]
mrjackv: They've been merged from [here](https://github.com/Cvelth/revng/commit/0b4636041b905f373b79c9918260afbf9d4faa7f), I can drop them if needed

include/revng/Yield/Support/PTML.h
mrjackv: Will fix, but shouldn't `clang-{format,tidy}` warn about this?

include/revng/Yield/Support/PTML.h
```diff
@@ -0,0 +1,158 @@
...
+    Attributes[Name] = llvm::join(Values, ",");
+    return *this;
+  }
+
+  template<typename... T>
+  requires(std::is_convertible_v<llvm::StringRef, T> and...)
+    Tag &add(llvm::StringRef Name, T &&...Values) {
+    Attributes[Name] = llvm::join(",", std::forward<T>(Values)...);
+    return *this;
+  }
```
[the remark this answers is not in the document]
mrjackv: I'd argue on the contrary, there should be only one way to represent a list of items in PTML (e.g. values concatenated with `,`), otherwise it'll lead to either:
* Fragmentation if at any point we decide that some attributes are concatenated differently
* Code duplication because each caller will exactly do the same thing copypasted

As for a name, I'd go with `addAttributeList` or something along those lines

include/revng/Yield/Support/PTML.h
```diff
@@ -0,0 +1,159 @@
...
+
+  std::string open() const {
+    llvm::SmallString<128> Out;
+    Out.append({ "<", TheTag });
+    for (auto &Pair : Attributes)
+      Out.append({ " ", Pair.first(), "=\"", Pair.second, "\"" });
+    //Out.pop_back();
```
aleclearmind, line 92: ?

include/revng/Yield/Support/PTML.h
```diff
@@ -0,0 +1,159 @@
...
+    Out.append({'>'});
+    return Out.str().str();
+  }
+
+  std::string close() const { return "</" + TheTag + ">"; }
+
+  std::string serialize() const { return open() + Content + close(); }
```
aleclearmind, line 99: Have you considered a RAII approach for this?

include/revng/Yield/Support/PTML.h
```diff
@@ -0,0 +1,159 @@
...
+inline std::string operator+(const Tag &LHS, const Tag &RHS) {
+  return LHS.serialize() + RHS.serialize();
+}
+
+inline llvm::raw_ostream &operator<<(llvm::raw_ostream &OS, const Tag &TheTag) {
+  OS << TheTag.serialize();
+  return OS;
+}
+
+class PTMLIndentedOstream : public llvm::raw_ostream {
```
aleclearmind, line 128: Move to its own header.

include/revng/Yield/Support/PTML.h
mrjackv: Oops, must have pushed a `wip` commit by accident. Squashed into the Tag commit

include/revng/Yield/Support/PTML.h
mrjackv: Done

include/revng/Yield/Support/PTML.h
```diff
@@ -0,0 +1,158 @@
...
+  explicit Tag(llvm::StringRef Tag) : TheTag(Tag.str()) {}
+  explicit Tag(llvm::StringRef Tag, llvm::StringRef Content) :
+    TheTag(Tag.str()), Content(Content.str()) {}
+
+  Tag &setContent(const llvm::StringRef Content) {
+    this->Content = Content.str();
+    return *this;
+  }
+
+  Tag &add(llvm::StringRef Name, llvm::StringRef Value) {
```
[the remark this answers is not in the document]
mrjackv: Done

lib/Yield/Support/PTML.cpp
```diff
@@ -0,0 +1,46 @@
+//
```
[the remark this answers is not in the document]
mrjackv: Done

lib/Yield/Assembly/HTML.cpp
mrjackv: Done

include/revng/Yield/Support/PTML.h
mrjackv: Fixed

include/revng/Yield/Support/PTML.h
aleclearmind: Can you paste an example output? I cannot appreciate the nuances otherwise.
All of this seems to have a lot of non-obvious implicit rules which would benefit a lot from being wrapped in a RAII object and we should really really strive for it.

mrjackv: Ok, I was mistaken, RAII works really well (dunno what happened the first time I tried it, probably did something wrong)
Added to PR

include/revng/Support/ErrorList.h
```diff
@@ -0,0 +1,96 @@
...
+    return failIf(std::move(Error), true, false);
+  }
+};
+
+/**
+ * Wrapper class to handle nullptr
+ * PipelineC can accept a nullptr as an ErrorList argument, this class wraps
+ * the pointer so that if it is nullptr it will point to the throwaway local
+ * ErrorList EL
+ */
```
aleclearmind, line 79: We don't use C-style doxygen, convert to `///`

include/revng/Yield/Support/PTML.h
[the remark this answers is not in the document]
aleclearmind: Drop everything that's unused.

include/revng/Yield/Support/PTML.h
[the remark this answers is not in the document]
aleclearmind: If you put this here, it's an actual extension of the "PTML" format so:
1. Document it in the pad (which will eventually become doc)
2. Assert no `,` in the values
3. Go for `addListAttribute`

I'm not a huge fan of this, but if it happens often that we need list of comma-separated things, let's go for it.

Are you sure we don't want to use `' '` as a separator? Would be similar to the `<div class="...">` attribute.

include/revng/PTML/Tag.h
```diff
@@ -0,0 +1,119 @@
...
+  public:
+    Scope(llvm::raw_ostream &OS, const Tag &TheTag, bool Newline) :
+      OS(OS), TheTag(TheTag) {
+      OS << TheTag.open();
+      if (Newline)
+        OS << "\n";
+    }
+
+    ~Scope() { OS << TheTag.close(); }
+  };
```
aleclearmind, line 45: Nice. I'd take it out of `class Tag` and document with an example on how to use it.

include/revng/Yield/Support/PTML.h
aleclearmind: Great!

include/revng/Yield/Support/PTML.h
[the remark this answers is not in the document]
mrjackv: Done

include/revng/PTML/Tag.h
mrjackv: Done

include/revng/Yield/Support/PTML.h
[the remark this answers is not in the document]
mrjackv: 1. will do
2. done
3. ok

> Are you sure we don't want to use `'  '` as a separator?

I'd stick with `,` since a list of `Target`s is already concatenated with `,` and it would be nice to remain consistent throughout the codebase. (` ` is still fine if we want to use it)

include/revng/Support/ErrorList.h
mrjackv: Done

lib/Yield/Assembly/HTML.cpp
mrjackv: Done

python/revng/cli/revng.py
```diff
@@ -15,14 +17,14 @@ def extend_list(paths, new_items):
...
-    from .override_by_name import ModelOverrideByName
-    from .pipeline import PipelineCommand
-    from .translate import TranslateCommand
+    with os.scandir(Path(__file__).parent / "_commands") as scan:
+        for entry in scan:
+            entry_path = Path(entry.path)
+            if entry.is_file():
+                import_module(f"._commands.{entry_path.stem}", "revng.cli")
+            elif entry.is_dir():
+                import_module(f"._commands.{entry_path.name}", "revng.cli")
```
aleclearmind, line 26: At this point we could also consider dropping `commands_registry.register_command(IRPipelineCommand())` and automatically register all those that inherit `Command`, but that's maybe too much, next time.

python/revng/cli/_commands/ptml.py
```diff
@@ -0,0 +1,273 @@
...
+            sys.exit(1)
+
+        def new_unraisablehook(arg):
+            if arg.exc_type != BrokenPipeError:
+                sys.__unraisablehook__(arg)
+
+        sys.unraisablehook = new_unraisablehook
+
+        with suppress(BrokenPipeError, KeyboardInterrupt):
+            options.parsed_args.func(options.parsed_args)
```
aleclearmind, line 270: This stuff is suspicious, document.

python/revng/cli/_commands/ptml.py
```diff
@@ -0,0 +1,273 @@
...
+
+def handle_file(
+    raw: str,
+    func_one: Callable[[str], T],
+    func_many: Callable[[Dict[str, str]], T],
+    filters: Union[str, List[str]],
+) -> T:
+    if is_ptml(raw):
+        if len(filters) > 0:
+            print("Cannot extract/filter a plain ptml file", file=sys.stderr)
```
aleclearmind, line 71: Never log to `stdout`

python/revng/cli/_commands/ptml.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+class PTMLCommand(Command):
+    def __init__(self):
+        super().__init__(("ptml",), "Tool to manipulate PTML files")
+
+    def register_arguments(self, parser: argparse.ArgumentParser):
+        subparsers = parser.add_subparsers(title="Subcommands", metavar="")
+
+        parser_cat = subparsers.add_parser("cat", help="Print the ptml on the console")
```
aleclearmind, line 197: No wait, we already have a way to handle subcommands, pretty please use that one.

Make distinct commands:
```python
super().__init__(("ptml","cat"), "Tool to manipulate PTML files")
```

python/revng/cli/revng.py
mrjackv: Done

python/revng/cli/_commands/ptml.py
mrjackv: Added docstring and isolated as `suppress_brokenpipe`

python/revng/cli/_commands/ptml.py
mrjackv: Done

include/revng/PTML/Constants.h
```diff
@@ -0,0 +1,35 @@
...
+
+inline constexpr auto Scope = "data-scope";
+inline constexpr auto Token = "data-token";
+inline constexpr auto LocationDefinition = "data-location-definition";
+inline constexpr auto LocationReferences = "data-location-references";
+inline constexpr auto ModelEditPath = "data-model-edit-path";
+
+} // namespace attributes
+
+namespace scopes {} // namespace scopes
```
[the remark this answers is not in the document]
mrjackv: It was previously populated but refactors emptied it, I opted to leave it for future use to include "generic" scopes if needed

python/revng/cli/_commands/ptml.py
mrjackv: Fixed, used `logging`

aleclearmind: Mmmh, `logging` sucks big time.
```
>>> import logging
>>> logging.warning("asd")
WARNING:root:asd
```
I see you introduced it in `revng-daemon`, but I'd prefer to stick with a simple `log` function as we're doing elsewhere.
If we decide to adopt something fancier, we need to adopt it *everywhere* and probably it won't be `logging`.

In order to avoid extending the scope of this PR, let's stick with a good old
```
def log(message):
    sys.stderr.write(message + "\n")
```
This will make refactoring way easier.

mrjackv:
```python
>>> import logging
>>> logging.basicConfig(format="%(message)s")
>>> logging.warning("asd")
asd
```
That said _sure_ the default isn't pretty, but besides that the `logging` library is pretty standard in terms of features (in fact it is used in `revng-daemon` since if we ever need to hook it to a LLVM logger it is _dead simple_)

I'll revert the commit and add `def log`, however I'd say to pick a logging library (I'd go with `logging` because it's in the standard library and pretty capable, but I'm open to using an external library) at some point, otherwise we'll end up with every module with their own `def log` that does things slightly differently and then it'll be a pain to switch.

aleclearmind: I'm in favor of switching, but it needs to be done globally. Opportunistic local improvements that deviate from the standard are a recipe for an incoherent code base that makes refactoring harder.
As a maintainer it's my duty to avoid this! `:)`.

I'd say spin up a new PR for this, but frankly we need to keep priorities straight and we already have a lot of infra projects in flight.

PR #206, opened 2022-07-26
Model: import data objects in `Segment.Type`

antoniofrighetto:


include/revng/Model/IRHelpers.h
```diff
@@ -58,3 +58,55 @@ inline model::TypePath createEmptyStruct(model::Binary &Binary, uint64_t Size) {
   NewStruct->Size = Size;
   return Path;
 }
+
+struct DataObjectSymbol {
```
aleclearmind, line 62:
```suggestion
struct DataSymbol {
```

lib/Model/Importer/Binary/ELFImporter.cpp
```diff
@@ -429,17 +437,33 @@ void ELFImporter<T, HasAddend>::parseSymbols(object::ELFFile<T> &TheELF,
...
+
+    if (IsCode) {
+      auto It = Model->Functions.find(Address);
+      if (It == Model->Functions.end()) {
+        model::Function &Function = Model->Functions[Address];
+        Function.Type = model::FunctionType::Regular;
+        if (MaybeName)
+          Function.OriginalName = *MaybeName;
+      }
+    }
```
aleclearmind, line 459: `else`

lib/Model/Importer/Binary/ELFImporter.cpp
```diff
@@ -429,17 +437,33 @@ void ELFImporter<T, HasAddend>::parseSymbols(object::ELFFile<T> &TheELF,
...
+
+    if (MaybeName and IsDataObject and !IsWeak and Size > 0) {
```
aleclearmind, line 461: * [ ] Why do *need* a name?
* [ ] Why do you ignore weak symbols? I think we said not to do this.
* [ ] `s/!/not/`

lib/Model/Importer/Binary/ELFImporter.cpp
```diff
@@ -429,17 +437,33 @@ void ELFImporter<T, HasAddend>::parseSymbols(object::ELFFile<T> &TheELF,
...
+      auto It = llvm::find_if(DataSymbols, [Address](const auto &E) {
```
aleclearmind, line 462: Use `llvm::count`.

lib/Model/Importer/Binary/ELFImporter.cpp
```diff
@@ -629,7 +652,10 @@ void ELFImporter<T, HasAddend>::parseDynamicSymbol(Elf_Sym_Impl<T> &Symbol,
       }
     } else {
       Address = relocate(fromGeneric(Symbol.st_value));
-      // TODO: create field in segment struct
+      auto &DS = DataSymbols;
```
aleclearmind, line 655: No abbreviations, reorganize code.

aleclearmind (changes requested):

include/revng/Model/IRHelpers.h
```diff
@@ -58,3 +58,55 @@ inline model::TypePath createEmptyStruct(model::Binary &Binary, uint64_t Size) {
...
+
+  bool operator==(const DataObjectSymbol &) const = default;
+};
+
+using DataSymbolVecTy = llvm::SmallVectorImpl<DataObjectSymbol>;
+
+inline void fillSegmentTypeStruct(model::Binary &Binary,
+                                  model::TypePath &Path,
+                                  model::Segment &Segment,
+                                  DataSymbolVecTy &DataSymbols) {
```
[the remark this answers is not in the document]
antoniofrighetto: It is only used by `ELFImporter` as of now, but it is meant to be used by `MachOImporter` as well as `PECOFFImporter` in the future. Sounds like a helper to me, although `BinaryImporterHelper.h` looks more suitable to place it. What do you think?

include/revng/Model/IRHelpers.h
```diff
@@ -58,3 +58,55 @@ inline model::TypePath createEmptyStruct(model::Binary &Binary, uint64_t Size) {
...
+      model::QualifiedType FieldType;
+      model::TypePath T;
+
+      if (Size == 2)
+        T = Binary.getPrimitiveType(model::PrimitiveTypeKind::Unsigned, 2);
+      else if (Size == 4)
+        T = Binary.getPrimitiveType(model::PrimitiveTypeKind::Unsigned, 4);
+      else if (Size == 8)
+        T = Binary.getPrimitiveType(model::PrimitiveTypeKind::Unsigned, 8);
+      else
```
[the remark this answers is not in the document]
antoniofrighetto: The reasoning was: if we have an arbitrary size, creating a generic of 1 byte and then specifying later that it's an array of size `Size` in the qualifiers of the `model::QualifiedType` should be good. Does it make sense to you? I feel like I should comment this properly.

include/revng/Model/IRHelpers.h
```diff
@@ -58,3 +58,55 @@ inline model::TypePath createEmptyStruct(model::Binary &Binary, uint64_t Size) {
...
+        T = Binary.getPrimitiveType(model::PrimitiveTypeKind::Generic, 1);
+
+      if (Size == 1 || Size == 2 || Size == 4 || Size == 8)
+        FieldType = { T, {} };
+      else
+        FieldType = { T, { { model::QualifierKind::Array, Size } } };
```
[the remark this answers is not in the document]
antoniofrighetto: OK, saw it now, let me reorganize the code taking into account the comment above.

include/revng/Model/IRHelpers.h
```diff
@@ -58,3 +58,55 @@ inline model::TypePath createEmptyStruct(model::Binary &Binary, uint64_t Size) {
...
+
+    if (Segment.contains(Address)) {
+      auto Offset = Address - Segment.StartAddress;
+      revng_assert(Offset.has_value());
+
+      model::QualifiedType FieldType;
+      model::TypePath T;
+
+      if (Size == 2)
+        T = Binary.getPrimitiveType(model::PrimitiveTypeKind::Unsigned, 2);
```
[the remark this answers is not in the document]
antoniofrighetto: I feel like I wanted to be the most conservative, and as a generic representation of both pointer and a value, for, e.g., 8 bytes, `uint64_t` seemed the right one to me. Maybe `PointerOrNumber` would be better here for register-size values?

lib/Model/Importer/Binary/ELFImporter.cpp
antoniofrighetto: `else` nothing, no? We only want to set `OriginalName` if `MaybeName` exists.

include/revng/Model/IRHelpers.h
```diff
@@ -58,3 +58,55 @@ inline model::TypePath createEmptyStruct(model::Binary &Binary, uint64_t Size) {
...
+        T = Binary.getPrimitiveType(model::PrimitiveTypeKind::Generic, 1);
+
+      if (Size == 1 || Size == 2 || Size == 4 || Size == 8)
+        FieldType = { T, {} };
+      else
+        FieldType = { T, { { model::QualifierKind::Array, Size } } };
+
+      model::StructField Field{ *Offset, {}, Name.str(), FieldType };
+      const auto &[_, Success] = Struct->Fields.insert(Field);
+      revng_assert(Success);
```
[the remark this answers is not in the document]
antoniofrighetto: Made clearer that the struct contains segments and segments only by creating an empty struct within this function, and returning it populated.

lib/Model/Importer/Binary/ELFImporter.cpp
antoniofrighetto: We may wish to ignore all symbols that overlap with previously introduced symbols (at the same address), but if a symbol is permitted to be overridden, we may want to use/show the strong one rather than the weak one (exactly like IDA and Ghidra do).

include/revng/Model/IRHelpers.h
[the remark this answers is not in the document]
aleclearmind: Make a header in the implementation directory with an appropriate name.
`SegmentImportHelpers.h`?

include/revng/Model/IRHelpers.h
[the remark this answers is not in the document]
aleclearmind: Use `Generic`, we don't know anything about this, so let's not make information up.

include/revng/Model/IRHelpers.h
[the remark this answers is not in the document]
aleclearmind: In `ELFImporter.h` you're replacing the type of `Segment` unconditionally, this is not OK.
You should create an empty struct if there's no `Type` yet, and introduce new fields in empty spaces otherwise.

lib/Model/Importer/Binary/ELFImporter.cpp
aleclearmind: Here you're not preferring the strong one if available, you're ignoring entirely the weak ones.

Make sure the other bullets in my comment are addressed too.

antoniofrighetto: > * [ ]  Why do _need_ a name?
> * [ ]  Why do you ignore weak symbols? I think we said not to do this.
> * [ ]  `s/!/not/`

1. We don't, removed.
2. Must have thought to handle only a specific set of cases that happened within revng tests. Removed `!IsWeak`, we cannot ignore weak symbols which may not have the strong one (linking phase might have not happened yet).
3. Sure.

share/revng/test/tests/model/import/types/types.c.model.yml
```diff
@@ -0,0 +1,207 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+---
+Segments:
+  - StartAddress:    "0x400000:Generic64"
```
aleclearmind, line 7: It seems you commited the model file as is, you need to clean it up from all the irrelevant things from the purpose of your test. In particular addresses, superfluous primitive types, size you're not interested in...

antoniofrighetto: I thought it was good to keep at least the key, `StartAddress` and `VirtualSize`; but OK, removed all of them and kept only the relevant types.

share/revng/test/configuration/revng/for-import-types.yml
```diff
@@ -0,0 +1,13 @@
...
+#
+
+commands:
+  - type: revng.test-import-types
+    from:
+      - type: revng-qa.compiled-with-debug-info
+        filter: for-import-types
+    command: |-
+      revng lift "$INPUT" -
+        | revng model dump
```
aleclearmind, line 12: No need to lift, you can just use `revng analyze ImportBinary "$INPUT"`

aleclearmind:
Merged in a910d3b. Thanks!

PR #207, opened 2022-08-01
revng-check-conventions v2

mrjackv:


libexec/revng/revng-check-conventions
```diff
@@ -1,498 +1,727 @@
...
-            -e '^include/revng/Support/Assert\.h$' \
-            -e '^include/revng/Support/ClassSentinel\.h$' \
-            -e '\.c$' \
-            <<< "$C_FILES_N"
+def get_config(filename: str) -> Path:
+    filepath = Path(__file__).parent / "../../share/revng" / filename
+    if filepath.is_file():
+        return filepath
+    else:
+        print(f"Could not find config file {filename}", file=sys.stderr)
```
aleclearmind, line 467: Let's have a `log` function as customary.

libexec/revng/revng-check-conventions
```diff
@@ -1,498 +1,727 @@
...
+        sys.exit(1)
```
aleclearmind, line 468: Can we have a `fail` function maybe?

libexec/revng/revng-check-conventions
```diff
@@ -1,498 +1,727 @@
...
+
+        return PassResult(
+            generate_error_lines("Whitespace at the end of line:", trailing_whitespace)
+            + generate_error_lines("Tabs present:", tabs)
+            + generate_error_files("Files that don't end in a newline:", no_end_newline),
+            empty(trailing_whitespace, tabs, no_end_newline),
+        )
+
+
+class RevngCPass(Pass):
```
aleclearmind, line 326: `RevngC` has a specific meaning in our context, find a better name.

libexec/revng/revng-check-conventions
```diff
@@ -1,498 +1,727 @@
...
-      See clang-format documentation for more details.
 
-    --print-clang-format-config
-      Print the clang-format configuration to stdout, ignoring all the other
-      arguments except for --use-local-clang-format-file.
-      The conventions are not checked.
+class Pass(abc.ABC):
+    def __init__(self, name: str, tag: str):
+        self.name = name
+        self.tag = tag
```
aleclearmind, line 62: Doc

libexec/revng/revng-check-conventions
```diff
@@ -1,498 +1,727 @@
...
+                normal_files.append(filepath)
+        runs = [super().process({f: "" for f in normal_files}, args)]
+        for script in script_files:
+            runs.append(super().process({script: ""}, args))
+        return PassResult.combine(*runs)
+
+
+class LicenseCheckPass(Pass):
+    def __init__(self):
+        super().__init__("license-check", "__all")
```
aleclearmind, line 248:
```suggestion
        super().__init__("license-check", "all")
```

libexec/revng/revng-check-conventions
```diff
@@ -1,498 +1,727 @@
...
+        if shebang != "":
+            first_line = content.splitlines()[0]
+            if re.search(r"#!.*" + shebang, first_line):
+                return True
+        return False
+
+    return matcher
+
+
+FILTERS = {
```
aleclearmind, line 452: Doc. What are `x, y`?

libexec/revng/revng-check-conventions
```diff
@@ -1,498 +1,727 @@
...
+        tag: {path: contents[path] for path in paths.keys()} for tag, paths in initial_dict.items()
+    }
+
+
+WRITE_PASSES: Tuple[Pass, ...] = (
+    ThreadedCommandPass(
+        "clang-format",
+        "c",
+        ("clang-format", "-Werror", "-style=${clang_style}", "-i"),
+        ("clang-format", "-Werror", "--dry-run", "-style=${clang_style}", "-i"),
```
aleclearmind, line 564:
```suggestion
        a="clang-format",
        b="c",
        c=("clang-format", "-Werror", "-style=${clang_style}", "-i"),
        d=("clang-format", "-Werror", "--dry-run", "-style=${clang_style}", "-i"),a
```

libexec/revng/revng-check-conventions
```diff
@@ -1,498 +1,727 @@
...
-      arguments except for --use-local-clang-format-file.
-      The conventions are not checked.
+class Pass(abc.ABC):
+    def __init__(self, name: str, tag: str):
+        self.name = name
+        self.tag = tag
 
-    --commit-range <RANGE>
-      Check files changed in the specified commit range
+    def __call__(self, files: Dict[str, Dict[Path, str]], args: Dict[str, str]) -> PassResult:
```
aleclearmind, line 64: Overriding `__call__` usually makes things a bit obscure.
Is there any real benefit?
What's the difference between `__call__` and `process`?

libexec/revng/revng-check-conventions
```diff
@@ -1,498 +1,727 @@
...
-    local MYPY_FILES
-    # mypy will complain that there are multiple __main__ files
-    # in the python/scripts directory, manually exclude them
-    # Will be fixed by https://github.com/python/mypy/pull/12543/files
-    readarray -t MYPY_FILES < <(printf '%s\n' "${PYTHON_FILES[@]}" | grep -vF 'python/scripts')
-    if [[ ${#MYPY_FILES[@]} -eq 0 ]]; then return; fi
-    mypy --scripts-are-modules --ignore-missing-imports --no-error-summary \
-         "${MYPY_FILES[@]}"
-}
+def startswith(string: str, prefixes: Iterable[str]):
```
aleclearmind, line 154:
```suggestion
def any_startswith(string: str, prefixes: Iterable[str]):
```
?

libexec/revng/revng-check-conventions
```diff
@@ -1,498 +1,727 @@
...
+        self.command = command
+        self.dry_command = dry_command if dry_command is not None else command
+
+    def process(self, files: Dict[Path, str], args) -> PassResult:
+        cmd_args = template_iterable(self.dry_command if args["dry_run"] else self.command, args)
+        process = run_with_capture([*cmd_args, *map(str, files)])
+        return PassResult.from_process(process)
+
+
+class ThreadedCommandPass(Pass):
```
aleclearmind, line 198:
```suggestion
class ParallelCommandPass(Pass):
```
?

libexec/revng/revng-check-conventions
```diff
@@ -1,498 +1,727 @@
...
-    done
-    printf '%s\n' "${RESULT[@]}"
-}
 
-# git ls-files wrapper
-# wrap git ls-files to exclude deleted files
-function git_ls_files_wrapper() {
-    comm -23 <(git ls-files "$@" | sort) <(git ls-files --deleted | sort)
-}
+def run_parallel(func, objs_list, func_params=()):
```
aleclearmind, line 84: * So this runs stuff in parallel but only tasks of the same type, correct? When this kind of stuff happens I feel like we're reinventing `ninja`. But this is likely not the case. At very least, we can provide better reporting.
* Can we take this occasion to experiment with thread-less asyncio?
  There's an interesting [SO answer](https://stackoverflow.com/questions/48483348/how-to-limit-concurrency-with-python-asyncio) that offers many alternative solutions.
  This has been a recurring topic and I really think we should learn The Right Way™ to do non-CPU intensive coroutines in Python.

libexec/revng/revng-check-conventions
```diff
@@ -1,498 +1,727 @@
...
+    ),
+)
+
+READ_PASSES: Tuple[Pass, ...] = (
+    SingleCommandPass("flake8", "python", ("flake8", "--config", "$flake8_config")),
+    MyPyPass(),
+    LicenseCheckPass(),
+    BashPass(),
+    RevngCPass(),
+    RevngGeneralPass(),
```
aleclearmind, line 595: Can you drop `Revng` prefix and get more descriptive names?

tests/pipeline/run_crash_test.sh
```diff
@@ -3,13 +3,8 @@
...
 
-# TODO: errexit and pipefail are intentionally not set. Here I'm setting and
-# and unsetting them just to silence check-conventions. Remove hack once
-# silencing functionality is added to check-conventions.
-set -euo pipefail
-
-set +o pipefail
-set +o errexit
+# rcc-ignore: bash-set-flags
+set -u
```
aleclearmind, line 7: rcc-2.0 is really really great, but we need a `README.md` documenting such features in detail.

libexec/revng/revng-check-conventions
```diff
@@ -1,498 +1,727 @@
...
+    )
+    parser_file_selection.add_argument(
+        "--HEAD", action="store_true", help="Alias for '--commit-range HEAD^..HEAD'"
+    )
+    parser_file_selection.add_argument(
+        "--cached", action="store_true", help="Only check staged files"
+    )
+    parser_file_selection.add_argument(
+        "--all", action="store_true", help="Check all files tracked by git"
+    )
```
aleclearmind, line 699: `--all` absolutely needs to be the default, in particular now that we're super-parallel. It has been a constant source of mistakes.
We can still benefit `--HEAD` in `orc check-branch`!

libexec/revng/revng-check-conventions
```diff
@@ -1,498 +1,727 @@
...
+                match = re.search(r"^\s*#include <.*\.hpp>", line)
+            return [match] if match else []
+
+        return include_matcher_inner
+
+    def process(self, files: Dict[Path, str], args) -> PassResult:
+        non_support_files = {
+            path: content
+            for path, content in files.items()
+            if str(path) not in ["share/revng/support.c", "lib/Support/Assert.cpp"]
```
aleclearmind, line 390: rcc is sprinkled with hardcoded strings. It'd be nice to have a configuration file in YAML for as many things as possible.
If we don't feel like going this route right now, at least please concentrate all the configuration (in particular, hardcoded strings), in a single place.

When I'll get unhappy about rcc, it'd be nice to have a single place where to change the configuration (maybe with docs), instead of having to search what I want among the program logic.

libexec/revng/revng-check-conventions
```diff
@@ -1,498 +1,727 @@
...
+    def process(self, files: Dict[Path, str], args: Dict[str, str]) -> PassResult:
+        shellcheck = super().process(files, args)
+        noset_files = [path for path, content in files.items() if not self.check_set(content)]
+        output = generate_error_files(
+            "There are script files without 'set -euo pipefail':", noset_files
+        )
+        return PassResult.combine(shellcheck, PassResult(output, len(noset_files) == 0))
+
+
+class RevngGeneralPass(Pass):
```
aleclearmind, line 300:
```suggestion
class WhitespaceChecksPass(Pass):
```
?

aleclearmind (changes requested):
This is pretty cool. Good job!

libexec/revng/revng-check-conventions
mrjackv: Done

libexec/revng/revng-check-conventions
mrjackv: Added

libexec/revng/revng-check-conventions
mrjackv: Renamed to `CPPCheckPass`

libexec/revng/revng-check-conventions
mrjackv: Renamed

libexec/revng/revng-check-conventions
mrjackv: Removed, passes are specified via config

libexec/revng/revng-check-conventions
mrjackv: Sure

libexec/revng/revng-check-conventions
mrjackv: I've reworked the passes a bit, now they're all `async` so a lot of threading code has been dropped
The only bit remaining that uses `run_parallel` is the one for reading the files, which does not have a `async` counterpart and *does* scale with python threads, so I've inlined the function and made it a bit less generic

libexec/revng/revng-check-conventions
mrjackv: Done

libexec/revng/revng-check-conventions
mrjackv: Done

libexec/revng/revng-check-conventions
mrjackv: Done

libexec/revng/revng-check-conventions
mrjackv: Added

libexec/revng/revng-check-conventions
mrjackv: Renamed

libexec/revng/revng-check-conventions
mrjackv: 1. I thought it was fitting because you "run" the pass given a filtered file list and a list of arguments, if you prefer it can be renamed to `run`
2.I've added a bit of doc but basically ` __call__` will be given all the files and in its default implementation will check if there are any files for the `tag` attribute and if so will forward to the process function, this is so that a `Pass` can operate on more than one tag but in most cases it will be inherited and `process` will be overridden

libexec/revng/revng-check-conventions
mrjackv: removed `--all`, it is now the default if no files are provided

tests/pipeline/run_crash_test.sh
mrjackv: Added a `README` in `share/revng`

libexec/revng/revng-check-conventions
aleclearmind: So `process` works on an individual `tag` while `__call__` works on a `Dict[str, List[str]]`?
Maybe a better name is `process_tag`.
As for `__call__`, operators override is scary and suspicious, I'd do it only if you need to interact with some library that expects an object to actually be callable. Otherwise a `.process` or `.run` method is simpler and more explicit.

share/revng/revng-check-conventions-README.md
```diff
@@ -0,0 +1,48 @@
+# Information about `revng-check-conventions`
```
aleclearmind, line 1: Move to `doc/revng-check-conventions.md`.
One day we'll make a `man` out of it or something.

libexec/revng/revng-check-conventions
aleclearmind: Great!

libexec/revng/revng-check-conventions
aleclearmind: Nice. Out of curiosity, did measure that the reading the files was a bottleneck? Interesting.

share/revng/revng-check-conventions-README.md
mrjackv: Moved

libexec/revng/revng-check-conventions
mrjackv: Renamed to `run`, kept `process`, `process_tag` sounds like you are doing an operation on the tags, not that you'll be given a filelist of a single tag

libexec/revng/revng-check-conventions
mrjackv: I've run a quick check, apparently `run_in_executor` has a bit of overhead, so when reading a lot of small files it's slower than reading them sequentially, whereas it's faster in the case of reading a few big files

Depau:
@mrjackv can you please add a check to catch if `InitRevng` is missing from entry points as pointed out in #219?

mrjackv:
> @mrjackv can you please add a check to catch if InitRevng is missing from entry points as pointed out in https://github.com/revng/revng/pull/219?

Added to the todo, will update once done

mrjackv:
@Depau Added the `InitRevng` rule

PR #209, opened 2022-08-03
Introduce `getInvalidations` for IsolatedKind

antoniofrighetto:


lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -50,6 +52,107 @@ using TaggedFK = TaggedFunctionKind;
...
+using IsolatedFK = IsolatedFunctionKind;
+void IsolatedFK::getInvalidations(const Context &Ctx,
+                                  TargetsList &ToRemove,
+                                  const GlobalTupleTreeDiff &Diff) const {
+  using PM = PathMatcher;
+  using Binary = model::Binary;
+
+  const auto &CurrentModel = getModelFromContext(Ctx);
+  if (not Ctx.containsReadOnlyContainer(BinaryCrossRelationsRole))
+    return;
```
aleclearmind, line 67: The `BinaryCrossRelationsRole` is only for refining the set of things you need to remove.
If you don't have it, you must invalidate *all* the targets. Here you're invalidating no target at all.

lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -50,6 +52,107 @@ using TaggedFK = TaggedFunctionKind;
...
+
+  static constexpr auto BCRR = BinaryCrossRelationsRole;
```
aleclearmind, line 69:
```suggestion
  static constexpr auto &BCRR = BinaryCrossRelationsRole;
```
You're copying it here,I think.

lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -50,6 +52,107 @@ using TaggedFK = TaggedFunctionKind;
 void TaggedFK::getInvalidations(const Context &Ctx,
                                 TargetsList &ToRemove,
                                 const GlobalTupleTreeDiff &Diff) const {
+  return;
+}
+
+using IsolatedFK = IsolatedFunctionKind;
+void IsolatedFK::getInvalidations(const Context &Ctx,
+                                  TargetsList &ToRemove,
+                                  const GlobalTupleTreeDiff &Diff) const {
```
aleclearmind, line 61: You're using `auto` way too much. Please change this habit now.

**Do not abuse `auto`**. Use `auto` only if:
    1. The line is very long and you can't figure any other way to make `check-conventions` happy.
    2. The type is obvious since it's expressedly
 stated on the same line (e.g., `auto *Lol = dyn_cast<CallInst>(...)`).
    3. The type is very ugly (typically, iterators).

lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -50,6 +52,107 @@ using TaggedFK = TaggedFunctionKind;
...
+  auto MaybeBuffer = llvm::MemoryBuffer::getFile(*MaybeContainerPath);
+  revng_assert(MaybeBuffer);
+  yaml::Input YAMLInput((*MaybeBuffer)->getBuffer());
+
+  yield::CrossRelations Relations;
+  YAMLInput >> Relations;
+
+  const auto *ModelDiff = Diff.getAs<Binary>();
+  if (not ModelDiff)
+    return;
```
aleclearmind, line 83: @drblallo Doesn't the diff include which global is subject to the change? We can't decide how to at only based on type.

lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -50,6 +52,107 @@ using TaggedFK = TaggedFunctionKind;
...
+
+  const Binary &Model = *CurrentModel.get();
+  auto EntryMatcher = PM::create<Binary>("/Functions/*/Entry").value();
```
aleclearmind, line 86: As we said multiple times, the key *cannot* change. You should never see this.
If someone changes the key, you will not see a change on `/Functions/0x1000/Entry` you will see the whole function (`/Functions/0x1000`) being removed and added again (`/Functions/0x2000`).
Did you to make such a change and see what `revng model diff` does?

lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -50,6 +52,107 @@ using TaggedFK = TaggedFunctionKind;
...
+  auto FSOMatcher = PM::create<Binary>("/Types/RawFunctionType-*/"
+                                       "FinalStackOffset")
+                      .value();
+  auto PresRegsMatcher = PM::create<Binary>("/Types/RawFunctionType-*/"
+                                            "PreservedRegisters")
+                           .value();
+  auto AttrMatcher = PM::create<Binary>("/Functions/*/Attributes").value();
+  auto DynFuncMatcher = PM::create<Binary>("/ImportedDynamicFunctions/"
+                                           "*/OriginalName")
+                          .value();
```
aleclearmind, line 96: Ibidem. This is a key, it will never match.

lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -50,6 +52,107 @@ using TaggedFK = TaggedFunctionKind;
...
+    return;
+
+  const Binary &Model = *CurrentModel.get();
+  auto EntryMatcher = PM::create<Binary>("/Functions/*/Entry").value();
+  auto FSOMatcher = PM::create<Binary>("/Types/RawFunctionType-*/"
+                                       "FinalStackOffset")
+                      .value();
+  auto PresRegsMatcher = PM::create<Binary>("/Types/RawFunctionType-*/"
+                                            "PreservedRegisters")
+                           .value();
```
aleclearmind, line 92: I'm afraid these won't do. We're not considering `CABIFunctionType`.

I'm starting to think that the best approach, at this point, is to ignore the diff and just do:

```c++
for (const model::Function &NewFunction : NewModel.Functions) {
  const auto &OldFunction = OldModel.Functions.at(NewFunction.Entry);
  Layout Old = makeLayout(OldFunction);
  Layout New = makeLayout(NewFunction);

  if (Old.FSO != New.FSO) {
    // ...
  }

}
```

In the future we could refine things reasoning on the diff to be more efficient.

lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -50,6 +52,107 @@ using TaggedFK = TaggedFunctionKind;
...
+  auto AttrMatcher = PM::create<Binary>("/Functions/*/Attributes").value();
+  auto DynFuncMatcher = PM::create<Binary>("/ImportedDynamicFunctions/"
+                                           "*/OriginalName")
+                          .value();
+
+  for (const auto &Change : ModelDiff->Changes) {
+    if (Change.Path.empty())
+      continue;
```
aleclearmind, line 100: What? Assert?

aleclearmind (changes requested):

lib/Pipes/TaggedFunctionKind.cpp
antoniofrighetto: The [upcoming structure](https://github.com/revng/revng/pull/209/commits/33268127353b18637111e97de73726f5c3c20988) of the Pipes guarantees the `BinaryCrossRelationsRole` is created if there at least a function has been isolated. If `CrossRelations` container does not exist, there are no isolated functions at all, thus returning should be reasonable.

lib/Pipes/TaggedFunctionKind.cpp
antoniofrighetto: I believe I have used `auto` appropriately so far. For instance, you can infer that the following line returns a ContainerType of `FileContainer` kind:
```cpp
const auto &Container = Ctx.getReadOnlyContainer<FileContainer>(BCRR);
```
Whereas, e.g., here:
```cpp
auto AttrMatcher = PM::create<Binary>("/Functions/*/Attributes").value();
```
It is necessary to make `check-conventions` happy. For the same, if the type returns a long `std::optional`, I needed to opt for `auto`. All the other cases are continuously being reviewed in order not to abuse `auto`.

lib/Pipes/TaggedFunctionKind.cpp
antoniofrighetto: I realize it, but the regex `"/Functions/*/Entry"` used to work correctly to match an addition and a removal until not a long ago. That said, since not correct, I had changed it. The approach used now should be correct.

lib/Pipes/TaggedFunctionKind.cpp
antoniofrighetto: Per above.

lib/Pipes/TaggedFunctionKind.cpp
antoniofrighetto: I don't understand. `CABIFunctionType` does not come with `FinalStackOffset`, the user cannot even change it, hence, why would we ever want to pass to `Layout` and check `FinalStackOffset` for `CABIFunctionType` prototypes?

lib/Pipes/TaggedFunctionKind.cpp
drblallo: sent to antonio the fix

lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -49,18 +55,223 @@ TaggedFunctionKind::symbolToTarget(const llvm::Function &Symbol) const {
...
+  if (not llvm::isa<TupleTreeGlobal<Binary>>(Before))
+    return;
+
+  const auto &OldModel = cast<TupleTreeGlobal<Binary>>(Before).get();
+  const auto &Model = cast<TupleTreeGlobal<Binary>>(After).get();
+
+  // If CrossRelations is not found, the most conservative approach is to
+  // invalidate all targets.
+  if (not Ctx.containsReadOnlyContainer(BinaryCrossRelationsRole)) {
+    ToRemove.emplace_back(Target(*this));
```
aleclearmind, line 121: Does this create a `*` target?
`*` is going to disappear soon, insert all the functions by hand.

lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -49,18 +55,223 @@ TaggedFunctionKind::symbolToTarget(const llvm::Function &Symbol) const {
...
+    }
+  }
+
+  // Because matching e.g., FinalStackOffset, for both RawFunctionType and
+  // CABIFunctionType is non-trivial, we will not leverage the diff for checking
+  // the change related to FSO; instead, we pass to `Layout` by employing the
+  // old model, and the new one (the one with the change).
+  for (const model::Function &NewFunction : Model->Functions) {
+    MetaAddress Entry = NewFunction.Entry;
+    const auto &OldFunction = OldModel->Functions.at(Entry);
```
aleclearmind, line 214: The function might have been removed. You can't use `.at`.

lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -49,18 +55,223 @@ TaggedFunctionKind::symbolToTarget(const llvm::Function &Symbol) const {
...
+        insertCallers(CrossRelations, Entry, Targets);
+      }
+    }
+
+    // A new dynamic function has been added or removed?
+    if (pathAsString<Binary>(Change.Path) == "/ImportedDynamicFunctions") {
+      if (Change.New) {
+        // Invalidate all targets
+        for (const model::Function &Function : Model->Functions)
+          Targets.insert(Function.Entry.toString());
```
aleclearmind, line 199: Postpone `.toString` for the final step in which you build the targets.

lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -49,18 +55,223 @@ TaggedFunctionKind::symbolToTarget(const llvm::Function &Symbol) const {
...
+      } else if (Change.Old) {
+        // Invalidate the function and its callers
+        auto Function = std::get<model::Function>(*Change.Old);
+        Targets.insert(Function.Entry.toString());
+        insertCallers(CrossRelations, Function.Entry, Targets);
+      }
```
aleclearmind, line 153:
```suggestion
      } else {
        revng_abort();
      }
```

lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -49,18 +55,223 @@ TaggedFunctionKind::symbolToTarget(const llvm::Function &Symbol) const {
...
+    }
+
+    // A new prototype has been added or removed?
+    if (pathAsString<Binary>(Change.Path) == "/Types") {
+      if (Change.New) {
+        // Invalidate all targets
+        for (const model::Function &Function : Model->Functions)
+          Targets.insert(Function.Entry.toString());
```
aleclearmind, line 161: Why should we invalidate all the functions when a new type is added?

lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -49,18 +55,223 @@ TaggedFunctionKind::symbolToTarget(const llvm::Function &Symbol) const {
...
+            insertCallers(CrossRelations, Function.Entry, Targets);
+          }
+        }
+
+        for (const model::DynamicFunction &Function : IDF) {
+          if (Prototype == Function.Prototype) {
+            // Invalidate callers of the model::DynamicFunction
+            insertCallers(CrossRelations, Function.OriginalName, Targets);
+          }
+        }
```
aleclearmind, line 178: Given what we do below, I don't think this is necessary.
Even if the type is completely changed, as long as the FSO and set of preserved registers is the same, we don't need to invalidate.

lib/Pipes/TaggedFunctionKind.cpp
antoniofrighetto: It should be OK now, we're invalidating all the targets (by specifying only the rank), sounds good?

lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -49,18 +55,223 @@ TaggedFunctionKind::symbolToTarget(const llvm::Function &Symbol) const {
...
+
+    // Attributes have changes?
+    auto MaybeChangedAttributes = AttrMatcher.match<MetaAddress>(Change.Path);
+    if (MaybeChangedAttributes) {
+      auto Entry = std::get<MetaAddress>(*MaybeChangedAttributes);
+      using FunctionAttribute = model::FunctionAttribute::Values;
+      bool NewAttribute = std::get<FunctionAttribute>(*Change.New);
+      if (NewAttribute == FunctionAttribute::NoReturn) {
+        // Invalidate callers
+        insertCallers(CrossRelations, Entry, Targets);
```
aleclearmind, line 190: Why not invalidate the function itself? Comment.

lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -49,18 +55,223 @@ TaggedFunctionKind::symbolToTarget(const llvm::Function &Symbol) const {
...
+      continue;
+
+    auto Old = abi::FunctionType::Layout::make(OldFunction.Prototype);
+    auto New = abi::FunctionType::Layout::make(NewFunction.Prototype);
+
+    if (Old.FinalStackOffset != New.FinalStackOffset
+        or Old.CalleeSavedRegisters != New.CalleeSavedRegisters) {
+      Targets.insert(Entry.toString());
+      insertCallers(CrossRelations, Entry, Targets);
+    }
```
aleclearmind, line 226: Factor out common code with the part below.

lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -49,18 +55,223 @@ TaggedFunctionKind::symbolToTarget(const llvm::Function &Symbol) const {
...
+        // Invalidate all targets
+        for (const model::Function &Function : Model->Functions)
+          Targets.insert(Function.Entry.toString());
+      } else if (Change.Old) {
+        // Invalidate callers of the model::DynamicFunction
+        for (const model::DynamicFunction &Function : IDF)
+          insertCallers(CrossRelations, Function.OriginalName, Targets);
+      }
+    }
+  }
```
aleclearmind, line 206: I think we want to do this for regular functions too.
You probably got it mixed up with the types and functions above.

aleclearmind (changes requested):

lib/Pipes/TaggedFunctionKind.cpp
antoniofrighetto: We invalidate all targets if we have an addition, or invalidate the function itself (not needed for if we have a `model::DynamicFunction` & its callers. This is what is happening currently both for `model::Function` and `model::DynamicFunction`. I didn't understand what is making you perplexed.

lib/Pipes/TaggedFunctionKind.cpp
antoniofrighetto: Uhmm, right, good catch, thanks. I drop it here, and likely reuse it for `ABIEnforced` Kind.

lib/Pipes/TaggedFunctionKind.cpp
antoniofrighetto: Wait sorry, I still have the function in the old model, no? Otherwise, what's the purpose of checking `FSO` & stuff if I do not have the function?

lib/Pipes/TaggedFunctionKind.cpp
antoniofrighetto: Indeed, we do not have too. Removed that part and kept it for `ABIEnforced` Kind, if anything.

lib/Pipes/TaggedFunctionKind.cpp
antoniofrighetto: Sounds good.

lib/Pipes/TaggedFunctionKind.cpp
aleclearmind: See the other comment

lib/Pipes/TaggedFunctionKind.cpp
aleclearmind: Uhm, no, I don't see why adding a new type should trigger an invalidation in general.

lib/Pipes/TaggedFunctionKind.cpp
aleclearmind: OK, put them close together, they are almost the same code.

antoniofrighetto: Sure.

antoniofrighetto:
Requires [`revng-qa#11`](https://github.com/revng/revng-qa/pull/11).

aleclearmind (changes requested):

aleclearmind:
NOTETOSELF: resume reviewing from "Pipeline: refine overestimated targets " included.

lib/Yield/CMakeLists.txt
```diff
@@ -85,3 +81,17 @@ target_tuple_tree_generator(
...
+  revngYieldPipes
+  SHARED
+  Pipes/AssemblyPipes.cpp
+  Pipes/CallGraphPipes.cpp
+  Pipes/CFGPipes.cpp)
+
+target_link_libraries(
+  revngYieldPipes
+  revngYield
+  revngFunctionIsolation
```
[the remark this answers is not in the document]
Cvelth: Yep, as far as I see now, `CrossRelations` only depends on `yield::Graph` (outside of the `model` and `efa` dependencies) so this should not be a problem.

My only concern is what I should rename the `yield::Graph` to then.

lib/Pipeline/Runner.cpp
```diff
@@ -383,23 +392,39 @@ const KindsRegistry &Runner::getKindsRegistry() const {
...
+                                  const Global &Before,
+                                  const Global &After,
                                   InvalidationMap &Map) const {
+  TargetsList OverestimatedTargets;
+  for (const Kind &Rule : getKindsRegistry())
+    Rule.getInvalidations(getContext(),
+                          OverestimatedTargets,
+                          Diff,
+                          Before,
+                          After);
```
aleclearmind, line 404:
```suggestion
  for (const Kind &Kind : getKindsRegistry()) {
    Kind.getInvalidations(getContext(),
                          OverestimatedTargets,
                          Diff,
                          Before,
                          After);
  }
```

include/revng/Pipeline/Option.h
```diff
@@ -48,15 +48,27 @@ template<typename T, size_t Index>
...
+template<typename InvokableType, size_t Index>
+constexpr bool IsConstCharPtr = std::is_same_v<OptionTypeImpl<InvokableType,
+                                                              Index>,
+                                               const char *>;
+
+template<typename InvokableType, size_t Index>
+using OptionType = std::conditional_t<IsConstCharPtr<InvokableType, Index>,
+                                      std::string,
+                                      OptionTypeImpl<InvokableType, Index>>;
+// clang-format on
```
aleclearmind, line 63: Why `clang-format off`?

include/revng/Pipeline/CLOption.h
```diff
@@ -73,8 +73,14 @@ class CLOptionWrapper : public CLOptionBase {
...
-  bool isSet() const override { return not Option.isDefaultOption(); }
-  llvm::StringRef get() const override { return Option.ValueStr; }
+  bool isSet() const override { return Option.getNumOccurrences() != 0; }
+  std::string get() const override {
+    std::string ToReturn;
+    llvm::raw_string_ostream SO(ToReturn);
+    SO << Option.getValue();
+    SO.flush();
+    return ToReturn;
+  }
```
aleclearmind, line 83: Commit message is too vague

lib/Yield/CrossRelations.cpp
```diff
@@ -143,19 +155,23 @@ yield::Graph yield::CrossRelations::toYieldGraph() const {
...
-
-    auto *FromNode = LookupHelper.at(std::get<0>(FromKey.value()));
-    FromNode->addSuccessor(LookupHelper.at(std::get<0>(ToKey.value())));
+      revng_assert(FromKey.has_value() && ToKey.has_value());
+
+      auto *FromNode = LookupHelper.at(std::get<0>(FromKey.value()));
+      auto *ToNode = LookupHelper.at(std::get<0>(ToKey.value()));
+      if (!llvm::is_contained(FromNode->successors(), ToNode))
+        FromNode->addSuccessor(ToNode);
+    }
```
aleclearmind, line 174: Very vague commit message

lib/Yield/Pipes/CallGraphPipes.cpp
```diff
@@ -35,7 +35,10 @@ void ProcessCallGraph::run(pipeline::Context &Context,
   SortedVector<efa::FunctionMetadata> Metadata;
   for (const auto &LLVMFunction : FunctionTags::Isolated.functions(&Module))
     Metadata.insert(*extractFunctionMetadata(&LLVMFunction));
-  revng_assert(Metadata.size() == Model->Functions.size());
+
+  // If some functions are missing, do not output anything
+  if (Metadata.size() != Model->Functions.size())
+    return;
```
aleclearmind, line 41: Commit message is very vague. If you got these commits from other people, please ask them to provide an accurate commit message.

tools/pipeline/print-invalidations/Main.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include "revng/Pipeline/LLVMGlobalKindBase.h"
+#include "revng/Pipeline/Loader.h"
+#include "revng/Pipeline/Step.h"
+#include "revng/Pipeline/Target.h"
+#include "revng/Pipes/ModelGlobal.h"
+#include "revng/Pipes/PipelineManager.h"
+#include "revng/Pipes/ToolCLOptions.h"
+#include "revng/Support/CommandLine.h"
+#include "revng/Support/InitRevng.h"
+#include "revng/TupleTree/TupleTree.h"
```
aleclearmind, line 35: I doubt we really need all of these includes. Please minimize the list.

tools/pipeline/print-invalidations/Main.cpp
```diff
@@ -0,0 +1,88 @@
...
+  Diff.apply(AfterModel->get(), EL);
+  cantFail(std::move(EL));
+
+  TargetsList OverestimatedTargets;
+  for (const Kind &Rule : Manager.getRunner().getKindsRegistry())
+    Rule.getInvalidations(Manager.context(),
+                          OverestimatedTargets,
+                          Diff,
+                          OriginalModel,
+                          *AfterModel);
```
aleclearmind, line 83: Drop all references to "Rule". Leftover of old naming.

tools/pipeline/print-invalidations/Main.cpp
```diff
@@ -0,0 +1,88 @@
...
+  const auto &Ctx = Manager.context();
+  auto OriginalModel = *AbortOnError(Ctx.getGlobal<BinaryRef>(ModelGlobalName));
+
+  const auto &Name = ModelGlobalName;
+  auto AfterModel = AbortOnError(Ctx.getGlobal<BinaryRef>(Name));
+
+  auto Diff = AbortOnError(deserializeFileOrSTDIN<DiffRef>(DiffPath));
+  revng::ErrorList EL;
+  Diff.apply(AfterModel->get(), EL);
+  cantFail(std::move(EL));
```
aleclearmind, line 75: What's going on here?

tools/pipeline/print-invalidations/Main.cpp
```diff
@@ -0,0 +1,88 @@
...
+  ParseCommandLineOptions(argc, argv);
+
+  Registry::runAllInitializationRoutines();
+
+  auto Manager = AbortOnError(BaseOptions.makeManager());
+  const auto &Ctx = Manager.context();
+  auto OriginalModel = *AbortOnError(Ctx.getGlobal<BinaryRef>(ModelGlobalName));
+
+  const auto &Name = ModelGlobalName;
+  auto AfterModel = AbortOnError(Ctx.getGlobal<BinaryRef>(Name));
```
aleclearmind, line 70: Why are we model-specific? Can we make this tool non model-specific?

include/revng/FunctionIsolation/IsolationFunctionKind.h
```diff
@@ -0,0 +1,188 @@
...
+
+inline IsolatedFunctionKind
+  Isolated("Isolated", &ranks::Function, FunctionTags::Isolated);
+
+inline TaggedFunctionKind
+  ABIEnforced("ABIEnforced", &ranks::Function, FunctionTags::ABIEnforced);
+
+using yield::CrossRelations;
+
+// Find callers of a given `model::Function` via CrossRelations
```
aleclearmind, line 59: Is there a real reason for functions below to be in a public header?

aleclearmind (changes requested):

tools/pipeline/Main.cpp
```diff
@@ -178,12 +178,23 @@ int main(int argc, const char *argv[]) {
...
+    auto &Global(*AbortOnError(Manager.context().getGlobals().get(ModelName)));
+
+    auto &Model = llvm::cast<TupleTreeGlobal<model::Binary>>(Global);
+    const auto Before = Model.clone();
+
     using Type = TupleTreeDiff<model::Binary>;
     auto Diff = AbortOnError(deserializeFileOrSTDIN<Type>(ApplyModelDiff));
+    revng::ErrorList EL;
+    Model.applyDiff(Diff, EL);
+    cantFail(std::move(EL));
```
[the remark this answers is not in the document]
drblallo: not sure what you mean. It is just asserting that diffs do not fail

include/revng/Pipes/TupleTreeContainer.h
```diff
@@ -0,0 +1,176 @@
...
+    Content = std::move(Other.Content);
+    return *this;
+  }
+
+  ~TupleTreeContainer() override = default;
+
+  std::unique_ptr<pipeline::ContainerBase>
+  cloneFiltered(const pipeline::TargetsList &Container) const final {
+    if (not Container.contains(pipeline::Target(*K)))
+      return std::make_unique<TupleTreeContainer>(this->name());
```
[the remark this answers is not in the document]
antoniofrighetto: If we do not use `this` pointer to access `name` base class method, we need to accomplish it through direct invocation (`pipeline::ContainerBase::name()`). Isn't the first case preferred, here?

include/revng/TupleTree/TupleTreePath.h
```diff
@@ -130,6 +135,21 @@ class ConcreteTupleTreeKeyWrapper : public TupleTreeKeyWrapper {
     }
   }
 
+  bool matches(const TupleTreeKeyWrapper &Other) const override {
```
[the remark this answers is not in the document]
antoniofrighetto: It definitely is. Must have inadvertently manually taken it and not cherry-picked it this time. Fixed, apologize for the inconvenience.

include/revng/Pipeline/Context.h
```diff
@@ -103,7 +103,7 @@ class Context {
 
   bool containsReadOnlyContainer(llvm::StringRef Name) const {
     return ReadOnlyContainers.count(Name)
-           and ReadOnlyContainers.find(Name)->second != nullptr;
+           and ReadOnlyContainers.find(Name)->second->second != nullptr;
   }
```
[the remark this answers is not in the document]
antoniofrighetto: Improved message for the bug fix.

lib/Yield/CrossRelations.cpp
antoniofrighetto: Improved message with better description and split it into two different commits.

include/revng/Pipeline/Kind.h
```diff
@@ -64,7 +65,9 @@ class Kind : public DynamicHierarchy<Kind> {
 
   virtual void getInvalidations(const Context &Ctx,
                                 pipeline::TargetsList &ToRemove,
-                                const GlobalTupleTreeDiff &Diff) const {}
+                                const GlobalTupleTreeDiff &Diff,
+                                const Global &Before,
+                                const Global &After) const {}
```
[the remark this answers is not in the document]
antoniofrighetto: We use `After` (Model on which the diff has been applied) in `IsolatedKind::getInvalidations` to perform the difference between the current functions of the model and the old one, for FSO/PreservedRegisters.

include/revng/FunctionIsolation/IsolationFunctionKind.h
antoniofrighetto: The idea was to expose those functions in a public header (but hide the inner implementation inside `detail` namespace) in order to make them available for `IsolatedFunctionKind.cpp` and for the upcoming `ABIEnforcedKind.cpp`.

tools/pipeline/print-invalidations/Main.cpp
antoniofrighetto: `AbortOnError` was replaced with `cantFail`, after the introduction of `ErrorList`, and more specifically, after [this](https://github.com/revng/revng/commit/0b80801b2b75b3584b811148994a16dc4f184bf2) jackv's change, whose `apply` method was modified to accept a `ErrorList` too.

include/revng/Pipes/TupleTreeContainer.h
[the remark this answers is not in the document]
aleclearmind: What stops you from just invoking `name()`?

```c++
struct X {
    void lel();
};

struct Y : public X {
   void mmm() { lel(); }
};
```

tools/pipeline/Main.cpp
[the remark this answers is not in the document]
aleclearmind: Gotcha. A couple of questions though:
1. Will it print the error messages within `ErrorList` nicely?
2. Is `cantFail` dropped in release?

lib/Yield/CMakeLists.txt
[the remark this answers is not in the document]
aleclearmind: Its core property seems to a graph where nodes have a position in the plain. `LaidOutGraph`?

tools/pipeline/print-invalidations/Main.cpp
aleclearmind: Some concerns has [here](https://github.com/revng/revng/pull/209#discussion_r1021809669).

tools/pipeline/Main.cpp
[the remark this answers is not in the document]
drblallo: 1) i think it will print just the first error
2) no

include/revng/FunctionIsolation/IsolationFunctionKind.h
aleclearmind: This might be true for some of them, but not all. Please review what you have here.
For those you think will be useful in other invalidations, create a new header with an appropriate name.
Also consider having a `.cpp` file for implementations.

include/revng/Pipes/TupleTreeContainer.h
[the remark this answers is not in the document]
antoniofrighetto: Right, however there must be some clash with names here, as I get:
```
/root/orchestra/sources/revng/include/revng/Pipes/TupleTreeContainer.h:94:51: error: use of undeclared identifier 'name'; did you mean 'LoadModelAnalysis::name'?
      return std::make_unique<TupleTreeContainer>(name());
                                                  ^~~~
                                                  LoadModelAnalysis::name
/root/orchestra/root/include/llvm/IR/PassManager.h:376:20: note: 'LoadModelAnalysis::name' declared here
  static StringRef name() {
                   ^
1 error generated.
```

lib/Yield/CMakeLists.txt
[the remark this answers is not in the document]
Cvelth: I guess that would work, but I'm not excited. Since the graph does not get created in `LaidOff` state, you need to run a `Layouter` separately. This could lead to confusion.

Also, where's the boundary we're cutting stuff on? I feel like wherever the graph goes, the layouter should go too, but I guess that's it.

Should Graph + layouters + relations should be a single library? Or two distinct ones (as in, graph stuff has nothing to do with the relations)?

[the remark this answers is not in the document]
aleclearmind: I feel like we should decouple the emission of graphs from the laying out.

Something like:
```c++
GenericGraph emitCrossRelations();
GenericGraph emitCFG();
yield::Graph layOut(GenericGraph);
```
Is there something preventing us to go this route?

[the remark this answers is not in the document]
Cvelth: This would make the layout function super-specialized. But maybe that's not such a bad thing, tbh.

But I'd still prefer for both graphs to be explicitly named so it's clear what they are (of course they are both `GenericGraph`s).

tools/pipeline/print-invalidations/Main.cpp
drblallo: yeah i guess when we will have the need to do so we can change the signature of the tool, but there is not great need at the moment.

include/revng/FunctionIsolation/IsolationFunctionKind.h
antoniofrighetto: Sounds good: I added `InvalidationHelpers.h` under `Pipes` and moved those functions not needed to be exposed in the implementation file.

share/revng/pipelines/isolate-translate.yml
```diff
@@ -2,6 +2,10 @@
 # This file is distributed under the MIT License. See LICENSE.md for details.
 #
 
+Containers:
+  - Name: cross-relations.yml
+    Type: BinaryCrossRelations
+    Role: CrossRelations
```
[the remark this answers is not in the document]
antoniofrighetto: Closing this thread, as there's one already ongoing on the same concern.

lib/Yield/CMakeLists.txt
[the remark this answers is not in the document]
aleclearmind: Can you link the pad?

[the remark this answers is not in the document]
Cvelth: I linked it in `~asm-and-graphs` header, but sure, here it is: https://pad.rev.ng/s/S3VedMpAo

lib/FunctionIsolation/IsolatedFunctionKind.cpp
```diff
@@ -0,0 +1,184 @@
...
+    // changed?
+    // Note that, since matching e.g., FinalStackOffset, for both
+    // RawFunctionType and CABIFunctionType is non-trivial, we will not leverage
+    // the diff for checking the change related to FSO. Instead, we move to
+    // `Layout` by employing the old model, and the new one (the one with the
+    // change).
+    for (const model::Function &Function : Model->Functions) {
+      MetaAddress Entry = Function.Entry;
+      if (OldModel->Functions.find(Entry) != OldModel->Functions.end())
+        if (haveFSOAndPreservedRegsChanged(OldModel->Functions.at(Entry)
```
aleclearmind, line 118: Don't use `.at`, use `auto It = OldModel->Functions.find(Entry)`.

lib/FunctionIsolation/IsolatedFunctionKind.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+    // A new `model::Function` has been added or removed?
+    if (pathAsString<Binary>(Change.Path) == "/Functions") {
+      if (Change.New) {
+        // Invalidate all targets
+        invalidateAllTargetsPerFunctionRank(Model, Targets);
+      } else if (Change.Old) {
+        // Invalidate function's callers
+        auto Function = std::get<model::Function>(*Change.Old);
+        insertCallers(CrossRelations, Function.Entry, Targets);
```
aleclearmind, line 88: Why not `insertCallersAndTransitiveClosureIfInline`?

lib/FunctionIsolation/IsolatedFunctionKind.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+    // A new `model::DynamicFunction` has been added or removed?
+    if (pathAsString<Binary>(Change.Path) == "/ImportedDynamicFunctions") {
+      if (Change.New) {
+        // Invalidate all targets
+        invalidateAllTargetsPerFunctionRank(Model, Targets);
+      } else if (Change.Old) {
+        // Invalidate model::DynamicFunction's callers
+        auto DynamicFunction = std::get<model::DynamicFunction>(*Change.Old);
+        insertCallers(CrossRelations, DynamicFunction.OriginalName, Targets);
```
aleclearmind, line 102: As above

lib/FunctionIsolation/IsolatedFunctionKind.cpp
```diff
@@ -0,0 +1,184 @@
...
+                                                    Name,
+                                                    Targets);
+      }
+    }
+
+    // Did attributes for a `model::Function` have changed?
+    auto MaybeFuncAttr = FuncAttrMatcher.match<MetaAddress>(Change.Path);
+    if (MaybeFuncAttr) {
+      auto Entry = std::get<MetaAddress>(*MaybeFuncAttr);
+      auto NewAttribute = std::get<FunctionAttribute>(*Change.New);
```
aleclearmind, line 149: You're assuming attributes can only be removed, it seems.

lib/FunctionIsolation/IsolatedFunctionKind.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+      // If the attribute of a function changes to `NoReturn`, we need to
+      // invalidate its callers (not the function itself as it would not provide
+      // any further information to the function itself).
+      if (NewAttribute == FunctionAttribute::NoReturn)
+        insertCallers(CrossRelations, Entry, Targets);
```
aleclearmind, line 155: Why not `insertCallersAndTransitiveClosureIfInline`?

include/revng/Pipes/InvalidationHelpers.h
```diff
@@ -0,0 +1,152 @@
...
+}
+
+namespace detail {
+
+inline void
+insertCallersAndTransitiveClosureInternal(const TupleTree<CrossRelations> &CR,
+                                          const TupleTree<model::Binary> &Model,
+                                          const std::string &Location,
+                                          std::set<MetaAddress> &Result) {
+  auto CRGraph = CR->toCallGraph();
```
aleclearmind, line 84: You need to hoist this so you do it only once per call to `getInvalidation`.

include/revng/Pipes/InvalidationHelpers.h
```diff
@@ -0,0 +1,152 @@
...
+
+  revng_assert(EntryNode != nullptr);
+
+  // Inverse DFS visit that aims at finding its callers, and its
+  // transitive closure, if they are all `Inline`.
+  llvm::df_iterator_default_set<const CRNode *> NoInlineFunctions;
+  auto VisitRange = llvm::inverse_post_order_ext(EntryNode, NoInlineFunctions);
+
+  for (const CRNode *Node : VisitRange) {
+    namespace p = pipeline;
```
aleclearmind, line 100: Mmmmh, avoid this.

include/revng/Pipes/InvalidationHelpers.h
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    auto Address = p::genericLocationFromString<0>(Node->data(),
+                                                   ranks::Function,
+                                                   ranks::BasicBlock,
+                                                   ranks::Instruction);
+    if (Address.has_value())
+      Result.insert(std::get<MetaAddress>(*Address));
+  }
+}
```
aleclearmind, line 123: Can't you simply pre-populate `NoInlineFunctions` with all the functions that are not inline (except for the starting point)?

share/revng/test/configuration/revng/for-invalidation.yml
```diff
@@ -0,0 +1,98 @@
...
+        --resume "$$TMP_FOLDER"
+        --produce Isolate/module.ll/*:Isolated;
+      for FILE in ${SOURCE}.overrides/*.yml; do
+        revng model override-by-name "$$BASELINE_MODEL_OVERRIDEN" "$$FILE"
+          | ./remove-candidate-functions-from-model.py /dev/stdin "$$FILE"
+          | (revng model diff "$$BASELINE_MODEL_OVERRIDEN" /dev/stdin || true) > "$$DIFF";
+        revng print-invalidations --resume "$$TMP_FOLDER" "$$DIFF"
+          | ./metaaddress-to-customname.py "$$BASELINE_MODEL_OVERRIDEN" /dev/stdin
+          | diff -B /dev/stdin <(grep -vE '^\s*#' "$$FILE".invalidations);
+      done
```
aleclearmind, line 37: Thoroughly comment what's going on, it's quite scary to look at as it is.

aleclearmind (changes requested):

include/revng/Pipes/InvalidationHelpers.h
antoniofrighetto: Yeah, good catch, missed it.

lib/FunctionIsolation/IsolatedFunctionKind.cpp
antoniofrighetto: As we have said, when removing a function, querying the `Relations` fields of `BinaryCrossRelations` should suffice in order to retrieve the callers, there should be no need to traverse the graph.

lib/FunctionIsolation/IsolatedFunctionKind.cpp
antoniofrighetto: You mean I'm assuming attributes can only be added?

lib/FunctionIsolation/IsolatedFunctionKind.cpp
aleclearmind: If a caller is inline, we need to recompute its callers too.

lib/FunctionIsolation/IsolatedFunctionKind.cpp
aleclearmind: Yep

include/revng/Pipes/InvalidationHelpers.h
antoniofrighetto: What's the point? It should be way more optimal to rely on `inverse_post_order_ext` to both traversal and filtering then first manually gathering non-inline functions (iterating over all the functions) and then do a conditional traversal. At least, that ought to be profiled to confirm which one of the two approaches is faster.

antoniofrighetto: If `Inline` attribute is rare, it might be better to just gather a set of all the inline functions instead of listing the non-inline ones, and then the inner loop would just be a lookup on that set?

aleclearmind: We're not discussing performance, we're discussing simplicity.
Using `_ext` without manually changing the set of visited nodes in the loop is much more intuitive and simpler to explain: "we mark all non-inline nodes as visited and then perform a simple inverse dfs".

If you're concerned about performance, you can reuse the initial set among various invocations of this method (copying it).

antoniofrighetto: Following your direction, pre-populating `NoInlineFunctions` implies that, no sooner `inverse_post_order_ext ` later sees a node that is already in the set, it abandons that branch of the traversal, not exploring further. If the callers of the entry function are added to the set `NoInlineFunctions`, the inverse POT won't visit them. This means we end up getting neither the transitive closure nor the callers. To make it work, we would need to use the complementary set, namely, something like:
```diff
+  llvm::df_iterator_default_set<const CRNode *> AnotherSet;
-  auto VisitRange = llvm::inverse_post_order_ext(EntryNode, NoInlineFunctions);
+  auto VisitRange = llvm::inverse_post_order_ext(EntryNode, AnotherSet);

  for (const CRNode *Node : VisitRange) {
+    if (!llvm::is_contained(Node, NoInlineFunctions)) {
+      AnotherSet.emplace(Node);
+      continue;
+    }
    using namespace pipeline;
    namespace ranks = revng::ranks;

    auto Address = genericLocationFromString<0>(Node->data(),
                                                ranks::Function,
                                                ranks::BasicBlock,
                                                ranks::Instruction);
    if (Address.has_value())
      Result.insert(std::get<MetaAddress>(*Address));
  }
}
```
This seemingly adds further complexity. Personally, I would refrain from doing such a change.

aleclearmind: It's more straightforward to read as opposed to "peeking" into successor's list while visiting nodes.
Also, you can pre-create once and for all the set of non-`Inline` functions and just copy it every time, so step 2) is not even there.

I already expressed these observations [in a previous comment](https://github.com/revng/revng/pull/209#discussion_r1029355688).

include/revng/Pipeline/CLOption.h
aleclearmind: The commit still does not address this part of the change.

lib/Yield/Pipes/CallGraphPipes.cpp
```diff
@@ -35,7 +35,10 @@ void ProcessCallGraph::run(pipeline::Context &Context,
   SortedVector<efa::FunctionMetadata> Metadata;
   for (const auto &LLVMFunction : FunctionTags::Isolated.functions(&Module))
     Metadata.insert(*::detail::extractFunctionMetadata(&LLVMFunction));
-  revng_assert(Metadata.size() == Model->Functions().size());
+
+  // If some functions are missing, do not output anything
+  if (Metadata.size() != Model->Functions().size())
+    return;
```
aleclearmind, line 41: Uhm, this doesn't seem right.

In order to produce the call graph we clearly need all of the functions or fail.

include/revng/TupleTree/TupleTreeDiff.h
```diff
@@ -514,6 +514,7 @@ TupleTreeDiff<T>::apply(TupleTree<T> &M, revng::ErrorList &EL) const {
       continue;
     }
     tupletreediff::detail::ApplyDiffVisitor<T> ADV{ &C, &EL };
+    M.evictCachedReferences();
```
aleclearmind, line 517: I'm confused, why do this at every iteration of the loop? And why before declaring `ADV` and using it?

Can we do it once at the beginning? Or even better, not do it at all.

I think we could push `evictCachedReferences` into `initializeReferences`.

lib/Yield/CrossRelations.cpp
```diff
@@ -83,11 +83,8 @@ static void conversionHelper(const yield::CrossRelations &Input,
     for (const auto &[RelationKind, TargetString] : Related) {
       switch (RelationKind) {
       case yield::RelationType::IsCalledFrom:
-        AddEdge(LocationString, TargetString, RelationKind);
-        break;
-
-      // TODO: handle dynamic function calls properly
       case yield::RelationType::IsDynamicallyCalledFrom:
+        AddEdge(LocationString, TargetString, RelationKind);
         break;
```
aleclearmind, line 94: * No fixup commits. "TODO: handle dynamic function calls properly" was introduced in an earlier commit.
* Do we still have "handle dynamic function calls properly" or are they all handled?

> location-dependent `toCallGraph`

The commit message is a bit confusing: `s/location-dependent/location-based/`.

> some minor call graph specific improvements

Vague

include/revng/EarlyFunctionAnalysis/CallEdge.h
```diff
@@ -46,15 +46,18 @@ class efa::CallEdge : public efa::generated::CallEdge {
...
+  using generated::CallEdge::CallEdge;
   CallEdge() : efa::generated::CallEdge() { Type() = AssociatedType; }
-
-  CallEdge(MetaAddress Destination, FunctionEdgeType::Values Type) :
-    efa::generated::CallEdge(Destination, Type) {}
+  CallEdge(MetaAddress Destination, efa::FunctionEdgeType::Values Type) :
+    efa::generated::CallEdge() {
+    this->Destination() = Destination;
+    this->Type() = Type;
+  }
```
aleclearmind, line 55: Why?
What's going on in this commit?

include/revng/Pipeline/Location.h
```diff
@@ -176,6 +172,18 @@ locationFromString(const Rank &, std::string_view String) {
...
 
+/// A helper interface for location conversion.
+///
+/// It discloses the static `convert` member in an easier-to-access fashion.
+template<typename ResultRank, typename InputRank>
+  requires(RankConvertibleTo<ResultRank, InputRank>)
+inline constexpr Location<ResultRank>
+convertLocation(const ResultRank &Result, const Location<InputRank> &Input) {
+  return Location<ResultRank>::convert(Input);
+}
```
aleclearmind, line 183: The argument is unused, this doesn't make much sense.

tools/pipeline/print-invalidations/Main.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+  auto Diff = AbortOnError(deserializeFileOrSTDIN<DiffRef>(DiffPath));
+  revng::ErrorList EL;
+  Diff.apply(AfterModel->get(), EL);
+  cantFail(std::move(EL));
+
+  TargetsList OverestimatedTargets;
```
aleclearmind, line 56: Why "overestimated"?

share/revng/test/tests/invalidation/invalidation.c.override.yml
```diff
@@ -0,0 +1,164 @@
...
+    Prototype: "/Types/RawFunctionType-12005182500180626567"
+  - CustomName: "main"
+    OriginalName: main
+    Prototype: "/Types/CABIFunctionType-17140048229552705196"
+ImportedDynamicFunctions:
+  - OriginalName: __libc_start_main
+    CustomName: "prefix___libc_start_main"
+    Prototype: "/Types/CABIFunctionType-3475954398804502039"
+    Relocations:
+      - Address: "0x402000:Generic64"
```
aleclearmind, line 27: * In tests we never ever commit hard-coded addresses.
* Drop all the stuff that you didn't write by hand such as `prefix___libc_start_main`

share/revng/test/tests/invalidation/invalidation.c.override.yml
```diff
@@ -0,0 +1,164 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+---
+Functions:
+  - CustomName: "bar"
+    OriginalName: bar
```
aleclearmind, line 8: No need to have both.

share/revng/test/tests/invalidation/invalidation.c.override.yml
```diff
@@ -0,0 +1,164 @@
...
+    Prototype: "/Types/RawFunctionType-13090241523659220176"
+  - CustomName: "foo"
+    OriginalName: foo
+    Prototype: "/Types/RawFunctionType-15357695279870196297"
+  - CustomName: "qux"
+    OriginalName: qux
+    Prototype: "/Types/RawFunctionType-6494433326490644482"
+  - CustomName: "baz"
+    OriginalName: baz
+    Prototype: "/Types/RawFunctionType-12005182500180626567"
```
aleclearmind, line 18: I guess each function is trying to test something. Put meaningful names depending on what they test.

share/revng/test/tests/invalidation/invalidation.c.overrides/CABIFunctionType-ABI.yml.invalidations
```diff
@@ -0,0 +1,5 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+plt___libc_start_main
```
aleclearmind, line 5: Are you actually testing `plt___libc_start_main`?
It seems to me that you're committing whatever output you got, which is not how writing ground truth works.

aleclearmind:
1. Double check commit messages, are they indented at 72?
2. "FunctionIsolation: execute CollectCFG earlier": why? Improve explanation in the commit message.
3. Move "revng.daemon: accept no arguments for analysis" before "[Introduce ApplyDiffAnalysis](https://github.com/revng/revng/pull/209/commits/9bda21f6cb8cff4fe3806a63330c3b91656d7e80)"
4. Postpone "Allow access to Global in invalidation" before "Introduce IsolatedKind::getInvalidations"
5. The testing is not right, it seems that you have basically committed the output of the tool. This is not how testing works.
  You should 1) decide what to test (e.g., dropping a preserved register), 2) design the smallest possible test to test that only that, 3) write the ground truth.
  Also, you need to document what you're trying to test, I don't see that in the source file or in the ground truth. I have no idea how to check if the ground truth makes sense or not.
  If you're hardcoding addresses, or your ground truth contains names that you didn't come up with (e.g., `libc_start_main`) there is for sure something wrong. Tests should not depend on names in the libc or addresses/offsets that can easily change with minor configuration changes/version update of the compiler/linker.

aleclearmind (changes requested):

aleclearmind:
1. Double check commit messages, are they indented at 72?
2. "FunctionIsolation: execute CollectCFG earlier": why? Improve explanation in the commit message.
3. Move "revng.daemon: accept no arguments for analysis" before "[Introduce ApplyDiffAnalysis](https://github.com/revng/revng/pull/209/commits/9bda21f6cb8cff4fe3806a63330c3b91656d7e80)"
4. Postpone "Allow access to Global in invalidation" before "Introduce IsolatedKind::getInvalidations"
5. The testing is not right, it seems that you have basically committed the output of the tool. This is not how testing works.
  You should 1) decide what to test (e.g., dropping a preserved register), 2) design the smallest possible test to test that only that, 3) write the ground truth.
  Also, you need to document what you're trying to test, I don't see that in the source file or in the ground truth. I have no idea how to check if the ground truth makes sense or not.
  If you're hardcoding addresses, or your ground truth contains names that you didn't come up with (e.g., `libc_start_main`) there is for sure something wrong. Tests should not depend on names in the libc or addresses/offsets that can easily change with minor configuration changes/version update of the compiler/linker.

include/revng/Pipeline/Location.h
Cvelth: (sorry for butting in into another person's PR, but)

It's exactly the same as the first argument of the function above it: this lets the caller to just pass `revng::ranks::Whatever` as an argument instead of trying to obtain its type (which is non-trivial) to pass as a template argument.

aleclearmind: I'd prefer to use `decltype` in the caller then having an unused argument

Cvelth: Are you sure you'd rather go
```cpp
auto NewLocation = pipeline::convertLocation<decltype(revng::ranks::Function)>(MyLocation);
```
over
```cpp
auto NewLocation = pipeline::convertLocation(revng::ranks::Function, MyLocation);
```

I'd say the second one is a lot easier to reason about.

aleclearmind:
The testing is not right, it seems that you have basically committed the output of the tool. This is not how testing works.
  You should 1) decide what to test (e.g., dropping a preserved register), 2) design the smallest possible test to test that only that, 3) write the ground truth.
  Also, you need to document what you're trying to test, I don't see that in the source file or in the ground truth. I have no idea how to check if the ground truth makes sense or not.
  If you're hardcoding addresses, or your ground truth contains names that you didn't come up with (e.g., `libc_start_main`) there is for sure something wrong. Tests should not depend on names in the libc or addresses/offsets that can easily change with minor configuration changes/version update of the compiler/linker.

include/revng/Pipeline/Location.h
aleclearmind: Fair, it's a bit misleading for the user, but until he doesn't open the hood it's all good.

lib/Yield/Pipes/CallGraphPipes.cpp
antoniofrighetto: It's simply checking if the number of metadata entries we have is the same as the number of functions; if not – the call graph is likely to end up broken – we fail. I remember we were previously hard-asserting this, but some issue either with the pipeline or the build tests prevented us from doing it.

include/revng/TupleTree/TupleTreeDiff.h
antoniofrighetto: IIRC pushing `evictCachedReferences` exactly before `callByPath` was decided jointly with @pfez, though I actually don't see why it couldn't be hoisted, in principle. It still needs to be done before `callByPath`, I guess. Maybe @pfez can suggest an improvement here?

include/revng/EarlyFunctionAnalysis/CallEdge.h
antoniofrighetto: @drblallo, for some reason, possibly due to some commit in `develop`, EFA CallGraph & yield CrossRelation got desync'd. Sync'ing back is the purpose of this commit.

lib/Yield/CrossRelations.cpp
antoniofrighetto: Took responsibility for this, as some git shenanigans were required, and wanted to the relieve the poor @drblallo from this. So, the "TODO: handle dynamic function calls properly" fixup was fixed by squashing that change into the commit that introduced it (Ivan's one), plus some further fix down in the commits (to reflect this). The commit message has also been improved.

tools/pipeline/print-invalidations/Main.cpp
drblallo: because kinds do not know what is the content of containers, so they overexistamate the ones that will really be killed, and provide instead all that may be killed.

lib/Pipeline/Runner.cpp
```diff
@@ -395,21 +395,27 @@ void Runner::getDiffInvalidations(const GlobalTupleTreeDiff &Diff,
                                   const Global &Before,
                                   const Global &After,
                                   InvalidationMap &Map) const {
+  TargetsList OverestimatedTargets;
```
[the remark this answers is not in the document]
drblallo: done, for some reason i can't mark this as resolved

tools/pipeline/print-invalidations/Main.cpp
drblallo: this will disappear with no error lists

include/revng/EarlyFunctionAnalysis/CallEdge.h
drblallo: i can't offer a explanation better than antonion's one

include/revng/FunctionIsolation/IsolationFunctionKind.h
```diff
@@ -0,0 +1,188 @@
...
+                        const pipeline::Global &After) const override;
+};
+
+inline IsolatedFunctionKind
+  Isolated("Isolated", &ranks::Function, FunctionTags::Isolated);
+
+inline TaggedFunctionKind
+  ABIEnforced("ABIEnforced", &ranks::Function, FunctionTags::ABIEnforced);
+
+using yield::CrossRelations;
```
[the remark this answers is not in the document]
drblallo: this issue can be closed

include/revng/TupleTree/TupleTreeDiff.h
drblallo: this can be closed

PR #210, opened 2022-08-08
pipeline analysis failure support

drblallo:
allow to return a llvm error from a analysis and have it propagate back all the way to main.

### DIFF
--- a/include/revng/Pipeline/Invokable.h
+++ b/include/revng/Pipeline/Invokable.h
@@ -29,8 +29,24 @@ concept HasName = requires() {
   { T::Name } -> convertible_to<const char *>;
 };
 
-// clang-format off
+template<typename InvokableType, typename... Args>
+constexpr bool
+invokableTypeReturnsErrorImpl(void (InvokableType::*F)(Args...)) {
+  return false;
+}
 
+template<typename InvokableType, typename... Args>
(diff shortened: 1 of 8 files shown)


include/revng/Pipeline/Invokable.h
```diff
@@ -29,8 +29,24 @@ concept HasName = requires() {
...
+invokableTypeReturnsErrorImpl(llvm::Error (InvokableType::*F)(Args...)) {
+  return true;
+}
 
+template<typename Invokable>
+constexpr bool invokableTypeReturnsError() {
+  return invokableTypeReturnsErrorImpl(&Invokable::run);
+}
+
+// clang-format off
```
aleclearmind, line 49: If I understand this correctly, we need this since we use the same machinery to invoke a pipe (returning void) and an analysis (returning `llvm::Error`).
I see from other changes you made that you're trying to let pipes return `llvm::Error` too.
Let's not do that. Pipes return `void` and never fail. Analysis always return `llvm::Error`.

aleclearmind (changes requested):

aleclearmind:
* `DetectABI` has not been changed to return `llvm::Error`, why?
* Do we have a sister PR for revng-c?

PR #213, opened 2022-08-22
Pipeline analysis arguments

drblallo:


include/revng/Pipeline/Invokable.h
```diff
@@ -13,23 +13,103 @@
...
+#include "revng/Pipeline/Container.h"
 #include "revng/Pipeline/ContainerSet.h"
 #include "revng/Pipeline/Context.h"
+#include "revng/Support/Assert.h"
 #include "revng/Support/Debug.h"
+#include "revng/Support/YAMLTraits.h"
 
 namespace pipeline {
 
+class CLOptionBase {
```
aleclearmind, line 33: This header is doing way too many things. Split it.

include/revng/Pipeline/Invokable.h
```diff
@@ -38,70 +118,264 @@ concept HasName = requires() {
...
-                InvokableType &Pipe,
-                void (InvokableType::*F)(const Context &, Args...),
-                ContainerSet &Containers,
-                const StringArrayRef &ArgsNames,
-                const std::integer_sequence<size_t, S...> &) {
-  using namespace std;
-  (Pipe.*F)(Ctx,
-            getContainerFromName<decay_t<Args>>(Containers, ArgsNames[S])...);
-}
+  template<typename Invokable, size_t Index>
```
aleclearmind, line 167: Why is this indented?
`clang-format` is behaving due to concepts.

include/revng/PipelineC/Prototypes.h
```diff
@@ -401,6 +403,27 @@ rp_analysis_get_argument_name(rp_analysis *analysis, int index);
...
+const char * /*owning*/
+rp_analysis_get_option_name(rp_analysis *analysis, int extra_argument_index);
+
+/**
+ * \return the type of the extra argument with the provided index. Returns
+ * nullptr if extra_argument_index < 0 or extra_argument_index >
+ * rp_analysis_get_extra_argument_count(analysis)
+ */
+const char * /*owning*/
+rp_analysis_get_option_type(rp_analysis *analysis, int extra_argument_index);
```
aleclearmind, line 425: You sure about `owning`? Can't we return a reference to a static string?

include/revng/PipelineC/Prototypes.h
```diff
@@ -658,3 +681,27 @@ rp_error_list_get_error_message(rp_error_list *error, uint64_t index);
...
 
 /** \} */
+
+/**
+ * \defgroup rp_string_map rp_string_map methods
+ * \{
+ */
+
+/**
+ * \returns a owing pointer to a new string map
```
aleclearmind, line 691: `s/returns/return/`

include/revng/PipelineC/Prototypes.h
```diff
@@ -658,3 +681,27 @@ rp_error_list_get_error_message(rp_error_list *error, uint64_t index);
...
+ * destroys the provided map
+ */
+void rp_string_map_destroy(rp_string_map *map);
+
+/**
+ * inserts the pair of key and value in the provided map
+ */
+void rp_string_map_insert(rp_string_map *map,
+                          const char *key,
+                          const char *value);
```
aleclearmind, line 705: Write only?

include/revng/Pipes/PipelineManager.h
```diff
@@ -149,10 +150,12 @@ class PipelineManager {
   runAnalysis(llvm::StringRef AnalysisName,
               llvm::StringRef StepName,
               const pipeline::ContainerToTargetsMap &Targets,
+              const llvm::StringMap<std::string> &ExtraArgs = {},
               llvm::raw_ostream *DiagnosticLog = nullptr);
 
   /// Run all analysis in reverse post order (that is: parents first),
-  llvm::Expected<pipeline::DiffMap> runAllAnalyses();
+  llvm::Expected<pipeline::DiffMap>
+  runAllAnalyses(const llvm::StringMap<std::string> &ExtraArgs = {});
```
aleclearmind, line 158: I'm confused by `Extra`. Also, elsewhere you call them options, not args.
I'd drop extra and pick one among arguments and options.

lib/PipelineC/PipelineC.cpp
```diff
@@ -804,3 +809,45 @@ void rp_error_list_destroy(rp_error_list *error_list) {
...
+  return new rp_string_map();
+}
+
+void rp_string_map_destroy(rp_string_map *map) {
+  delete map;
+}
+
+/**
+ * inserts the pair of key and value in the provided map
+ */
```
aleclearmind, line 847: Documentation is already in the header. Drop.

lib/Pipes/Pipes.cpp
```diff
@@ -155,6 +155,9 @@ class DetectABIAnalysis {
 public:
   static constexpr auto Name = "DetectABI";
 
+  static constexpr std::tuple ArgumentInfos = { ArgumentInfo("random_name",
+                                                             0) };
```
aleclearmind, line 159: The `Info` suffix is almost never a good idea. Drop or go for `ArgumentTypes`.
At very least go for `ArgumentsInfo`.

lib/Pipes/Pipes.cpp
```diff
@@ -165,7 +168,9 @@ class DetectABIAnalysis {
     Implementation.print(Ctx, OS, ContainerNames);
   }
 
-  void run(const pipeline::Context &Ctx, pipeline::LLVMContainer &Container) {
+  void run(const pipeline::Context &Ctx,
+           pipeline::LLVMContainer &Container,
+           bool RandomName) {
```
aleclearmind, line 173: What?

include/revng/Pipeline/Invokable.h
```diff
@@ -392,9 +393,10 @@ concept Printable = requires(InvokableType Pipe) {
 
 class InvokableWrapperBase {
 public:
-  virtual void run(Context &Ctx,
-                   ContainerSet &Containers,
-                   const llvm::StringMap<std::string> &ExtraArgs = {}) = 0;
+  virtual void 
+  run(Context &Ctx,
+      ContainerSet &Containers,
+      const llvm::StringMap<std::string> &ExtraArgs = {}) = 0;
```
aleclearmind, line 421: Uh? Same?
Did you run `orc check-branch`?

include/revng/Pipes/LiftPipe.h
```diff
@@ -39,7 +39,9 @@ class LiftPipe {
 
   void run(pipeline::Context &Ctx,
            const FileContainer &SourceBinary,
-           pipeline::LLVMContainer &TargetsList);
+           pipeline::LLVMContainer &Output);
+
+  llvm::Error precondition(const pipeline::Context &Ctx) const;
```
aleclearmind, line 44: Shouldn't we call this `checkPrecondition`?

lib/Pipeline/Runner.cpp
```diff
@@ -305,6 +305,12 @@ Error Runner::run(llvm::StringRef EndingStepName,
   if (ToExec.size() <= 1)
     return Error::success();
 
+  for (auto &StepGoalsPairs : llvm::drop_begin(ToExec)) {
+    auto &Step = *StepGoalsPairs.ToExecute;
+    if (auto Error = Step.precondition(getContext()); Error)
+      return Error;
```
aleclearmind, line 316: Shall we augment the error with step and pipe names?

include/revng/Pipeline/Invokable.h
```diff
@@ -367,16 +389,16 @@ void invokePipeFunction(Context &Ctx,
...
+  return detail::invokeImpl<ContextT>(Ctx,
+                                      Pipe,
+                                      F,
+                                      Containers,
+                                      ArgsNames,
+                                      static_cast<ContainersTypes *>(nullptr),
+                                      static_cast<ExtraArgsTypes *>(nullptr),
+                                      Indexes,
+                                      ExtraArgs,
+                                      ExtraArgsIndexes);
```
aleclearmind, line 401: Invokable is getting more and more complex. Can you please add doxygen to it detailing all of its features?

aleclearmind:
* Can we a test for analysis arguments?
* Please drop the last commit, I noted it down.

aleclearmind (changes requested):

include/revng/Pipeline/Invokable.h
drblallo: i have not idea, if deindent it, it gets indented again

include/revng/PipelineC/Prototypes.h
drblallo: the issue is that it is implemented as getting a vector of strings, each rappresenting the type or name of the argument. To make it return a non-owing pointer would require two virtual fuctions in the type erased analysis, one to get the number of options and the other to retrieve it. It is not impossible, just uglier.

include/revng/Pipeline/Invokable.h
drblallo: some what? not sure what this refers to

include/revng/Pipes/LiftPipe.h
drblallo: i called it precondition so that the writer of a pipe must write a  function name precondition, which i think in that context

include/revng/Pipeline/Invokable.h
aleclearmind: It's because of the `concept`. We use [this trick](https://github.com/revng/revng/blob/develop/include/revng/ADT/Concepts.h#L37) elsewhere.

include/revng/Pipeline/Invokable.h
aleclearmind: The changeset seems to be just reformatting the code, which should not happen if `revng-check-conventions` passes on all commits.

include/revng/Pipes/LiftPipe.h
aleclearmind: I think you didn't finish the sentence.

drblallo: in that context is more explanatory than checkPipe

lib/Pipeline/Runner.cpp
drblallo: about that, i am usually kind of against augmenting errors, because it requires for the inner error to be converted to string. Do we have a policy on this?

include/revng/Pipes/LiftPipe.h
aleclearmind: A method called `precondition` makes me think that's the same as `getPrecondition`, not that you're checking it.

lib/Pipeline/Runner.cpp
aleclearmind: Eh we could wrap them instead of converting them to strings, but that gets Java-ish quite rapidly.
Maybe we can have the error with step and pipe fields that are filled in while returning?
It's rather important to know who generated the error.

include/revng/Pipes/LiftPipe.h
drblallo: mhh yeah but what is it a precondition if not a function from something to bool? check precondition would be itself a function from something to bool, and do the exact same as the precondition.

aleclearmind: It is a function from something to `bool`, but here we're not returning a function, we're running it.
If this was to return a function `something -> bool`, the name would be great, but here we're *running it*, not returning it.

drblallo: yeah but it is the syntax Pipe::precondition that returns the precondition, Pipe::precondition() is the syntax that actually checks it.

aleclearmind: By this logic the function `Instruction::dump` should be called `Instruction::dumper`.

drblallo: the difference is that dump is a verb first and you can extract a name (dumper). While here we have a name first and we need to extract a verb. a more fitting example is a java interface called ConditionChecker that has a single function from void to bool. That is not a ConditionChecker, it is a Condition, because there is nothing to remove to extract a Condition from a ConditionChecker.

Furthermore this name refers to the the well known concept of preconditions and post conditions, which should hint to the user "this is run first and must be true."

aleclearmind: I think nouns make sense for fields, using a noun for a method is quite confusing.
Sometimes one elides `is`, or `has` prefixes (still sad), but eliding `check` is a bit too much to me.
But I guess we're overdiscussing this, proceed as you think it's more fit.

drblallo: ok, i will had check

include/revng/Pipes/PipelineManager.h
aleclearmind: Not fixed.

include/revng/PipelineC/Prototypes.h
```diff
@@ -658,3 +681,27 @@ rp_error_list_get_error_message(rp_error_list *error, uint64_t index);
...
 
 /** \} */
+
+/**
+ * \defgroup rp_string_map rp_string_map methods
+ * \{
+ */
+
+/**
+ * \return a owing pointer to a new string map
```
aleclearmind, line 691:
```suggestion
 * \return a owning pointer to a new string map
```

tests/unit/Pipeline.cpp
```diff
@@ -1344,4 +1346,35 @@ BOOST_AUTO_TEST_CASE(MultiStepInvalidationTest) {
...
 
+class ArgumentTestPipe {
+public:
+  constexpr static const char *Name = "dont_care";
+
+  std::vector<ContractGroup> getContract() const { return {}; }
+
+  constexpr static std::tuple Options = { pipeline::Argument("first", 10),
+                                          pipeline::Argument("second",
+                                                             "something") };
```
aleclearmind, line 1357: Options or arguments? Take a decision and make them coherent everywhere.

lib/Pipeline/Runner.cpp
```diff
@@ -303,6 +310,14 @@ Error Runner::run(llvm::StringRef EndingStepName,
   if (ToExec.size() <= 1)
     return Error::success();
 
+  for (auto &StepGoalsPairs : llvm::drop_begin(ToExec)) {
+    auto &Step = *StepGoalsPairs.ToExecute;
+    if (auto Error = Step.checkPrecondition(getContext()); Error)
+      return llvm::make_error<AnnotatedError>(std::move(Error),
+                                              "while scheduling step: "
+                                                + Step.getName());
```
aleclearmind, line 318: Please add the pipe name too.

aleclearmind (changes requested):

include/revng/Pipeline/Invokable.h
```diff
@@ -13,95 +13,303 @@
...
+/// * a RetT run(T...) method where the first argument must be a Context& or
+///   const Context&, arguments after the first must be K& or const K& where
+///   K is derived from a container.
+///
+///   Argumens after the last container can be of any type, but for each of
+///   them there must exists a entry in a constexpr tuple named Options that
+///   defines their name and default value.
+///
+///   RetT can either be llvm::Error or void, if it is void then the invokable
+///   never fails.
```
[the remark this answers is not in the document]
drblallo: why deindenting there? it still refers to the previous *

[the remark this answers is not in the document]
aleclearmind: Oh OK!

include/revng/Pipes/PipelineManager.h
drblallo: ahh forgot to change argument names outside of invokables.h

aleclearmind: Can you fix these? I'd like to merge this PR.

PR #218, opened 2022-08-30
Stop using `git ls-files` for header discovery

Cvelth:
This fixes a long standing (and pretty annoying) issue or having to stage any newly headers before they can be used.


Cvelth:
Merged in https://github.com/revng/revng/commit/4c08f2153dfa5defdaf35ff5c52d74cbb517bdfd.

PR #221, opened 2022-09-07
Pipeline: rework containers

drblallo:


include/revng/Yield/Pipes/YieldControlFlowPipe.h
```diff
@@ -12,9 +12,39 @@
...
 #include "revng/Pipes/FunctionStringMap.h"
+#include "revng/Pipes/Kinds.h"
 
 namespace revng::pipes {
 
+inline constexpr char FunctionAssemblyMIMEType[] = "application/"
+                                                   "x.yaml.function-assembly."
+                                                   "internal";
+
+using FunctionAssemblyStringMap = FunctionStringMap<
```
aleclearmind, line 23: Some of these classes have `Container` in the name, some don't. Can we have a namespace?

aleclearmind (changes requested):

PR #222, opened 2022-09-10
Add `fetch-debuginfo` tool

dwarfyresearch:


python/revng/cli/_commands/fetch_debuginfo/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+        path = os.path.abspath(args.input)
+        with open(args.input, "rb") as f:
+            try:
+                the_elffile = ELFFile(f)
+                if not args.urls:
+                    args.urls.append(default_elf_debug_server)
+
+                result = fetch_dwarf(the_elffile, path, args.urls, self.verbose)
+            except ELFError as elf_error:
+                if str(elf_error) == "Magic number does not match":
```
aleclearmind, line 61: * Does `ELFError` has `.message` or something like that?
* Consider either dropping this check and just go to pefile in case any `ELFError` arises, or maybe manually check for the ELF header. Manually comparing error message seems a bit fragile.

python/revng/cli/_commands/fetch_debuginfo/common.py
```diff
@@ -0,0 +1,44 @@
...
+
+import gzip
+import os
+import shutil
+import sys
+
+import requests
+
+
+def log(message, verbose):
```
aleclearmind, line 14: The `verbose` argument is effectively a global variable here.
Consider making a global variable that's private to the module.

python/revng/cli/_commands/fetch_debuginfo/common.py
```diff
@@ -0,0 +1,44 @@
...
+            # TODO: Add support for other types of compressed files.
+            if r.headers["Content-Encoding"] == "gzip":
+                gzip_filename = local_filename + ".gz"
+                with gzip.open(gzip_filename, "wb") as gzip_file:
+                    shutil.copyfileobj(r.raw, gzip_file)
+                with gzip.open(gzip_filename, "rb") as gzip_file:
+                    content = gzip.decompress(gzip_file.read())
+                with open(local_filename, "wb") as debug_file:
+                    debug_file.write(content)
+                os.remove(gzip_filename)
```
aleclearmind, line 32: Are you sure this is required?

> The gzip and deflate transfer-encodings are automatically decoded for you.

https://requests.readthedocs.io/en/latest/user/quickstart/#binary-response-content

python/revng/cli/_commands/fetch_debuginfo/common.py
```diff
@@ -0,0 +1,44 @@
...
+            else:
+                with open(local_filename, "wb") as debug_file:
+                    shutil.copyfileobj(r.raw, debug_file)
+
+            log("Downloaded...", verbose)
+            return True
+        else:
+            log("URL was not found...", verbose)
+            # Remove empty file.
+            os.remove(local_filename)
```
aleclearmind, line 42: Why remove if the file was never created?

python/revng/cli/_commands/fetch_debuginfo/elf.py
```diff
@@ -0,0 +1,195 @@
...
+    global_debug_package_list_path = "/etc/apt/sources.list.d/ddebs.list"
+    if os.path.exists(global_debug_package_list_path):
+        log("dbgsym packages are enabled...", verbose)
+        return True
+
+    log(
+        "warning: Please enable dbgsym packages by following "
+        "https://wiki.ubuntu.com/Debug%20Symbol%20Packages",
+        verbose,
+    )
```
aleclearmind, line 59: This is Ubuntu-specific, also the user might have called the file something different from `ddebs.list`.

I think fastest way to check this and be compatible with both Ubuntu and Debian is to check if any repos has `-debug` suffix (which seems to be the [Debian's way](https://wiki.debian.org/AutomaticDebugPackages)) or the repo URL contains `ddebs.ubuntu.com` (which seems to be [Ubuntu's way](https://wiki.ubuntu.com/Debug%20Symbol%20Packages)).

We need to do this for `/etc/apt/sources.list` and for each file in `/etc/apt/sources.list.d/`, considering each line starting with `deb`.

The alternative would be `apt-cache search '\-dbgsyms' | grep -- '-dbgsyms '` but that's soooo slow.

python/revng/cli/_commands/fetch_debuginfo/elf.py
```diff
@@ -0,0 +1,195 @@
...
+#   4. use Debuginfod like apporach by finding it on web
+#   5. apt debug packages
+def fetch_dwarf(file: ELFFile, file_path, urls, verbose):
+    log("Looking for Debugging Information for an ELF...", verbose)
+
+    # 1) Check file itself.
+    section = file.get_section_by_name(".debug_info")
+    if section is not None:
+        log("Binary has debug info within itself", verbose)
+        return file_path
```
aleclearmind, line 76: Can a file have both `.debug_info` and external debug information?

python/revng/cli/_commands/fetch_debuginfo/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+    else:
+        log("Result: No debug info found", verbose)
+
+
+class FetchDebugInfoCommand(Command):
+    def __init__(self):
+        super().__init__(("model", "fetch-debuginfo"), "Fetch Debugging Information.")
+
+    def register_arguments(self, parser):
+        parser.add_argument("input", help="The input file.")
```
aleclearmind, line 36: In debug info, do we usually get debug info for imported dynamic functions? That is, do we get the signature of `qsort` if we dynamically link the libc?
If not, it might make sense to have `--dependencies=3` which recursively fetches debug info from linked libraries too up to depth 3. We could leave as a default `1`.

What do you think?

python/revng/cli/_commands/fetch_debuginfo/elf.py
```diff
@@ -0,0 +1,195 @@
...
+    absolute_path_debuginfo = os.path.join(build_id_dir, first_two_chars, debug_file_name)
+    if os.path.exists(absolute_path_debuginfo):
+        log("Debug File found: " + absolute_path_debuginfo, verbose)
+        return absolute_path_debuginfo
+    else:
+        log(f"Could not find {absolute_path_debuginfo}", verbose)
+
+    # 4) Find debug info on the web as `debuginfod` does.
+    if "XDG_CACHE_HOME" not in os.environ or os.environ["XDG_CACHE_HOME"] == "":
+        log("warning: XDG_CACHE_HOME was not set. Using ~/.cache/debuginfod_client/...", verbose)
```
aleclearmind, line 156: Consider making a `warning` function that prefixes `Warning:`. One day we might use a colorful logger and will make things easier.

python/revng/cli/_commands/fetch_debuginfo/elf.py
```diff
@@ -0,0 +1,195 @@
...
+    build_id_dir = global_debug_directory + ".build-id/"
+    if not os.path.exists(build_id_dir):
+        log(f"Could not find {build_id_dir}", verbose)
+        return None
+
+    # 3) Build ID.
+    #  If build-id is `abcdef1234`, we look for:
+    #   - /usr/lib/debug/.build-id/ab/cdef1234.debug
+    log("Parsing build-id...", verbose)
+    biuld_id = parse_build_id(file)
```
aleclearmind, line 136: Typo, `build`.

python/revng/cli/_commands/fetch_debuginfo/elf.py
```diff
@@ -0,0 +1,195 @@
...
+        log(f"Trying to download from {debug_info_url}...", verbose)
+        if download_file(debug_info_url, debug_file_to_download, verbose) is True:
+            # TODO: This can be compressed, if the file was compressed. Do we need to uncopress it?
+            return debug_file_to_download
+
+    # 5) Find debug symbols with Ubuntu Debug Symbol Packages.
+    log("Trying to find on the web with apt-file...", verbose)
+    if check_dbgsym_packages(verbose):
+        if shutil.which("dpkg") is not None:
+            proc = subprocess.run(["dpkg", "-S", file_path], capture_output=True)
```
aleclearmind, line 181: * Check the exit code of the program
* Fix the comment, we're not using `apt-file`.

python/revng/cli/_commands/fetch_debuginfo/pe.py
```diff
@@ -0,0 +1,66 @@
...
+            continue
+        pdb_file_name = e.entry.PdbFileName.decode("utf-8")
+        pdb_file_name = pdb_file_name.replace("\x00", "")
+        # A hash for the PDB file.
+        pdb_id = (
+            str(hex(e.entry.Signature_Data1)[2:])
+            + str(hex(e.entry.Signature_Data2)[2:])
+            + str(hex(e.entry.Signature_Data3)[2:])
+            + e.entry.Signature_Data4.hex()
+            + str(hex(e.entry.Age)[2:])
```
aleclearmind, line 38: * No need for `str`
* Make an helper function for `hex(...)[2:]`.

python/revng/cli/_commands/fetch_debuginfo/pe.py
```diff
@@ -0,0 +1,66 @@
...
+        )
+        log("PDBID: " + pdb_id + "...", verbose)
+
+        if "XDG_CACHE_HOME" not in os.environ or os.environ["XDG_CACHE_HOME"] == "":
+            log(
+                "warning: XDG_CACHE_HOME was not set. Using ~/.cache/debuginfod_client/...", verbose
+            )
+            path_to_download = str(Path.home()) + "/.cache/debuginfod_client/" + pdb_id + "/"
```
aleclearmind, line 46: * Elsewhere we `import xdg`, which should do this for you.
* Use `~/.cache/revng/debug-symbols/{elf,pe}`.

python/revng/cli/_commands/fetch_debuginfo/pe.py
```diff
@@ -0,0 +1,66 @@
...
+        else:
+            path_to_download = os.environ["XDG_CACHE_HOME"] + "/debuginfod_client/" + pdb_id + "/"
+
+        if not os.path.exists(path_to_download):
+            os.makedirs(path_to_download)
+
+        pdb_file_to_download = path_to_download + "/" + pdb_file_name
+        if os.path.exists(pdb_file_to_download):
```
aleclearmind, line 54: We're trying to migrate away from `os.path` and concatenating strings to [`from pathlib import Path`](https://docs.python.org/3/library/pathlib.html), which is nicer and more portable. Please switch.

python/revng/cli/_commands/fetch_debuginfo/pe.py
```diff
@@ -0,0 +1,66 @@
...
+            log("Already downloaded debug file from web...", verbose)
+            return pdb_file_to_download
+
+        for symbol_server_url in urls:
+            pdb_url = symbol_server_url + pdb_file_name + "/" + pdb_id + "/" + pdb_file_name
+            log("Trying to download PDB from URL: " + pdb_url + "...", verbose)
+            if download_file(pdb_url, pdb_file_to_download, verbose) is True:
```
aleclearmind, line 61: `is True`?

python/revng/cli/_commands/fetch_debuginfo/pe.py
```diff
@@ -0,0 +1,66 @@
...
+                return pdb_file_to_download
+
+        break
```
aleclearmind, line 64: This a bit of a weird for loop:
```python
    for e in pe.DIRECTORY_ENTRY_DEBUG:
        if not isinstance(e.entry, pefile.Structure):
            continue

        # ...

        break
```
Maybe you can reorganize it?

python/revng/cli/_commands/fetch_debuginfo/common.py
```diff
@@ -0,0 +1,44 @@
...
+                    debug_file.write(content)
+                os.remove(gzip_filename)
+            else:
+                with open(local_filename, "wb") as debug_file:
+                    shutil.copyfileobj(r.raw, debug_file)
+
+            log("Downloaded...", verbose)
+            return True
+        else:
+            log("URL was not found...", verbose)
```
aleclearmind, line 40: Please drop `...` from all messages.

tools/model/import/debuginfo/Main.cpp
```diff
@@ -0,0 +1,102 @@
...
+    Model->DefaultPrototype = abi::registerDefaultFunctionPrototype(*Model);
+
+    const llvm::object::pe32_header *PE32Header = TheBinary->getPE32Header();
+    if (PE32Header) {
+      ImageBase = MetaAddress::fromPC(LLVMArchitecture, PE32Header->ImageBase);
+    } else {
+      const llvm::object::pe32plus_header
+        *PE32PlusHeader = TheBinary->getPE32PlusHeader();
+      if (not PE32PlusHeader)
+        return 1;
```
aleclearmind, line 86: `EXIT_FAILURE`

aleclearmind (changes requested):

lib/Model/Importer/DebugInfo/DwarfImporter.cpp
```diff
@@ -893,14 +896,56 @@ ArrayRef<uint8_t> getSectionsContents(StringRef Name, T &ELF) {
   return {};
 }
 
-static StringRef getAltDebugLinkFileName(const object::Binary *B) {
+static std::string getBuildID(const object::Binary *B) {
```
aleclearmind, line 899: NOTETOSELF: review the last commit.

python/revng/cli/_commands/fetch_debuginfo/__init__.py
dwarfyresearch: I will check it manually. I agree.

python/revng/cli/_commands/fetch_debuginfo/common.py
dwarfyresearch: I saw it, but without this I get ```.gzip``` files from the elfutils server.

python/revng/cli/_commands/fetch_debuginfo/common.py
dwarfyresearch: In some situations we end up with an empty file with the ```local_filename```.

python/revng/cli/_commands/fetch_debuginfo/elf.py
dwarfyresearch: I think no... I'd consider that as a hack and non-standard way. In theory, you can use ```dd``` and copy whatever part of the ELF you want, but I guess we don't want to count on that.

There are 2 ways of generating separate debug files (on Linux):

- Separate Debug FIles -- ```.debug``` files, most common way, but not Standardized (let's say supported as ```GNU``` extensions)
- Split DWARF and .debug_sup``` -- Standardized way of keeping separate debug info files; part of DWARF 5;
  - In the case of Split DWARF, we do have some DWARF sections, but they are marked with ```.dwo``` suffix (e.g. ```debug_info.dwo```)

dwarfyresearch: Also, I am not able to add support for Mac OS, since I don't have a machine, but they use ```dsym``` files, very similar approach to these ```.debug``` files.

python/revng/cli/_commands/fetch_debuginfo/__init__.py
dwarfyresearch: Hm, it makes sense. But I think we agreed that the ```dependecies``` part we should resolve on a higher level, possibly from another tool. I think this should be very simple and work with one binary only.

python/revng/cli/_commands/fetch_debuginfo/common.py
dwarfyresearch: Dropped :) ```GNU``` likes those in their tools a lot.

python/revng/cli/_commands/fetch_debuginfo/elf.py
dwarfyresearch: Ah....yes.... I actually don't like either way of these two. Can we just do it unconditionally (without checking if ```check_dbgsym_packages``` is ```true```)? Just be specific in the message that it will work if the ```service``` was enabled?

python/revng/cli/_commands/fetch_debuginfo/common.py
dwarfyresearch: I've resolved this one, but I am having this ```check-convention``` error:

```
$ orc shell revng check-conventions --force-format python/revng/cli/_commands/fetch_debuginfo/__init__.py 
python/revng/cli/_commands/fetch_debuginfo/__init__.py:49: error: Name "common" is not defined
```
It is strange I guess. @mrjackv, you may help me with this? Thanks!

python/revng/cli/_commands/fetch_debuginfo/elf.py
aleclearmind: * What `service`?
* So you want to simply run `dpkg` (check if it's available)? Works!

python/revng/cli/_commands/fetch_debuginfo/elf.py
aleclearmind: Let's postpone macOS.
OK, for the rest.

python/revng/cli/_commands/fetch_debuginfo/__init__.py
aleclearmind: If you want we can postpone this, but it's rather important to have it. I think it's preferable to have a single tool with `--dependencies=3` than a distinct one just spawning a subprocess.
If I said something different in the past, I changed my mind `:)`.

python/revng/cli/_commands/fetch_debuginfo/elf.py
dwarfyresearch: Yeah, exactly that, without making the hacking/observation of users ```sources.list```.

python/revng/cli/_commands/fetch_debuginfo/__init__.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+        parser.add_argument(
+            "-d",
+            "--dependencies",
+            help="Fetch debug info for dependencies as well.",
+            type=int,
+            default=0,
+        )
+
+    def run(self, options: Options):
```
aleclearmind, line 52: Too deep, can you split in multiple smaller functions?

python/revng/cli/_commands/fetch_debuginfo/__init__.py
```diff
@@ -0,0 +1,147 @@
...
+        self.verbose = args.verbose
+        logger.verbose = self.verbose
+
+        the_input_path = Path(args.input)
+        if not the_input_path.exists():
+            log_error("Could not find " + args.input)
+            return 1
+
+        path = str(the_input_path.resolve())
+        with open(args.input, "rb") as f:
```
aleclearmind, line 64: I think `Path` has an `open` method.
You almost never want to convert it to a `str`.

python/revng/cli/_commands/fetch_debuginfo/pe.py
```diff
@@ -0,0 +1,150 @@
...
+    processed_dependencies_pe,
+    current_level,
+    dependencies_level,
+):
+    log("Getting dependencies for " + path + " current level " + str(current_level))
+    processed_dependencies_pe.add(path)
+    # TODO: All dlls from Windows we are searching for in the current directory only.
+    for e in file.DIRECTORY_ENTRY_IMPORT:
+        dll_name = e.dll.decode("utf8").lower()
+        full_dll_path = str(PurePath(directory_of_input_file).joinpath(dll_name))
```
aleclearmind, line 37: Why pass through `str`? You can use the `/` operator.

aleclearmind (changes requested):

aleclearmind (changes requested):

lib/Model/Importer/Binary/ELFImporter.cpp
```diff
@@ -326,9 +328,76 @@ Error ELFImporter<T, HasAddend>::import(bool FetchDebugInfo) {
...
+void ELFImporter<T, HasAddend>::findMissingTypes(object::ELFFile<T> &TheELF,
+                                                 unsigned FetchDebugInfo) {
+  ModelMap ModelsOfLibraries;
+
+  LDDTree Dependencies;
+  lddtree(Dependencies, TheBinary.getFileName().str(), FetchDebugInfo);
+  for (auto &Library : Dependencies) {
+    revng_log(ELFImporterLog,
+              "Importing Models for dependencies of " << Library.first
+                                                      << ":\n");
```
aleclearmind, line 348: No need for `\n`, unless you really want a blank line.

lib/Model/Importer/Binary/ELFImporter.cpp
```diff
@@ -326,9 +328,76 @@ Error ELFImporter<T, HasAddend>::import(bool FetchDebugInfo) {
...
+      if (ModelsOfLibraries.count(DependencyLibrary))
+        continue;
+      revng_log(ELFImporterLog,
+                " Importing Model for: " << DependencyLibrary << "\n");
+      auto BinaryOrErr = llvm::object::createBinary(DependencyLibrary);
+      if (not BinaryOrErr) {
+        llvm::consumeError(BinaryOrErr.takeError());
+        revng_log(ELFImporterLog,
+                  "Can't create object for " << DependencyLibrary);
+        return;
```
aleclearmind, line 359: Why bail out entirely? Can't we just skip this one?

lib/Model/Importer/Binary/ELFImporter.cpp
```diff
@@ -326,9 +328,76 @@ Error ELFImporter<T, HasAddend>::import(bool FetchDebugInfo) {
...
+      }
+
+      auto &Object = *cast<llvm::object::ObjectFile>(BinaryOrErr->getBinary());
+      auto *TheBinary = dyn_cast<ELFObjectFileBase>(&Object);
+      if (!TheBinary)
+        continue;
+
+      using model::Binary;
+      using std::make_unique;
+      ModelsOfLibraries[DependencyLibrary] = make_unique<TupleTree<Binary>>();
```
aleclearmind, line 369: Why `unique_ptr`?

lib/Model/Importer/Binary/ELFImporter.cpp
```diff
@@ -326,9 +328,76 @@ Error ELFImporter<T, HasAddend>::import(bool FetchDebugInfo) {
...
+      TupleTree<model::Binary> &DepModel = *ModelsOfLibraries[DependencyLibrary]
+                                              .get();
+      DepModel->Architecture = Model->Architecture;
+
+      if (auto E = importELF(DepModel,
+                             *TheBinary,
+                             BaseAddress,
+                             1 /*FetchDebugInfo*/)) {
```
aleclearmind, line 377: Do we have recursion here?
Could one craft a binary leading to infinite recursion?

lib/Model/Importer/Binary/CrossModelFindTypeHelper.h
```diff
@@ -0,0 +1,268 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include <map>
+#include <memory>
+
+#include "revng/Model/Binary.h"
+
+namespace {
+using ModelMap = std::map<std::string,
```
aleclearmind, line 13: NOTETOSELF: review this file

lib/Model/Importer/Binary/PECOFFImporter.cpp
```diff
@@ -311,6 +320,124 @@ void PECOFFImporter::parseDelayImportedSymbols() {
   }
 }
 
+// NOTE: For the PE/COFF, we are assuming that the libraries are in the current
+// directory.
```
aleclearmind, line 324: Consider using Doxygen `/// \note For...`

lib/Model/Importer/Binary/PECOFFImporter.cpp
```diff
@@ -311,6 +320,124 @@ void PECOFFImporter::parseDelayImportedSymbols() {
...
+static void getDependenciesHelper(StringRef FileName,
```
aleclearmind, line 325: NOTETOSELF: review this commit

aleclearmind (changes requested):

python/revng/cli/commands_registry.py
```diff
@@ -125,7 +125,7 @@ def run(self, arguments, options: Options):
                     "perf",
                     "record",
                     "--call-graph",
-                    "dwarf",
+                    "debuginfo",
```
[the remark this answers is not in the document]
dwarfyresearch: ahhh, I m stupid... I just used ```grep/sed``` :/ tnx

lib/Model/Importer/DebugInfo/DwarfImporter.cpp
```diff
@@ -932,32 +977,185 @@ static void error(StringRef Prefix, std::error_code EC) {
...
+                                           StringRef DebugFileName,
+                                           llvm::object::ObjectFile *ELF) {
+  // Let's find it in canonical places, where debug info was fetched.
+  //  1) Look for a .gnu_debuglink/.gnu_debugaltlink/.debug_sup section.
+  //  The .debug file should be in canonical places.
+  //  E.g., if the executable is `/usr/bin/ls`, we look for:
+  //     - /usr/bin/ls.debug (current dir of exe)
+  //     - /usr/bin/.debug/ls.debug
+  //     - /usr/lib/debug/usr/bin/ls.debug
+  std::string Result;
```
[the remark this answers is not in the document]
dwarfyresearch: Right, moving to ```llvm::sys::path::append```.

python/revng/cli/_commands/fetch_debuginfo/elf.py
```diff
@@ -0,0 +1,229 @@
...
+    section = file.get_section_by_name(".gnu_debuglink")
+    if section is not None:
+        log("Found a .gnu_debuglink")
+        # TODO: Use file.get_dwarf_info().parse_debugsupinfo(), since it supports
+        # `.gnu_debugaltlink’ and ‘.debug_sup’ sections.
+        debug_file_name = Path(parse_debug_link(section.data()))
+
+        dir_from_exe = Path(PurePath(file_path).parent)
+        if debug_file_name is not None:
+            # 2A) Look in the directory where executable is located.
```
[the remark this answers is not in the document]
dwarfyresearch: We are just checking if it is on the device, so we do not invoke the `downloading` process if there is no need for it.

lib/Model/Importer/DebugInfo/PDBImporter.cpp
```diff
@@ -270,34 +268,172 @@ void PDBImporter::loadDataFromPDB(std::string PDBFileName) {
...
+        auto ExitCode = ::Runner.run("revng",
+                                     { "model",
+                                       "fetch-debuginfo",
+                                       TheBinary.getFileName().str() });
+        if (ExitCode != 0) {
+          revng_log(DILogger,
+                    "Failed to find debug info with `revng model "
+                    "fetch-debuginfo`.");
+          return;
+        }
```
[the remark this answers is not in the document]
dwarfyresearch: Yep.

lib/Model/Importer/DebugInfo/PDBImporter.cpp
```diff
@@ -270,34 +268,172 @@ void PDBImporter::loadDataFromPDB(std::string PDBFileName) {
...
+// the `~/.local/share/revng/debug-symbols/pe/`.
+std::string PDBImporter::getCachedPDBFilePath(std::string PDBFileID,
+                                              StringRef PDBFilePath,
+                                              StringRef InputFileName) {
+  std::string Result;
+  // Convert the path to a real Unix path if it was specified with `Z:`.
+  if (PDBFilePath.startswith("Z:")) {
+    Result = PDBFilePath.substr(2);
+    std::replace(Result.begin(), Result.end(), '\\', '/');
+  }
```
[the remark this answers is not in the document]
aleclearmind: I mean: is having `Z:\` hardcoded a convention in PDBs?
Does `z:\` represent somehow `/` in unix?
I'm not sure how much sense does it make to convert, say `z:\Program Files\lol.pdb` into `/Program Files/lol.pdb`, if that's what's happening.

python/revng/cli/_commands/fetch_debuginfo/elf.py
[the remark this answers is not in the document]
aleclearmind: Uhm, OK, but what I mean is: we are duplicating the logic to find the location of files (in Python and in C++).
Didn't we say that Python should call the C++ tool and drop the duplication?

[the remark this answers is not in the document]
dwarfyresearch: Hmmm, it was for ```lddtree``` functionality, but I can think of reducing the duplication even for this...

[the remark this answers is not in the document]
dwarfyresearch: How about something like this:

``` revng find-debuginfo INPUTFILE```

It returns ```0``` if it was able to find the debug info. The ```fetch-debuginfo``` just downloads the debug file from the web.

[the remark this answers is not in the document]
Comment (dwarfyresearch) [reply python/revng/cli/_commands/fetch_debuginfo/elf.py:120]: >The fetch-debuginfo just downloads the debug file from the web.
For just a one input. The ```dependency``` level will be handled from C++.

File: include/revng/Support/LDDTree.h [hunk 34]
```diff
@@ -0,0 +1,14 @@
...
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include <map>
+#include <string>
+#include <vector>
+
+using LDDTree = std::map<std::string, std::vector<std::string>>;
```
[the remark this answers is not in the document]
dwarfyresearch: ```SmallVector``` is good option, thanks!

lib/Model/Importer/DebugInfo/PDBImporter.cpp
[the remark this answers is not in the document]
dwarfyresearch: Well, you are right, it is not a convention...

lib/Model/Importer/Binary/ELFImporter.cpp
dwarfyresearch: it makes sense

include/revng/Model/Importer/Binary/BinaryImporter.h
```diff
@@ -18,8 +18,8 @@ class ObjectFile;
 llvm::Error importBinary(TupleTree<model::Binary> &Model,
                          llvm::object::ObjectFile &BinaryHandle,
                          uint64_t PreferredBaseAddress,
-                         bool FetchDebugInfo);
+                         unsigned FetchDebugInfo);
```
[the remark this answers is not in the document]
dwarfyresearch: Squashed && renamed, thanks

lib/Model/Importer/Binary/ELFImporter.cpp
dwarfyresearch: Well, I am not sure if it is possible. At least, we are just importing the binary, without looking for missing types (```1 /*FetchDebugInfo*/```).

python/revng/cli/_commands/fetch_debuginfo/elf.py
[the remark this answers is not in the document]
dwarfyresearch: Addressed.

lib/Model/Importer/Binary/ELFImporter.cpp
aleclearmind: Due to `if (FetchDebugInfoWithLevel > 1)` you mean?

OK, but then `FetchDebugInfoWithLevel == 1` means "don't look in any dependency", right?
If so, make sure this is documented, it's not obvious.

dwarfyresearch: Documented in the option description.

python/revng/cli/_commands/fetch_debuginfo/common.py
```diff
@@ -0,0 +1,57 @@
...
+                with open(local_filename, "wb") as debug_file:
+                    shutil.copyfileobj(r.raw, debug_file)
+
+            log("Downloaded")
+            return True
+        else:
+            log("URL was not found")
+            # Remove empty file if it was created.
+            if os.path.exists(local_filename):
+                os.remove(local_filename)
```
aleclearmind, line 55: How would this be possible?

python/revng/cli/_commands/fetch_debuginfo/common.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python3
```
aleclearmind, line 1: Drop shebang from all files that are not supposed to be run directly (i.e., have no `main`).

python/revng/cli/_commands/fetch_debuginfo/elf.py
```diff
@@ -0,0 +1,65 @@
...
+    if build_id is None:
+        # If we cannot parse the build id, we cannot fetch debug info.
+        log("Parsing build-id failed")
+        return None
+
+    log("BUILD-ID: " + build_id)
+
+    # Find debug info on the web as `debuginfod` does.
+    # Ensure that we have created `$xdg_data_home/revng/debug-symbols/elf` directory.
+    path_to_revng_data = Path(str(xdg_data_home()) + "/revng/")
```
aleclearmind, line 39: `xdg_cache_home` would probably be more appropriate. What do you think?

python/revng/cli/_commands/fetch_debuginfo/pe.py
```diff
@@ -0,0 +1,77 @@
...
+    # Parse the file.
+    try:
+        pe = pefile.PE(file_path)
+    except Exception as exception:
+        log_error("Unable to parse the input file.")
+        log_error(str(exception))
+        return None
+
+    # Ensure that we have created `$xdg_data_home/revng/debug-symbols/pe` directory.
+    path_to_revng_data = Path(str(xdg_data_home()) + "/revng/")
```
aleclearmind, line 28: As before

lib/Support/LDDTree.cpp
```diff
@@ -0,0 +1,210 @@
...
+#include "revng/Support/Debug.h"
+#include "revng/Support/LDDTree.h"
+
+Logger<> Log("lddtree");
+
+using namespace llvm;
+
+static std::optional<std::string>
+findLibrary(llvm::StringRef Lib, StringRef RunPath) {
+  auto FindInDir = [Lib](std::string WhereToLook) {
```
aleclearmind, line 30: Prefer a regular function here

aleclearmind (changes requested):

python/revng/cli/_commands/fetch_debuginfo/common.py
dwarfyresearch: Hm, cannot reproduce it anymore. It may have been some other issue that I resolved meanwhile. Removing it. Thanks.

python/revng/cli/_commands/fetch_debuginfo/elf.py
dwarfyresearch: We have used ```xdg_cache_home``` at first, and decided to move to ```xdg_data_home``` :) But I guess that ```cache``` is more appropriate.

aleclearmind: If it's something we can fetch/recreate again (unless the apocalypse ensues), let's go for cache.

lib/Support/LDDTree.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+    if (llvm::sys::fs::exists(ResultPath.str()))
+      return std::string(ResultPath.str());
+
+    std::error_code EC;
+    llvm::sys::fs::directory_iterator DirIt(WhereToLook, EC);
+    llvm::sys::fs::directory_iterator DirEnd;
+    if (EC) {
+      llvm::consumeError(llvm::errorCodeToError(EC));
+      return std::string();
```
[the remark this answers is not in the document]
dwarfyresearch: yep, you are right, thanks. I've missed this..

lib/Model/Importer/Binary/ELFImporter.cpp
```diff
@@ -324,11 +327,96 @@ Error ELFImporter<T, HasAddend>::import() {
...
+      using model::Binary;
+      using std::make_unique;
+      ModelsOfLibraries[DependencyLibrary] = TupleTree<Binary>();
+      TupleTree<model::Binary> &DepModel = ModelsOfLibraries[DependencyLibrary];
+      DepModel->Architecture = Model->Architecture;
+      if (auto E = importELF(DepModel,
+                             *TheBinary,
+                             BaseAddress,
+                             1 /*FetchDebugInfoWithLevel*/)) {
+        llvm::consumeError(std::move(E));
```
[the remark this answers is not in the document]
dwarfyresearch: I agree.

aleclearmind (changes requested):

lib/Model/Importer/Binary/CrossModelFindTypeHelper.h
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  return Result;
+}
+
+template<typename T>
+std::optional<model::TypePath>
+findPrototype(T &Functions, llvm::StringRef FunctionName) {
+  for (auto &Function : Functions) {
+    if (Function.OriginalName != FunctionName)
```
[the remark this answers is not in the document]
dwarfyresearch: Yes, I agree.

lib/Model/Importer/Binary/CrossModelFindTypeHelper.h
```diff
@@ -0,0 +1,103 @@
...
+  llvm::df_iterator_default_set<Node *> VisitedFromThePrototype;
+  for (Node *N :
+       depth_first_ext(TypeToNode.at(Prototype.get()), VisitedFromThePrototype))
+    ;
+
+  for (auto &P : FromModel->Types) {
+    if (VisitedFromThePrototype.contains(TypeToNode.at(P.get()))) {
+      // Clone the pointer.
+      auto NewType = P;
+      NewType->CustomName = "";
```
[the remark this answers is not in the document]
dwarfyresearch: The latter is fine with me. Thanks.

aleclearmind:
Merged in 329f610. Thanks!

PR #191, opened 2022-06-27
Emit call graphs in SVG

Cvelth:

PR #224, opened 2022-09-14
`SubGraph`: make `getEntryNode` more `const`-correct

pfez:
`GraphTraits<SubGraph<InnerNodeType>>::getEntryNode` now takes a `const GraphType &`, not just a `GraphType &`.

This is necessary to be compliant with some fixes for `ReversePostOrderTraversal` I've just backported in `llvm/next-master`, because otherwise compilation fails.

This is also a blocker for https://github.com/revng/revng/pull/191

### DIFF
--- a/lib/Lift/SubGraph.h
+++ b/lib/Lift/SubGraph.h
@@ -122,8 +122,7 @@ struct GraphTraits<SubGraph<InnerNodeType>> {
   using ChildIteratorType = typename GraphType::ChildIteratorType;
   using nodes_iterator = typename GraphType::nodes_iterator;
 
-  // TODO: here G should be const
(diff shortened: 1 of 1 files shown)


aleclearmind:
Any particular reason why you made the PR from `revng/revng`?

pfez:
> Any particular reason why you made the PR from `revng/revng`?

Mainly that this PR depends on on https://github.com/revng/llvm-project/pull/18. I wanted the CI to run, and doing the PR from my clone didn't seem to make the CI happy (the new CI workflow is still broken). So I thought that I could try the old workflow of pushing branches to this repo instead of my own clone to trigger the CI, because I hoped that the old flow was still working. But this seems to be broken too, and it turns out that it's a wider problem with the CI. Given how it went, I could have done it from my clone.

PR #225, opened 2022-09-21
revng ptml: merge cat and strip

mrjackv:
Merge the cat and strip subcommands within `revng ptml` Refactor codebase so that both color and plaintext output are handled in a unified manner.


python/revng/cli/_commands/ptml/text.py
```diff
@@ -0,0 +1,120 @@
...
+        output = args.input
+    else:
+        output = args.output
+
+    if not args.color:
+        console = PlainConsole(output)
+    else:
+        try:
+            # rich takes a while to import, do it if needed
+            from rich.console import Console
```
aleclearmind, line 101: Do we have a PR for adding this as a dep?

python/revng/cli/_commands/ptml/text.py
```diff
@@ -0,0 +1,120 @@
...
+
+            console = Console(markup=False, highlight=False, force_terminal=True, file=output)
+            console.options.no_wrap = True
+            color = True
+        except ImportError:
+            console = PlainConsole(output)
+            color = False
+
+    if args.color and not color:
+        print("Module 'rich' not found, please install it to use color mode")
```
aleclearmind, line 111:
```suggestion
        log("Module 'rich' not found, please install it to use color mode")
```

python/revng/cli/_commands/ptml/text.py
```diff
@@ -0,0 +1,120 @@
...
+                console.print(content, end="", style=COLOR_CONVERSION[metadata["data-token"]])
+            else:
+                console.print(content, end="")
+        elif node.nodeType == Node.ELEMENT_NODE:
+            new_metadata = {**metadata}
+            for key, value in node.attributes.items():
+                new_metadata[key] = value
+            _parse_ptml_node(node, console, indent, new_metadata)
+        else:
+            pass
```
aleclearmind, line 78: ?
```suggestion
```

python/revng/cli/_commands/ptml/text.py
```diff
@@ -0,0 +1,120 @@
...
+
+def parse_ptml_plain(content: str, console):
+    dom = parseString(content)
+    _parse_ptml_node(dom, console, "", {})
+
+
+def parse_ptml_yaml(content: Dict[str, str], console):
+    for key, value in content.items():
+        console.print(f"{key}:", style="yellow1", end="")
+        console.print(" |-")
```
aleclearmind, line 57: Are you sure that using `|-` you can dump whatever you want and things works just by prefixing with `  `?
I'd feel safer if we'd just modify the document in place and use `yaml.dump`.

python/revng/cli/_commands/ptml/text.py
```diff
@@ -0,0 +1,120 @@
...
+    if args.color and not color:
+        print("Module 'rich' not found, please install it to use color mode")
+        sys.exit(1)
+
+    handle_file(
+        content,
+        lambda x: parse_ptml_plain(x, console),
+        lambda x: parse_ptml_yaml(x, console),
+        filters,
+    )
```
aleclearmind, line 119: Handle file returns a value that you seem to be ignoring.
If I'm looking at the wrong function... consider giving different names, it'd be a `make_manager` situation.

aleclearmind (changes requested):

python/revng/cli/_commands/ptml/text.py
mrjackv: This was already added in the first `revng ptml` PR
https://github.com/revng/orchestra/commit/8378db3c539f1fbd482f2f0613ada8342e724ee4

python/revng/cli/_commands/ptml/text.py
mrjackv: There are 4 main issues:
1. By using yaml.dump the keys can't be colored as easiliy
2. Currently pyyaml needs a [slight hack](https://github.com/yaml/pyyaml/issues/121#issuecomment-1018117110) to make flow scalars work
3. `rich` works mainly on the terminal, to make it work with yaml I'd have to have a fake file-like object where rich prints in it and then repackage it in a dictionary, which is inefficient and would bloat the codebase
4. If the situation was more generic (i.e. we don't know ahead of time the structure of the output) I'd 100% agree with using `yaml.dump`, however the structure of the output is fixed, so there (as far as I see) no situations where the output would get broken

python/revng/cli/_commands/ptml/text.py
mrjackv: Fixed

python/revng/cli/_commands/ptml/text.py
mrjackv: Dropped

python/revng/cli/_commands/ptml/text.py
mrjackv: Fixed, `handle_file` is mean to forward the exit code from the `parse_*` function

python/revng/cli/_commands/ptml/text.py
aleclearmind: OK, please add a comment about these reasons.

mrjackv: Added

aleclearmind:
Please rebase

mrjackv:
Rebased

aleclearmind:
Merged in df68b2c.

PR #227, opened 2022-10-04
Drop * from targets

drblallo:


include/revng/Pipeline/LLVMContainer.h
```diff
@@ -126,9 +126,11 @@ class LLVMContainerBase
         Other->addMetadata(MD.first, *llvm::MapMetadata(MD.second, Map));
     }
 
-    return std::make_unique<ThisType>(this->name(),
-                                      this->Ctx,
-                                      std::move(Cloned));
+    auto ToReturn = std::make_unique<ThisType>(this->name(),
+                                               this->Ctx,
+                                               std::move(Cloned));
+    /*ToReturn->enumerate().dump();*/
```
aleclearmind, line 132: Drop

include/revng/Pipes/FunctionKind.h
```diff
@@ -6,43 +6,44 @@
...
+/*Output.push_back(Input);*/
+/*return;*/
+/*}*/
+
+/*for (const auto &Function : Model.Functions) {*/
+/*auto Component = pipeline::PathComponent(Function.Entry.toString());*/
+/*pipeline::Target ToInsert({ Component }, Kind);*/
+/*Output.emplace_back(std::move(ToInsert));*/
+/*}*/
+/*}*/
```
aleclearmind, line 35: Drop

include/revng/Pipeline/Kind.h
```diff
@@ -69,27 +68,29 @@ class Kind : public DynamicHierarchy<Kind> {
   /// As an example, the expansion of Input equal to /Root/* of isolated
   /// functions kind will create a target equivalent to /Root/FX for each FX
   /// container in the revng model.
-  virtual void expandTarget(const Context &Ctx,
-                            const Target &Input,
-                            TargetsList &Output) const;
```
aleclearmind, line 74: Drop Doxygen too

include/revng/Pipeline/Kind.h
```diff
@@ -69,27 +68,29 @@ class Kind : public DynamicHierarchy<Kind> {
...
 
   virtual void getInvalidations(const Context &Ctx,
                                 pipeline::TargetsList &ToRemove,
                                 const GlobalTupleTreeDiff &Diff) const {}
 
+  virtual void allTargets(const Context &Ctx, TargetsList &Out) const = 0;
```
aleclearmind, line 76:
```suggestion
  virtual void appendAllTargets(const Context &Ctx, TargetsList &Out) const = 0;
```

include/revng/Pipeline/Kind.h
```diff
@@ -69,27 +68,29 @@ class Kind : public DynamicHierarchy<Kind> {
...
+  TargetsList all(const Context &Ctx) const;
```
aleclearmind, line 77: Push to `Rank`, the relationship is between ranks and parts of the model.

include/revng/Pipeline/Kind.h
```diff
@@ -69,27 +68,29 @@ class Kind : public DynamicHierarchy<Kind> {
...
+
 public:
   template<typename RankDefinitionType>
   static Kind &deadKind(RankDefinitionType &Rank);
 };
 
+class UniqueKind : public Kind {
+  using Kind::Kind;
+  void allTargets(const Context &Ctx, TargetsList &Out) const override;
+};
```
aleclearmind, line 87: * All private?
* What do you think of `SingletonKind`?

include/revng/Pipeline/Kind.h
```diff
@@ -69,27 +68,29 @@ class Kind : public DynamicHierarchy<Kind> {
...
+
 class DeadKind : public Kind {
 
 public:
   DeadKind(Rank *R) : Kind("Dead", R) {}
-
-  void expandTarget(const Context &Ctx,
-                    const Target &Input,
-                    TargetsList &Output) const final {}
+  void allTargets(const Context &Ctx, TargetsList &Out) const override {}
```
aleclearmind, line 93: Abort?

include/revng/Pipeline/LLVMContainer.h
```diff
@@ -126,9 +126,11 @@ class LLVMContainerBase
...
+    return ToReturn;
```
aleclearmind, line 133: Drop this change

lib/Pipes/RootKind.cpp
```diff
@@ -62,3 +62,11 @@ void RootKind::getInvalidations(const Context &Ctx,
...
     ToRemove.emplace_back(*this);
 }
+
+void RootKind::allTargets(const Context &Ctx, TargetsList &Out) const {
+  Out.push_back(Target(*this));
+}
+
+void IsolatedRootKind::allTargets(const Context &Ctx, TargetsList &Out) const {
+  Out.push_back(Target(*this));
+}
```
aleclearmind, line 72: Isn't this the default implementation of `UniqueKind`? Can it be?

lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -66,3 +59,18 @@ void TaggedFK::getInvalidations(const Context &Ctx,
 
   // ToRemove.push_back(Target("0x1000", *this));
```
aleclearmind, line 60: Do not commit commented code!

lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -66,3 +59,18 @@ void TaggedFK::getInvalidations(const Context &Ctx,
...
+
+  for (const auto &Function : Model->Functions) {
+    Out.push_back(Target(Function.Entry.toString(), K));
+  }
+}
+
+void TaggedFunctionKind::allTargets(const pipeline::Context &Ctx,
+                                    pipeline::TargetsList &Out) const {
+  const auto &Model = getModelFromContext(Ctx);
+  getAllFunctions(Model, Out, *this);
```
aleclearmind, line 75: Inline function call

include/revng/Pipeline/Target.h
```diff
@@ -45,68 +43,42 @@ class ContainerBase;
 /// its extension.
 class Target {
```
aleclearmind, line 44: Update docs

include/revng/Pipeline/Target.h
```diff
@@ -191,7 +151,24 @@ class TargetsList {
...
   TargetsList() = default;
-  TargetsList(List C) : Contained(std::move(C)) { removeDuplicates(); }
+  TargetsList(List C) : Contained(std::move(C)) {}
+  static TargetsList allTargets(const Context &Ctx, const Kind &K) {
+    TargetsList ToReturn;
+    K.allTargets(Ctx, ToReturn);
+    return ToReturn;
+  }
+
+  bool operator==(const TargetsList &Other) const {
```
aleclearmind, line 161: Why do you need to implement this?

include/revng/Pipeline/Target.h
```diff
@@ -161,7 +119,7 @@ class Target {
   void dumpPathComponents(OStream &OS) const debug_function {
```
aleclearmind, line 119: The `dump` method is not dumping the kind.

Let's have a proper `fromString` and `toString` routines for targets.
Targets are an important actor in the pipeline, let's give them the right importance.

Also, we should either push to the users (e.g., CLI) the parser that understands `*` or, at least, give them a proper name to make sure people does not think that's the standard way to deserialize a target.

include/revng/Pipeline/Target.h
```diff
@@ -235,21 +212,73 @@ class TargetsList {
...
+  }
+
+  TargetsList copyEntriesOfKind(const Kind &K) const {
+    auto [b, e] = std::equal_range(begin(), end(), K, Comp());
+    TargetsList ToReturn;
+    copy(b, e, std::back_inserter(ToReturn));
+    return ToReturn;
+  }
+
+  llvm::iterator_range<iterator> entriesOfKind(const Kind &K) {
```
aleclearmind, line 260:
```suggestion
  llvm::iterator_range<iterator> filterByKind(const Kind &K) {
```

include/revng/Pipeline/Target.h
```diff
@@ -235,21 +212,73 @@ class TargetsList {
...
+  }
+
+  llvm::SmallVector<const Kind *, 4> getContainedKinds() const {
+    llvm::SmallVector<const Kind *, 4> ToReturn;
+
+    auto Current = begin();
+    while (Current != end()) {
+      ToReturn.push_back(&Current->getKind());
+      Current = std::upper_bound(Current, end(), Current->getKind(), Comp());
+    }
```
aleclearmind, line 277: I feel like this has the same complexity as simply going through the list and adding if not the same kind as the previous one, but with extra steps. Can we make it dumber?

include/revng/Pipeline/Target.h
```diff
@@ -235,21 +212,73 @@ class TargetsList {
...
+    Contained.erase(b, e);
+    return ToReturn;
+  }
+
+  TargetsList copyEntriesOfKind(const Kind &K) const {
+    auto [b, e] = std::equal_range(begin(), end(), K, Comp());
+    TargetsList ToReturn;
+    copy(b, e, std::back_inserter(ToReturn));
+    return ToReturn;
+  }
```
aleclearmind, line 258: I feel like these are too coupled with its users.
I'd consider simply having `removeByKind` and let users do `llvm::copy(entriesOfKind())`.

lib/Pipeline/Contract.cpp
```diff
@@ -10,325 +10,276 @@
...
+    return find(Kinds, this->TargetKind) != Kinds.end();
+  }
+
+  if (Source->depth() < TargetKind->depth()) {
+    return true;
+  }
+
+  if (Source->depth() > TargetKind->depth()) {
+    return true;
+  }
```
aleclearmind, line 209:
```suggestion
  if (Source->depth() == TargetKind->depth()) {
    auto Kinds = List.getContainedKinds();
    return find(Kinds, this->TargetKind) != Kinds.end();
  } else {
    return true;
  }
```

include/revng/Pipeline/Target.h
```diff
@@ -45,68 +43,42 @@ class ContainerBase;
 /// its extension.
 class Target {
 private:
+  using PathComponents = std::vector<std::string>;
   PathComponents Components;
```
aleclearmind, line 47: Can you take a look at how much effort would it be to switch to `Location`?
It'd be the Right Thing To Do™.

aleclearmind (changes requested):

include/revng/Pipeline/Target.h
drblallo: if there is one kind for each target it has linear complexity yes. if there are thousands of isolated functions and only 1 or 2 kinds, this has logarithmic complexity.I would like to keep it since it used in the most intensive part of the pipeline, which is contract.cpp

lib/Pipes/TaggedFunctionKind.cpp
drblallo: this was an example for antonio when  it started to work on invalidation. It is already gone in his branch

lib/Pipes/TaggedFunctionKind.cpp
drblallo: not sure what you mean. Inlining this function will do nothing, it is virtual

include/revng/Pipeline/Target.h
drblallo: contracts sometimes needs to check if a target list is exactly the one they expected. These lists are sorted, so it is like set equivalence.

lib/Pipes/RootKind.cpp
drblallo: unfortunately the other virtual function, getInvalidations, will be customized by the invalidation mr

lib/Pipes/TaggedFunctionKind.cpp
aleclearmind: Uh? I'm looking at this branch and it's going in first.

lib/Pipes/TaggedFunctionKind.cpp
aleclearmind: `getAllFunctions` is defined right above but introduces like 10 lines that could be inlined.

include/revng/Pipeline/Target.h
aleclearmind: How is this different from the `= default`?

include/revng/Pipeline/Target.h
aleclearmind: OK, but keep in mind that computing the set of things to compute will never be the bottleneck, so let's avoid early optimization.

lib/Pipes/RootKind.cpp
aleclearmind: You seem to assume the invalidations PR will go in first. This PR seems quite ready to go in to me.

In general, I didn't understand your answer. If these are `UniqueKind`, we can make them have this as the default implementation and then let people override it if needed.

include/revng/Pipeline/Target.h
drblallo: forgot it was possible to do in cpp20

lib/Pipes/RootKind.cpp
drblallo: ahh, i thought you wanted to replace RootKind with UniqueKind entirelly. What you propose works.

drblallo: RootKind extends LLVMKind, which is unrelated to UniqueKind. we can introduce a UniqueLLVMKind that implements allTargets, but this two would be the only two users. I don't see great utility in having them.

include/revng/Pipeline/Target.h
aleclearmind: Right! My bad.

PR #229, opened 2022-10-05
`TupleTree`: `visitReferences` with non-const visitors

pfez:


aleclearmind:
I trust you on this one, but is more of this coming?
Shall we have a PR for the caching/uncaching of TTRs?

pfez:
> I trust you on this one, but is more of this coming? Shall we have a PR for the caching/uncaching of TTRs?

This is unrelated with caching/uncaching, so if you don't have anything against it, I'd keep them separated.

The caching/uncaching needs a little more love, so I was planning to do a separate PR. I wouldn't block this from getting merged to wait for caching/uncaching. But your call.

aleclearmind:
If no one is revng/revng-c is using this, I'd attach this PR to the first PR doing that.

pfez:
Makes sense. I'll make a single one with the cache/uncache thing then.

pfez:
What about switching it to "wip"? I'll update it with the rest of the commits later on and re-request a review.

aleclearmind:
The branch name is not appropriate, I think.
In any case, more than `wip` you can draft it.

pfez:
I'm gonna close this. I hoped github would let me change the source branch of the PR but apparently it does not. I'll open a new one with the correct branch.

PR #231, opened 2022-10-05
Expose Invalidations to GraphQL

mrjackv:


aleclearmind:
I'm confused by this PR. It has a non-negligible overlap with #215, which is going in soon.
If you want to have a PR based on another one, state in the PR message so I can push the branch of the other PR and we can let this PR target that branch.
Otherwise reviewing becomes confusing.

python/revng/api/__init__.py
```diff
@@ -12,6 +11,8 @@
 from .manager import Manager
 from .rank import Rank
 
+__all__ = ["Manager", "Rank", "initialize", "shutdown"]
+
```
aleclearmind, line 15: I guess this is to expose less stuff to users?

include/revng/Pipeline/Global.h
```diff
@@ -32,15 +32,22 @@ class Global {
...
   virtual GlobalTupleTreeDiff diff(const Global &Other) const = 0;
-
   virtual llvm::Error applyDiff(const llvm::MemoryBuffer &Diff) = 0;
+  virtual llvm::Error applyDiff(const GlobalTupleTreeDiff &Diff) = 0;
+
   virtual llvm::Error serialize(llvm::raw_ostream &OS) const = 0;
   virtual llvm::Error deserialize(const llvm::MemoryBuffer &Buffer) = 0;
+  virtual llvm::Expected<GlobalTupleTreeDiff>
+  deserializeDiff(const llvm::MemoryBuffer &Diff) = 0;
+  virtual void assign(const Global &NewGlobal) = 0;
```
aleclearmind, line 42: `operator=`?

include/revng/Pipeline/Global.h
```diff
@@ -114,6 +126,16 @@ class TupleTreeGlobal : public Global {
     return llvm::Error::success();
   }
 
+  llvm::Error applyDiff(const GlobalTupleTreeDiff &Diff) override {
+    Diff.getAs<Object>()->apply(Value);
```
aleclearmind, line 130: No return value? Cannot fail?

include/revng/PipelineC/Prototypes.h
```diff
@@ -8,14 +8,25 @@
...
+/**
+ * INVALIDATIONS
+ * If a user of the pipeline wishes to cache the output of
+ * rp_manager_produce_targets it can do so, however some modifications can lead
+ * to a produced output to be invalid. To help with this there is
+ * rp_invalidations, which if provided to the methods that have it in their
+ * argument, will fill in all the targets that have been invalidated by the
+ * operation. If the user is not interested in having invalidation information
+ * the parameter can be set to nullptr.
+ */
```
aleclearmind, line 37: Move this to the doxygen of one of the functions that has a `rp_invalidations *invalidations` argument and refer to it in the other functions.

mrjackv:
The situation was: I need uvicorn on this branch as well because websockets did not work with hypercorn, but I wasn't sure if #215 would get merged before this, so I left the uvicorn commit in.
The first commit is a fix otherwise rcc would fail.
The "real" PR starts with the third commit, if you prefer I can draft this and undraft it once #215 goes in

lib/PipelineC/PipelineC.cpp
```diff
@@ -30,6 +31,35 @@
 using namespace pipeline;
 using namespace ::revng::pipes;
 
+template<typename T>
+concept default_constructible = std::is_default_constructible<T>::value;
+
+template<default_constructible T>
+class Wrapper {
```
aleclearmind, line 38: Not a great name. `ExistingOrTemporary`? `ExistingOrNew`?

lib/Pipes/PipelineManager.cpp
```diff
@@ -382,19 +390,50 @@ llvm::Expected<DiffMap>
...
+llvm::Expected<InvalidationMap>
+PipelineManager::invalidateFromDiff(const llvm::StringRef Name,
+                                    const pipeline::GlobalTupleTreeDiff &Diff) {
+  InvalidationMap Map;
+  if (auto ApplyError = getRunner().apply(Diff, Map); !!ApplyError)
+    return std::move(ApplyError);
+
+  // TODO: to remove once invalidations are working, uncomment line below
+  return invalidateAllPossibleTargets();
+  // return Map;
```
aleclearmind, line 438: No commented code

python/revng/api/invalidations.py
```diff
@@ -0,0 +1,30 @@
...
+class Invalidations:
+    def __init__(self, invalidations=None):
+        if invalidations is None:
+            self._invalidations = _api.rp_invalidations_create()
+        else:
+            self._invalidations = invalidations
+
+    def __str__(self):
+        res = _api.rp_invalidations_serialize(self._invalidations)
+        return make_python_string(res)
```
aleclearmind, line 21: Would be nice to have a list.

include/revng/PipelineC/Prototypes.h
```diff
@@ -7,6 +7,7 @@
 #include "PipelineC.h"
 
 /**
+ * OWNING POINTERS
```
aleclearmind, line 10: There's probably a Doxygen way to do this other than all caps. Can you look it up?

lib/PipelineC/PipelineC.cpp
```diff
@@ -176,13 +176,6 @@ rp_manager *rp_manager_create(uint64_t pipelines_count,
                                 execution_directory);
 }
 
-rp_manager *rp_manager_create_memory_only(const char *pipeline_path,
-                                          uint64_t flags_count,
-                                          const char *flags[]) {
-  const char *paths[1] = { pipeline_path };
-  return rp_manager_create(1, paths, flags_count, flags, "");
-}
-
```
aleclearmind, line 155: Own commit?

python/revng/daemon/static_handlers.py
```diff
@@ -128,6 +129,15 @@ async def resolve_targets(_, info, *, pathspec: str):
     return result
 
 
+@query.field("target")
+async def resolve_target(_, info, *, step: str, container: str, target: str) -> Optional[Target]:
+    manager: Manager = info.context["manager"]
+    result = await run_in_executor(
+        lambda: manager.deserialize_target(f"{step}/{container}/{target}")
+    )
+    return result.as_dict() if result is not None else None
```
aleclearmind, line 138: Not sure what this does.

aleclearmind (changes requested):

python/revng/daemon/static_handlers.py
mrjackv: It's the handler to retrieve a single target, I use it to detect if a binary has been uploaded by doing:
```
{
  target(step: "begin", container: "input", target: ":Binary") {
    ready
  }
}
```
if ready is `true` then the target exists otherwise I get `target: null`

python/revng/api/__init__.py
mrjackv: It does 2 things:
* flake8 will not complain if those symbols are unused
* if you do `from revng.api import *` it will only import those symbols

lib/PipelineC/PipelineC.cpp
mrjackv: `ExistingOrDefault`?

lib/Pipes/PipelineManager.cpp
mrjackv: Dropped

include/revng/PipelineC/Prototypes.h
mrjackv: Writing here for this issue and the block with `INVALIDATIONS`
I've pushed changes that should generate a separate page for revng's PipelineC conventions, so they are all in one place, together with this all functions that use invalidation and/or errors now have a doxygen link that, if clicked, will bring the user to the conventions page, highlighting the appropriate heading.

python/revng/api/invalidations.py
mrjackv: Eh, rp_invalidations under the hood is implemented as:
```
llvm::StringMap<llvm::StringMap<std::vector<pipeline::Target>>>
```
So having an enumeration is, while not impossible, somewhat tricky.
Moreover I'd argue that it's useless, if you want validations you want them all, so you call `serialize`
Moreover I'd be against serializing on the python side, that would turn a single ffi call in (in a trivial case) a few thousands

include/revng/Pipeline/Global.h
mrjackv: Added

python/revng/api/invalidations.py
aleclearmind: Mmmh, OK.

include/revng/PipelineC/Prototypes.h
```diff
@@ -7,29 +7,52 @@
 #include "PipelineC.h"
 
 /**
+ * \page pipelineC_convention PipelineC API conventions
```
aleclearmind, line 10: We're not really using Doxygen for this kind of stuff. If you want to have a "page" we should move this stuff to `mkdocs`.

Can you add a `// TODO: move to mkdocs` so once that gets in we can move it there?

mrjackv: Ok, altough it'd be convenient if there was a way to link the mkdocs stuff from doxygen (so that we'd keep the links in the function descriptions)

aleclearmind: As you wish, but I think our goal is to use Doxygen strictly for documenting the API. For all the rest I want to use `mkdocs`. We can references stuff from Doxygen to mkdocs, eventually.

include/revng/Pipeline/Global.h
mrjackv: Reworked to populate `ErrorList` wit appropriate errors

PR #234, opened 2022-10-21
Fix pipeline drivers targetless invocation

drblallo:
Analyze and artifact crashed when targets were not specified


aleclearmind:
Merged in 6a86973. Thanks!

PR #238, opened 2022-10-26
Add cache to avoid parsing metadata in hotloop

drblallo:


aleclearmind:
We need a pass caching stuff in itself for all those interested in `extractFunctionMetadata`.

include/revng/EarlyFunctionAnalysis/IRHelpers.h
```diff
@@ -31,70 +32,126 @@ extractFunctionMetadata(llvm::MDNode *MD) {
...
-  efa::FunctionMetadata FM = *extractFunctionMetadata(ParentFunction).get();
-  efa::BasicBlock Block = FM.ControlFlowGraph.at(BlockAddress);
-
-  // Find the call edge
-  efa::CallEdge *ModelCall = nullptr;
-  for (auto &Edge : Block.Successors) {
-    if (auto *CE = dyn_cast<efa::CallEdge>(Edge.get())) {
-      revng_assert(ModelCall == nullptr);
-      ModelCall = CE;
+class MetadataCache {
```
aleclearmind, line 49: * `FunctionMetadataCache`?
* Dedicated header

aleclearmind:
> under all practical aspects using metaadress a key is strictly worst, i just used llvm::Value instead. Should we ever notice that the extra copies are a issue we can always use the right key.

OK, but consider that we *will* have to change this once we will move this to its own container. If we only expose a `MetaAddress` API we won't have to change it and we might (at least in some places) reuse the `MetaAddress` floating around (no need to extract it from the IR).

This is more of a concern to me than performance, unless you have proof that this actually impacts performance significantly.

Anyway, we can reconsider this in the future.

aleclearmind:
Merged in 0bfb73f. Thanks!

PR #239, opened 2022-10-26
Changes for revng-distributable-revamp

mrjackv:


python/revng/cli/_commands/daemon_self_test.py
```diff
@@ -83,9 +83,8 @@ def log(string: str):
     def fail(self, ex: Exception):
         if self.process is not None:
             self.process.terminate()
-            self.process.communicate()
-            if self.process.stdout is not None:
-                self.log(f"\n\n Daemon Stacktrace:\n{self.process.stdout.read()}")
+            stdout = self.process.communicate()[0]
+            self.log(f"\n\nDaemon Stacktrace:\n{stdout}")
```
aleclearmind, line 87: `"\n\nrevng-daemon output:\n`

lib/Support/PathList.cpp
```diff
@@ -35,7 +35,11 @@ std::string getCurrentRoot() {
   Maps.split(MapsLines, "\n");
 
   for (const auto &Line : MapsLines) {
-    llvm::StringRef File = std::get<1>(Line.rsplit(" "));
+    llvm::SmallVector<llvm::StringRef> Parts;
+    Line.split(Parts, ' ', 5);
+    revng_assert(Parts.size() == 6, "Malformed /proc/self/maps");
+
+    llvm::StringRef File = Parts[5].ltrim();
```
aleclearmind, line 42: Mmmh, this does not fix the problem described in the commit message.
You should get anything that comes after the 4th space.

aleclearmind (changes requested):

lib/Support/PathList.cpp
aleclearmind: Ah-ah! Didn't know that! My bad!

python/revng/cli/_commands/daemon_self_test.py
mrjackv: Fixed

aleclearmind:
Merged in 91c4bdc. Thanks!

PR #240, opened 2022-10-28
Fix pipeline logger crash

drblallo:


aleclearmind:
Merged in 10ab396. Thanks!

PR #232, opened 2022-10-06
Improve alignment support when converting function types

Cvelth:

PR #241, opened 2022-10-31
Fix a plethora of minor layout issues blocking cabi prototype adoption

Cvelth:


include/revng/EarlyFunctionAnalysis/ControlFlowGraph.h
```diff
@@ -4,6 +4,8 @@
 // This file is distributed under the MIT License. See LICENSE.md for details.
 //
 
+#pragma clang optimize off
+
```
aleclearmind, line 8: Aaaah! I'm a monster!

include/revng/Model/Register.h
```diff
@@ -681,6 +682,242 @@ fromCSVName(llvm::StringRef Name, model::Architecture::Values Architecture) {
...
+  case f7_systemz:
+  case f8_systemz:
+  case f9_systemz:
+  case f10_systemz:
+  case f11_systemz:
+  case f12_systemz:
+  case f13_systemz:
+  case f14_systemz:
+  case f15_systemz:
+    return model::PrimitiveTypeKind::Float;
```
aleclearmind, line 909: NOTETOSELF: cry

include/revng/ADT/STLExtras.h
```diff
@@ -339,3 +339,23 @@ replaceAll(std::string &Input, const std::string &From, const std::string &To) {
...
+constexpr decltype(auto) find(R &&Range, const T &Value) {
+  return std::find(std::begin(Range), std::end(Range), Value);
+}
+
+/// \note use `llvm::is_contained` instead after it's made `constexpr`.
+template<typename R, typename T> // NOLINTNEXTLINE
+constexpr bool is_contained(R &&Range, const T &Value) {
+  return revng::find(Range, Value) != std::end(Range);
+}
+
```
aleclearmind, line 360: Can I have two `static_assert` quick tests here?

include/revng/ABI/Trait.h
```diff
@@ -237,9 +237,6 @@ concept TraitLike = requires(TraitType<ABI> Trait) {
   /// ARM ABI), the value of `StackAlignment` should be equal to 4.
   { Trait.StackAlignment } -> convertible_to<size_t>;
 
-  /// States the required size for a single stack argument in bytes.
-  { Trait.MinimumStackArgumentSize } -> convertible_to<size_t>;
-
```
aleclearmind, line 242: Can you briefly explain in the commit message why we're doing this?

lib/ABI/FunctionType.cpp
```diff
@@ -727,31 +725,46 @@ class ConversionHelper {
...
 
       constexpr bool CanSplit = AT::ArgumentsCanBeSplitBetweenRegistersAndStack;
       if (Argument.Type.isFloat()) {
+        // The conventional non-position based approach is not applicable for
+        // vector registers since it's rare for multiple registers to be used
+        // to pass a single argument.
+        //
+        // For now, just return the next free register.
+        //
+        // TODO: handle this properly.
```
aleclearmind, line 743: I don't understand. Can you make an example of what situation we're not handling?

aleclearmind (changes requested):

include/revng/Model/Register.h
Cvelth: I thought for a really long time for a better way, sadly this is what it came down to 😢

lib/ABI/FunctionType.cpp
Cvelth: For example, `float128`s under `s390x` are supposed to be returned on the stack instead of the register.

Same for more typical architectures. If, for example under SystemV_x64, you try to return `float128` on an architecture that doesn't have vector registers big enough to hold it (most of the floating point stuff are declared as extensions after all), the behaviour should be the same, as in - use stack.

But since for now we have no way to declare something as specific as available extensions (within the model, abi trait or whatnot), I decided to just go with "handle the most common situation, fail miserable with the rest" plus "leave a TODO" approach.

include/revng/EarlyFunctionAnalysis/ControlFlowGraph.h
Cvelth: I didn't even notice it was there until you pointed it out :)

include/revng/ADT/STLExtras.h
Cvelth: Done: https://github.com/revng/revng/pull/241/commits/e0595113289db4bf506c6f2b2c9792463fa1c730

include/revng/ABI/Trait.h
Cvelth: Something like this? https://github.com/revng/revng/pull/241/commits/8443c3196320618f238689ed0473481826371937

lib/ABI/FunctionType.cpp
aleclearmind: Uhm, I think the best approach would be to assume a certain extension being present and that's it.
In the future, we'll enable variants, but for now I think we should just pick one and stick with it.

For instance, as of now, we assume we have `xmm0`, so we're already choosing an ABI, I think?

Cvelth: But the "real" size of the `xmm` is also extension dependent. It could be 128 bit (`xmm`), or 256 (`ymm`), or 512 (`zmm`). If we try to return a value we have no register for under the current architecture (including extensions), we return it on the stack instead.

Cvelth: Also, under the `SystemV_x86_64` there is a `long double` (80-bit) beast which is returned in `stN` registers :cry:

aleclearmind: We say `xmm` so we assume 128-bits in `Register.h`. Let's stick with that.

> Also, under the SystemV_x86_64 there is a long double (80-bit) beast which is returned in stN (x87) registers cry

I would say we emit a message on the logger and pretend that return value doesn't exist.
If it's not in `Register.h`, we're ignoring it anyway.

include/revng/EarlyFunctionAnalysis/ControlFlowGraph.h
aleclearmind: Still here

include/revng/ADT/STLExtras.h
```diff
@@ -339,3 +339,26 @@ replaceAll(std::string &Input, const std::string &From, const std::string &To) {
     Start += To.length();
   }
 }
+
```
aleclearmind, line 342: Typo in commit message `s/constepxr/constexpr/`

aleclearmind (changes requested):

aleclearmind:
This PR is pending fixing all the issues that the changes in the first two commits triggered.
The ball is now in @pfez and my hands.

include/revng/EarlyFunctionAnalysis/ControlFlowGraph.h
Cvelth: It's gone now.

include/revng/ADT/STLExtras.h
Cvelth: Fixed: https://github.com/revng/revng/pull/241/commits/7992f1ccb944148fb0d84ff480a543d974d71bd8

lib/ABI/FunctionType.cpp
Cvelth: Since this is still unresolved when the branch got merged, let's move it to #232. I'll recreate the thread there and link to this one, allowing this one to get closed.

aleclearmind:
Merged in 0ed1c6f. Thanks!

PR #242, opened 2022-10-31
Fix artifact emission crash

drblallo:
When artifact was requested to produce the functions of a binary that had 0 functions it crashed. This commit fixes it by checking if something has been actually being produced.


aleclearmind:
Merged in e24fdbd. Thanks!

PR #244, opened 2022-11-02
Refactor PTML C

pfez:
This PR contains a few preliminary changes that are necessary for a larger refactor of the PTML C generation, that happens later in revng-c decompilation pipeline.

### DIFF
--- a/lib/Model/Type.cpp
+++ b/lib/Model/Type.cpp
@@ -727,6 +727,23 @@ bool QualifiedType::isPointer() const {
   return isPointerImpl(*this);
 }
 
+static RecursiveCoroutine<bool> isConstImpl(const model::QualifiedType &QT) {
+  if (not QT.Qualifiers.empty()) {
+    // If there are qualifiers, just look at the first
+    rc_return Qualifier::isConst(QT.Qualifiers.front());
+  } else if (auto *TD = dyn_cast<model::TypedefType>(QT.UnqualifiedType.get())) {
+    // If there are no qualifiers, but it's a typedef, traverse it
+    rc_return rc_recur isConstImpl(TD->UnderlyingType);
+  }
+
+  // If there are no qualifiers, and it's not a typedef, it's not const.
+  rc_return false;
+}
+
+bool QualifiedType::isConst() const {
+  return isConstImpl(*this);
+}
+
 static RecursiveCoroutine<bool>
 isPrimitiveImpl(const model::QualifiedType &QT,
                 model::PrimitiveTypeKind::Values V) {
--- a/include/revng/ADT/UpcastablePointer.h
+++ b/include/revng/ADT/UpcastablePointer.h
@@ -57,7 +57,7 @@ concept NotUpcastablePointerLike = not UpcastablePointerLike<T>;
 // clang-format off
 template<typename ReturnT, typename L, UpcastablePointerLike P, size_t I = 0>
   requires(not std::is_void_v<ReturnT>)
-ReturnT upcast(P &&Upcastable, const L &Callable, const ReturnT &IfNull) {
+ReturnT upcast(P &&Upcastable, L &&Callable, const ReturnT &IfNull) {
   // clang-format on
   using pointee = std::remove_reference_t<decltype(*Upcastable)>;
   using concrete_types = concrete_types_traits_t<pointee>;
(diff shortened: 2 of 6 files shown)


include/revng/PTML/Constants.h
```diff
@@ -23,7 +23,12 @@ inline constexpr auto ModelEditPath = "data-model-edit-path";
 
 } // namespace attributes
 
-namespace scopes {} // namespace scopes
+namespace scopes {
+
+// Default name for a foldable scope in PTML with no other meaning
+inline constexpr auto Generic = "ptml.generic.scope";
```
aleclearmind, line 29: 1. Why introduce this on the `revng` side?
2. You sure you can't come up with a name for a scope? What's the use case?

lib/Model/Type.cpp
```diff
@@ -727,6 +727,23 @@ bool QualifiedType::isPointer() const {
   return isPointerImpl(*this);
 }
 
+static RecursiveCoroutine<bool> isConstImpl(const model::QualifiedType &QT) {
+  if (not QT.Qualifiers.empty()) {
+    // If there are qualifiers, just look at the first
+    rc_return Qualifier::isConst(QT.Qualifiers.front());
+  } else if (auto *TD = dyn_cast<model::TypedefType>(QT.UnqualifiedType.get())) {
+    // If there are no qualifiers, but it's a typedef, traverse it
+    rc_return rc_recur isPointerImpl(TD->UnderlyingType);
```
aleclearmind, line 736: `isPointerImpl`?

aleclearmind (changes requested):

include/revng/PTML/Constants.h
aleclearmind: 2. What's the *specific* usage you need it for as of now? If we accept this, we could drop scope *names* (not scope themselves) entirely, which might actually be a good idea.
   We already have `data-location-*` to assign meaning to things. What's the use of having a `asm.basic-block` scope? It seems redundant to the associated `data-location-definition`. @mrjackv what's your take?

mrjackv: My general take on defining scopes was: "we already are putting the attribute, might as well describe what the scope is about"
Right now (with the exception of `asm.instruction`) the're all treated equally, but maybe we might want to have things like "Collapse all unions" while in `model_header.h` which does require knowing which scope is which

aleclearmind: @pfez On a second thought, let's keep the PTML scopes and let's ban any "generic" scope.
If you think there are scopes that are really really "generic", please bring an example. I suspect every scope can get a decent name.

pfez: The scope I'm using this for is just a bunch of lines in C code that I want to be foldable, in particular in the decompiled C header we use for types, but I'm planning to use it also in the body of the functions. I have a few cases that could all fall under this umbrella like: "includes", "typedefs for primitive types", "forward declarations", "body of a struct definition", but it could even be used for "loop body". Do we really need to use different names for all these?

aleclearmind: If you want to be able to let the user say "these type of foldable areas need to be collapsed by default", yes.
I think I'd like that for "includes", "typedefs for primitive types" and "forward declarations".
"body of a struct definition": we have it for function bodies, why not for struct bodies?
"loop body": could just be a `c.scope` or a `c.control-statement`.

Point is: either we drop them entirely, or we stick with giving them sensible names. Going halfway is a sad route.

pfez: I agree, but at this point I'm quite in favor of dropping them entirely. Having names right now is "just in case". I'm not sure it's worth it. What's your take?

aleclearmind: https://mattermost.rev.ng/revng/pl/615hb7rejprdfphqwj65yuj4do

lib/Model/Type.cpp
pfez: Fixed!

include/revng/PTML/Constants.h
pfez: I dropped the generic scope. I'll add the sensibly-named scopes in revng-c.

aleclearmind:
Merged in 764c7ef. Thanks!

PR #245, opened 2022-11-03
Fix tools empty input crash

drblallo:
Fix various edge cases when a binary with no functions was analyzed


drblallo:
related to https://rev.ng/gitlab/revng-private/revng-c/-/merge_requests/120
and https://rev.ng/gitlab/revng-private/revng-c/-/merge_requests/120

aleclearmind:
Please replace `FileContainer` with the new `StringBufferContainer` wherever appropriate.

drblallo:
> Please replace `FileContainer` with the new `StringBufferContainer` wherever appropriate.

i looked into it, the only uses that are left are InputFileContainer, ObjectFileContainer and TranslatedFileContainer. The first is useful to keept  it around that way, and the latter two are needed because we link on disk.

CrossRelations will be replaced by antonio's merge request, so in the end the only one worth replacing was the SVG one.

aleclearmind:
Merged in 0440714. Thanks!

PR #246, opened 2022-11-09
Fix revng analysis crash

drblallo:
Ensures that only content that is actually inside a container is elegible to be removed


aleclearmind:
Cherry-picked on my private branch. Thanks!

PR #248, opened 2022-11-22
Remove ErrorList and introduce DiffError

drblallo:


include/revng/Support/Errors.h
```diff
@@ -0,0 +1,51 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Support/Error.h"
+
+namespace revng {
+
+class DiffError : public llvm::ErrorInfo<DiffError> {
```
aleclearmind, line 13: Move to `include/revng/TupleTree/DiffError.h`

include/revng/ADT/UpcastablePointer.h
```diff
@@ -57,7 +57,30 @@ concept NotUpcastablePointerLike = not UpcastablePointerLike<T>;
...
+                                          std::forward<ReturnT>(IfNull));
+    }
+  } else {
+    revng_abort();
+  }
+}
+
+template<typename L, UpcastablePointerLike P, size_t I = 0>
+llvm::Error upcast(P &&Upcastable, const L &Callable, llvm::Error IfNull) {
   // clang-format on
```
aleclearmind, line 84: Can we have this change in its own commit?

lib/Support/Errors.cpp
```diff
@@ -0,0 +1,46 @@
...
+  if (!Error)
+    return;
+
+  std::string S;
+  llvm::raw_string_ostream OS(S);
+  OS << Error;
+  OS.flush();
+  llvm::consumeError(std::move(Error));
+  addReason(S);
+}
```
aleclearmind, line 46: Here we're slurping *any* errors AFAIU. I think we should slurp only errors of the same type.
Otherwise this is not `DiffError` but `ErrorList` again.

aleclearmind (changes requested):

lib/Support/Errors.cpp
aleclearmind: Uhm, seems like your design is to reuse this for non-diff-related purposes.
Do you want to use it for, say, reporting validation errors of `Binary::verify`? If so, I'm fine with keeping this in `librevngSupport` but it needs to be renamed to... `StringErrorList`?

drblallo: probably jack has some opinions too about this

include/revng/TupleTree/DiffError.h
```diff
@@ -0,0 +1,162 @@
...
+    RevngErrorBase(Ids) {
+    addReason(Reason, ReasonLocation);
+  }
+
+  void addReason(llvm::StringRef NewReason, const Location &ReasonLocation) {
+    Reasons.push_back(NewReason.str());
+    Locations.push_back(std::make_unique<Location>(ReasonLocation));
+  }
+
+  void consumeAndAppend(llvm::Error Error, const Location &ReasonLocation) {
```
aleclearmind, line 139: I see no reason to eat other `llvm::Error`s, this is not a collection of `llvm::Error`s but one error with multiple reasons and locations.
If the caller wants to do this, it should be his responsibility (maybe with an helper function), but I don't see this as a feature of `RevngError`.

lib/PipelineC/PipelineC.cpp
```diff
@@ -617,6 +617,24 @@ const char *rp_manager_get_global_name(rp_manager *manager, int index) {
   return nullptr;
 }
 
+static void errorToStrings(llvm::Error Error, std::vector<rp_error> &Strings) {
+  // clang-format off
+  llvm::handleAllErrors(std::move(Error),
+    [&](const revng::DiffError &Error) {
```
aleclearmind, line 623: Why `DiffError`? We did all these tricks to be able to use a base class here, don't we?

include/revng/PipelineC/ForwardDeclarations.h
```diff
@@ -18,6 +26,6 @@ typedef const pipeline::Target rp_target;
 typedef const pipeline::TargetsList rp_targets_list;
 typedef const pipeline::Step::AnalysisValueType rp_analysis;
 typedef const pipeline::DiffMap rp_diff_map;
-typedef revng::ErrorList rp_error_list;
+typedef std::vector<rp_error> rp_error_list;
```
aleclearmind, line 29: In the [design pad](https://pad.rev.ng/vK82o4YtRXKv2A_ctMjHKg#), `rp_error_list` also exposed `getLocationTypeName` and `getErrorTypeName`. Can we do that?

include/revng/TupleTree/DiffError.h
```diff
@@ -0,0 +1,162 @@
...
+    llvm::consumeError(std::move(Error));
+    addReason(S, ReasonLocation);
+  }
+
+public:
+  // Returns the class ID for this type.
+  static const void *classID() { return &Derived::ID; }
+};
+
+class DiffError : public RevngError<DiffError, DiffLocation<model::Binary>> {
```
aleclearmind, line 156: `RevngError` deserves its own header.

include/revng/TupleTree/DiffError.h
```diff
@@ -0,0 +1,162 @@
...
+
+  [[nodiscard]] const_iterator begin() const { return Reasons.begin(); }
+
+  [[nodiscard]] const_iterator end() const { return Reasons.end(); }
+
+  size_t size() const { return Reasons.size(); }
+
+  const std::string &getMessage(size_t I) const { return Reasons[I]; }
+  std::string getLocation(size_t I) const { return Locations[I]->toString(); }
+
```
aleclearmind, line 94: Can we have the [two string discriminators](https://pad.rev.ng/vK82o4YtRXKv2A_ctMjHKg#) of the error type?
```c++
  virtual string getLocationTypeName();
  virtual string getErrorTypeName();
```

aleclearmind (changes requested):

include/revng/ADT/UpcastablePointer.h
```diff
@@ -52,27 +52,50 @@ template<typename T>
...
+                                          Callable,
+                                          std::forward<ReturnT>(IfNull));
+    }
+  } else {
+    revng_abort();
+  }
+}
+
+template<typename L, UpcastablePointerLike P, size_t I = 0>
+llvm::Error upcast(P &&Upcastable, const L &Callable, llvm::Error IfNull) {
```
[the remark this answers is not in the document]
drblallo: i don't mind but doing so will affect half of the visits of the codebase, do i do it in this mr?

include/revng/TupleTree/DiffError.h
```diff
@@ -0,0 +1,162 @@
...
+};
+
+class RevngErrorBase : public llvm::ErrorInfo<RevngErrorBase> {
+public:
+  inline static char ID = '0';
+
+  using iterator = llvm::SmallVector<std::string, 4>::iterator;
+  using const_iterator = llvm::SmallVector<std::string, 4>::const_iterator;
+
+  using IDContainer = llvm::SmallVector<const void *, 4>;
```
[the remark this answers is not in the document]
drblallo: i can't, classof returns a void*

include/revng/TupleTree/DiffError.h
```diff
@@ -0,0 +1,162 @@
...
+  [[nodiscard]] const_iterator end() const { return Reasons.end(); }
+
+  size_t size() const { return Reasons.size(); }
+
+  const std::string &getMessage(size_t I) const { return Reasons[I]; }
+  std::string getLocation(size_t I) const { return Locations[I]->toString(); }
+
+  virtual bool isA(const void *const ClassID) const {
+    return llvm::find(Ids, ClassID) != Ids.end();
+  }
```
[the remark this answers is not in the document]
drblallo: i can't do the range trick, it works only if your hierarchy is not extensible from downstream. Same issue of standard dyn_cast.

lib/PipelineC/PipelineC.cpp
drblallo: yeah, this was an example, we can handle revngErrorBase directly if that looks correct to you

aleclearmind: Yes! This kinda is point of all these castles `:)`

include/revng/TupleTree/DiffError.h
drblallo: i am fiddling with this one but there is a annoying thing. When you create a error you do `llvm::createError<ErrorType>`  which returns a llvm::Error not a ErrorType, thus it's very uncomfortable to accumulate errors, because  you can't cast a error without consuming it.

aleclearmind: I didn't express myself clearly: I don't want this method not to consume the error, I don't want this method at all `:D`

`RevngError` is not a bin of arbitrary errors, it's a single error multiple reasons and locations.
I'd like to keep this very simple and only let people add reasons and locations in a plain way, without involving any other `llvm::Error`.

drblallo: the issue is that we do need it something like this

if you have a situation like 

```cpp
llvm::Error maybeError = llvm::Error::success();

while (something())
{
  auto maybeError2 = somethingFailable();
  if (maybeError2)
    maybeError = compose(std::move(maybeError), std::move(maybeError2));

}

if (maybeError)
  return std::move(maybeError);
```

aleclearmind: You can just use a vector of reasons and create the error at the end.
Our error is not composable with other errors, it's just an error with multiple reasons.

We could force to be composable only with error with the same type, but it's just simpler to have a vector and construct the error at the end.

The way we will use this is to have some sort of diagnostics class where we accumulate problems, we won't be combining things as you're suggesting.

include/revng/TupleTree/TupleTreeDiff.h
```diff
@@ -299,10 +304,9 @@ struct llvm::yaml::MappingTraits<T> {
       IO.mapRequired("Path", SerializedPath);
       auto MaybePath = stringAsPath<Model>(SerializedPath);
       if (!MaybePath.has_value()) {
-        ::ErrorList *EL = static_cast<::ErrorList *>(IO.getContext());
+        auto *EL = static_cast<revng::DiffError *>(IO.getContext());
         std::string ErrorMessage = "Path " + SerializedPath + " is invalid";
-        EL->push_back(llvm::createStringError(llvm::inconvertibleErrorCode(),
-                                              ErrorMessage));
+        EL->addReason(ErrorMessage, revng::DiffError::LocationType(*MaybePath));
```
aleclearmind, line 309: You can't inspect `MaybePath` in an `if` where you verified that it does not have a value.

aleclearmind (changes requested):

PR #250, opened 2022-12-01
Private model accessors

drblallo:


include/revng/Support/Assert.h
```diff
@@ -199,3 +210,15 @@ void setAbortHook(AbortHook Hook);
 #ifdef __cplusplus
 }
 #endif
+
+#if REVNG_ASSERT_IS_MACRO == 0 && defined(__cplusplus)
+template<typename Arg>
+inline void revng_assert(Arg &&Cond) {
```
aleclearmind, line 216: * Why not `bool` instead of `Arg`?
* No abbreviations

include/revng/Support/Assert.h
```diff
@@ -199,3 +210,15 @@ void setAbortHook(AbortHook Hook);
...
+  revng_assert_hidden(Cond);
+}
+
+template<typename Arg, typename Message>
+inline void revng_assert(Arg &&Cond, Message &&Msg) {
```
aleclearmind, line 221: Inner functions assume `const char *` instead of `Message`.

include/revng/Support/Assert.h
```diff
@@ -28,6 +28,8 @@ extern "C" {
 #define __builtin_unreachable(x)
 #endif
 
+#define REVNG_ASSERT_IS_MACRO 0
```
aleclearmind: Make a CMake option?

include/revng/TupleTree/TupleTreeReference.h
```diff
@@ -133,7 +133,6 @@ class TupleTreeReference {
 
 public:
   const T *get() const {
-    static_assert(TupleTreeCompatible<RootType>);
```
aleclearmind, line 136: Can you comment in the commit message the ratio for this?

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
```diff
@@ -26,24 +26,33 @@ struct /*= struct | fullname =*/
   /**- if struct.inherits **/
   static constexpr const /*= struct.inherits.name =*/Kind::Values AssociatedKind = /*= struct.inherits.name =*/Kind::/*= struct.name =*/;
   /**- endif **/
+  static const constexpr bool MODIFY_THIS_ONE = true;
```
aleclearmind, line 29: ?

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
```diff
@@ -32,9 +32,13 @@ struct /*= struct | fullname =*/
...
   /*= field.doc | docstring =*/
   /**- if field.const **/const /** endif -**/
+  private:
   /*= field | field_type =*/ The/*= field.name =*/ = /*= field | field_type =*/{};
   static_assert(Yamlizable</*= field | field_type =*/>);
 
+  public:
+  using /*= field.name =*/Type = /*= field | field_type =*/;
+
   const /*= field | field_type =*/ & /*= field.name =*/() const {
```
aleclearmind, line 42: Add another public here to group things nicely

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
```diff
@@ -32,9 +32,13 @@ struct /*= struct | fullname =*/
   /**- for field in struct.fields **/
   /*= field.doc | docstring =*/
   /**- if field.const **/const /** endif -**/
+  private:
   /*= field | field_type =*/ The/*= field.name =*/ = /*= field | field_type =*/{};
   static_assert(Yamlizable</*= field | field_type =*/>);
 
+  public:
```
aleclearmind, line 39: De-indent `private` and `public`

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct_late.h.tpl
```diff
@@ -94,7 +94,7 @@ struct llvm::yaml::ScalarTraits</*= struct | user_fullname =*/::Key>
 /** if struct.keytype == "simple" **/
 template<>
 struct KeyedObjectTraits</*= struct | user_fullname =*/> {
-  static /*= struct.key_fields[0] | field_type =*/ key(const /*= struct | user_fullname =*/ &Obj) { return Obj.The/*= struct.key_fields[0].name =*/; }
+  static /*= struct.key_fields[0] | field_type =*/ key(const /*= struct | user_fullname =*/ &Obj) { return Obj./*= struct.key_fields[0].name =*/(); }
```
aleclearmind, line 97: This seems like a fixup change, squash?
I see other instances of this.

aleclearmind (changes requested):

aleclearmind:
1. Cleanup commit messages
2. revng is not the right place for the rewriter, we can discuss what to do with it, it's quite useful.
3. Can you anticipate the commit introducing `private` fields?
4. How does YAMLTraits work without public fields? In MappingTratis you simply pass `CustomName()` instead of `CustomName`?
5. In `Introduce model accessors` I see changes to `scripts/tuple_tree_generator/tuple_tree_generator/templates/*.tpl`. Please isolate infrastructural changes into the commit
   introducing `private` fields.
6. `CustomName()` does not respect our coding conventions, I'd expect to see `clang-tidy off` somewhere, but I can't find any. Does this compile without warnings?

include/revng/Support/Assert.h
drblallo: not sure exactly why, but std optional required a explicit bool conversion that i guess happened in the macro. if we accept bool it will  not compile

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
drblallo: not sure i understand the rationale of this, currently it is being printed as
```
private:
 FieldType Field = {};
public:
 const FieltType& field() const ...
 FieltType& field()  ...
```

do you want another public inbetween the two?

drblallo:
wait, what do you mean anticipate the private field? that can't be done before fixing all uses. i can squash it toh

include/revng/Support/Assert.h
aleclearmind: Uhm, you sure? I can understand the other one because we explicitly cast to `bool` in the macro, but here?

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
aleclearmind: We use `public`/`private` to group things (see [B.16](https://pad.rev.ng/s/SJNUet_ku#B-CC)).
Group `using`s vs accessors with an additional `public`.

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
aleclearmind: Please don't silently mark as solved issues. I still see this.
If you mark solved issues silently, every time I need to go through all the solved issues to check if they have actually been solved.

aleclearmind:
> what do you mean anticipate the private field? that can't be done before fixing all uses

"It can't be done" as in "it doesn't compile"? OK, makes sense.

6. The `CustomName` *method* does not respect our conventions (methods should start with a lowercase letter), so I'd expect `clang-tidy` to bitch about this (since we're running it as a compiler plugin). Does it not? If so, why?

aleclearmind:
Please handle `revng_check` too

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
drblallo: this one will get fixed by squashing, which i wanted to do as last thing. i will keep it open until that point

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
drblallo: ahh it is a using, i thought it was btween the two getters

include/revng/Support/Assert.h
drblallo: i think i will delay this commit into a unrelated rewriter mr so this one can go in

aleclearmind:
Merged in a15428b. Thanks!

PR #251, opened 2022-12-16
Added analisys tags

drblallo:


aleclearmind:
Unfortunately this does not work. We need to be able to schedule analysis that require nothing (pure model) *after* DLA.
Tags do not have enough expressive power for this. We need lists `:'(`

drblallo:
can't you just put them as dla step analysis?

aleclearmind:
That means that every time we want to run a pure model analysis, we need to prepare all the inputs for DLA. It doesn't solve the problem we were trying to solve.

We need to express in distinct ways *what is needed to run the analysis* and *when do we want to run the analysis*. The first thing is fixed, the second is different depending on what you're trying to do (initial auto-analysis vs later analyses).

aleclearmind:
Can you implement the `initial-auto-analysis` analysis pipeline? Knowing that works enables me to more easily review bottom up.

include/revng/PipelineC/Prototypes.h
```diff
@@ -289,6 +289,51 @@ rp_manager_run_analysis(rp_manager *manager,
...
+ */
+uint64_t rp_manager_get_analyses_list_count(rp_manager *manager);
+
+/**
+ * \return the name of a give analyses list
+ */
+const char *rp_analyses_list_get_name(rp_analyses_list *list);
+
+/**
+ * \return the leght of the give analyses list
```
aleclearmind, line 303: `s/leght/length/`

include/revng/PipelineC/Prototypes.h
```diff
@@ -289,6 +289,51 @@ rp_manager_run_analysis(rp_manager *manager,
                         rp_invalidations *invalidations,
                         const rp_string_map *options);
 
+/**
+ * \return the number of analyses lists loaded in the pipeline
+ */
+uint64_t rp_manager_get_analyses_list_count(rp_manager *manager);
+
+/**
+ * \return the name of a give analyses list
```
aleclearmind, line 298: `s/give/given/`

lib/PipelineC/PipelineC.cpp
```diff
@@ -846,6 +846,47 @@ const rp_kind *rp_analysis_get_argument_acceptable_kind(rp_analysis *analysis,
...
+}
+
+uint64_t rp_analyses_list_count(rp_analyses_list *list) {
+  return list->size();
+}
+
+rp_analysis *rp_manager_get_analysis(rp_manager *manager,
+                                     rp_analyses_list *list,
+                                     uint64_t index) {
+  return &manager->getAnalysis(list->at(index));
```
aleclearmind, line 869: It's a bit weird that you need `manager` to do this, but OK.

lib/Pipes/PipelineManager.cpp
```diff
@@ -374,6 +374,39 @@ llvm::Error PipelineManager::produceAllPossibleTargets(bool ExpandTargets) {
...
+                             const llvm::StringMap<std::string> &Options,
+                             llvm::raw_ostream *DiagnosticLog) {
+  auto Result = Runner->runAnalyses(List, Map, Options);
+  if (not Result)
+    return Result.takeError();
+
+  recalculateAllPossibleTargets();
+
+  // TODO: to remove once invalidations are working
+  if (auto Invalidations = invalidateAllPossibleTargets(); !!Invalidations)
```
aleclearmind, line 402: Mmmh, is this the same behavior of running an individual analysis?
Do we say "always everything invalidated" even if we run a single analysis?

share/revng/pipelines/analyses-lists.yml
```diff
@@ -0,0 +1,10 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+AnalysesLists:
+  - Name: initial-auto-analysis
+    Analyses:
+      - ImportBinary
+      - AddPrimitiveTypes
+      - DetectABI
```
aleclearmind, line 10: What's the story for the `revng-c` side?
Do we want to support augmenting an existing list with additional analyses or do we simple run a new one from scratch?

IIRC the second one. If so, please make a PR for revng-c too.
Once that's done we can pass things to @mrjackv so he can fix `GraphQL` API and revng-ui.

aleclearmind (changes requested):

lib/Pipes/PipelineManager.cpp
drblallo: yeah, but this is to be dropped when the invalidation works correclty

share/revng/pipelines/analyses-lists.yml
drblallo: https://rev.ng/gitlab/revng-private/revng-c/-/merge_requests/131/diffs?commit_id=9dadc43af0bc14c2154c80efa9b4c800dc0051fa added here

lib/PipelineC/PipelineC.cpp
aleclearmind: Nah, maybe in the future.

aleclearmind:
Please rebase.

mrjackv:
> Please rebase.

Done

python/revng/cli/_commands/daemon_self_test.py
```diff
@@ -119,7 +119,18 @@ async def run_self_test(self, executable_path: str):
...
+
+            if "revng-initial-auto-analysis" in list_names:
+                await client.execute(
+                    gql("""mutation { runAnalysesList(name: "revng-initial-auto-analysis") }""")
+                )
+            if "revng-c-initial-auto-analysis" in list_names:
+                await client.execute(
+                    gql("""mutation { runAnalysesList(name: "revng-c-initial-auto-analysis") }""")
+                )
             self.log("Autoanalysis complete")
```
aleclearmind, line 134: We should assert if there's neither `revng-initial-auto-analysis` nor `revng-c-initial-auto-analysis`, now we silently proceed.

PR #253, opened 2022-12-21
Fix the level for FetchDebugInfo

dwarfyresearch:


lib/Model/Importer/Binary/CrossModelFindTypeHelper.h
```diff
@@ -102,18 +114,36 @@ findPrototypeHelper(T &Functions, llvm::StringRef FunctionName) {
   return std::nullopt;
 }
 
+template<typename T>
```
aleclearmind, line 117: No need for a template here

lib/Model/Importer/DebugInfo/DwarfImporter.cpp
```diff
@@ -1358,11 +1384,35 @@ inline void detectAliases(const llvm::object::ObjectFile &ELF,
...
+          // For example in the case of importing `__libc_calloc` from
+          // libc.so.6.
+          TheFunction = It->second;
+        }
+      }
+      // It is a local function. Populate the ExportedNames.
+      if (TheFunction) {
+        for (auto &Name : PotentialExportedNamesToBeAdded)
+          TheFunction->ExportedNames().insert(Name);
+        continue;
```
aleclearmind, line 1406: Restructure code to avoid `continue`

lib/Model/Importer/DebugInfo/DwarfImporter.cpp
```diff
@@ -1333,16 +1348,27 @@ inline void detectAliases(const llvm::object::ObjectFile &ELF,
                           TupleTree<model::Binary> &Model) {
   EquivalenceClasses<StringRef> Aliases = computeEquivalentSymbols(ELF);
   auto &ImportedDynamicFunctions = Model->ImportedDynamicFunctions();
+  auto &Functions = Model->Functions();
+
+  std::unordered_map<std::string, model::Function *> FunctionsByName;
+  // Map functions by names, so we have faster lookup bellow.
+  for (auto &Function : Functions) {
+    if (Function.OriginalName().size()) {
```
aleclearmind, line 1356: `OriginalName` doesn't have to be unique.

lib/Model/Importer/DebugInfo/DwarfImporter.cpp
```diff
@@ -1333,16 +1348,27 @@ inline void detectAliases(const llvm::object::ObjectFile &ELF,
...
+      FunctionsByName[Function.OriginalName()] = &Function;
+    }
+  }
 
   for (auto AliasesIt = Aliases.begin(), E = Aliases.end(); AliasesIt != E;
        ++AliasesIt) {
+    llvm::SmallVector<std::string, 4> CurrentAliases;
```
aleclearmind, line 1363: This can be pushed in  the `if`

lib/Model/Importer/DebugInfo/DwarfImporter.cpp
```diff
@@ -1358,11 +1384,35 @@ inline void detectAliases(const llvm::object::ObjectFile &ELF,
...
+      // Consider it as a Dynamic function.
       if (Prototype.isValid()) {
         for (const std::string &Name : UnprototypedFunctionsNames) {
           auto It = ImportedDynamicFunctions.find(Name);
           if (It == ImportedDynamicFunctions.end())
             It = ImportedDynamicFunctions.insert({ Name }).first;
+
           It->Prototype() = Prototype;
         }
       }
```
aleclearmind, line 1418: I'm confused by this piece of code. We're iterating over symbols that *have* an address (enforced in `computeEquivalentSymbols`) but we're created new `ImportedDynamicFunctions`. This doesn't sound right at all.
I think all of this function should not touch `ImportedDynamicFunctions` at all.

We have, one non-null address, many names.
So we should:

1. Check if `Binary.Functions` contains the address, if not create it.
2. Assign to such `model::Function` the prototype.
3. If the symbol is dynamic (`computeEquivalentSymbols` only considers static symbols AFAIU), we should add the name of the dynamic symbol to `ExportedNames`.

I think this basically needs to be rewritten.

aleclearmind (changes requested):

lib/Model/Importer/DebugInfo/DwarfImporter.cpp
dwarfyresearch: I think that we should generate ```ExportedNames``` for local functions only. The rest of the comment makes perfectly sense to me. I've rewritten the code.

aleclearmind: We need to add `ExportedNames` for `model::Function` but only for names coming from dynamic symbols.
If we have an alias in `.symtab` it's good to know, but, unless it's also in `.dynsym`, it's not an "exported name", you can't refer to a symbol with a name that's not in `.dynsym`.

aleclearmind:
Merged in 4a7ccfd, thanks!

PR #254, opened 2022-12-22
 hitman-vscode integration

mrjackv:


lib/PipelineC/PipelineC.cpp
```diff
@@ -807,7 +807,9 @@ const char *rp_container_get_mime(rp_container *container) {
 
 const char *
 rp_container_extract_one(rp_container *container, rp_target *target) {
-  revng_check(container->second->enumerate().contains(*target));
+  if (!container->second->enumerate().contains(*target)) {
+    return nullptr;
+  }
```
aleclearmind, line 811: > Fix PipelineC aborting when it should return nullptr

The docs says otherwise:

```
include/revng/PipelineC/Prototypes.h-/**
include/revng/PipelineC/Prototypes.h- * \return the serialized content of the element associated to the provided
include/revng/PipelineC/Prototypes.h- * target,
include/revng/PipelineC/Prototypes.h- *
include/revng/PipelineC/Prototypes.h- * \note Target must be already present in container
include/revng/PipelineC/Prototypes.h- */
include/revng/PipelineC/Prototypes.h-const char * /*owning*/
include/revng/PipelineC/Prototypes.h:rp_container_extract_one(rp_container *container, rp_target *target);
```

Do we have an API to check if the target is available before calling this?
If you really want to return `nullptr`, we need to update the doc.

I know you don't like `abort`s, but remember that this is an internal API, it's not going to be used by users, so forcing ourselves to do extra checks is not too bad.

python/revng/cli/_commands/daemon.py
```diff
@@ -68,6 +68,8 @@ def run(self, options: Options):
             elif bind.startswith("unix:"):
                 _, path = bind.split(":", 1)
                 args.extend(["--uds", path])
+            elif bind == "none":
+                pass
```
aleclearmind, line 75: Can you doc this in the help of the `-b` argument?

python/revng/daemon/event_manager.py
```diff
@@ -0,0 +1,135 @@
...
+            return result
+
+        return inner
+
+    return decorator
+
+
+class EventHandler(Protocol):
+    def __call__(self, manager: Manager, event_type: EventType, token: str):
+        ...
```
aleclearmind, line 45: Is this equivalent to `pass`?

python/revng/api/listenable_manager.py
```diff
@@ -1,62 +0,0 @@
...
-
-class EventType(Enum):
-    BEGIN = auto()
-    CONTEXT = auto()
-
-
-EventHandler = Callable[[Manager, EventType], Any]
-
-
-class ListenableManager(Manager):
```
aleclearmind, line 23: Aaaah nice

python/revng/daemon/event_manager.py
```diff
@@ -0,0 +1,135 @@
...
+        begin_step.save(tmpdir)
+        self.write_fifo(f"PUSH begin {tmpdir} {token}\n")
+
+
+class ContextListener(BaseListener):
+    def __call__(self, manager: Manager, type_: EventType, token: str):
+        assert type_ == EventType.CONTEXT
+        tmpdir = mkdtemp()
+        manager.save_context(tmpdir)
+        self.write_fifo(f"PUSH context {tmpdir} {token}\n")
```
aleclearmind, line 135: Is it really necessary to have two of them? Let's merge them.

python/revng/daemon/event_manager.py
```diff
@@ -0,0 +1,135 @@
...
+
+from revng.api import Manager
+
+
+class EventType(Enum):
+    BEGIN = auto()
+    CONTEXT = auto()
+
+
+def emit_event(event_type: EventType):
```
aleclearmind, line 28: This is not straightforward, it does some magic with the first argument.
Docs!

python/revng/daemon/__init__.py
```diff
@@ -93,19 +94,35 @@ def shutdown():
...
-app = Starlette(
-    debug=DEBUG,
-    middleware=[
+def get_middlewares() -> List[Middleware]:
+    extra_middlewares: List[Middleware] = []
+    if "STARLETTE_MIDDLEWARES" in os.environ:
+        for element in os.environ["STARLETTE_MIDDLEWARES"].split(","):
+            module_path, attribute_name = element.split(":", 1)
+            module = import_module(module_path)
+            if not hasattr(module, attribute_name):
```
aleclearmind, line 44: 1. Can we go by convention? That is, force middlewares to implement a function/object whatever with a specific name.
2. Have you considered making a canonical path where to put middlewares so we don't need an additional `STARLETTE_MIDDLEWARES`.

Conventions are good!

include/revng/PipelineC/Prototypes.h
```diff
@@ -765,3 +766,25 @@ const char * /* owning */
...
+ * \return the pointer at the start of the buffer
+ */
+const char *rp_buffer_data(const rp_buffer *buffer);
+
+/**
+ * Free a rp_buffer
+ */
+void rp_buffer_destroy(const rp_buffer *buffer);
+
+/** \} */
```
aleclearmind, line 790: Nice!

include/revng/PipelineC/ForwardDeclarations.h
```diff
@@ -21,3 +21,4 @@ typedef const pipeline::DiffMap rp_diff_map;
 typedef ErrorList rp_error_list;
 typedef llvm::StringMap<std::string> rp_string_map;
 typedef pipeline::InvalidationMap rp_invalidations;
+typedef llvm::SmallString<0> rp_buffer;
```
aleclearmind, line 24: Why `SmallString`? It's not a string in the general case.

aleclearmind (changes requested):

aleclearmind:
Split `Miscellaneous fixes`, many small commits are good!

include/revng/PipelineC/ForwardDeclarations.h
mrjackv: I can put `llvm::SmallVector<char, 0>` if it makes it clearer on the intentions

python/revng/daemon/util.py
```diff
@@ -64,3 +66,12 @@ def project_workdir() -> Optional[Path]:
 
 def target_dict_to_graphql(target_dict: Dict[str, str]):
     return {"pathComponents": target_dict["path_components"], **target_dict}
+
+
+def produce_serializer(input_: Dict[str, str | bytes]) -> str:
+    return json.dumps(
+        {
+            key: (value if isinstance(value, str) else b64encode(value).decode("utf-8"))
```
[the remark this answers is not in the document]
mrjackv: Yes

https://pad.rev.ng/emEtVjiLTz2dRCnf9avSng?both#MIME-types

> Mimes that are not text/* or image/svg will be transmitted over GraphQL via Base64 encoding

python/revng/daemon/event_manager.py
mrjackv: Yes
In the wild I've (anectdotally) seen `...` used more for ABCs and Protocols but nothing more

lib/PipelineC/PipelineC.cpp
mrjackv: There's `rp_target_is_ready` but I'd argue that it's more consistent with the rest of the API where `nullptr` says it's not present

python/revng/daemon/__init__.py
mrjackv: 1. We can do that, but then it's a bit annoying if you have a package providing more than one middleware
2. That's possible but __really__ annoying, e.g. right now the auth middleware it's packaged and can declare dependencies and be cleanly packaged

python/revng/cli/_commands/daemon.py
mrjackv: Done

python/revng/daemon/event_manager.py
mrjackv: Added

python/revng/daemon/event_manager.py
mrjackv: Eh, I'd say it'd only make it harder and more complicated to read if they were merged
Also this snippet makes it pretty explicit what they do (and it's one and one thing only)
```python
            if string_type == "begin":
                self.add_event_listener(EventType.BEGIN, BeginListener(Path(fifo_path)))
            elif string_type == "context":
                self.add_event_listener(EventType.CONTEXT, ContextListener(Path(fifo_path)))
```

lib/PipelineC/PipelineC.cpp
aleclearmind: OK, then update the doc.

python/revng/daemon/event_manager.py
aleclearmind: To me it seems a lot of boiler plate to add a new event for little benefit.

> more complicated to read

Why?

lib/PipelineC/PipelineC.cpp
mrjackv: Updated

python/revng/daemon/__init__.py
aleclearmind: OK, let's import everything in a path automatically and then let's import but fully qualified name as django does.

include/revng/PipelineC/ForwardDeclarations.h
aleclearmind: `uint8_t` works.

python/revng/daemon/util.py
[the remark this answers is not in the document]
aleclearmind: Did we commit this document somewhere in the end?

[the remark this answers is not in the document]
mrjackv: In `revng-ui` under `docs/mime_handling.md`

include/revng/PipelineC/ForwardDeclarations.h
mrjackv: Not really, `llvm::raw_svector_ostream` wants a `SmallVectorImpl<char>` as the constructor argument

python/revng/daemon/util.py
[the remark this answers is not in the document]
mrjackv: Added `docs/mime_handling.md` to this PR

python/revng/daemon/__init__.py
mrjackv: After having implemented `STARLETTE_MIDDLEWARE_{EARLY,LATE}` I'd really push just to use those, the special folder is just a special case and it can lead to unexpected surprises  later on

python/revng/daemon/event_manager.py
mrjackv: Because the resulting class would look like
```python
class Listener:
  def __init__(self, listen_type: Begin | Context):
    self.listen_type = listen_type

  def __call__(self, ...):
    if self.listen_type == Begin:
      <do begin stuff>
    if self.listen_type == Context:
      <do context stuff>
 ```
 
 Instead here you know that each class does exactly one kind of operation

File: python/revng/daemon/__init__.py (hunk 7 above)
Comment (aleclearmind) [reply python/revng/daemon/__init__.py]: For example?

In the end they'll end up one on the side of the other, why leave a degree of freedom we don't need?
We can have `share/revng/daemon/middlewares/` and be done with it.

File: python/revng/daemon/event_manager.py (hunk 5 above)
Comment (aleclearmind) [reply python/revng/daemon/event_manager.py:139]: As an implementor of a listener I just want to write a function that takes `Begin`/`Context` as an argument and act based on that. I don't really care about the rest.

Can you figure out enough infrastructure to make this possible?

File: python/revng/daemon/__init__.py (hunk 7 above)
Comment (mrjackv) [reply python/revng/daemon/__init__.py]: Because it's an incredibly poor way to handle them, it's really better to have each middleware/middleware group packaged as a proper python package that specifies its dependencies and it's installed the proper way via pip.
Also folders are just annoying, it usually leads to things like `<folder>/01_early_middleware.py` etc. and `<folder>/middleware.py.disabled`

Comment (aleclearmind) [reply python/revng/daemon/__init__.py]: Having an environment variable which is a *list* of *pairs* of paths + names is not great either. Doing configuration in environment variables is already bad enough, let's not put JSONs in there please.

If you want, we can assume they are in `sys.path` and just have the environment variables specify the names.
No new magic paths, just some place already in `sys.path`.

Comment (mrjackv) [reply python/revng/daemon/__init__.py]: Right now there's no magic paths, only imports

e.g. the auth middleware is in `revng_middleware_auth.AuthMiddleware`, that's it
It's the equivalent of `from revng_middleware_auth import AuthMiddleware`

I know that comma-separated environment variables are not the greatest, but this is the most python way to handle things IMHO (and that would be extremely intuitive to anybody that has used a similar mechanism in the past)

Comment (aleclearmind) [reply python/revng/daemon/__init__.py]: Then the middleware is already installed in someplace in `sys.path`?

Comment (mrjackv) [reply python/revng/daemon/__init__.py]: Yes

File: python/revng/daemon/event_manager.py (hunk 5 above)
Comment (aleclearmind) [reply python/revng/daemon/event_manager.py]: aaaah *massages nipples*

File: python/revng/daemon/__init__.py [hunk 11]
```diff
@@ -93,19 +94,44 @@ def shutdown():
...
+        module = import_module(module_path)
+        if not hasattr(module, attribute_name):
+            raise ValueError(f"Middleware not found: {element}")
+        result.append(getattr(module, attribute_name))
+
+    return result
+
+
+app = Starlette(
+    debug=DEBUG,
```
aleclearmind: What are the implications of `DEBUG`?
Don't we want a `--production` (or something like that) argument?

aleclearmind (changes requested):

python/revng/daemon/__init__.py
mrjackv: [We have one](https://github.com/revng/revng/blob/develop/python/revng/cli/_commands/daemon.py#L47)

PR #257, opened 2023-01-10
Fix idb importer

dwarfyresearch:


lib/Model/Pass/FixModel.cpp
```diff
@@ -31,6 +31,21 @@ void model::fixModel(TupleTree<model::Binary> &Model) {
       }
     }
 
+    // Filter out empty arrays.
+    bool IsEmptyArray = false;
+    for (const model::QualifiedType &QT : T->edges()) {
+      auto IsArray = [](const model::Qualifier &Q) {
+        return Q.Kind() == model::QualifierKind::Array;
```
aleclearmind, line 64: You can check size right in here and simplify the code.

lib/Model/Pass/FixModel.cpp
```diff
@@ -31,6 +31,21 @@ void model::fixModel(TupleTree<model::Binary> &Model) {
...
+      };
+      auto Iterator = llvm::find_if(QT.Qualifiers(), IsArray);
+      if (Iterator != QT.Qualifiers().end() and !Iterator->Size()) {
+        ToDrop.insert(T.get());
```
aleclearmind, line 68: I feel like this function could be restructured as

```c++
  for (UpcastablePointer<model::Type> &T : Model->Types())
    if (shouldDrop(T))
      ToDrop.insert(T.get());
```

Outlining most of the logic will make the function simpler and reduce the amount of `continue`.

python/revng/cli/_commands/idb_converter.py
```diff
@@ -401,22 +401,63 @@ def _convert_idb_type_to_revng_type(
...
+                else:
+                    return self.idb_types_to_revng_types[type.type_details.ref.type_details.ordinal]
+            else:
+                revng_type = m.UnionType(OriginalName=type_name, Fields=[])
+                if len(type.type_details.members) == 0:
+                    self.log(f"warning: Ignoring empty union {type_name}, typedef it to void*")
+                    revng_void_type = self.unwrap_qualified(
+                        self._get_primitive_type(m.PrimitiveTypeKind.Void, 0)
+                    )
+                    # TODO: The pointer size is hardcoded.
```
aleclearmind, line 448: Please fix this

python/revng/cli/_commands/idb_converter.py
```diff
@@ -401,22 +401,63 @@ def _convert_idb_type_to_revng_type(
                 else:
                     return self.idb_types_to_revng_types[type.type_details.ref.type_details.ordinal]
             else:
-                revng_type = m.StructType(OriginalName=type_name, Size=0, Fields=[])
```
aleclearmind, line 404: Can you describe in the commit message what was the problem?

aleclearmind (changes requested):

lib/Model/Pass/FixModel.cpp
```diff
@@ -19,16 +19,39 @@ static RegisterModelPass
...
     // Filter out empty structs and unions.
-    if (!T->size()) {
-      if (isa<StructType>(T.get()) or isa<UnionType>(T.get())) {
+    if (isa<StructType>(T.get()) or isa<UnionType>(T.get())) {
+      if (!T->size()) {
         ToDrop.insert(T.get());
         continue;
       }
+      if (isa<StructType>(T.get()))
+        if (filterZeroSizedElements(ToDrop, cast<StructType>(T.get())))
```
aleclearmind:
```c++
      if (auto *Struct = dyn_cast<StructType>(T.get()))
        if (filterZeroSizedElements(ToDrop, Struct))
```

lib/Model/Pass/FixModel.cpp
```diff
@@ -20,86 +20,83 @@ static Logger<> ModelFixLogger("model-fix");
...
+  return false;
+}
+
+void model::fixModel(TupleTree<model::Binary> &Model) {
+  std::set<const model::Type *> ToDrop;
+
+  for (UpcastablePointer<model::Type> &T : Model->Types()) {
+    if (shouldDrop(T))
+      ToDrop.insert(T.get());
+  }
```
aleclearmind, line 98: Next time you can use `llvm::copy_if()` `:)`
No biggie.

aleclearmind (changes requested):

lib/Model/Pass/FixModel.cpp
dwarfyresearch: I am not the biggest fun of the paradigm (I don't think it improves readability always :) ), but I understand part of the motivation. The code would be a bit messy, since ```ToDrop``` expects ```Type *```, so we cannot do simply as:
```
  llvm::copy_if(Model->Types(),
                std::back_inserter(ToDrop),
                [](UpcastablePointer<model::Type> &T) {
                  return shouldDrop(T);
                });
```

aleclearmind:
Merged in 925fae4, thanks!

PR #258, opened 2023-01-15
Introduce progress reporting

aleclearmind:
This PR depends on [revng/llvm-project#21](https://github.com/revng/llvm-project/pull/21).

### DIFF
--- a/lib/Support/Progress.cpp
+++ b/lib/Support/Progress.cpp
@@ -1,0 +1,320 @@
+/// \file Progress.cpp
+/// \brief
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include <chrono>
+
+#include "llvm/Support/Mutex.h"
+#include "llvm/Support/Progress.h"
+
+#include "revng/Support/Debug.h"
+
+class TraceProgressListener : public llvm::ProgressListener {
+private:
+  std::error_code EC;
+
+  // TODO: consider buffering in thread-local buffers
+  llvm::sys::Mutex OutputMutex;
+  llvm::raw_fd_ostream Output;
+
+public:
+  static constexpr bool AllThreads = true;
+
+public:
+  TraceProgressListener(llvm::StringRef OutputPath) : Output(OutputPath, EC) {
+    revng_assert(!EC);
+    Output << "[\n";
+  }
+
+  ~TraceProgressListener() override { Output << "{}]\n"; }
+
+public:
+  void handleNewTask(const llvm::Task *T) override {
+    llvm::sys::ScopedLock Lock(OutputMutex);
+    emitEvent(T->name(), "task", "B");
+  }
+
+  void handleTaskCompleted(const llvm::Task *T) override {
+    llvm::sys::ScopedLock Lock(OutputMutex);
+    if (T->stepIndex() != -1) {
+      emitEvent(T->stepName(), "task", "E");
+    }
+    emitEvent(T->name(), "task", "E");
+  }
+
+  void handleTaskAdvancement(const llvm::Task *T,
+                             llvm::StringRef PreviousStepName) override {
+    llvm::sys::ScopedLock Lock(OutputMutex);
+    if (T->stepIndex() != 0)
+      emitEvent(PreviousStepName, "task", "E");
+    emitEvent(T->stepName(), "task", "B");
+  }
+
+  void emitEvent(llvm::StringRef Name,
+                 llvm::StringRef Category,
+                 llvm::StringRef Phase) {
+    Output << "{";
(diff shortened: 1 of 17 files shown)


lib/Support/Progress.cpp
```diff
@@ -0,0 +1,255 @@
...
+    LastStepName.pop_back();
+    emitEvent(T->name(), "task", "E");
+  }
+
+  void handleTaskAdvancement(const llvm::Task *T,
+                             llvm::StringRef NewStepName) override {
+    endStep();
+    emitEvent(NewStepName, "task", "B");
+    LastStepName.back() = NewStepName;
+  }
```
pfez, line 48: These three methods are advertised  in the sister llvm MR to be callable by multiple threads at the same time. 

If this happens, there are 3 issues:
- `emitEvent` does not guard against concurrent access to `Output` so you could end up printing garbled output;
- `LastStepName` is not guarded against concurrent access, so it could end up in invalid state.
- `endStep` is also not guarded against concurrent access, so it's not safe.

I think you should put a `llvm::sys::SmartRWMutex` in `TraceProgressListener` and use it to make this API safe for concurrent use.

pfez: BTW, all other `ProgressListeners` in this file have the same flaw.

aleclearmind: Yep, I'm aware. This is still WIP.

* `LastStepName` has to go. I changed the API on LLVM side so you no longer need it.
* My plan is to lock all writes to the output for `PlainProgressListener` and `TerminalBarsProgressListener` (once every 50ms). While I thinking of buffering output in thread-specific buffers for `TraceProgressListener`. The ratio is that the `PlainProgressListener` might be useful for debugging a crash, so delaying output is not helpful. `TerminalBarsProgressListener` is already amortized. `TraceProgressListener` is more for performance profiling than debugging stuff, so I think buffering the output is fine.

lib/Support/Progress.cpp
```diff
@@ -0,0 +1,255 @@
...
+  std::chrono::time_point<std::chrono::high_resolution_clock> LastDraw;
+
+public:
+  TerminalBarsProgressListener(llvm::raw_ostream &Output)
+    : Output(Output) {}
+
+public:
+  void handleNewTask(const llvm::Task *T) override {
+    if (T->stack().Tasks.size() > MaxProgressBars) {
+      ++MaxProgressBars;
```
pfez, line 134: :thinking: `MaxProgressBars` is not safe to be accessed concurrently. What happens if 2 threads call `handleNewTask` on the same listener concurrently?

lib/Support/Progress.cpp
pfez: It works for this listener, but in general all listeners that have internal state need to be built to be safe to be called concurrently.

lib/Support/Progress.cpp
aleclearmind: Please do not review this, this is draft and WIP, I still need to work out multi-threading details of these guys.

lib/Support/Progress.cpp
aleclearmind: I put a lock in front of `TraceProgressListener` and introduced a `bool` to state that the other listeners are interested in the main thread only, which bypasses the threading problem entirely.

lib/Support/Progress.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+      auto MaybeStepsCount = Task->totalSteps();
+      if (MaybeStepsCount) {
+        auto StepsCount = *MaybeStepsCount;
+        revng_assert(Task->stepIndex() < StepsCount);
+
+        float Addendum = 0.0;
+
+        // auto *InnerTask = Stack.Tasks[I + 1];
+        // if (InnerTask->isSingleSubtask())
```
pfez, line 178: Leftover comment I guess.

lib/Support/Progress.cpp
```diff
@@ -0,0 +1,265 @@
...
+};
+
+static auto RTBPLCallback = callback(RegisterTerminalBarsProgressListener);
+
+static opt<bool> ProgressBars("progress", RTBPLCallback);
+
+static auto RegisterProgressPlain = callback([](const bool &Value) {
+  if (Value)
+    llvm::ProgressReport->registerListener<PlainProgressListener>(llvm::errs());
+});
```
pfez, line 263: Can't you make these regular functions instead of lambdas so that you can avoid the awkward `callback` dance?

aleclearmind: I'll double check but IIRC the caller explicitly uses `operator()` instead of performing a simple function call, so regular functions won't work.

PR #261, opened 2023-01-31
TypeSystemPrint: fix print of function types without stack type

pfez:


aleclearmind:
Merged in 04d877a. Thanks!

PR #262, opened 2023-01-31
Improve daemon test log handling

mrjackv:


aleclearmind:
Merged in f3f04ef. Thanks!

PR #264, opened 2023-02-02
Propagate prototypes to callers in DetectABI procedure

kolodziej:


include/revng/EarlyFunctionAnalysis/CFGAnalyzer.h
```diff
@@ -31,17 +31,21 @@ class CallSummarizer : public CallHandler {
   llvm::Function *PostCallHook = nullptr;
   llvm::Function *RetHook = nullptr;
   llvm::GlobalVariable *SPCSV = nullptr;
+  const FunctionSummaryOracle &Oracle;
   OpaqueFunctionsPool<llvm::StringRef> RegistersClobberedPool;
+  OpaqueFunctionsPool<llvm::StringRef> WeakReadWritePools;
```
aleclearmind, line 36: Your project is about adding **non-**weak reads and writes.

include/revng/EarlyFunctionAnalysis/FunctionSummaryOracle.h
```diff
@@ -140,7 +140,12 @@ class FunctionSummaryOracle {
     return LocalFunctions.at(PC);
   }
 
-  /// \return a description of the call and boolean indicating whether the call
+  /// \param Function caller function address
+  /// \param CallerBlockAddress caller basic block address
+  /// \param CalledLocalFunction callee function address
+  /// \param CalledSymbol name of dynamic symbol we are calling (can be
+  /// extracted from llvm::Value with extractFromConstantStringPtr) \return a
+  /// description of the call and boolean indicating whether the call
```
aleclearmind, line 148: Mind the indentation
```suggestion
  /// \param CalledSymbol name of dynamic symbol we are calling (can be
  ///         extracted from llvm::Value with extractFromConstantStringPtr)
  /// \return a description of the call and boolean indicating whether the call
```

include/revng/MFP/SetLattices.h
```diff
@@ -21,7 +21,7 @@ struct SetUnionLattice {
     return Result;
   }
 
-  static bool
+  [[nodiscard]] static bool
```
aleclearmind, line 24: Nice

lib/EarlyFunctionAnalysis/CFGAnalyzer.cpp
```diff
@@ -3,6 +3,7 @@
 //
 // This file is distributed under the MIT License. See LICENSE.md for details.
 //
+// #pragma clang optimize off
```
aleclearmind, line 6: Avoid committing these thing, they tend to slip in. Keep them in your stash.

lib/EarlyFunctionAnalysis/CFGAnalyzer.cpp
```diff
@@ -964,21 +967,29 @@ CallSummarizer::CallSummarizer(llvm::Module *M,
...
+  Oracle(Oracle),
+  RegistersClobberedPool(M, false),
+  WeakReadWritePools(M, false) {
   RegistersClobberedPool.addFnAttribute(llvm::Attribute::ReadOnly);
   RegistersClobberedPool.addFnAttribute(llvm::Attribute::NoUnwind);
   RegistersClobberedPool.addFnAttribute(llvm::Attribute::WillReturn);
+
+  WeakReadWritePools.addFnAttribute(llvm::Attribute::ReadOnly);
+  WeakReadWritePools.addFnAttribute(llvm::Attribute::NoUnwind);
+  WeakReadWritePools.addFnAttribute(llvm::Attribute::WillReturn);
```
aleclearmind, line 986: No need for a new pool, the only goal of these functions is to produce an opaque value to use to clobber the CSV. No  other special meaning is attached to them. I'd just rename `RegistersClobberedPool` to `OpaqueValuesPool` or something along those lines.

I think I suggested this on Mattermost too.

lib/EarlyFunctionAnalysis/CFGAnalyzer.cpp
```diff
@@ -1001,8 +1012,34 @@ void CallSummarizer::handleCall(MetaAddress CallerBlock,
...
+  const auto &[CalleeSummary, IsTail] = Oracle.getCallSite(Caller,
+                                                           CallerBlock,
+                                                           Callee,
+                                                           SymbolName);
+  const auto &ABIResults = CalleeSummary->ABIResults;
+
+  for (const auto &[Variable, Values] : ABIResults.ArgumentsRegisters) {
+    if (Values == abi::RegisterState::Yes) {
+      auto Register = M->getNamedGlobal(Variable->getName());
+      Builder.CreateLoad(Register);
```
aleclearmind, line 1025: This load will be unused. Consider that if we run any optimization pipeline after this, this will be DCE'd away.

lib/EarlyFunctionAnalysis/DetectABI.h
```diff
@@ -0,0 +1,102 @@
...
+    GCBI(GCBI),
+    Binary(Binary),
+    Oracle(Oracle),
+    Analyzer(Analyzer) {}
+
+public:
+  void run();
+  ABIAnalyses::ABIAnalysesResults analyzeABI(llvm::BasicBlock *Entry);
+  ABIAnalyses::ABIAnalysesResults
+  analyzeABI(MetaAddress Addr) { // TODO: Tidy this
```
aleclearmind, line 53: Use `// WIP:` for comments that should not be merged. `check-conventions` will complain about them.

lib/EarlyFunctionAnalysis/UsedRegisters.h
```diff
@@ -0,0 +1,152 @@
...
+struct FunctionABI {
+  using SetOfRegisters = std::set<model::Register::Values>;
+  using SUL = SetUnionLattice<SetOfRegisters>;
+  SetOfRegisters ArgumentRegisters;
+  SetOfRegisters ReturnRegisters;
+};
+
+// using LatticeElement = std::map<MetaAddress, FunctionABI>;
+
+struct LatticeElement : std::map<MetaAddress, FunctionABI> {
```
aleclearmind, line 29: I'd avoid inheritance here, better to use composition (i.e., have a `std::map` field) and expose only the methods you need.

lib/EarlyFunctionAnalysis/UsedRegisters.h
```diff
@@ -0,0 +1,152 @@
...
+      return at(Addr).ArgumentRegisters.count(V);
+    }
+
+    return false;
+  }
+};
+
+struct ExtremalLatticeElement : LatticeElement {
+  [[nodiscard]] bool
+  hasRetReg(MetaAddress Addr, model::Register::Values V) const {
```
aleclearmind, line 53: No abbreviations. Same for `hasArgReg`.

lib/EarlyFunctionAnalysis/CFGAnalyzer.cpp
```diff
@@ -1001,8 +1012,34 @@ void CallSummarizer::handleCall(MetaAddress CallerBlock,
 
   Builder.CreateCall(PreCallHook, Args);
 
+  const auto SymbolName = extractFromConstantStringPtr(SymbolNamePointer);
+  const auto &[CalleeSummary, IsTail] = Oracle.getCallSite(Caller,
+                                                           CallerBlock,
+                                                           Callee,
+                                                           SymbolName);
```
aleclearmind, line 1019: You should not use the `Oracle` here. Currently this works, because we're cloning the monotone framework state in `Oracle`:

```c++
Oracle.getLocalFunction(EntryAddress).ABIResults = ABIResults
```

But this should go away, it's confusing, we should update the `Oracle` only once at the end and use the state of the monotone framework directly instead.

You need to get access to the state of the monotone framework.
You need to basically get access to the lattice element and query it. See if you can have a reference field to some part o the monotone framework state in the class and initialize it in the constructor of `CallSummarizer`, as we do for `Oracle`.

lib/EarlyFunctionAnalysis/CFGAnalyzer.cpp
```diff
@@ -1001,8 +1012,34 @@ void CallSummarizer::handleCall(MetaAddress CallerBlock,
...
+  for (const auto &[Variable, Values] : ABIResults.ArgumentsRegisters) {
+    if (Values == abi::RegisterState::Yes) {
+      auto Register = M->getNamedGlobal(Variable->getName());
+      Builder.CreateLoad(Register);
+    }
+  }
+
   clobberCSVs(Builder, ClobberedRegisters);
 
+  for (auto &[Variable, Values] : ABIResults.FinalReturnValuesRegisters) {
```
aleclearmind, line 1031:
```suggestion
  for (auto &[CSV, Values] : ABIResults.FinalReturnValuesRegisters) {
```

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -3,6 +3,7 @@
 //
 // This file is distributed under the MIT License. See LICENSE.md for details.
 //
+#pragma clang optimize off
```
aleclearmind, line 6: Do not commit

lib/EarlyFunctionAnalysis/UsedRegisters.h
```diff
@@ -0,0 +1,152 @@
...
+  using SetOfRegisters = std::set<model::Register::Values>;
+  using SUL = SetUnionLattice<SetOfRegisters>;
+  SetOfRegisters ArgumentRegisters;
+  SetOfRegisters ReturnRegisters;
+};
+
+// using LatticeElement = std::map<MetaAddress, FunctionABI>;
+
+struct LatticeElement : std::map<MetaAddress, FunctionABI> {
+  virtual ~LatticeElement() = default;
```
aleclearmind, line 30: I don't think this class needs to be `virtual`

lib/EarlyFunctionAnalysis/UsedRegisters.h
```diff
@@ -0,0 +1,152 @@
...
+
+  [[nodiscard]] virtual bool
+  hasRetReg(MetaAddress Addr, model::Register::Values V) const {
+    if (count(Addr)) {
+      return at(Addr).ReturnRegisters.count(V);
+    }
```
aleclearmind, line 36: Here you're performing a lookup twice.
It's annoying the first time to do the right thing, but once you get used to it it's a no brainer:

```suggestion
    auto It = find(Addr);
    if (It != end()) {
      return It->ReturnRegisters.count(V);
    }
```

lib/EarlyFunctionAnalysis/UsedRegisters.h
```diff
@@ -0,0 +1,152 @@
...
+  hasArgReg(MetaAddress Addr, model::Register::Values V) const {
+    if (count(Addr)) {
+      return at(Addr).ArgumentRegisters.count(V);
+    }
+
+    return false;
+  }
+};
+
+struct ExtremalLatticeElement : LatticeElement {
```
aleclearmind, line 51: This does not need to be a class. It just need to be a specific instance of `LatticeElement`.

lib/EarlyFunctionAnalysis/UsedRegisters.h
```diff
@@ -0,0 +1,152 @@
...
+
+  DetectABI &DA;
+
+  explicit UsedRegistersMFI(DetectABI &DA) : DA{ DA } {}
+
+  [[nodiscard]] LatticeElement
+  combineValues(const LatticeElement &E1, const LatticeElement &E2) const {
+    LatticeElement Out;
+    for (const auto &[Address, E1Abi] : E1) {
+      if (E2.count(Address)) {
```
aleclearmind, line 77: Use iterators to avoid multiple lookups.

lib/EarlyFunctionAnalysis/UsedRegisters.h
```diff
@@ -0,0 +1,152 @@
...
+    for (auto &[CSV, State] : Results.FinalReturnValuesRegisters) {
+      if (State == abi::RegisterState::YesOrDead) {
+        Result[Addr].ReturnRegisters.insert(Reg(CSV));
+      }
+    }
+
+    return Result;
+  }
+
+  static ExtremalLatticeElement ExtremalLattice;
```
aleclearmind, line 149: It's an *element* of the lattice.

```suggestion
  static ExtremalLatticeElement ExtremalLatticeElement;
```

lib/EarlyFunctionAnalysis/UsedRegisters.h
```diff
@@ -0,0 +1,152 @@
...
+  using GraphType = llvm::Inverse<BasicBlockNode *>;
+  using LatticeElement = LatticeElement;
+  using ExtremalLatticeElement = ExtremalLatticeElement;
+
+  DetectABI &DA;
+
+  explicit UsedRegistersMFI(DetectABI &DA) : DA{ DA } {}
+
+  [[nodiscard]] LatticeElement
+  combineValues(const LatticeElement &E1, const LatticeElement &E2) const {
```
aleclearmind, line 74: Implement `FunctionABI::combineValues` and `LatticeElement::combineValues`: `UsedRegistersMFI::combineValues` should use `LatticeElement::combineValues`, which in turn should use `FunctionABI::combineValues`.

Same of `isLessThenOrEqual`.

lib/EarlyFunctionAnalysis/UsedRegisters.h
```diff
@@ -0,0 +1,152 @@
...
+
+  [[nodiscard]] LatticeElement
+  applyTransferFunction(Label L, const LatticeElement &E2) const {
+    LatticeElement Result = E2;
+    auto Reg = [this](const llvm::GlobalVariable *V) {
+      return model::Register::fromCSVName(V->getName(),
+                                          DA.getBinary()->Architecture());
+    };
+
+    const auto &Addr = L->Address;
```
aleclearmind, line 129: No abbreviations

aleclearmind (changes requested):

lib/EarlyFunctionAnalysis/UsedRegisters.h
aleclearmind: You're right, but you're describing the "top" element more than the extremal value.

But in this case, I think we can have "bottom" as the extremal value.
I don't see a reason to start from "top".

lib/EarlyFunctionAnalysis/CFGAnalyzer.cpp
aleclearmind: OK, double check and let's see.

include/revng/EarlyFunctionAnalysis/FunctionSummaryOracle.h
kolodziej: I think it was clang-format's idea, but I'll change it.

lib/EarlyFunctionAnalysis/CFGAnalyzer.cpp
kolodziej: I've verified and not it works. Probably it was caused by some other thing and during further changes it has been fixed.

lib/EarlyFunctionAnalysis/CFGAnalyzer.cpp
```diff
@@ -1001,8 +1006,41 @@ void CallSummarizer::handleCall(MetaAddress CallerBlock,
...
+
+  for (const auto &[CSV, Values] : ABIResults.ArgumentsRegisters) {
+    if (Values == abi::RegisterState::Yes) {
+      auto Register = M->getNamedGlobal(CSV->getName());
+      auto Name = llvm::formatv("read_{0}", CSV->getName());
+      auto F = OpaqueValuesPool.get(Name.str(),
+                                    llvm::Type::getVoidTy(Context),
+                                    { Register->getValueType() },
+                                    Name);
+      auto Load = Builder.CreateLoad(Register->getValueType(), Register);
```
kolodziej, line 1025: @aleclearmind could you take a look on it? Now `Load` is used in some function call. Is it enough to limit optimizing out this load?

kolodziej: 1. You suggested in review that it will be eliminated.
2. But in the review you told me to eliminate separate pool and put everything into one pool. Or you thought only about `write_` functions?

aleclearmind: 1. OK, but did you check it? `:)` It was an hypothesis after some issues you mentioned on mattermost, if it's not DCE'd I wouldn't bother doing any of this.
2. The idea is to have one pool per goal: one goal is to generate opaque values to store in CSVs, a distinct goal is to prevent DCE of a load.

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -500,6 +528,117 @@ DetectABI::tryGetRegisterState(model::Register::Values RegisterValue,
...
+        Node::AnalysisType::URVOFC);
+
+    Nodes.UAOF->addSuccessor(Nodes.Args.get());
+
+    return Nodes;
+  };
+
+  GraphType G;
+
+  for (auto F : Binary->Functions()) {
```
aleclearmind, line 608: `auto` abuse ([B.15](https://pad.rev.ng/YwADn1W7QZGKGbY7yaVg_Q?view#))

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -500,6 +528,117 @@ DetectABI::tryGetRegisterState(model::Register::Values RegisterValue,
...
+    Nodes.UAOF = std::make_unique<ForwardNode<Node>>(Entry,
+        Node::AnalysisType::UAOF);
+    Nodes.RAOFC = std::make_unique<ForwardNode<Node>>(Entry,
+        Node::AnalysisType::RAOFC);
+    Nodes.URVOF = std::make_unique<ForwardNode<Node>>(Entry,
+        Node::AnalysisType::URVOF);
+    Nodes.URVOFC = std::make_unique<ForwardNode<Node>>(Entry,
+        Node::AnalysisType::URVOFC);
+
+    Nodes.UAOF->addSuccessor(Nodes.Args.get());
```
aleclearmind, line 601: 1. The API of `GenericGraph` was not really designed to add edges on nodes that are not attached to the graph, it might work but it's not the intended usage.
2. I don't understand why we're adding edges between UAOF and Args? What about all the other edges?

lib/EarlyFunctionAnalysis/DetectABIMFI.h
```diff
@@ -0,0 +1,212 @@
...
+  std::map<MetaAddress, OutlinedFunction> &OutlinedFunctions;
+  CFGAnalyzer &Analyzer;
+  TupleTree<model::Binary> &Binary;
+
+public:
+  using Label = BasicBlockNode *;
+
+  explicit DetectABIMFI(
+    GeneratedCodeBasicInfo &GCBI,
+    std::map<MetaAddress, OutlinedFunction> &OutlinedFunctions,
```
aleclearmind, line 90:
```suggestion
    const std::map<MetaAddress, OutlinedFunction> &OutlinedFunctions,
```

lib/EarlyFunctionAnalysis/DetectABIMFI.h
```diff
@@ -0,0 +1,212 @@
...
+
+    return SUL::isLessOrEqual(ReturnRegisters, A2.ReturnRegisters);
+  }
+};
+
+class LatticeElement {
+private:
+  using map_t = std::map<MetaAddress, FunctionABI>;
+  using iterator_t = map_t::iterator;
+  using const_iterator_t = map_t::const_iterator;
```
aleclearmind, line 49: Types are `CamelCase`, no `_t` suffix

lib/EarlyFunctionAnalysis/DetectABIMFI.h
```diff
@@ -0,0 +1,212 @@
...
+        Result[Address] = E1Abi;
+      }
+    }
+
+    for (const auto &[Address, E2Abi] : E2) {
+      auto ResultIt = Result.find(Address);
+      if (ResultIt == Result.end()) {
+        Result[Address] = E2Abi;
+      }
+    }
```
aleclearmind, line 117: You can rewrite this in a much more compact way using `zipmap_range`.
`git grep` its usages to see how it works or ping me if you struggle with it.

lib/EarlyFunctionAnalysis/DetectABIMFI.h
```diff
@@ -0,0 +1,212 @@
...
+
+    return Result;
+  }
+
+  [[nodiscard]] bool
+  isLessOrEqual(const LatticeElement &E1, const LatticeElement &E2) const {
+    for (const auto &[Address, E1Abi] : E1) {
+      auto It = E2.find(Address);
+      if (It != E2.end()) {
```
aleclearmind, line 126: Are you sure it's OK to ignore the situation in which `Address` is not in `E2`?
What should the default value be in case of absence?

lib/EarlyFunctionAnalysis/DetectABIMFI.h
```diff
@@ -0,0 +1,212 @@
...
+
+    const auto &Address = L->Address;
+    if (Address.isInvalid()) {
+      return Result;
+    }
+
+    auto Entry = GCBI.getBlockAt(Address);
+    if (!Entry) {
+      return Result;
+    }
```
aleclearmind, line 153: This looks like an assertion

lib/EarlyFunctionAnalysis/DetectABIMFI.h
```diff
@@ -0,0 +1,212 @@
...
+
+    return true;
+  }
+
+  [[nodiscard]] LatticeElement
+  applyTransferFunction(Label L, const LatticeElement &E2) const
+  {
+    LatticeElement Result = E2;
+
+    auto Reg = [this](const llvm::GlobalVariable *V) {
```
aleclearmind, line 141: No abbreviations

lib/EarlyFunctionAnalysis/DetectABIMFI.h
```diff
@@ -0,0 +1,212 @@
...
+      return model::Register::fromCSVName(V->getName(), Binary->Architecture());
+    };
+
+    const auto &Address = L->Address;
+    if (Address.isInvalid()) {
+      return Result;
+    }
```
aleclearmind, line 148: This looks like an assertion

lib/EarlyFunctionAnalysis/DetectABIMFI.h
```diff
@@ -0,0 +1,212 @@
...
+    if (!Entry) {
+      return Result;
+    }
+
+    using namespace ABIAnalyses;
+
+    ABIAnalysesResults ABIResults;
+
+    OutlinedFunction &OutlinedFunction = getOutlinedFunction(Address);
+    auto F = OutlinedFunction.Function.get();
```
aleclearmind, line 160: Reduce usage of `auto`

lib/EarlyFunctionAnalysis/DetectABIMFI.h
```diff
@@ -0,0 +1,212 @@
...
+
+    // Find registers that may be target of at least one store. This helps
+    // refine the final results.
+    auto WrittenRegisters = findWrittenRegisters(F);
```
aleclearmind, line 164: Unused?

lib/EarlyFunctionAnalysis/DetectABIMFI.h
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  OutlinedFunction &getOutlinedFunction(MetaAddress Address) const {
+    auto It = OutlinedFunctions.find(Address);
+    if (It != OutlinedFunctions.end()) {
+      return It->second;
+    }
+
+    return OutlinedFunctions[Address];
+  }
```
aleclearmind, line 209: This function is the same as simply doing `OutlinedFunctions[Address]`

lib/EarlyFunctionAnalysis/DetectABIMFI.h
```diff
@@ -0,0 +1,212 @@
...
+  [[nodiscard]] bool isLessOrEqual(const FunctionABI &A2) const {
+    if (!SUL::isLessOrEqual(ArgumentRegisters, A2.ArgumentRegisters)) {
+      return false;
+    }
+
+    return SUL::isLessOrEqual(ReturnRegisters, A2.ReturnRegisters);
+  }
+};
+
+class LatticeElement {
```
aleclearmind, line 45: How is this different from `using LatticeElement = std::map<MetaAddress, FunctionABI>`?

aleclearmind:
1. Comments on commit messages:
  * `Add EntryHook`: explain why we suddenly need it
2. Remember to import the design document in `docs/`.
3. Something fundamental seems to missing here. I expected hook functions (entry, pre call, post call, return) to be handled in a special way by the various analyses. For instance, when UAOF meets precall, it should act as if it met a read from all the registers that are arguments of the callee.

aleclearmind (changes requested):

PR #267, opened 2023-02-02
Add --binary argument to llvm_pipeline.py

kolodziej:


include/revng/Model/RawBinaryView.h
```diff
@@ -52,6 +52,15 @@ class RawBinaryView {
     return getByOffset(*Offset, Size);
   }
 
+  std::optional<llvm::StringRef>
+  getStrByAddress(MetaAddress Address, uint64_t Size) const {
```
aleclearmind, line 56: No abbreviations `getStringByAddress`

lib/Support/IRHelpers.cpp
```diff
@@ -78,6 +78,7 @@ static std::string mangleName(StringRef String) {
 }
 
 Constant *getUniqueString(Module *M, StringRef String, StringRef Namespace) {
+  revng_assert(not Namespace.empty());
```
aleclearmind, line 81: I'm OK with this, maybe add docs. But frankly I was thinking about dropping `Namespace` altogether and have a fixed one.

I suppose you triggered this because you're explicitly specifying it (as opposed to using the default value for this argument). Do you *really* care about having a distinct global variable if there's another user of `getUniqueString` requesting the same string? Do you need them to have different addresses? Do you need to change its content?

If not, let's collapse them all in a single namespace. If you're serializing `MetaAddress` or a `QualifiedType` or the like there's a good chance we get deduplication for free and save a lot of space.

lib/Support/IRHelpers.cpp
```diff
@@ -79,19 +79,35 @@ static std::string mangleName(StringRef String) {
...
 Constant *getUniqueString(Module *M, StringRef String, StringRef Namespace) {
   revng_assert(not Namespace.empty());
+
   LLVMContext &Context = M->getContext();
   std::string GlobalName = (Twine(Namespace) + mangleName(String)).str();
   auto *Global = M->getGlobalVariable(GlobalName);
-  ConstantDataSequential *Initializer = nullptr;
 
-  if (Global != nullptr) {
+  if (Global) {
```
aleclearmind, line 87: Explicit is better than implicit, I prefer to have `!= nullptr`

lib/Support/IRHelpers.cpp
```diff
@@ -79,19 +79,35 @@ static std::string mangleName(StringRef String) {
...
-    Initializer = cast<CDA>(CDA::getString(Context,
-                                           String,
-                                           /* AddNull */ true));
+    // This may return a ConstantAggregateZero in case of empty String.
+    Constant *Initializer = ConstantDataArray::getString(Context,
+                                                         String,
+                                                         /* AddNull */ true);
+    revng_assert(isa<ConstantDataArray>(Initializer)
+                 or isa<ConstantAggregateZero>(Initializer));
+    if (not String.empty()) {
```
aleclearmind, line 103: Drop `not` and flip

aleclearmind (changes requested):

lib/Support/IRHelpers.cpp
pfez: We're using this for string literals to be emitted in C, so no deduplication for `MetaAddress` and `QualifiedType`. On the other hand I expect potential deduplication due to the same string literal being used possibly in many places in C.

For what concerns namespaces, I don't really care about them though. For what I am concerned we can just get rid of the `Namespace` argument alltogether, and only always use the same namespace.

aleclearmind: I've some FUD that one day we'll need to distinguish strings, let's this feature survive for a little longer.
In the meantime, let's use the default value instead of explicitly choosing a namespace.

kolodziej: We already have default value for this, so I think for now it's resolved.

aleclearmind: > We already have default value for this

Not sure what you mean.

I'm suggesting to drop the manually specified namespace name in `revng-c` when invoking `getUniqueString`.

pfez: > I'm suggesting to drop the manually specified namespace name in revng-c when invoking getUniqueString.

@kolodziej you can do this. I'm sure it doesn't affect any other part of your PR in revng-c

aleclearmind: Sure

pfez: Ok so this is solved. I'm closing this conversation.

pfez:
Merged in 2159dbbdc7b51b40facc8e87b32bc665232dbe52

Thanks!

PR #268, opened 2023-02-02
no error list preliminaries

drblallo:


aleclearmind:
Merged in b89c3e9. Thanks!

PR #269, opened 2023-02-03
LLVMContainerBase::cloneFiltered: fix dbg metadata

pfez:
Before this commit the cloneFiltered method was duplicating debug metadata on cloned functions, causing the Module not to verify after cloning.
The reason is a nasty combination of how llvm::cloneModule and llvm::Function::clearMetadata work internally:
- llvm::cloneModule already clones debug metadata, so the cloned functions have debug metadata
- llvm::Function::clearMetadata does not clear debug metadata
- so it's wrong to manually copy debug metadata from old functions to cloned functions

This commit fixes the problem by skipping the copy of the debug metadata alltoghether, since they're already in place.

### DIFF
--- a/include/revng/Pipeline/LLVMContainer.h
+++ b/include/revng/Pipeline/LLVMContainer.h
@@ -127,15 +127,19 @@ public:
       Other->clearMetadata();
       llvm::SmallVector<std::pair<unsigned, llvm::MDNode *>, 2> MDs;
       Function.getAllMetadata(MDs);
-      for (auto &MD : MDs) {
-        // The !dbg attachment from the function defintion cannot be attached to
-        // its declaration.
-        if (Other->isDeclaration() && isa<llvm::DISubprogram>(MD.second))
+      for (auto &[MDKindID, MetaData] : MDs) {
+        // CloneModule clone already clones debug metadata, and clearMetadata
+        // does not clear debug metadata from functions.
+        // We don't want to duplicate debug metadata because that causes the
+        // model to not verify.
+        if (llvm::isa<llvm::DINode>(MetaData))
           continue;
 
-        Other->addMetadata(MD.first, *llvm::MapMetadata(MD.second, Map));
+        Other->addMetadata(MDKindID, *llvm::MapMetadata(MetaData, Map));
       }
     }
+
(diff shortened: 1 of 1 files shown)


drblallo:
i think we may still need to manually copy debug metadata too if the destination function is a declaration. That too is kind of arbitrary and up to the particular version 
 of LLVM.

pfez:
> i think we may still need to manually copy debug metadata too if the destination function is a declaration.

Why do you think so? AFAIK `cloneModule` already copies the debug metadata. Isn't this the case for the specific LLVM version we're using?

include/revng/Pipeline/LLVMContainer.h
```diff
@@ -122,10 +122,18 @@ class LLVMContainerBase
       Other->clearMetadata();
       llvm::SmallVector<std::pair<unsigned, llvm::MDNode *>, 2> MDs;
       Function.getAllMetadata(MDs);
-      for (auto &MD : MDs)
-        Other->addMetadata(MD.first, *llvm::MapMetadata(MD.second, Map));
+      for (auto &[MDKindID, MetaData] : MDs) {
+        // CloneModule clone already clones debug metadata, and clearMetadata
+        // does not clear debug metadata, so we don't want to duplicate debug
+        // metadata because that causes the model to not verify.
```
aleclearmind, line 128:
```suggestion
        // metadata because that causes the module to not verify.
```

aleclearmind:
Please rebase, I get conflicts. We might already have a fix for this.

pfez:
> Please rebase, I get conflicts. We might already have a fix for this.

Will do!

aleclearmind:
Under which circumstance was this creating problems?

I'm fiddling with this piece of code.

pfez:
> Under which circumstance was this creating problems?

I'm sorry but I don't remember exactly, it was already 2 months ago. I was debugging a pass downstream in `revng-c` and the module I was getting did not pass `verify` for this reason. So overall this may hamper debugging.

aleclearmind:
I'll try this post LLVM 16 I guess.

pfez:
Now that we've rebased on top of llvm-16 I'm retesting this. I'll mark this as draft until it's confirmed or it can be closed.

PR #273, opened 2023-02-28
Prepare LLVM IR for MLIR

dwarfyresearch:


lib/FunctionIsolation/StripDebugInfoFromHelpers.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "llvm/IR/DebugInfo.h"
+#include "llvm/Pass.h"
+
+#include "revng/Pipeline/RegisterLLVMPass.h"
+#include "revng/Support/FunctionTags.h"
+
+using namespace llvm;
+
+// Debug Info on helpers is not very useful after this phase, since the binary
+// cannot be run after this phase.
```
aleclearmind, line 14: Use doxygen for this kind of comments

aleclearmind:
> Up until now, we did not need different !dbg/DISubprogram attached to each isolated llvm::Function, but for the purpose of MLIR lowering, we need it.

Why again?

pfez:
@aleclearmind I dropped the "ready" label temporarily, so that this doesn't get merged by mistake.

This is due to a commit @dwarfyresearch has pushed here recently, that I've asked him to remove from the patchset since we'll fix the issue directly in `revng-c`.

As soon as @dwarfyresearch drops the last commit I'll re-add the "ready" label.

PR #274, opened 2023-03-07
Add dynamic call support to the call graph

Cvelth:


include/revng/GraphLayout/Graphs.h
```diff
@@ -7,7 +7,7 @@
 #include <unordered_map>
 
 #include "revng/ADT/GenericGraph.h"
-#include "revng/Yield/Support/GraphLayout/Traits.h"
+#include "revng/GraphLayout/Traits.h"
```
aleclearmind, line 10: Yes please

include/revng/EarlyFunctionAnalysis/ControlFlowGraph.h
```diff
@@ -75,17 +75,22 @@ inline ParsedSuccessor parseSuccessor(const T &Edge,
...
-/// using the `IsBasicBlock` concept. These blocks are required to have start
-/// and end addresses as well as a list of their successors. It's expected for
-/// the graph it represents to be self contained, as in "no block can ever
-/// reference another block that is not listed in this container".
+/// \param BB An arbitrary container of basic blocks that are verified
+///        using the `IsBasicBlock` concept. These blocks are required to have
+///        a start and an end addresses as well as a list of their successors.
+///        It's expected for the graph it represents to be self contained, as
+///        in "no block can ever reference another block that is not listed in
+///        this container".
```
aleclearmind, line 83: Comment on the commit:

>  EFA: upgrade buildControlFlowGraph
>
> It now assumes that the nodes of the graph it constructs also require
> the function entry point on top of the basic block ID.

Why?
Later on we switch to `pipeline::Location` which makes this change obsolete. No fixups!
Squash?

include/revng/Yield/ControlFlow/Graph.h
```diff
@@ -5,20 +5,44 @@
...
-
-  explicit Node(BasicBlockID Address = BasicBlockID::invalid()) :
-    Address(std::move(Address)) {}
-  explicit Node(BasicBlockID Address, const MetaAddress &) :
-    Address(std::move(Address)) {}
-
-  bool isEmpty() const { return !Address.isValid(); }
+class Node {
+private:
+  std::string Location = {};
```
aleclearmind, line 16: Why store a string given that we seem to assume a specific location type (i.e., `revng::ranks::BasicBlock`)?

include/revng/Yield/CrossRelations/RelationTarget.h
```diff
@@ -13,11 +13,12 @@ type: struct
 fields:
   - name: Kind
     type: RelationType
-  - name: Location
-    type: string
+  - name: Locations
+    sequence:
+      type: SortedVector
+      elementType: string
```
aleclearmind: Why do we have a set of locations?

lib/Yield/CrossRelations.cpp
```diff
@@ -52,21 +53,19 @@ CR::CrossRelations::CrossRelations(const MetadataContainer &Metadata,
               //       `CallLocation` after metadata starts providing it.
 
               auto CalleeAddress = Callee.notInlinedAddress();
-              auto L = pipeline::location(ranks::Function, CalleeAddress)
```
aleclearmind, line 55: A `using namespace pipeline` here would increase readability significantly.

include/revng/Yield/CrossRelations/RelationType.h
```diff
@@ -11,7 +11,6 @@ doc: Enum for identifying different instruction markup tag types
 type: enum
```
aleclearmind: >  CrossRelations: merge IsCalledFrom flavors
>
> I had FUD about using the same `Type` for both, but now looking into
> how similar working with both is, there's no reason to keep them
> separated.

No first person in commit messages

include/revng/Yield/CrossRelations/RelationDescription.h
```diff
@@ -14,10 +13,11 @@ type: struct
 fields:
   - name: Location
     type: string
-  - name: Related
+  - name: IsCalledFrom
     sequence:
       type: SortedVector
-      elementType: RelationTarget
+      elementType: string
+    optional: true
```
aleclearmind, line 20: This is a fixup over the previous change.

No fixup commits. Please don't do this, it makes my review longer and in the end I always have to ask you to drop them.

lib/Model/Binary.cpp
```diff
@@ -156,6 +156,16 @@ bool Binary::verify(VerifyHelper &VH) const {
       return VH.fail();
   }
 
+  // Make sure no segments overlap
+  if (!Segments().empty()) {
+    for (const auto &[LHS, RHS] :
+         llvm::zip(skip_back(Segments()), skip_front(Segments()))) {
```
aleclearmind, line 162: `:O` `:O` `:O` `:O`

lib/Model/Binary.cpp
```diff
@@ -156,6 +156,16 @@ bool Binary::verify(VerifyHelper &VH) const {
...
+      revng_assert(LHS.StartAddress() <= RHS.StartAddress());
+      if (LHS.endAddress() > RHS.StartAddress())
+        return VH.fail("Segments must not overlap.");
```
aleclearmind, line 165: Pass to fail at least one of the two, so it will get printed

aleclearmind (changes requested):

include/revng/Yield/ControlFlow/Graph.h
Cvelth: Two reasons:
1. for consistency with the other node type, which can in fact point to two distinct location types.
2. because we're likely to want to extend this in the future with different types of nodes.

include/revng/Yield/CrossRelations/RelationDescription.h
Cvelth: Ah wait, I thought I squashed this one. It's my bad.

lib/Model/Binary.cpp
Cvelth: It's your idea actually. We had the same construct in the ABI PR, which was super easy to adopt here as well.

P. S. We probably should also introduce a named helper for this, something like
```cpp
template<typename Range>
constexpr auto pairwise_view(Range R) { return llvm::zip(skip_back(R), skip_front(R)); }
```

include/revng/Yield/CrossRelations/RelationTarget.h
Cvelth: Before now the `cross-relations` used to look like
```yaml
Relations:
  - Location:        '/function/0x401040:Code_x86_64'
    Related:
      - Kind:            IsCalledFrom
        Location:        '/instruction/0x401100:Code_x86_64/0x401100:Code_x86_64/:Invalid'
      - Kind:            IsCalledFrom
        Location:        '/instruction/0x402100:Code_x86_64/0x402100:Code_x86_64/:Invalid'
      - Kind:            IsCalledFrom
        Location:        '/instruction/0x403100:Code_x86_64/0x403100:Code_x86_64/:Invalid'
  - Location:        '/dynamic-function/my_cool_function'
    Related:
      - Kind:            IsDynamicallyCalledFrom
        Location:        '/instruction/0x401149:Code_x86_64/0x401149:Code_x86_64/:Invalid'
```

Now it looks like
```yaml
Relations:
  - Location:        '/function/0x401040:Code_x86_64'
    Related:
      - Kind:            IsCalledFrom
        Locations:
          - '/instruction/0x401100:Code_x86_64/0x401100:Code_x86_64/:Invalid'
          - '/instruction/0x402100:Code_x86_64/0x402100:Code_x86_64/:Invalid'
          - '/instruction/0x403100:Code_x86_64/0x403100:Code_x86_64/:Invalid'
  - Location:        '/dynamic-function/my_cool_function'
    Related:
      - Kind:            IsCalledFrom
        Locations:       
          - '/instruction/0x401149:Code_x86_64/0x401149:Code_x86_64/:Invalid'
```

which is a lot easier to work with.

include/revng/GraphLayout/Graphs.h
Cvelth: Yeah, we couldn't do this until now because it depended on the `yield::Graph`, so it would have resulted in a dependency cycle. But thanks to the #278 effort that's no longer the case.

include/revng/EarlyFunctionAnalysis/ControlFlowGraph.h
Cvelth: Ah, wait, no, I mixed pieces up. This change IS necessary exactly because we switch stuff relying on this to use locations later.

Before now we were using `BasicBlockID`s, so all we needed to build one was an ID. Now that the locations are built from the stuff this function returns, we also need to know the function entry point for each basic block we receive. Hence, this change.

Cvelth: It could still be squashed, technically, since it's done for a similar purpose. BUT consider that this commit affects `efa` part of the codebase, while being the only commit to do so. As such, I'd rather keep it as self-contained as possible.

aleclearmind: 1. Commit X: don't just use `BasicBlockID` but add function entry `MetaAddress`
2. Commit Y: drop `BasicBlockID` and `MetaAddress` in favor of `Location<rank::BasicBlock>` which embeds both

What prevents you from merging these two commits?

include/revng/Yield/ControlFlow/Graph.h
aleclearmind: 1. Can we factor them out?
2. If so, add a comment.

include/revng/EarlyFunctionAnalysis/ControlFlowGraph.h
Cvelth: It seems like you're misunderstanding the situation a bit.

The focus of the first commit is to make it so `buildControlFlowGraph` thing also throws a function address at the caller. (as of now it's a super-generic graph-builder that works with any `GenericGraph` you pass to it as long as its node is constructible from `{ BasicBlockID Block, MetaAddress Function }`)

And then the second commit does not really interact with it directly, it only reworks the internals of the graphs, but still keeps the constructor signature the same.

So, in essence, `X` is "patch **efa** to also give me function addresses" and `Y` is "update the **yield** graphs to store locations internally instead of whatever is used now".

include/revng/Yield/ControlFlow/Graph.h
Cvelth: 1. I tried to avoid unnecessary inheritance, there's no guarantee that we always want to keep unrelated node types linked. But if you want, it's not hard to separate common parts of the interface.

2. Sure thing.

include/revng/EarlyFunctionAnalysis/ControlFlowGraph.h
aleclearmind: Gotcha, I'll check this again. I might have misunderstood.

include/revng/Yield/ControlFlow/Graph.h
aleclearmind: 1. No, OK. I'm persuaded.
2. :+1:

include/revng/EarlyFunctionAnalysis/ControlFlowGraph.h
aleclearmind: I was referring to the following change in the "EFA: upgrade buildControlFlowGraph" commit:

```patch
 struct Node {
   BasicBlockID Address;
   BasicBlockID NextAddress;

   explicit Node(BasicBlockID Address = BasicBlockID::invalid()) :
     Address(std::move(Address)) {}
+  explicit Node(BasicBlockID Address, const MetaAddress &) :
+    Address(std::move(Address)) {}
 };
```
Which is then basically completely overridden by a complete rewrite of the `struct Node` in "ControlFlow: switch to pipeline::Location".

Cvelth: Ah, sorry for misunderstanding you then.

That's just an ad-hoc fix so that the code compiles after the change to the `efa`'s interface.

Cvelth: I'm not sure I agree with "when changing an interface, all the users currently using it need to be update to use the new interface" qualifying as a trick.

The fact that we might want to rework some of these users later in the same PR doesn't feel like a reason enough to squash two logically-unrelated changes into a single commit.

Cvelth: I guess that's a fair point.

I really didn't want to have to reorder these two commits (because then everything between them emits broken locations), but I guess it's still better than just squashing them.

aleclearmind: I appreciate your effort to make small commits and have a great history, but I'd like to get yourself tuned a lil0 bit more towards having larger commits.

Cvelth: I'm doing it for a selfish reason: I often use `blame` view when looking at my old code to recover context. Having these things smaller means, well, reading less code to get the picture.

lib/Yield/CrossRelations.cpp
Cvelth: Done.

include/revng/Yield/CrossRelations/RelationType.h
Cvelth: Reworded.

lib/Model/Binary.cpp
Cvelth: I went with a bit more explicit error message, but I can't imagine that's controversial. So I'm closing this.

aleclearmind:
Please retarget to `develop`

include/revng/Yield/ControlFlow/Graph.h
aleclearmind: Works

Cvelth: Ah, there's a typo in the second one. Keeping this open so I don't forget to fix it.

Cvelth: Done.

aleclearmind:
Merged in 3c77bb5. Thanks!

PR #276, opened 2023-03-09
Introduce tracking of TTGs.

drblallo:
TTGs can now be tracked, that you can query the path of all fields that are accessed.


include/revng/ADT/TrackingContainer.h
```diff
@@ -222,7 +252,8 @@ class TrackingContainer {
   size_type count(const key_type &Key) const {
     auto Quantity = Content.count(Key);
     if (Quantity == 0) {
-      NonExisting.back().insert(Key);
+      if (IsTracking)
```
aleclearmind, line 255: Hoist out of `if (Quantity == 0) {`

include/revng/TupleTree/Visits.h
aleclearmind, line 1: NOTETOSELF: review

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -165,7 +165,7 @@ class DetectABI {
 private:
   CSVSet computePreservedCSVs(const CSVSet &ClobberedRegisters) const;
 
-  SortedVector<model::Register::Values>
+  TrackingSortedVector<model::Register::Values>
```
aleclearmind, line 168: Isolate this type of changes in its own commit

lib/EarlyFunctionAnalysis/FunctionSummaryOracle.cpp
```diff
@@ -165,18 +165,23 @@ void importModel(Module &M,
...
+        .registerCallSite(Function.Entry(),
+                          BasicBlockID(CallSite.CallerBlockAddress()),
+                          importPrototype(M,
+                                          ABICSVs,
+                                          AttributesSet(CallSite.Attributes()
+                                                          .begin(),
+                                                        CallSite.Attributes()
+                                                          .begin()),
+                                          CallSite.Prototype()),
+                          CallSite.IsTailCall());
```
aleclearmind, line 178: 1. Split in multiple statements, same for next changes
2. You're using `.begin` twice!

aleclearmind (changes requested):

include/revng/ADT/MutableSet.h
```diff
@@ -78,6 +78,14 @@ class MutableSet {
     }
   }
 
+  template<typename Iter>
+  MutableSet(Iter Begin, Iter End) {
+    auto Inserter = batch_insert_or_assign();
+    for (const T &Element : llvm::make_range(Begin, End)) {
+      Inserter.insert_or_assign(Element);
+    }
+  }
```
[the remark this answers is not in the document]
drblallo: we need insert_or_asign because the source range may be a vector, and the user is trying to deduplicate it

include/revng/ADT/TrackingContainer.h
```diff
@@ -76,6 +82,11 @@ class TrackingContainer {
                              ToReturn.InspectedKeys.begin()));
 
     ToReturn.Exact = Exact.isSet();
+    if (Exact.isSet())
+      for (auto &Entry : Content) {
+        using KOT = KeyedObjectTraits<value_type>;
+        ToReturn.InspectedKeys.insert(KOT::key(Entry));
+      }
```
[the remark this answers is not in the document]
drblallo: i do it because i find out that is easier for some parts of the code on being able to see that we rely on all paths by enumerating all of them, instead of pushing the if that checks if the container is Exact in all those locations.

include/revng/ADT/TrackingContainer.h
```diff
@@ -109,6 +120,20 @@ class TrackingContainer {
...
+  TrackingContainer &operator=(TrackingContainer &&) = default;
+  TrackingContainer &operator=(const TrackingContainer &) = default;
+  TrackingContainer &operator=(const T &Other) {
+    Content = Other;
+    return *this;
+  }
+  TrackingContainer &operator=(T &&Other) {
+    Content = std::move(Other);
+    return *this;
+  }
```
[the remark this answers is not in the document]
drblallo: ahh right, well i do need to allow copies somehow, if i don't put them it will break compatibility with the underlying container signatures

include/revng/ADT/TrackingContainer.h
drblallo: not sure what you mean, i can't just hoist it out, the semantic is not the same

include/revng/ADT/TrackingContainer.h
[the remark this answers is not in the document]
aleclearmind: That doesn't sound safe, you'd be ignoring the fact that you're also relying on the *non*-existence of al those *not* listed.

include/revng/ADT/TrackingContainer.h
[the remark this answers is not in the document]
aleclearmind: You have a regular copy constructor, why do you need a constructor taking `T`?

include/revng/ADT/TrackingContainer.h
aleclearmind: I'm suggesting

```c++
    if (TrackingIsActive) {
      if (Quantity == 0) {
       NonExisting.back().insert(Key);
      } else {
        markExistingKey(Key);
      }
    }
```

drblallo: ahh right, yeah make sense

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
```diff
@@ -142,9 +155,33 @@ public:
...
+  }
+
+  llvm::SmallVector<revng::AccessTracker*, 3> getKeyTrackers() const {
+    return {
+      /**- for key_field in struct.key_fields -**/
+      &/*= key_field.name =*/Tracker,
+      /**- endfor -**/
+    };
+  }
+  /** endif -**/
```
aleclearmind, line 183: 1. You can return a `std::array`
2. Return `const revng::AccessTracker *`
3. Add space before `*`
4. Are you sure you want raw access to these? Maybe we should have these return a set of indexes representing the filed IDs that are changed and keep the fact we're using `AccessTracker` an implementation detail.

include/revng/ADT/TrackingContainer.h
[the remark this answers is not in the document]
drblallo: no wait, what do you mean this enables the user to bypass tracking? You can't invoke this when they are const?

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct_late.h.tpl
```diff
@@ -46,6 +46,18 @@ template <> struct TupleLikeTraits</*=- struct | user_fullname =*/> {
...
+  /**- if emit_tracking **/
+template <int I> revng::AccessTracker &getTracker(const /*= struct.name =*/ &x) {
+  if constexpr (false)
+    return *std::declval<revng::AccessTracker*>();
+  /**- for field in struct.all_fields **/
+  else if constexpr (I == /*= loop.index0 =*/)
+    return x./*= field.name =*/Tracker;
+  /**- endfor **/
+}
+  /** endif -**/
```
aleclearmind, line 59: If we were to generate a `std::tuple<...>` for both fields and access trackers we could generate less code.
This would just become a `std::get<I>(TrackerTuple)`.

We didn't use `std::tuple` for fields since in the past the fields were public fields, but since we switched to accessors this is no longer required.

What do you think?

aleclearmind (changes requested):

include/revng/ADT/TrackingContainer.h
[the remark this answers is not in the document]
aleclearmind: Uhm, right!

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct_late.h.tpl
drblallo: i did it this way to have simmetry with the get of the field, so you can do get<5> and getTracker<5>, which helps writing easier visits, the tuple idea is viable, but it complicates other generated code because then one has to write get<TTFunctionThatGoesFromFieldToFieldIndex> rather than fieldNameTracker.

no strong opinions toh

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
```diff
@@ -35,15 +37,26 @@ private:
   /*= field | field_type =*/ The/*= field.name =*/ = /*= field | field_type =*/{};
   static_assert(Yamlizable</*= field | field_type =*/>);
 
+  /**- if emit_tracking **/
+public:
```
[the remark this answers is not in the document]
drblallo: this was intentional, but it is better discussed in the issue below this one

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
drblallo: 4. We need to provide access to them since the visits are in cpp code. We can return the id the field i guess, but i don't see much reason to generate that code here rather than the cpp. Nobody but the visits has a interest in reading this stuff.

In particular we need at the very least the getTracker<Index> method, that method is outside of the class itself, and thus requires that the tracker itself is accessible, for that reason i made the public. I can write a generate for each toh. This accessors that return the list of them are just here for simplicity.

not sure which design you like most, but i really doubt we will encounter issue in the future regardless of which path we take.

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct_late.h.tpl
aleclearmind: Next time

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
aleclearmind: > We can return the id the field i guess, but i don't see much reason to generate that code here rather than the cpp

I'd be much happier to hide the implementation detail and make things private.
Let's return a set of indices. If you don't like generating too much code here, write an helper function in a header and use it here, but I don't see a reason to make any of this public.

include/revng/ADT/TrackingContainer.h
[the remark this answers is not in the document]
drblallo: i thought about this, i think this is correct and they are two different things.
From a high level point of view we will have a visit that returns all read existing read fields of a model, and then a visit tha give a set of paths will tell you which ones added a entry to a vector that was marked as exact. This is the only implementation that makes sense now that all tracking primitives are private.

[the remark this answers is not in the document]
drblallo: what do you mean? it is not irrelevant. If you are considering a diff that is just a removal you don't care about exactness, you only care about the list all read nodes.

lib/EarlyFunctionAnalysis/DetectABI.cpp
drblallo: made a commit containing changes to the generator and consequences to other parts of the code base

include/revng/ADT/TrackingContainer.h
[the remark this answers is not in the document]
aleclearmind: I still see this in first commit. Please review the history before asking for reviews.

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
```diff
@@ -142,9 +155,76 @@ public:
...
+      ;
+    /**- for field in struct.all_fields **/
+    else if constexpr (I == /*= loop.index0 =*/)
+        /*= field.name =*/Tracker.push();
+    /**- endfor -**/
+    else
+      revng_assert(false);
+  }
+
+  template<size_t I> void clearTracker() const {
```
aleclearmind, line 215: Let's be consistent with terminology. Elsewhere we are using `reset`. Let's stick with `clear` everywhere.

aleclearmind (changes requested):

tools/efa/extractcfg/DecoratedFunction.h
```diff
@@ -23,7 +23,7 @@ type: struct
     type: efa::FunctionMetadata
   - name: Attributes
     sequence:
-      type: MutableSet
+      type: TrackingMutableSet
```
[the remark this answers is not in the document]
drblallo: not intended

lib/EarlyFunctionAnalysis/FunctionSummaryOracle.cpp
```diff
@@ -165,18 +165,24 @@ void importModel(Module &M,
     for (const model::CallSitePrototype &CallSite :
          Function.CallSitePrototypes()) {
 
+      AttributesSet Attributes;
+      for (auto &ToCopy : CallSite.Attributes())
+        Attributes.insert(ToCopy);
```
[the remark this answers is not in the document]
drblallo: not sure i understand the question. The model contains TrackingVectors, and the other datastructures unrelated to the model contain SortedVectors, so they can't be trivially copied. We can add a constructor to sortedvector and mutable set if you prefer.

include/revng/ADT/TrackingContainer.h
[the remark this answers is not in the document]
aleclearmind: Ping

lib/EarlyFunctionAnalysis/FunctionSummaryOracle.cpp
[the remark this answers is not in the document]
aleclearmind: My question was why do we have to do this, it wasn't obvious to me that it was due to model containers having become tracking.

STL data structures have a constructor taking `.begin()` and `.end()`, what do you think?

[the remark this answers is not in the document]
drblallo: this ends up in the discussion about mutable set being unclear about which instance of values that compare equal ends up picking. It is the second issue from the top in this MR.

include/revng/ADT/TrackingContainer.h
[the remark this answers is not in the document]
drblallo: if i understood correctly your concern this should be gone. i did `git reset --soft` are recreated manually each commit

include/revng/TupleTree/TrackingVisits.h
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  template<NotTupleTreeCompatible T>
+  void popImpl(const T &LHS) {}
+};
+
+} // namespace revng::detail
+
+template<TupleSizeCompatible T>
+ReadFields collectReadFields(const T &Object) {
```
aleclearmind, line 218: Can we promote this to methods?

This header contains quite a bit of private stuff, but these functions are very much to be used by users.
I feel like users should ignore the existence of `TrackingVisits.h` and just use some very simple methods on the `struct`s themselves.

include/revng/TupleTree/TrackingVisits.h
aleclearmind, line 1: This file is called `TrackingVisits.h` but it visits *without* tracking.

Given that the plan is to unify all classes in a single one (`TrackersManager`?), rename this files as that class.

include/revng/TupleTree/TrackingVisits.h
```diff
@@ -0,0 +1,239 @@
...
+  std::set<TupleTreePath> ExactVectors;
+};
+
+namespace revng::detail {
+
+template<typename M>
+struct ReadFieldsCollector {
+  TupleTreePath Stack;
+
+  ReadFields Info;
```
aleclearmind, line 30: Promote to local variable of `collect` and pass them down as arguments to `collectImpl`s.

include/revng/TupleTree/TrackingVisits.h
```diff
@@ -0,0 +1,239 @@
...
+
+  ReadFields collect(const M &LHS) {
```
aleclearmind, line 32: Every other class has a `run` method, make sure we're consistent.

aleclearmind (changes requested):

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
```diff
@@ -31,19 +68,30 @@ struct /*= struct | fullname =*/
   /**- for field in struct.fields **/
   /*= field.doc | docstring =*/
   /**- if field.const **/const /** endif -**/
-private:
+protected:
```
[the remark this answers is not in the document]
drblallo: we need to let inheritors access fields bypassing tracking because upcastable points uses parts of the base class to compose their key.

lib/EarlyFunctionAnalysis/FunctionSummaryOracle.cpp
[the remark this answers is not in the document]
aleclearmind: The destination is being written so the tracking is entirely irrelevant.
Best thing would not to copy any form of tracking.

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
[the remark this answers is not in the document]
aleclearmind: Uhm, no, we can't inheritors bypass tracking. Users *will* f- up.

Not sure what you're referring to about upcastable pointers.

lib/EarlyFunctionAnalysis/FunctionSummaryOracle.cpp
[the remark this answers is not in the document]
drblallo: it did copied before too, generated classes cannot have references. what changed is not the contained stuff, just that one field is a sorted vector and the source was a tracking vector. 

anyway this is unrelated to the issue of begin() end() constructor. I had that constructor in one of the earlier version of this pr and then it got removed because  the semantic was unclear when the input range had multiple copies of the same element.

include/revng/TupleTree/TrackingVisits.h
drblallo: i tried this, but it can't be done with the current hierarchy, to be a method it must be a method of the user defined class, not of the base generated class, otherwise you fail to be able to tell apart if the current generated class is TupleTreeCompatible. Anyway it should not really matter, there will be just one user of this ideally

include/revng/TupleTree/TrackingVisits.h
drblallo: i went for TupleTree/Tracking.h, since the manager is just a implementation detail and it is hidden from the true api

aleclearmind: In 99% of the cases the file name should reflect the main class it contains.
If there are multiple main classes, make multiple headers. I often find myself struggling to find where is a class defined in pipeline and I'd like to reduce this effect.

drblallo: there is one class yes, but that class is inside detail namespace and should not never be used outside of that header. The real declarations that should be used are the 4 visits, which are free functions.

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
```diff
@@ -31,19 +68,30 @@ struct /*= struct | fullname =*/
...
+  /** endif -**/
+
 public:
   using /*= field.name =*/Type = /*= field | field_type =*/;
 
 public:
   const /*= field | field_type =*/ & /*= field.name =*/() const {
+  /**- if emit_tracking **/
+    /*= field.name =*/Tracker.access();
+  /** endif -**/
```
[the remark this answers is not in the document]
aleclearmind: Nested too

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
```diff
@@ -142,9 +172,54 @@ public:
...
+  /**- endfor **/
+  /** endif **//*# End UpcastablePointer stuff #*/
+
+  template<size_t I>
+  const auto& untrackedGet() const {
+    if constexpr (false)
+      return 0;
+    /**- for field in struct.all_fields **/
+    else if constexpr (I == /*= loop.index0 =*/)
+        return The/*= field.name =*/;
```
aleclearmind, line 205:
```suggestion
      return The/*= field.name =*/;
```

include/revng/TupleTree/Tracking.h
```diff
@@ -0,0 +1,225 @@
...
+
+struct ReadFields {
+  std::set<TupleTreePath> Read;
+  std::set<TupleTreePath> ExactVectors;
+};
+
+namespace revng::detail {
+
+template<typename M>
+struct TrackingImpl {
```
aleclearmind, line 27: 1. Make all methods `static`
2. `s/TrackingImpl/Tracking/`
3. Drop methods below

include/revng/TupleTree/Tracking.h
```diff
@@ -0,0 +1,225 @@
...
+  void clearTuple(const T &LHS) {
+    if constexpr (I < std::tuple_size_v<T>) {
+
+      clearImpl(LHS.template untrackedGet<I>());
+      LHS.template getTracker<I>().clear();
+
+      // Recur
+      clearTuple<I + 1>(LHS);
+    }
+  }
```
aleclearmind, line 107: Collapse implementations of `clear`, `push` and `pop` in a single one taking a lambda or something.

aleclearmind (changes requested):

scripts/tuple_tree_generator/tuple-tree-generate-cpp.py
```diff
@@ -20,6 +20,7 @@
 argparser.add_argument("--root-type", required=True, help="Schema root type")
 argparser.add_argument("--scalar-type", action="append", default=[], help="Scalar type")
 argparser.add_argument("--include-path-prefix", required=True, help="Prefixed to include paths")
+argparser.add_argument("--tracking", type=bool, help="Emits tracking sequences", default=False)
```
aleclearmind, line 23:
```suggestion
argparser.add_argument("--tracking", action="store_true", help="Emits tracking sequences")
```

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
```diff
@@ -142,9 +157,54 @@ public:
...
+    /**- for field in struct.all_fields **/
+    if constexpr (I == /*= loop.index0 =*/)
+        return /*= field.name =*/Tracker;
+    /**- endfor -**/
+  }
+  /** if upcastable **/
+  /**- for child_type in upcastable|sort(attribute="user_fullname") **/
+  friend /*= child_type | fullname =*/;
+  /**- endfor **/
+  /** endif **//*# End UpcastablePointer stuff #*/
```
aleclearmind, line 195: Put "End UpcastablePointer stuff" on a new line

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
```diff
@@ -142,9 +157,54 @@ public:
...
+  Key untrackedKey() const {
+    return {
+      /** for key_field in struct.key_fields -**/
+      The/*= key_field.name =*/
+      /**- if not loop.last **/,
+      /** endif **/
+      /**- endfor **/
+    };
+  }
+  /** endif **/
```
aleclearmind, line 217: Drop

share/revng/cmake/TupleTreeGenerator.cmake
```diff
@@ -232,19 +234,26 @@ function(
...
+  SCALAR_TYPES
+  EMIT_TRACKING)
 
   set(SCALAR_TYPE_ARGS)
   foreach(ST ${SCALAR_TYPES})
     list(APPEND SCALAR_TYPE_ARGS --scalar-type "'${ST}'")
   endforeach()
 
+  if(${EMIT_TRACKING})
+    set(TRACKING "--tracking=true")
```
aleclearmind, line 246:
```suggestion
    set(TRACKING "--tracking")
```

include/revng/TupleTree/Visits.h
```diff
@@ -712,3 +712,12 @@ constexpr bool validateTupleTree(L Check) {
 
   return true;
 }
+
+namespace revng::detail {
```
aleclearmind, line 716: If something it's used outside of the file where it's defined, let's not put it in a `detail` namespace.

include/revng/TupleTree/Tracking.h
```diff
@@ -0,0 +1,186 @@
...
+#include "revng/Support/YAMLTraits.h"
+#include "revng/TupleTree/TupleLikeTraits.h"
+#include "revng/TupleTree/TupleTreeCompatible.h"
+#include "revng/TupleTree/TupleTreePath.h"
+#include "revng/TupleTree/Visits.h"
+
+struct ReadFields {
+  std::set<TupleTreePath> Read;
+  std::set<TupleTreePath> ExactVectors;
+};
```
aleclearmind, line 22: Docs

aleclearmind (changes requested):

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
drblallo: i think i am going to just remove the comment, i copy pasted it from somewhere else and did not noticed it.

aleclearmind:
Merged in dba6f8d. Thanks!

PR #278, opened 2023-03-16
Decouple the sugiyama layouter from the `yield::Graph`

Cvelth:


include/revng/Yield/Support/GraphLayout/Graphs.h
```diff
@@ -0,0 +1,140 @@
...
+struct InputNode : public DataType {
+  Size Size = { 0, 0 };
+
+  using DataType::DataType;
+  InputNode(const DataType &Another) : DataType(Another){};
+  InputNode(DataType &&Another) : DataType(Another){};
+};
+
+template<typename DataType>
+struct OutputNode : public DataType {
```
pfez, line 26: Shouldn't this inherit from `InputNode`? In a sense the `OutputNode` is just the same as the input, but with an assigned "placement" i.e. `Center`.

pfez: Or do you see reasons not to do this?

include/revng/Yield/Support/GraphLayout/Graphs.h
```diff
@@ -0,0 +1,140 @@
...
+
+template<typename DataType>
+struct InputEdge : public DataType {
+  using DataType::DataType;
+  InputEdge(const DataType &Another) : DataType(Another){};
+  InputEdge(DataType &&Another) : DataType(Another){};
+};
+
+template<typename DataType>
+struct OutputEdge : public DataType {
```
pfez, line 43: Same here. Shouldn't this inherit from `InputEdge`?

include/revng/Yield/CallGraphs/Graph.h
```diff
@@ -0,0 +1,30 @@
...
+
+struct Edge {
+  bool IsBackwards = false;
+};
+
+using PreLayoutGraph = layout::InputGraph<Node, Edge>;
+using PreLayoutNode = PreLayoutGraph::Node;
+
+using PostLayoutGraph = layout::OutputGraph<Node, Edge>;
+using PostLayoutNode = PostLayoutGraph::Node;
```
pfez, line 28: Couldn't all these just be the output version and then we do the layout in-place? AFAICT the actual layout engine only works on internal graph, so it could update the place/route info in place.

include/revng/Yield/ControlFlow/Graph.h
```diff
@@ -0,0 +1,31 @@
...
+enum class EdgeType { Unconditional, Call, Taken, Refused };
+struct Edge {
+  EdgeType Type = EdgeType::Unconditional;
+};
+
+using PreLayoutNode = layout::InputNode<Node, Edge>;
+using PreLayoutGraph = layout::InputGraph<Node, Edge>;
+
+using PostLayoutNode = layout::OutputNode<Node, Edge>;
+using PostLayoutGraph = layout::OutputGraph<Node, Edge>;
```
pfez, line 29: Couldn't all these be the output version and then we do the layouting in place? AFAICT the layout engine work on its own internal graph, so this should be feasible.

include/revng/Yield/CallGraphs/Graph.h
pfez, line 28: Couldn't all these be the output version and then we do the layouting in place? AFAICT the layout engine work on its own internal graph, so this should be feasible.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/Compute.h
```diff
@@ -71,8 +72,47 @@ struct Configuration {
...
+/// \param Graph An input graph
+/// \param Configuration An object configuring the specifics of the layout
+///
+/// \return The laid out version of the graph corresponding to \ref Graph
+template<typename Node, typename Edge = Empty>
+inline std::optional<layout::OutputGraph<Node, Edge>>
+compute(const layout::InputGraph<Node, Edge> &Graph,
+        const Configuration &Configuration) {
+  // TODO: rename into `compute` once `llvm::GraphTraits` has a concept support
+  //       for checking whether it's defined for a given type or not.
```
pfez, line 108: This TODO seems outdated

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/Compute.h
```diff
@@ -71,8 +72,47 @@ struct Configuration {
...
+
+  using OutputGraph = layout::OutputGraph<Node, Edge>;
+  std::optional<OutputGraph> Result = layout::detail::convert(Graph);
```
pfez, line 111: Are there any places where we want to do it not in-place? Because if we always want to do it in place I don't see the point of having the not-in-place version.

include/revng/Yield/Support/GraphLayout/Graphs.h
```diff
@@ -0,0 +1,140 @@
...
+    revng_assert(Success);
+  }
+
+  for (const InputNode *From : Input.nodes())
+    for (auto [To, EdgeLabel] : From->successor_edges())
+      Lookup.at(From)->addSuccessor(Lookup.at(To),
+                                    EdgeDataType(std::move(*EdgeLabel)));
+
+  return Result;
+}
```
pfez, line 136: Do we really need conversion? Given that the layout algorithm only works on its internal data structure, and we are already copying the input in there, can't we just update the input in-place at the end of the layout? This would save us an entire copy of the input.

Instead of input -> copy -> internal -> copy -> output we'd have inputoutput -> copy -> internal -> update -> inputoutput.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
```diff
@@ -0,0 +1,234 @@
...
+  layout::Point Center;
+  layout::Size Size;
+
+  bool IsVirtual;
+
+private:
+  std::size_t Index;
+
+private:
+  friend InternalGraph;
```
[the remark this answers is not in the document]
pfez: Why do you need this?

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
```diff
@@ -0,0 +1,234 @@
...
+  InternalNode(std::size_t Index, layout::Size &&Size) :
+    Center({ 0, 0 }), Size(std::move(Size)), IsVirtual(false), Index(Index) {}
+  InternalNode(std::size_t Index) :
+    Center({ 0, 0 }), Size({ 0, 0 }), IsVirtual(true), Index(Index) {}
+
+public:
+  InternalNode(const InternalNode &) = delete;
```
pfez, line 36: I guess this is non-copyable for ensuring we maintain the unique `Index` right?

Please explicitly `delete` also the copy-assignment for https://en.cppreference.com/w/cpp/language/rule_of_three

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
```diff
@@ -0,0 +1,234 @@
...
+  InternalNode(InternalNode &&) = default;
+  InternalNode &operator=(InternalNode &&) = default;
+
+  std::size_t index() const { return Index; }
+};
+
+struct InternalEdge {
+  bool IsRouted;
```
pfez, line 44: Can't this be a method that checks if the `Path` is empty?

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
```diff
@@ -0,0 +1,234 @@
...
+
+public:
+  InternalEdge(const InternalEdge &) = delete;
+  InternalEdge(InternalEdge &&) = default;
+  InternalEdge &operator=(const InternalEdge &) = delete;
+  InternalEdge &operator=(InternalEdge &&) = default;
+
+  std::size_t index() const { return Index; }
+  bool isVirtual() const {
+    return std::get_if<PathOwnership>(&Path) == nullptr;
```
pfez, line 93: `std::holds_alternative` perhaps?

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
```diff
@@ -0,0 +1,234 @@
...
+  std::size_t Index;
+
+  using PathOwnership = std::unique_ptr<layout::Path>;
+  using PathReference = layout::Path *;
+  std::variant<PathOwnership, PathReference> Path;
+
+private:
+  // Only allow `InternalGraph` to construct these objects.
+  // This helps to keep the indexing consistent.
+  friend InternalGraph;
```
pfez, line 57: Why are they all friends? I'm very suspicious about `friend` here. Moving the declarations of `InternalNode` and `InternalEdge` inside `InternalGraph` is better.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
```diff
@@ -0,0 +1,234 @@
...
+                                              InternalLookup.at(Edge.Neighbor));
+        revng_assert(NewEdge->Index == Lookup.Edges.size());
+        Lookup.Edges.emplace_back(Edge);
+      }
+    }
+
+    struct ResultType {
+      InternalGraph Graph;
+      const OutputLookups<NodeRef, EdgeRef> Lookup;
+    };
```
pfez, line 183: Give more dignity to this `struct` definition. It's part of your public API, you can't just hide the definition inside here, in a function whose return type is `auto`. Drop `auto` and make this struct visible outside. Perhaps `InternalGraph` itself should have a `Lookup` field and `ResultType` (or whatever you want to call it) should just go away.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
```diff
@@ -0,0 +1,234 @@
...
+    return ResultType{ .Graph = std::move(Result),
+                       .Lookup = std::move(Lookup) };
+  }
+
+  template<layout::HasLayoutableGraphTraits GraphType,
+           typename NodeR = typename llvm::GraphTraits<GraphType>::NodeRef,
+           typename EdgeR = typename llvm::GraphTraits<GraphType>::EdgeRef>
+  void exportInto(const OutputLookups<NodeR, EdgeR> &Lookup) {
```
pfez, line 191: Can you document this method?

lib/Yield/Support/SugiyamaStyleGraphLayout/GraphPreparation.cpp
```diff
@@ -219,40 +169,6 @@ void partition(const std::vector<EdgeType> &Edges,
...
-    }
-
-    auto EntryPoint = Graph.addNode(nullptr);
-    for (auto *Node : Graph.nodes())
-      if (!Node->hasPredecessors() && Node->Index != EntryPoint->Index)
-        EntryPoint->addSuccessor(Node, nullptr);
-    Graph.setEntryNode(EntryPoint);
-  }
-}
-
```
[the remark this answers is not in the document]
Cvelth: Sure thing.

lib/Yield/Support/GraphLayout/SugiyamaStyle/InternalCompute.h
```diff
@@ -169,6 +173,10 @@ bool computeInternal(InternalGraph &Graph, const Configuration &Configuration) {
   if (Configuration.UseOrthogonalBends)
     Prerouted = routeBackwardsCorners(Graph, Ranks, Lanes, Margin, EdgeGap);
 
+  // Restore the original edge directions.
+  // The graph will no longer be a DAG.
+  restoreEdgeDirections(Graph);
+
   // To simplify routing, order the edges first. This consumes the prerouted
```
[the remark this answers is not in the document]
Cvelth: We used to consume BOTH the graph and the "prerouted edges" data-structure.

We still consume the edges, but not the graph.

So technically, both are correct. Except it might indeed be confusing. How do you think we should reword this?

include/revng/Yield/Support/GraphLayout/Traits.h
```diff
@@ -0,0 +1,79 @@
...
+  //
+  // const Size &getNodeSize(typename LLVMTrait::NodeRef Node);
+
+  // Elements to provide to qualify as an output graph:
+  //
+  // void setNodePosition(typename LLVMTrait::NodeRef Node, Point &&Point);
+  // void setEdgePath(typename LLVMTrait::EdgeRef Edge, Path &&Path);
+
+  // Note that a single graph is allowed to provide everything, so it can
+  // qualify as both (for computing the layout in-place).
```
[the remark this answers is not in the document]
Cvelth: We could (see the later commits for how it's done (actually, you probably saw them by now)). I'll keep this open until we deal with that part of the review.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/Compute.h
Cvelth: Ah, whoops, I forgot I switched the names around. This TODO is for `computeInPlace`.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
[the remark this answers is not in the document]
Cvelth: Because they require an index to be created. And since I was unable to keep them `const` (we require an assignment operator), I softened the restriction to "only `InternalGraph` (the one who keeps track of the used indices) can make these).

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
```diff
@@ -0,0 +1,234 @@
...
+struct InternalEdge {
+  bool IsRouted;
+  bool IsBackwards;
+
+private:
+  std::size_t Index;
+
+  using PathOwnership = std::unique_ptr<layout::Path>;
+  using PathReference = layout::Path *;
+  std::variant<PathOwnership, PathReference> Path;
```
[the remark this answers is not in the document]
Cvelth: These could be two separate field, but we never need both of them at the same time - an edge either owns a path, or links to one owned by another edge.

And since converting these two fields into a `variant` brings barely any complexity, there's not much reason not to do it here.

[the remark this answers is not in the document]
pfez: Ok but what are the two scenarios where an edge owns the path and where it doesn't?
Can you describe all these things in comments that document this data structure?
Otherwise these subtleties are not self-evident for someone looking at the code for the first time.
Take this as a general rule, when you do a non-obvious design decision, anyone reading your code will be left wondering, so it's your job to comment and document your design decisions.

lib/Yield/Support/GraphLayout/SugiyamaStyle/InternalCompute.h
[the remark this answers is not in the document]
pfez: Just use the exact wording you've used here in this answer, an put it in the commit message. In this way even for the non-attentive reader, like I've been in this case, things are immediately obvious.

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
```diff
@@ -243,9 +243,21 @@ RankContainer partitionLongEdges(InternalGraph &Graph,
   // Remove an artificial entry node if it was ever added.
   revng_assert(HasSingleEntryPoint(Graph));
   if (Graph.getEntryNode() != nullptr) {
-    if (Graph.getEntryNode()->IsVirtual) {
-      Ranks.erase(Graph.getEntryNode());
-      Graph.removeNode(Graph.getEntryNode());
+    if (InternalNode *Entry = Graph.getEntryNode(); Entry->IsVirtual) {
+      std::vector<NodeView> ToRemove;
+      llvm::df_iterator_default_set<InternalNode *> Visited;
+      for (InternalNode *Current : llvm::depth_first_ext(Entry, Visited)) {
```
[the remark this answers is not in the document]
pfez: If you want to see examples look in the tests or in revng-c

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
pfez: Ok, strictly speaking you're right. At this point I'd consider one of the two following alternatives:
- if you can drop all the `friends` alltogether, leaving the nest of type declarations as they are now, I think that's the best option. At the moment I haven't tried, but what fails if you drop the `friend`?
- otherwise I think you can figure out something along with my comment on `ResultType` below

What do you think?

Cvelth: Dropping `friend`ship would mean having to make the constructors public.

I guess it's not THAT big of a deal that something named `InternalX` is constructed in non-obvious way, so I'm not totally opposed, but it does feel wrong to let anyone at any point say "make me a new node with this arbitrary index I chose", especially considering how much care went into making sure no two nodes ever get the same index.

include/revng/Yield/Support/GraphLayout/Graphs.h
Cvelth: Now, that you have the context about not wanting the `InputNode` and `OutputNode` to be related from the user's point of view, are you also on the same page that they shouldn't be implicitly convertible, only using a separate method hidden away somewhere?

include/revng/Yield/Support/GraphLayout/Traits.h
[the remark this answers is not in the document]
Cvelth: Closing this as per mattermost discussion.

include/revng/Yield/Support/GraphLayout/Graphs.h
Cvelth: Closing this in favour of [another thread](https://github.com/revng/revng/pull/278#discussion_r1146082077).

include/revng/Yield/CallGraphs/Graph.h
Cvelth: > the actual layout engine only works on internal graph

It only works on the output graph, but this trickery is more to allow users not having to deal with output graphs until the layouter is done, as per mattermost discussion. Closing this.

include/revng/Yield/ControlFlow/Graph.h
Cvelth: This looks like a duplicate of [another thread](https://github.com/revng/revng/pull/278#discussion_r1146086216), so I'm closing it as well.

include/revng/Yield/CallGraphs/Graph.h
Cvelth: Wait, are these three the same? Is something weird going on with github today?

(closing this one too)

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/Compute.h
Cvelth: Closing this as per mattermost discussion.

include/revng/Yield/Support/GraphLayout/Graphs.h
Cvelth: Closing this as per mattermost discussion.

lib/Yield/Support/SugiyamaStyleGraphLayout/GraphPreparation.cpp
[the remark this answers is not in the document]
Cvelth: Done: https://github.com/revng/revng/pull/278/commits/63e69f59bd18fcaee6b55b258738eb8055861032.

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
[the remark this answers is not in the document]
Cvelth: Done :tada:

Please take a look at `stripArtificialEntryPoint` function.

lib/Yield/Support/SugiyamaStyleGraphLayout/GraphPreparation.cpp
```diff
@@ -60,112 +60,62 @@ static SelfLoopContainer extractSelfLoops(InternalGraph &Graph) {
...
+    // BUT if the currently set entry node is already virtual, remove it first:
+    // this prevents the possibility of chaining virtual entry nodes when this
+    // function is invoked on a slightly-modified graph multiple times.
+    if (Graph.getEntryNode() != nullptr) {
+      if (Graph.getEntryNode()->isVirtual()) {
+        Graph.removeNode(Graph.getEntryNode());
+        if (MaybeRanks != nullptr)
+          MaybeRanks->erase(Graph.getEntryNode());
       }
     }
```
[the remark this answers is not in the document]
Cvelth: Cleaned this mess up, please take a fresh look at the `Sugiyama: rework artifical entry computation` commit.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/Compute.h
Cvelth: Done.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
```diff
@@ -0,0 +1,234 @@
...
+  }
+  Node::Edge *makeEdge(Node *From, Node *To) {
+    return From->addSuccessor(To, Node::Edge(EdgeIndexCounter++)).Label;
+  }
+  Node::Edge makeVirtualEdge(Node::Edge &Edge, bool IsBackwards) {
+    return Node::Edge(EdgeIndexCounter++, Edge, IsBackwards);
+  }
+  Node::Edge makeVirtualEdge(Node::Edge &Edge) {
+    return makeVirtualEdge(Edge, Edge.IsBackwards);
+  }
```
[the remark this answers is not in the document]
Cvelth: Dropped.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
Cvelth: Done.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
Cvelth: Ah, wait, I'm now seeing why I went with such an interior approach: since this struct includes the `InternalGraph` object as a member, it cannot be declared before the `InternalGraph` is.

Either way, please take a look at it now (`InternalGraphWithLookupTable`).

Cvelth: > Perhaps `InternalGraph` itself should have a `Lookup` field

The entire point of this PR is to move the information `Lookup` represents outside of the `InternalGraph`. This way `InternalGraph` has no relation to whatever form the data is represented in (as in, it doesn't need to be a template - because `Lookup` is), and the graph itself can just refer to specific nodes by their indices in the `Lookup`.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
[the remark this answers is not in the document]
Cvelth: What do you think?
```cpp
/// Represents an edge in the internal graph.
///
/// Because the layouter heavily relies on splitting edges into pieces
/// (partitioning them, see `GraphPreparation.cpp` for more details), we need
/// a way to let multiple paths accumulate into a single path that can
/// eventually be exported for the original edge.
///
/// To achieve this, each edge holds EITHER a real path (`PathOwner`) OR just
/// a reference to a path held by some other edge (`PathReference`).
///
/// To limit the opportunities for breaking stuff by using this non-intuitive
/// system incorrectly, access to the path is heavily limited to just two
/// methods:
/// - `appendPoint` - adds a point to the current path without regard of whether
///                   the path is owned by this edge, or some other one.
/// - `getPath`     - returns the path reference (which allows it to be moved
///                   out), but is only defined for the edge owners. This
///                   guarantees that each path can only be returned at most
///                   once (the only accessor asserts it).
///
/// There's also a `isVirtual` helper for determining whether the path is owned
/// or not. The expected usage pattern is
///
/// for (auto &Edge : allTheEdgesOfMyGraph())
///   if (Edge.isVirtual())
///     acceptThePath(std::move(Edge.getPath()));
```
(this doxygen is attached to `InternalEdge` class definition).

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
Cvelth: (first see my comment here for additional context: https://github.com/revng/revng/pull/278#discussion_r1149204523)

> What does it even mean to have a not-yet-routed edge with a non-empty path?

Edges can share paths. There is exactly one "`PathOwner`" edge per each real edge in the original graph.

For example, imagine that we have a long edge `A -> B` where rank of `A` is `0`, and rank of `B` is `4`, we need to insert 3 additional nodes in-between, something like
```
A (0) -> A' (1) -> A'' (2) -> A''' (3) -> B (4)
```
Except in the real graph we still only have a single edge for all of these, so only one `LayoutableEdgeTraits::setEdgePath()` call is expected.

So, the approach I chose is to let only one of these edges to own the path (the last one to be specific (`A''' (3) -> B (4)`)), while all the other just refer to it.

By the way the layouter is structured, we are guaranteed that all these edges are going to be routed in the correct order (that's what `orderEdges` (see `EdgeRouting.cpp`) is for), but we still have the problem of making sure we every real edge exactly once. Which is easy to achieve by guaranteeing that only "`PathOwner`"s are exported.

> What happens then when the non-routed-edge-with-non-empty-path gets routed?

By definition, when an edge is routed it simply appends points to the path it's set to. Since we take extra steps to guarantee the proper edge routing order, this let's us route edges without regard to where it's actually stored - by only exposing `appendPoint` which does "the right thing" (tm).

>  

P. S. As for the `IsRouted` flag, we only use to make sure no edge is routed twice, which is significant for "pre-routed" backedges (which we have to route separately because to get them right we need to look at multiple edges at once).

Cvelth: P. P. S. I'm guessing I should in fact fit this comment somewhere into the sources as well. What do you think?

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
Cvelth: Done.

lib/Yield/Support/GraphLayout/SugiyamaStyle/InternalCompute.h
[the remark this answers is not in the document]
Cvelth: Done.

lib/Yield/Support/GraphLayout/SugiyamaStyle/NodeRanking.h
```diff
@@ -16,7 +16,7 @@ RankContainer rankNodes(InternalGraph &Graph, int64_t DiamondBound);
 
 /// Updates node ranking after the graph was modified.
 ///
-/// Garanties ranking consistency i.e. that each node has
+/// Guaranties ranking consistency i.e. that each node has
```
pfez, line 19: Guarantees

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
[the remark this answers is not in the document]
pfez: Great!

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
pfez: Perfect!

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
```diff
@@ -47,32 +48,63 @@ static SelfLoopContainer extractSelfLoops(InternalGraph &Graph) {
...
-          MaybeRanks->erase(Graph.getEntryNode());
+    InternalNode *EntryNode = Graph.getEntryNode();
+    if (EntryNode != nullptr) {
+      // If there is already an entry node present for the graph,
+      if (EntryNode->IsVirtual) {
+        // and it's virtual, reuse it,
+      } else {
+        // otherwise, ignore it.
+        EntryNode = nullptr;
       }
```
pfez: I don't this can ever happen. If it could happen we'd have a graph such that:
- `entryPoints(&Graph)` returns many entry points
- `Graph.getEntryNode` returns a single entry point that is not virtual

Intuitively, if this happens there must be something very wrong, don't you think? Have you seen cases where it happens in practice? If so, can you share them? Otherwise I'd just assert that whenever you have many `EntryPoints` then `EntryNode == nullptr or EntryNode->IsVirtual`.

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
```diff
@@ -47,32 +48,63 @@ static SelfLoopContainer extractSelfLoops(InternalGraph &Graph) {
...
-    for (InternalNode *Node : Graph.nodes())
-      if (!Node->hasPredecessors() && Node->index() != EntryPoint->index())
-        Graph.makeEdge(EntryPoint, Node);
-    Graph.setEntryNode(EntryPoint);
+    if (EntryNode == nullptr) {
+      // If there's no node to reuse, make a new one.
+      EntryNode = Graph.makeVirtualNode();
+    }
+
+    // Add all the entry points entry node is not aware off as successors.
```
pfez, line 76: `s/off/of`

lib/Yield/Support/SugiyamaStyleGraphLayout/GraphPreparation.cpp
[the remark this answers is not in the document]
pfez: Ok, this is much better now. I left a couple of comments on the new code, but I'm closing this.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
pfez: Yeah no biggie in making `public` the constructors for the `InternalX`. It's not that they're are really public.

If you want to add some safeguards against misuse I think you can move the associated headers away from `include` and move them in `lib/Support/GraphLayout/SugiyamaStyle` (or something closer to that point). This effectively makes them "public" but only in that part of the code where it makes sense and where you really need to know what you're doing to use them. Or is there something preventing you to do that?

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
```diff
@@ -0,0 +1,290 @@
...
+/// graph and the lookup table necessary for the results of the layouter to be
+/// merged back into the original graph.
+template<typename NodeRef, typename EdgeRef>
+struct InternalGraphWithLookupTable;
+
+} // namespace detail
+
+class InternalGraph : public GenericGraph<InternalNode, 16, true> {
+public:
+  using GenericGraph<InternalNode, 16, true>::GenericGraph;
```
pfez, line 164: About the `ResultType` comments here's what I propose:
- add here `using GraphWithLookupTable = detail::InternalGraphWithLookupTable<NodeRef, EdgeRef>;`
- in the prototype of `make` drop the `auto` return type and use `GraphWithLookupTable` explicitly. It makes it much more readable
- in the body of `make`,  at the end, you can drop the `using` and the `ResultType`, and just use `GraphWithLookupTable`.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
pfez: See my other new comment of `ResultType`. If you agree with that and do it, you can then close both conversations.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
pfez: Ok, this now makes sense, but I have to say it seems more unintuitive and convoluted than necessary. Is this also the only case why you have the `variant` in the `Path`?

Overall I think that if we can remove this asymmetry and just let all edges own their path (allowing duplicated paths of course for all the edges that "share" the path in the current implementation), everything becomes clearer and more intuitive.

The only thing that would be left over would be the asymmetry in how you apply these things back to the output graph, but that's an intrinsic asymmetry we'll never get rid of (there's a many-to-one mapping from internal graph to output graph, we can't do anything about it).

> By definition, when an edge is routed it simply appends points to the path it's set to. Since we take extra steps to guarantee the proper edge routing order, this let's us route edges without regard to where it's actually stored - by only exposing appendPoint which does "the right thing" (tm).

:thinking: this makes me think that the routing of multi-node-edges happens iteratively and not all at once. Is that the case? Would that prevent or complicate the implementation of what I just proposed above in this message?

> P. S. As for the IsRouted flag, we only use to make sure no edge is routed twice, which is significant for "pre-routed" backedges (which we have to route separately because to get them right we need to look at multiple edges at once).

I see, but this can also be done in my proposal by just pre-filling the `Path` for pre-routed edges. So that if a edge has non-empty path it means that is routed already (possibly pre-routed) and we don't route it again.

> P. P. S. I'm guessing I should in fact fit this comment somewhere into the sources as well. What do you think? Any ideas for an optimal location?

Yes, all this information needs to be put in some comments. Let's just finalize the design, and then we can come up with where's the best place to document this.

lib/Yield/Support/GraphLayout/SugiyamaStyle/NodeRanking.h
pfez: I think this was lost across force-pushes. Please double check

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
Cvelth: Since we split back-edges into `V`-shapes, they add new entry points. Hence the trickery and the need to "fixup" the entry point.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
Cvelth: Yep, the `make` and `exportInto` methods of the `InternalGraph` have to be templates. And they have to be able to access the class they create and dismantle.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
pfez: Anyway, independently of how this discussion turns out, the proper place to document this is on the `variant`. Unless we manage to get rid of the `variant`, but at that point also the `IsRouted` thing and all this discussion is outdated. Which BTW seems like a good reason to get rid of the `variant` if we can, because all this non-obvious information becomes irrelevant.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
Cvelth: 2. I'd gladly, BUT the `InternalGraphWithLookupTable` requires non-trivial template arguments that are deduced from `make`'s template parameters, so that non-`auto` return type would have to look something like `detail::InternalGraphWithLookupTable<typename llvm::GraphTraits<GraphType>::NodeRef, typename llvm::GraphTraits<GraphType>::EdgeRef>` with no place to alias those names (unless we do it inside the `template`, which I'm always opposed - it opens the possibility for somebody to explicitly override the aliases, which leads to FUD, which leads to `requires` clauses double-checking that, which mean that four letter return type (`auto`) gets bloated to four dense lines of templates.

lib/Yield/Support/GraphLayout/SugiyamaStyle/NodeRanking.h
Cvelth: Re-cherry-picked that commit.

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
Cvelth: Done.

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
pfez: Ok then, please document this.

pfez: Or better, can you assert that this happens only for the nasty V-shaped ones?

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
pfez: Uff. Ok. But please at least change `ResultType` to `GraphWithLookupTable`. It's much more meaningful.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
Cvelth: > Yeah no biggie in making `public` the constructors for the `InternalX`. It's not that they're are really public.

Actually it's a lot bigger than it seems. Because when they are `public`, they are constructible using the "natural" means, as in `InternalGraph.addNode()`. Which complains about not specifying index, BUT, if someone were to call `InternalGraph.addNode(0)` for some reason, that would probably be quite an issue to debug.

(for reference, the "canonical" ways to create nodes are `InternalGraph.makeNode()` and `InternalGraph.makeVirtualNode()`).

I guess it's not really sustainable as is - so we should probably leave this as a comment somewhere and/or also rename `makeX` things into something more natural.

There's also an option of preventing the constructor from being called using a `tag`-type in the constructor. But at that point, I'd rather just declare friendship and be done with it.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
Cvelth: Sadly that wouldn't work, see the response to the other comment.

P. S. Since this "- see the other comment", "- no, YOU see the other comment" approach kind of sucks, I'll close this thread and let the other one track this issue.

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
Cvelth: Actually, that's a really good idea!

Cvelth: But wait, now that I actually think about it, what is somebody passes a graph with the entry point already set into the layouter, except it's not the only entry point. That case would also trigger this branch. Should we also assert that the entry point is not set when we get the hold of the graph? Or maybe quietly override it?

Cvelth: The more I think about it, the more I'm starting to feel like this should be left as is.

Consider the broad picture:
```
1. find all the entry nodes.
2. if there's only one, set it as "official" and get out.
3. otherwise, check if we already have an "official" entry point set.
5. if there is one, and it's marked as "artificial", just update it.
6. if there is one, but it's not marked as "artificial", ignore it and make a new one.
7. if there's none - also make a new one.
8. fill the successors of the new node and get out.
```

which implicitly makes sense to me - we make as few assumptions about the graph as we can, while also making sure the algorithm still works correctly when it's called multiple times on a graph that was modified in-between (mainly, new entry points appeared).

Is it that advantageous to make sure the newly added nodes are always V-shaped? So much so as to compromise how general the check is?

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
Cvelth: I'm closing this as per the mattermost discussion.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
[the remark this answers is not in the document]
Cvelth: I'm closing this as per the mattermost discussion.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
Cvelth: Renamed it.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
pfez: How about having the constructor of the node take a reference to the `InternalGraph` you want to insert it into?

Then you can have:
- have a public method on `InternalGraph` that gives you the next available ID and increments the current ID: `getAndIncrementID() { return NodeIndexCounter++; }`
- All the constructors of `InternalNode` have to take an additional `InternalGraph &` argument. Something along this line: `InternalNode(InternalGraph &G) : ID(G.getAndIncrementID) { /* do stuff if necessary */ }` (I haven't specified `Size` explicitly, but you get the idea). In this way you cannot construct a node if you're not holding a graph, which makes it pretty obvious that the node should be part of that graph.
- The constructors for `InternalNode` that explicitly take the ID as parameter have to go away, so that nobody from outside can mess up the invariant that indices are unique.
- `InternalGraph::makeNode()` don't call the node constructors specifying the Index (since they are gone) but they use `InternalNode(*this)` (where `this` points to the `InternalGraph` you are inserting them to). What's nice is that this can now be done without them being `friend`s.
- So everything else is left unchanged, except that now someone can construct a `InternalNode` in the wild, as long as it has a `InternalGraph &` to provide to the constructor. This doesn't really strike me like something that can be heavily misused, and even if it was misused it would be very suspicious both for who writes it and for who reviews it.

Then something similar can be done for edges, and nobody is `friend` of anyone else anymore.

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
pfez: Not sure. I'm pretty concerned about 2. and 6.

In both cases you might be getting a graph where the "official" entry point has been set from outside but you ignore it. This is very very weird from a user's standpoint. Imagine if the user is trying to set the "official" entry point from outside, so that the routing starts from there, but we actually ignore it and route the way we like it considering another entry point. As a user I would be pissed off.

I'd be much happier to assert that we never get a graph from the user that has an "official" entry point already set.

Then if it happens we can detect it and decide what to do.

But in principle, until I don't see such a case, I'd like to rule this out and make this as part of the "contract" of the API: you just don't send in a graph with an "official" entry point already set, or the implementation will assert.

Or you're saying that you already have in mind a situation where we might wanna do exactly that?

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
Cvelth: Why haven't I thought of this! Yeah, this is definitely the way to go.

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
pfez: It's not that your proposal can't work, but why bothering if we don't even envision users to provide us an entry point that is decided by them?

Let's just decide that the layouter API has a contract: the place and route algorithm works **only** with an entry point that **we** decide. If someone tries to be a smartass and sets the entry point from outside, they will just get an assertion in their face and learn to respect the API's contract.

I don't want to be stuck supporting a half-baked counterintuitive scenario unless we find at least a real use case for which this is a real game changer. Do you have one of these uses cases in mind? Or you're just speculating?

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
Cvelth: Please take a fresh look. I'm really happy about how this turned out! Thanks to easier to access constructor, I was also able to get rid of `makeNode` and `makeEdge` outright! I still decided to keep the "virtual" counterparts though - I really like how explicit
```
auto *Node = Graph.makeVirtualNode();
// as opposed to
auto *Node = Graph.addNode(Graph);
```

lib/Yield/Support/GraphLayout/SugiyamaStyle/NodeRanking.cpp
```diff
@@ -133,11 +130,12 @@ static bool areDisjoint(NodeView A, NodeView B, size_t MaxDepth = -1) {
...
-RankContainer rankNodes<DDFS>(InternalGraph &Graph, int64_t DiamondBound) {
+RankContainer
+rankNodes<DDFS>(InternalNode *GraphEntryPoint, int64_t DiamondBound) {
   RankContainer Ranks;
 
   std::vector<NodeView> Stack;
-  for (auto *Node : Graph.nodes()) {
+  for (auto *Node : llvm::depth_first(GraphEntryPoint)) {
     if (!Node->hasPredecessors() && !Ranks.contains(Node)) {
       Stack.emplace_back(Node);
```
pfez, line 140: This is weird. Now you're doing a DFS but you only do stuff when `!Node->hasPredecessors()`. So in my understanding you'd be only doing stuff when `Node` is `GraphEntryPoint`. At this point why doing the DFS in the first place?

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
```diff
@@ -42,39 +43,66 @@ static SelfLoopContainer extractSelfLoops(InternalGraph &Graph) {
...
-      if (Graph.getEntryNode()->IsVirtual) {
-        Graph.removeNode(Graph.getEntryNode());
-        if (MaybeRanks != nullptr)
-          MaybeRanks->erase(Graph.getEntryNode());
+/// Adds a virtual node to the graph to serve as an entry point, while
+/// also cleanly removing it when this object is being destroyed.
+class VirtualEntryPointGuard {
+private:
+  InternalGraph &Graph;
+  InternalNode &Node;
```
pfez, line 51: So this is the virtual entry point right? Can you call it `VirtualEntryPoint` please?

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
```diff
@@ -42,39 +43,66 @@ static SelfLoopContainer extractSelfLoops(InternalGraph &Graph) {
...
+    };
+    using FilteredGraphType = NodePairFilteredGraph<InternalNode *, Filter>;
+
+    std::vector<NodeView> ToRemove;
+    for (InternalNode *Current : llvm::depth_first(FilteredGraphType(&Node)))
+      ToRemove.emplace_back(Current);
+
+    for (NodeView Node : ToRemove) {
+      if (MaybeRanks != nullptr)
+        MaybeRanks->erase(Node);
```
pfez, line 101: Then some ranks might become empty. Might this be a problem?

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
```diff
@@ -213,15 +220,15 @@ RankContainer partitionLongEdges(InternalGraph &Graph,
...
       } else {
         ++Iterator;
       }
     }
   }
 
   // Calculate real ranks for the remainder of the graph.
-  ensureSingleEntry(Graph, &Ranks);
-  revng_assert(HasSingleEntryPoint(Graph));
   Ranks = rankNodes<Strategy>(Graph.getEntryNode());
```
pfez, line 232: Here and later in this function you call `Graph.getEntryNode`. Now that you have the `GraphEntry` guard, wouldn't it make more sense to use that instead of calling `Graph.getEntryNode()`?

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
```diff
@@ -213,15 +220,15 @@ RankContainer partitionLongEdges(InternalGraph &Graph,
         revng_assert(Edge != nullptr);
         SavedLongEdges.emplace_back(From, To, std::move(*Edge));
         Iterator = From->removeSuccessor(Iterator);
+        if (!To->hasPredecessors())
+          GraphEntry->addSuccessor(To, InternalNode::Edge(Graph));
```
pfez, line 224: So here we've removed a long edge that was going into `To`, and `To` might become orphan. If it does, it becomes unreachable from the entry point, so we're adding a placeholder edge from the virtual entry point to `To`, so that it remains reachable.

Intuitively it makes sense, but I have a question: at the end of `partitionLongEdges` the virtual entry guard goes out of scope, so it deletes the virtual entry point, along with all the outgoing edges from it. At that point `To` would become orphan again. Can't this cause problems down the line?

lib/Yield/Support/GraphLayout/SugiyamaStyle/NodeRanking.cpp
Cvelth: Ah, I didn't really think about it. My thought process was:
- this just wants a list of nodes, while all the other ranking algorithms just want an entry.
- but wait! There's an easy way to get list of nodes from the entry. And it's not like it cares about the order they are in.

To summarize, does just pulling this preparation step outside the loop make sense to you?

Cvelth: TL;DR: this ranking algorithm probably wants to be redesigned, to be consistent with the rest, but I'd guess today is not the day.

include/revng/Yield/Support/GraphLayout/SugiyamaStyle/InternalGraph.h
pfez: Nice! No `friend`s anymore. And it brought improvements also in adjacent parts of the codebase!

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
Cvelth: As per the mattermost discussion, I'm closing this thread.

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
Cvelth: But, said that, to answer the original question, when `Ranks` get converted into the `Layers`, there's an extensive clean-up to ensure the "problem" you are worrying about here can never happen (again).

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
Cvelth: Whoops, I didn't pay attention to this, I guess. But also consider that we still have to use `setEntryNode` within the `GraphEntry` no matter what. `getEntryPoints` from `GraphAlgorithms.h` uses graph-level iteration - i. e. `GenericGraph`'s `llvm::GraphTraits`, which relies on the `getEntryNode`. As such, there's no avoiding it one way or the other.

Cvelth: Note that the guard destructor even does `revng_assert(Graph.getEntryNode() == &VirtualEntry)`.

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
Cvelth: Every top-corner-style V-shape is an orphan by definition. So it is just going to be one of many.

This is the entire point of going with the guard approach - we only care about a clean entry point at limited points in the algorithm (and only because LLVM does), so if we just wrap these limited points in the guard - everyone is happy - and it's easy to see where we are allowed to rely on that (because there's a guard in the scope), and where we can't.

Am I missing something?

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
Cvelth: I went with `VirtualEntry` (80 character columns are brutal).

Closing this, feel free to reopen if you don't like it.

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
Cvelth: Either way, I replaced them.

lib/Yield/Support/GraphLayout/SugiyamaStyle/GraphPreparation.cpp
pfez: Ok cool thanks.

lib/Yield/Support/GraphLayout/SugiyamaStyle/NodeRanking.cpp
pfez: > this ranking algorithm probably wants to be redesigned, to be consistent with the rest, but I'd guess today is not the day.

That's definitely true, but I'd not focus on it in this PR.

PR #280, opened 2023-03-21
Start testing 32-bit microsoft ABIs

Cvelth:


include/revng/ABI/Definition.h
```diff
@@ -379,7 +379,15 @@ class Definition : public generated::Definition {
...
 
     llvm::SmallVector<model::Register::Values, 8> Result;
-    for (auto Register : GeneralPurposeArgumentRegisters())
+
+    const auto &GPRs = GeneralPurposeArgumentRegisters();
+    const model::Register::Values &RVL = ReturnValueLocationRegister();
+    constexpr model::Register::Values Invalid = model::Register::Invalid;
+    bool NoRVLInGPRs = RVL != Invalid && !llvm::is_contained(GPRs, RVL);
+    if (NoRVLInGPRs && Lookup.contains(RVL))
+      Result.emplace_back(RVL);
```
aleclearmind, line 388: I don't understand what's going on here. What does the return value location has to do with arguments?

share/revng/abi/Microsoft_x86_cdecl.yml
```diff
@@ -33,7 +33,9 @@ ScalarTypes:
   - Size: 2
   - Size: 4
   - Size: 8
+    AlignedAt: 4
```
aleclearmind, line 36: Is this by deduction or was it properly documented?

share/revng/test/configuration/revng/test-abi.yml
```diff
@@ -31,6 +36,11 @@ commands:
       - [i386-regparm-3-systemv-abi]
       - [i386-regparm-2-systemv-abi]
       - [i386-regparm-1-systemv-abi]
+      - [i386-cdecl-microsoft-abi]
+      - [i386-fastcall-microsoft-abi]
+      - [i386-stdcall-microsoft-abi]
+      - [i386-thiscall-microsoft-abi]
+      - [i386-vectorcall-microsoft-abi]
       - [arm-aapcs-abi]
```
aleclearmind, line 44: `repeat-for` is for definition of `sources` only. Drop it.
Same above.

We'd need a YAML schema.

aleclearmind (changes requested):

aleclearmind:
Nice!

include/revng/ABI/Definition.h
Cvelth: Return value location is where the "first shadow argument" is placed. Some ABIs (i.e. AArch64) dedicate a register specifically for it (x8 in that case). When going `cabi->raw->cabi` we don't want to lose the fact that the register in question is argument-like.

share/revng/abi/Microsoft_x86_cdecl.yml
Cvelth: The documentation for these is not as well structured as I'd hope. So this came from observation first, but I did manage to find a confirmation of the fact that they are aligned at most at register-sized boundaries after the fact. Do you want me to look for it again?

share/revng/test/configuration/revng/test-abi.yml
Cvelth: What do you mean? I always assumed this was the exact use case `repeat-for` was introduced for 🤔.

include/revng/ABI/Definition.h
aleclearmind: OK, in the snipped of code there's not mention of this.
Mention somewhere `ShadowPointerToAggregateReturnValue`, which is our conventional name for this thing.

share/revng/abi/Microsoft_x86_cdecl.yml
aleclearmind: No, just curiosity.

share/revng/test/configuration/revng/test-abi.yml
aleclearmind: No, it's just for `sources`.
`repeat-for` does exactly nothing here (and you don't need it).

Cvelth: I'll try, if it does indeed changes nothing, then my bad on misunderstanding how it works.

include/revng/ABI/Definition.h
Cvelth: Added this comment near the snippet:
```cpp
    // Some ABIs (for example, AArch64) designite a special register to be used
    // in the cases where `ShadowPointerToAggregateReturnValue` is necessary.
    // The register behaves in an argument-like fashion, so we want to account
    // for it while sorting argument registers.
```

Feel free to close the thread if it sounds good to you.

share/revng/test/configuration/revng/test-abi.yml
Cvelth: Dropped.

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.cpp
```diff
@@ -11,25 +11,101 @@
...
+                          std::move(Arguments));
+    } else {
+      revng_assert(isReturnValueTest());
+      revng_assert(ReturnValueAddress.size() == 1);
+      revng_assert(ReturnValue.size() == 1);
+      revng_assert(ExpectedReturnValue.size() == 1);
+      return ReturnValueTest(std::move(*this).extractTheState(Architecture),
+                             std::move(ReturnValueAddress[0]),
+                             std::move(ReturnValue[0]),
+                             std::move(ExpectedReturnValue[0]));
```
aleclearmind, line 97: Drop all the `std::move`, they make the code unreadable and obscure.
Drop them all.

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.cpp
```diff
@@ -11,25 +11,101 @@
...
+    if (isArgumentTest()) {
+      revng_assert(!isReturnValueTest());
+      return ArgumentTest(std::move(*this).extractTheState(Architecture),
+                          std::move(Arguments));
+    } else {
+      revng_assert(isReturnValueTest());
+      revng_assert(ReturnValueAddress.size() == 1);
+      revng_assert(ReturnValue.size() == 1);
+      revng_assert(ExpectedReturnValue.size() == 1);
+      return ReturnValueTest(std::move(*this).extractTheState(Architecture),
```
aleclearmind, line 94: Whatever this is, do not do it.

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.cpp
```diff
@@ -11,25 +11,101 @@
...
+namespace abi::runtime_test {
+
+struct RawState {
+  std::vector<RawWord> Registers;
+  std::vector<std::byte> Stack;
+
+  bool empty() const { return Registers.empty() && Stack.empty(); }
+
+  State extract(model::Architecture::Values Architecture) && {
+    SortedVector<Register> ResultRegisters;
```
aleclearmind, line 23: Do not use `SortedVector`. Use `std::set`. `batch_insert` is fragile and confusing and this is not performance critical code.

Do not treat testing code as performance critical code. The readability of this file is seriously compromised by the attempts to make it performant.

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.cpp
```diff
@@ -11,25 +11,101 @@
...
+      I.emplace(std::move(*Iterator), Architecture);
+    }
 
-using namespace abi::artifact;
+    return State(std::move(ResultRegisters), std::move(Stack));
+  }
+};
 
-struct SingleRun {
+struct SingleIteration {
```
aleclearmind, line 49: "Single" is the opposite of "multiple".
This is just an iteration.

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.h
```diff
@@ -4,86 +4,137 @@
...
   fromKey(const model::Register::Values &Key) {
     return { Key };
   }
 };
 
-namespace abi::artifact {
+namespace abi::runtime_test {
+
+struct State {
+  SortedVector<Register> Registers;
```
aleclearmind, line 68: Do not use `SortedVector` in place of a `std::set`. It's not its purpose.

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.h
```diff
@@ -4,86 +4,137 @@
...
-  key(const abi::artifact::ModelRegister &Object) {
+  key(const abi::runtime_test::Register &Object) {
     return Object.Name;
   }
-  static abi::artifact::ModelRegister
+  static abi::runtime_test::Register
   fromKey(const model::Register::Values &Key) {
     return { Key };
   }
 };
```
aleclearmind: We don't really need KOT. Drop.

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.cpp
```diff
@@ -202,14 +272,25 @@ State::Parsed abi::artifact::parse(llvm::StringRef RuntimeArtifact,
...
+    }
+    void operator()(ReturnValueTest &&Test) {
+      Result.ReturnValueTests[Name].Iterations.emplace_back(std::move(Test));
+    }
+  };
+  ResultInsertionHelper Helper{ Result };
+  for (auto &Iteration : Deserialized.Iterations) {
+    Helper.Name = Iteration.Function;
+    std::visit(Helper, std::move(Iteration).extract(Architecture));
   }
```
aleclearmind: This is very convoluted.

Rewrite as

```c++
abi::runtime_test::Parsed Result;

for (Iteration &Iteartion : Deserialized.Iterations) {
  // Inline here all the extract* methods and simply add stuff here to Result
}
```

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.h
```diff
@@ -4,86 +4,137 @@
...
+struct KeyedObjectTraits<abi::runtime_test::Named<Iteration>> {
+  using Type = abi::runtime_test::Named<Iteration>;
+  static llvm::StringRef key(const Type &Object) { return Object.Name; }
+  static Type fromKey(const llvm::StringRef &Key) { return { Key }; }
 };
 
-namespace abi::artifact {
+namespace abi::runtime_test {
 
 struct Parsed {
```
aleclearmind, line 130: 1. Just call this `TestCase` or something. The fact that its the result of parsing is not relevant to its users.
2. This seems like a pretty plain format what prevents us from making *this* the YAML format?

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.h
```diff
@@ -4,86 +4,137 @@
...
-template<>
-struct KeyedObjectTraits<abi::artifact::FunctionArtifact> {
-  static llvm::StringRef key(const abi::artifact::FunctionArtifact &Object) {
-    return Object.Name;
-  }
-  static abi::artifact::FunctionArtifact fromKey(const llvm::StringRef &Key) {
-    return { Key };
-  }
+template<typename Iteration>
+struct KeyedObjectTraits<abi::runtime_test::Named<Iteration>> {
```
aleclearmind, line 122: Do not use these. You don't need them. Use a `std::map`.

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.cpp
```diff
@@ -129,65 +208,56 @@ struct KeyedObjectTraits<VerificationHelper> {
...
 };
 
-static bool verify(const State::Deserialized &Data, bool ShouldAssert) {
+static bool verify(const abi::runtime_test::Deserialized &Data,
+                   bool ShouldAssert) {
```
aleclearmind, line 212: Drop this argument

aleclearmind (changes requested):

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
```diff
@@ -17,52 +17,49 @@ void verifyABI(const TupleTree<model::Binary> &Binary,
                model::ABI::Values ABI);
```
aleclearmind, line 17: Why are you declaring a function in a `.cpp` file?

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
```diff
@@ -230,114 +227,299 @@ void VH::verify(const abi::FunctionType::Layout &FunctionLayout,
...
+  std::size_t ArgumentTestCount = 0, ReturnValueTestCount = 0;
   for (auto &Function : Binary->Functions()) {
-    const std::string &Name = Function.OriginalName();
-    auto CurrentFunction = ParsedArtifact.Functions.find(Name);
-    if (CurrentFunction == ParsedArtifact.Functions.end()) {
-      // Ignore types not present in the artifact.
+    Helper.FunctionName = Function.OriginalName();
+    if (Helper.FunctionName.take_front(5) == "test_")
+      Helper.FunctionName = Helper.FunctionName.drop_front(5);
+    if (auto Test = llvm::find_if(Parsed.ArgumentTests, LookupHelper);
```
aleclearmind, line 491: `count_if`. Review all usages.

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
```diff
@@ -92,7 +89,7 @@ void VH::verify(const abi::FunctionType::Layout &FunctionLayout,
         std::size_t CurrentSize = ExtractedStackBytes.size();
         ExtractedStackBytes.resize(CurrentSize + Argument.Stack->Size);
```
aleclearmind, line 90: Remove all usages of resize. Early optimization.

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
```diff
@@ -17,52 +17,49 @@ void verifyABI(const TupleTree<model::Binary> &Binary,
...
 using VH = VerificationHelper;
-void VH::verify(const abi::FunctionType::Layout &FunctionLayout,
-                const abi::artifact::FunctionArtifact &Artifact) const {
-  for (std::size_t Index = 0; Index < Artifact.Iterations.size(); ++Index) {
-    IterationAccessHelper Helper(Artifact.Iterations[Index], Architecture);
-
-    // Sort the argument registers
+void VH::arguments(const abi::FunctionType::Layout &FunctionLayout,
+                   const ATest &Iterations) const {
+  for (const abi::runtime_test::ArgumentTest &Test : *Iterations) {
```
aleclearmind, line 56: Outline this for and earn two spaces!

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
```diff
@@ -17,52 +17,49 @@ void verifyABI(const TupleTree<model::Binary> &Binary,
...
 
 using VH = VerificationHelper;
-void VH::verify(const abi::FunctionType::Layout &FunctionLayout,
-                const abi::artifact::FunctionArtifact &Artifact) const {
-  for (std::size_t Index = 0; Index < Artifact.Iterations.size(); ++Index) {
-    IterationAccessHelper Helper(Artifact.Iterations[Index], Architecture);
-
-    // Sort the argument registers
+void VH::arguments(const abi::FunctionType::Layout &FunctionLayout,
+                   const ATest &Iterations) const {
```
aleclearmind, line 55: What this function does is intuitively clear but it's hard to check for the details.
In substance, it compares two things. Please rename all variables in something like `ExpectedXYZ` and `ActualXYZ` or `LeftXYZ` and `RightXYZ`.

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
```diff
@@ -230,114 +227,299 @@ void VH::verify(const abi::FunctionType::Layout &FunctionLayout,
...
+    //
+    // Also, considering we do multiple iterations of each test, the chance
+    // of having false positives is extremely low as well.
+    if (MatchingByteCount * 2 < Found.Bytes.size()) {
+      std::string Error = "Return value of '" + FunctionName.str()
+                          + "' was lost: something went *very* wrong.";
+      revng_abort(Error.c_str());
+    }
+
+    // Because we have no way to represent functions that use SPTAR in `rft`
```
aleclearmind, line 282: Wait, why are we testing RFTs at all?

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
```diff
@@ -230,114 +227,299 @@ void VH::verify(const abi::FunctionType::Layout &FunctionLayout,
...
+    //   // invisible 3 bytes here (4 - 1).
+    // };
+    // ```
+    // Which adds up to 3/8 (~37.5%). In the case where maximum alignment is
+    // 16 bytes - this jumps to 15/32 (~46.9%), and so on. Knowing that,
+    // we can assume that, at the very least, _more_ than half of the bytes
+    // will match.
+    //
+    // Also, considering we do multiple iterations of each test, the chance
+    // of having false positives is extremely low as well.
```
aleclearmind, line 275: Reading this I cannot understand much.
Make an explicit, specific example referencing a specific architecture, ABI, return value.

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
```diff
@@ -230,114 +227,299 @@ void VH::verify(const abi::FunctionType::Layout &FunctionLayout,
...
-    }
+  if (ArgumentTestCount != Parsed.ArgumentTests.size()) {
+    std::string Error = std::to_string(ArgumentTestCount)
+                        + " functions from the binary were tested, but "
+                          "artifact contains "
+                        + std::to_string(Parsed.ArgumentTests.size())
+                        + " argument test functions.\n"
+                          "Does the binary match the artifact?";
+    revng_abort(Error.c_str());
   }
```
aleclearmind, line 514: My main suggestions on Mattermost boils down to this: instead of hardcoding all these assertions in the code, we could emit a new file for the *actual* and *expected* data and then diff them.

Using assertions requires you to use GDB to debug stuff, while using diff you can look at the bigger picture and see all the differences altogether.

On the other hand 1) I'm not really excited at the idea of introducing yet another file format for testing purpose, 2) I really want to get this done.

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
aleclearmind, line 1: At the top of the file, explain why we need this.
Explain that we cannot do a 1-to-1 comparison, we need to *consume* stuff, skip padding and whatnot.

aleclearmind (changes requested):

share/revng/test/configuration/revng/runtime.yml
```diff
@@ -12,6 +12,22 @@ commands:
...
           qemu-$QEMU_NAME $INPUT $$ARGUMENTS > $OUTPUT/$$NAME.stdout || true;
       done
+  - type: revng.abi-test-run
+    from:
+      - type: revng-qa.compiled-with-debug-info
+        filter: for-abi-runtime and (x86-64 or i386)
+      - type: revng-qa.compiled-with-debug-info-copy
+        filter: x86-64 or i386
+    suffix: /
+    command: $INPUT1 $INPUT2 > $OUTPUT.stdout
```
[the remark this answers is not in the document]
Cvelth: `stdout` is just an extension. Ah, wait, there's `suffix` right there! My bad.

share/revng/test/configuration/revng/runtime.yml
```diff
@@ -12,6 +12,22 @@ commands:
...
+  - type: revng.abi-test-run
+    from:
+      - type: revng-qa.compiled-with-debug-info
+        filter: for-abi-runtime and !x86-64 and !i386
+      - type: revng-qa.compiled-with-debug-info-copy
+        filter: "!x86-64 and !i386"
+    suffix: /
+    command: qemu-$QEMU_NAME $INPUT1 $INPUT2 > $OUTPUT.stdout
```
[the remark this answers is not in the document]
Cvelth: What do you mean about QEMU? We need two things because there are two binaries - one (`$INPUT1`) `mmap`s the other (`$INPUT2`).

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.h
Cvelth: 1. Sounds good.
2. The need to redesign it (I _really_ didn't want to add even more). In the early iterations we were _very_ limited as to what we can do when emitting stuff. Now, thanks to the interrupt handlers, the possibilities are limitless.

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
Cvelth: To not create a header specifically for it (since this prototype is the only thing that needs sharing). Not a big deal either way - your call.

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
Cvelth: I'd argue that it's helpful semantically to say: "hey, this next vector is pretty much an array with X elements), but I don't feel like we need to have that conversation today. Will remove.

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
Cvelth: To make sure they are not broken: it adds very little complexity (since everything is based on a layout anyway) but gives me a lot of piece of mind.

On top of that, it also serves as the test that layout works well for RFTs too.

P. S. This comment might be slightly misleading, it doesn't only affects RFTs: it's exactly the same for CFTs after they have been converted down, I was just trying to stress that "we need this extra logic because CFT->RFT loses information".

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
Cvelth: To be fair, I found that the use of GDB really reduced after I made all of these messages really explicit: In ~95% of the case looking at the error message + `yaml` dumped by the runner is enough to pinpoint the exact issue.

The only thing I'd also add to cover ~4.5% more is to add a logger here dumping function types and layouts everywhere, but maybe not today. I'll probably just do at the first instance when I find that I want to `GDB` into this.

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
Cvelth: Ah, wait, I was wrong. This is not an optimization: it's used for shrinking here!

Do you want an assert kind of like
```cpp
revng_assert(CurrentSize + Argument.Stack->Size > ExtractedStackBytes.size());
```
to bring that point home?

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
Cvelth: Wait, what do you mean? Something like
```diff
+revng_assert(llvm::count_if(Parsed.ArgumentTests, LookupHelper) <= 1);
if (auto Test = llvm::find_if(Parsed.ArgumentTests, LookupHelper);
    Test != Parsed.ArgumentTests.end()) {
```
?

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
Cvelth: Sorry for the wait. While at it I also split it into pieces. Should be a breeze to review it now: please take a look.

share/revng/test/configuration/revng/runtime.yml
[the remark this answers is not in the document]
Cvelth: Done.

share/revng/test/configuration/revng/runtime.yml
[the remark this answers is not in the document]
Cvelth: Done.

lib/ABI/FunctionType/Layout.cpp
```diff
@@ -118,7 +118,7 @@ class ToRawConverter {
                 bool AllowPuttingPartOfAnArgumentOnStack) const;
 
 public:
-  std::uint64_t finalStackOffset(const DistributedArguments &Arguments) const;
+  std::uint64_t finalStackOffset(std::uint64_t SizeOfArgumentsOnStack) const;
```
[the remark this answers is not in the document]
Cvelth: Done (see the first commit).

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
Cvelth: Done.

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
Cvelth: Because of needing to debug error I introduced today while shuffling `arguments` around, I needed to debug it 😅. So, now there's a `fail` function that dumps tested function name and layout on top of the `revng_abort` message. Please take a look.

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
Cvelth: I feel the newly cleaned up `arguments` function kind of covers this. Please take a look. Do you feel like I should repeat the same thing on the top of the file too?

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
Cvelth: How does something like this sound?
```cpp
  // Sadly, we cannot do a full comparison since return value structs often
  // contain padding, value of which is not guaranteed to be preserved,
  // causing the test to fail. We either need a way to tell if a specific byte
  // belongs to the padding or not (which would restrict stuff we can test,
  // for example, nested structs) or assume that non-perfect matches are
  // acceptable.
  // An example of a struct that exhibits this behaviour (i386):
  // struct ReturnValue {
  //   uint8_t b; // _Alignof(uint8_t) == 1
  //   // 3 invisible bytes here
  //   // (_Alignof(ReturnValue) - _Alignof(uint8_t) = 4 - 1 = 3)
  //   uint32_t a; // _Alignof(uint32_t) == 4
  // };
  //
  // So, when looking at the return value we detect, it looks something like
  // `[ 0x11, 0xXX, 0xXX, 0xXX, 0x22, 0x33, 0x44, 0x55 ]`
  // where `0xXX` is unuinitialized value that is not stable and changes from
  // execution to execution.
  //
  // But, luckily, by the definition of how such padding is introduced, we are
  // guaranteed that in the worst case it's going to occupy `half_the_size - 1`
  // bytes. So, we can at least enforce that. And, thanks to doing multiple
  // iterations for each tests, the chance of false positives is extremely low.
  ```

File: tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.h (hunk 11 above)
Comment (aleclearmind) [reply tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.h:130]: 2. OK, leave a TODO.

File: tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp (hunk 14 above)
Comment (aleclearmind) [reply tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp:17]: So you're declaring it twice?
Since we never do this, I'd push for uniformity and create a header.

File: tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp (hunk 15 above)
Comment (aleclearmind) [reply tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp:467]: `find_if(...) != end` is the same as `count_if(...) != 0`. Prefer the second syntax.

File: tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp (hunk 16 above)
Comment (aleclearmind) [reply tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp]: What's wrong with `back_inserter`?
Please, do not try to make the code performant unless it really matters.

File: tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp (hunk 15 above)
Comment (Cvelth) [reply tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp:467]: The `Test` variable is used inside the body of the loop. It's doing the search for a reason, not to just check if it exists.

P. S. consider that after dropping `SortedVector` this becomes `std::map::find`.

File: tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp (hunk 14 above)
Comment (Cvelth) [reply tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp:17]: It's declared in a different file.

👌 for the header.

File: tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp (hunk 19 above)
Comment (aleclearmind) [reply tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp]: 1. I do not understand the workflow for testing RFTs and I do not understand if we have any tests that actually does that.
Please, this is already complex enough, I really see no point in testing RFTs.
2. The purpose of this tool is to ensure that a `Layout` create from CFT correctly predicts where arguments/return values will be, correct?
3. If 2) is correct, what's the `CFG -> RFT` conversion matters at all?

Please remove all the complexity that's not strictly necessary for the goals of this tool.

Comment (Cvelth) [reply tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp]: 2. The original purpose of this tool is to make sure that no matter what weirdness happens inside either conversion, the prototype is still valid - as in it doesn't contradict the stuff we got by ~~ab~~using the compiler.

Since then it also shifted somewhat to also serve as the test for the `abi::FunctionType::Layout`  infrastructure too. Consider that we use Layouts all over the place, having them untested doesn't sit right with me.

But that's beside a point. Here's the only place where such information loss is even relevant. And it's fixed by a really simple workaround, which is what I was trying to say by the comment above (I guess using the term "RFT" there was a mistake). Here's a copy of literally all the "extra" complexity introduced for this:
```cpp
  bool UsesSPTAR = FunctionLayout.returnsAggregateType();
  if (!UsesSPTAR) {
    bool SingleArgument = FunctionLayout.Arguments.size() == 1;
    bool SingleReturnValue = FunctionLayout.ReturnValues.size() == 1;
    if (SingleArgument && SingleReturnValue) {
      const auto &ArgumentType = FunctionLayout.Arguments[0].Type;
      const auto &ReturnValueType = FunctionLayout.ReturnValues[0].Type;
      bool TypesMatch = ArgumentType == ReturnValueType;
      bool ReturnValueIsAPointer = ReturnValueType.isPointer();

      // Case for a register SPTAR - types must match.
      if (TypesMatch && ReturnValueIsAPointer)
        UsesSPTAR = true;

      // Case for a stack SPTAR - allow any type as the argument as long as
      // it's pointer-sized.
      if (*ArgumentType.size() == PointerSize && ReturnValueIsAPointer)
        UsesSPTAR = true;
    }
  }
```
It's a simple check that if the first argument is a pointer, and so is the return value, and that their types match.

Not to say, it's also perfectly safe, since tests currently only _either_ have arguments or return values. So having both - is a really good sign that it's a `SPTAR`-function.

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.cpp
Cvelth: I simplified the data structures here a lot. Please take a fresh look when you have a moment.

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.cpp
Cvelth: The `rvalue` members are gone.

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.cpp
Cvelth: `SortedVector` is gone.

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.cpp
Cvelth: Done.

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.h
Cvelth: Done.

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.cpp
Cvelth: What do you think about it now?

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.h
Cvelth: 1. I went with `TestedFunctions` since `Case` sounds weird when the data structure contains _all_ the tests.

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.h
Cvelth: Done.

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
aleclearmind: If you want, you can make a helper function to put in some public header that can be reused. Otherwise, let's go with back inserter.

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.h
Cvelth: Done.

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
Cvelth: Done.

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
Cvelth: Done.

aleclearmind (changes requested):

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
Cvelth: It looks like you have already taken a look at this (since you closed a related thread). Should this also be closed?

share/revng/test/configuration/revng/runtime.yml
```diff
@@ -12,6 +12,21 @@ commands:
       ( grep RUN $SOURCE || true ) | sed 's|/\* RUN-\(.*\): \(.*\) \*/|\1,\2|' | while IFS=',' read -r NAME ARGUMENTS; do
           qemu-$QEMU_NAME $INPUT $$ARGUMENTS > $OUTPUT/$$NAME.stdout || true;
       done
+  - type: revng.runtime-abi-test
+    from:
+      - type: revng-qa.runtime-abi-tests
+        filter: "!x86-64 and !i386"
+    suffix: .stdout.yml
+    command: qemu-$QEMU_NAME ${INPUT}/runner-binary ${INPUT}/mmapped-binary > $OUTPUT
```
Cvelth, line 20: @aleclearmind Do names `runner-binary` and `mmapped-binary` sound good to you?
(this is the first time they explicitly appeared outside of docs)

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+    Value = static_cast<std::byte>(Temporary);
+    return StringRef{};
+  }
+
+  static QuotingType mustQuote(StringRef Scalar) {
+    return llvm::yaml::ScalarTraits<uint8_t>::mustQuote(Scalar);
+  }
+};
+LLVM_YAML_IS_FLOW_SEQUENCE_VECTOR(std::byte);
```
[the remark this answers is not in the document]
aleclearmind: OK

share/revng/test/configuration/revng/runtime.yml
aleclearmind: I propose:
* `runner-binary -> runner`
* `mmapped-binary -> compiler-artifact` to stress that we do not intend to run `compiler-artifact`.

Cvelth: `runner`: 👌 

`compiler-artifact`: I just weeded out the `artifact` naming from the suite - since it seems to be quite an overloaded term (kind of like `data`). But I do agree with your sentiment. Maybe something along that venue of `test-bodies` or `tested-instructions/functions`, would work better?

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.cpp
```diff
@@ -0,0 +1,263 @@
...
+  Fs.IsLittleEndian = Deserialized.IsLittleEndian;
+
+  for (auto &I : Deserialized.Iterations) {
+    llvm::StringRef Name = I.Function;
+
+    revng_assert(I.isArgumentTest() != I.isReturnValueTest());
+    if (I.isArgumentTest())
+      Fs.ArgumentTests[Name].emplace_back(I.extactArgumentTest(Architecture));
+    else
+      Fs.ReturnValueTests[Name].emplace_back(I.extractRVTest(Architecture));
```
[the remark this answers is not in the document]
Cvelth: This is done (forgot to close it).

share/revng/test/configuration/revng/runtime.yml
aleclearmind: > artifact [...] seems to be quite an overloaded term

Right.

What about `run-natively` and `foreign-executable`?
So you have `./run-natively foreign-executable`, which kinda succinctly conveys what's happening.

Cvelth: Sounds perfect!

Cvelth: I guess the only complaint I have now is about the `natively` wording. I'd rather it convey the fact that it modifies the control flow by injecting interrupts instead. Maybe something like `run-with-breakpoints` or `run-while-dumping-the-state`? I agree that it's very verbose, but it better conveys what's hapenning.

aleclearmind: I think `run-natively` is good enough, otherwise we can have `trace-functions` or `probe-functions`.

Cvelth: `trace/probe-calls`?

aleclearmind: I'd go for `probe-functions` or `probe-calls`.

tests/abi/tools/revng-check-compatibility-with-abi/ABIRuntimeTestResultParser.cpp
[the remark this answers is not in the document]
Cvelth: Done.

aleclearmind:
Merged in c6cb97c. Thanks.

PR #284, opened 2023-04-06
Model To Header - for "Edit Type"

dwarfyresearch:


include/revng/Pipeline/CLOption.h
```diff
@@ -55,6 +55,11 @@ inline llvm::StringRef typeNameImpl<int>() {
   return "int";
 }
 
+template<>
+inline llvm::StringRef typeNameImpl<uint64_t>() {
+  return "uint64_t";
+}
```
aleclearmind, line 61: Drop "Patch by drblallo"

include/revng/Model/VerifyHelper.h
```diff
@@ -28,6 +28,7 @@ class VerifyHelper {
   std::map<const model::Type *, uint64_t> AlignmentCache;
   std::set<const model::Type *> InProgress;
   bool AssertOnFail = false;
+  mutable std::string ReasonBuffer;
```
aleclearmind, line 31: 1. This is an hack, we should return an `llvm::Error`, but given that this part of the code will eventually also need to report *where* the error happened and report multiple errors, I'm OK with this hack for now.
   Add a TODO about this though.
2. In any case, no `mutable`

aleclearmind (changes requested):

include/revng/Model/VerifyHelper.h
dwarfyresearch: Yeah, I've added a comment and removed `mutable`. Methods should return an `llvm::Error`, but it is not trivial to port whole project to it now - so it can be a future work.

PR #285, opened 2023-04-06
Add opaque function to LLVM test helpers

pfez:
This commits adds an opaque function that can be called by simple LLVM IR snippets used in unit tests.


aleclearmind:
Merged in ca1fd10. Thanks!

PR #290, opened 2023-05-08
NodeJS-support fixups

mrjackv:


aleclearmind:
Merged in fcc0d21. Thanks!

PR #294, opened 2023-05-16
Miscellaneous improvements unlocked by libc++-16

Cvelth:
This PR proposes some style changes. It also drops most of the stuff that mirrors something from STL that is available now. As well as makes our containers more STL-like.

`llvm` sister PR: https://github.com/revng/llvm-project/pull/27
`revng-qa` sister PR: https://github.com/revng/revng-qa/pull/18
`revng-c` sister PR: https://rev.ng/gitlab/revng-private/revng-c/-/merge_requests/157


share/revng/clang-format-style-file.yml
```diff
@@ -115,7 +115,7 @@
   PenaltyBreakTemplateDeclaration: 10,
   PenaltyExcessCharacter: 751362,
   PenaltyIndentedWhitespace: 0,
-  PenaltyReturnTypeOnItsOwnLine: 0,
+  PenaltyReturnTypeOnItsOwnLine: 21,
```
aleclearmind: I've no strong opinions on this.  Both before and after look good to me.

How did you come up with 21?

share/revng/clang-format-style-file.yml
```diff
@@ -110,7 +110,7 @@
   PenaltyBreakBeforeFirstCallParameter: 800000,
   PenaltyBreakComment: 0,
   PenaltyBreakFirstLessLess: 120,
-  PenaltyBreakOpenParenthesis: 0,
+  PenaltyBreakOpenParenthesis: 800000,
```
aleclearmind: Uhm OK. Not sure how this will pan out in practice.

include/revng/ADT/Concepts.h
```diff
@@ -21,9 +21,8 @@ concept equality_comparable = requires(T &&LHS, T &&RHS) {
   { LHS == RHS } -> std::convertible_to<bool>;
 };
 
-// clang-format off
```
aleclearmind, line 24: Did we get rid of all of them?
If so, please ban `clang-format off` in `rcc.yml`

include/revng/ADT/Concepts.h
```diff
@@ -67,18 +68,17 @@ static_assert(not TupleLike<std::vector<int>>);
...
+concept SpecializationOf = requires(Type &Value) {
+  []<typename... Ts>(Ref<Ts...> &) {
+  }(const_cast<std::remove_const_t<Type> &>(Value));
+};
+
+static_assert(SpecializationOf<std::pair<int, long>, std::pair>);
+static_assert(SpecializationOf<const std::pair<int, long>, std::pair>);
+static_assert(SpecializationOf<std::string, std::basic_string>);
+static_assert(SpecializationOf<std::ostream, std::basic_ios>);
+static_assert(not SpecializationOf<std::string, std::basic_string_view>);
```
aleclearmind: Great!

lib/Support/IRHelpers.cpp
```diff
@@ -345,11 +345,9 @@ findJumpTarget(llvm::BasicBlock *&Result,
                  "This block leads to multiple jump targets");
     Result = BB;
   } else {
-    for (BasicBlock *Predecessor : predecessors(BB)) {
-      if (Visited.count(Predecessor) == 0) {
+    for (BasicBlock *Predecessor : predecessors(BB))
+      if (!Visited.contains(Predecessor))
```
aleclearmind, line 349: Please double check again all the `!`, `!=` and friends.
I want to be sure we don't f- this up and introduce super-subtle bugs.

CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ set(CMAKE_SHARED_LIBRARY_LINK_C_FLAGS)
 # cmake-format: off
 # set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address -shared-libasan")
 # cmake-format: on
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++2a")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++2b")
```
aleclearmind, line 31: Is this needed? What benefits do we get?

aleclearmind (changes requested):

include/revng/Support/Assert.h
```diff
@@ -63,7 +63,7 @@ extern "C" {
 
 #define boolcast(what) (bool) (what)
 
-#if __STDC_VERSION__ >= 201112L
+#if __STDC_VERSION__ >= 201'112L
```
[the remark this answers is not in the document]
Cvelth: Yep, I kind of expected that. What about hexes only though?

share/revng/clang-format-style-file.yml
Cvelth: Binary search :). I noticed that llvm uses 60 and the behaviour is different from ours (0), so I tried values until I found the breaking point. It seems to be `>20` - hence 21.

share/revng/clang-format-style-file.yml
Cvelth: > Not sure how this will pan out in practice.

What do you mean?

This will make clang-format a lot less likely to break nested function calls (as in, passing the result of one call as the first argument of the next), especially when function names are short-ish.

It has been quite frustrating in the past, since clang-format seemingly breaks legit looking code.

include/revng/Support/Assert.h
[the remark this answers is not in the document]
aleclearmind: Those are the ones I want to grep the most. Nope nope.

share/revng/clang-format-style-file.yml
```diff
@@ -13,7 +13,7 @@
   AlignTrailingComments: false,
   AllowAllArgumentsOnNextLine: true,
   AllowAllParametersOfDeclarationOnNextLine: false,
-  AllowShortBlocksOnASingleLine: Never,
+  AllowShortBlocksOnASingleLine: Empty,
```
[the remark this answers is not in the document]
aleclearmind: That might be good: such a loop would benefit from a comment in the body.

share/revng/clang-format-style-file.yml
```diff
@@ -15,7 +15,7 @@
   AllowAllParametersOfDeclarationOnNextLine: false,
   AllowShortBlocksOnASingleLine: Empty,
   AllowShortCaseLabelsOnASingleLine: false,
-  AllowShortEnumsOnASingleLine: true,
+  AllowShortEnumsOnASingleLine: false,
```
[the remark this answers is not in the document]
aleclearmind: I'm OK with it. Up to you.

share/revng/clang-format-style-file.yml
aleclearmind: Scary but OK.

include/revng/ADT/Concepts.h
pfez: I envision problems if we ban `clang-format off`. Let me explain why.

In `revng-c` we're gonna use MLIR, and MLIR requires certain headers to be included in a specific order (isn't that the case @drblallo @andrealinux1 ?).

In our `clang-format` configuration we force the reordering of headers, and that would break the ordering required from MLIR. So at the moment @andrealinux1 had to use `clang-format off` at least around those nasty `#include`s. This means that either we find a way to solve this include ordering problem, or it's already NOGO to ban `clang-format off` because we already know that we're gonna need it at least there.

There's of course the option of explicitly changing `check-conventions` so that `clang-fomat off` is allowed only in the specific MLIR files, but then `check-conventions` will be "polluted" by filenames, potentially causing MLIR-related changesets in `revng-c` to require changes in `revng` too. It can be done, but I'd honestly prefer to avoid this solution.

Cvelth: I got read of what I could. But some pain-points are still there:

- chaining `requires` in the concept definition sucks. `clang-format` will definitely fix it, but who knows when (There's still no option for it, even in trunk).
- when separate statements inside `requires` block grows beyond certain size, it gets really ugly.
- Stuff like this:
```
using rp_error = std::variant<std::monostate,
                              rp_simple_error,
                              rp_document_error>;
```
- there's a macro in `Assert.h` that look really broken when formatted.

lib/Support/IRHelpers.cpp
Cvelth: I went through them three times, but sounds good, I'll do one extra pass-through right before the merge.

include/revng/Support/Assert.h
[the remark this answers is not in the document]
Cvelth: Gotcha.

share/revng/clang-format-style-file.yml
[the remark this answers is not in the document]
Cvelth: I guess that makes sense. I'm convinced.

share/revng/clang-format-style-file.yml
aleclearmind: :+1:

share/revng/clang-format-style-file.yml
Cvelth: I'd assume this `20` is some value hardcoded somewhere in `clang-format` sources, but I never got a chance to look for it :)

include/revng/ADT/Concepts.h
aleclearmind: Can you post a `diff` of things you don't like when you drop all of `clang-format off`?

Cvelth: (moved to mattermost, but keeping the thread open as a reminder)

Cvelth: done :tada:

lib/Support/IRHelpers.cpp
Cvelth: Did an extra look-through.

include/revng/ADT/GenericGraph.h
```diff
@@ -1463,4 +1505,49 @@ struct GraphTraits<llvm::Inverse<T *>>
...
+namespace llvm {
+
+template<typename T>
+struct GraphTraits<T *>
+  : public decltype(revng::detail::forwardsSpecializationTrick<T>()) {};
+
+template<typename T>
+struct GraphTraits<Inverse<T *>>
+  : public decltype(revng::detail::backwardsSpecializationTrick<T>()) {};
+
```
aleclearmind, line 1552: Mmmh, I'm not very happy with the final form of the changes to `GenericGraph.h`.

It became strictly more difficult and hackish to read and what do we get back?

include/revng/ABI/Definition.h
```diff
@@ -380,10 +380,10 @@ class Definition : public generated::Definition {
 
     llvm::SmallVector<model::Register::Values, 8> Result;
     for (auto Register : GeneralPurposeArgumentRegisters())
-      if (Lookup.count(Register) != 0)
+      if (Lookup.contains(Register))
```
aleclearmind, line 383: Wiiiiiiiiiiiii

aleclearmind (changes requested):

include/revng/ADT/Concepts.h
```diff
@@ -7,6 +7,7 @@
 #include <array>
 #include <concepts>
 #include <iterator>
+#include <ostream>
```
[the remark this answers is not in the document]
Cvelth: That's the smallest standard header with inheritance - used to test `SpecializationOf`. I guess the way to go is to pick something from llvm instead. Any ideas?

include/revng/ADT/GenericGraph.h
Cvelth: > what do we get back?

No need to keep the "SomethingSomethingTag" within the classes. Also, it's a bit more generic (these things can be used on unrelated classes as long as they are sufficiently `GenericGraph`-like).

But yeah, I get what you mean - the bug makes it quite ugly. Consider, though, that it's a temporary solution that would eventually disappear.

Cvelth: Either way, I'm not opposed from stashing this in a branch somewhere and restoring this approach once we update to a clang version that doesn't suffer from this anymore.

include/revng/ADT/Concepts.h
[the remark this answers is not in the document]
aleclearmind: Make something in `tests/unit`.

include/revng/ADT/GenericGraph.h
aleclearmind: This PR has been open too long, let's try to get it down to non-controversial stuff.

Cvelth: Sounds good.

include/revng/ADT/Concepts.h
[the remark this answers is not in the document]
aleclearmind: ACK, let's keep them here.

> That's the smallest standard header with template-adjacent inheritance

Make up some classes using inheritance for the sole purpose of testing.
You can probably get away with a class inheriting `std::vector`.

aleclearmind:
Can you rebase one last time? I'd like to get this in next since it spans many many files.

include/revng/ADT/Concepts.h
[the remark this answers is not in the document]
Cvelth: Done, please check it out (`Concepts.h` header)

include/revng/ADT/GenericGraph.h
Cvelth: Stashed it into here for now:
https://github.com/Cvelth/revng/tree/stash/non-strict-generic-graph-concepts

Cvelth:
Merged in `79ad7ff5ea5fd12d04be5d8f532c3ea4f0446372` :tada:

PR #296, opened 2023-05-23
Tier 1 improvements 2

mrjackv:


python/revng/cli/_commands/fetch_debuginfo/common.py
```diff
@@ -25,6 +27,13 @@ def log_warning(self, message):
 log_warning = logger.log_warning
 
 
+def cache_directory() -> Path:
+    if "REVNG_CACHE_DIR" in os.environ:
```
aleclearmind, line 31: I don't think we need this environment variable.
Let's not overcustomize things, if user wants to customize such things there's an XDG-way to do it..

python/revng/cli/_commands/fetch_debuginfo/common.py
```diff
@@ -29,25 +27,15 @@ def log_warning(self, message):
...
-                with gzip.open(gzip_filename, "rb") as gzip_file:
-                    content = gzip.decompress(gzip_file.read())
-                with open(local_filename, "wb") as debug_file:
-                    debug_file.write(content)
-                os.remove(gzip_filename)
-            else:
-                with open(local_filename, "wb") as debug_file:
-                    shutil.copyfileobj(r.raw, debug_file)
-
+    with requests.get(url, stream=True) as req:
```
aleclearmind, line 39: `s/req/request/`

python/revng/cli/_commands/fetch_debuginfo/common.py
```diff
@@ -29,25 +27,15 @@ def log_warning(self, message):
...
+        if req.status_code == 200:
+            with open(local_filename, "wb") as debug_file:
+                for chunk in req.iter_content(chunk_size=64 * 1024):
```
aleclearmind, line 42: Ah! Nice!

aleclearmind (changes requested):

python/revng/cli/_commands/fetch_debuginfo/common.py
aleclearmind:
```
python/revng/cli/_commands/daemon.py:REVNG_DATA_DIR set, REVNG_PROJECT_ID unset: use '$REVNG_DATA_DIR' as persistence folder
python/revng/cli/_commands/daemon.py:REVNG_PROJECT_ID set, REVNG_DATA_DIR unset: use '$XDG_DATA_HOME/revng/$REVNG_PROJECT_ID'
python/revng/cli/_commands/daemon.py:REVNG_DATA_DIR and REVNG_PROJECT_ID set: use '$REVNG_DATA_DIR/$REVNG_PROJECT_ID'
```

`revng-daemon` should not be aware of the existence of "project IDs". We should just have `REVNG_DATA_DIR` or, better, a command line argument.

> Doing things via xdg in e.g. K8s is really annoying, you have to work things backwards from "I want revng to write its cache-y things in this directory".

Just set `XDG_CACHE_HOME=/tmp`? Let's not reinvent `XDG_CACHE_HOME`.

> Also it's fragile, because suddenly the downstream component needs to know "revng will write in $XDG_CACHE_HOME/revng"

An application (revng) cannot be aware of all the possible ways it will be deployed, it should do something reasonable and do something standard.
Beyond that, it's a duty of the deployer of taking care of additional constraints it introduces (e.g., `$HOME/.cache` not writeable by default).
XDG is a well known and it's a good standard decoupling point. If we fail because we're not handling such a thing as appropriate, it's a bug in hitman.

mrjackv: Let me address the matter in a few parts:

> `revng-daemon` should not be aware of the existence of "project IDs"

That was a really old design and the code just sat there, we can remove it. `hob` writes other files to `$XDG_DATA_HOME` anyways, so there aren't many advantages

> We should just have `REVNG_DATA_DIR` or, better, a command line argument.

The interaction between uvicorn (the asgi server) and `revng.daemon` (our asgi application) does not allow for command line arguments to be passed, this would mean that the cli handler (`revng.cli._commands.daemon`) has to serialize the options to an env variable, I'd just cut the middleman altogether

> Just set `XDG_CACHE_HOME=/tmp`? Let's not reinvent XDG_CACHE_HOME

First of all, I'm OK with using xdg, but as a default.
I've had a look and all application that write to `~/.cache` also have a dedicated environment variable to move it (e.g. [yarn](https://classic.yarnpkg.com/lang/en/docs/cli/cache/#toc-change-the-cache-path-for-yarn) [pip](https://stackoverflow.com/a/64180835) [black](https://black.readthedocs.io/en/stable/contributing/reference/reference_functions.html#black.cache.get_cache_dir) [node-gyp](https://github.com/nodejs/node-gyp/blob/main/README.md#command-options) [helm](https://helm.sh/docs/helm/helm/) [pylint](https://pylint.pycqa.org/en/latest/faq.html#where-is-the-persistent-data-stored-to-compare-between-successive-runs))
I think the basic idea is "if there's no defaul XDG_CACHE_HOME is good enough", but I imagine people wanting to use a different directory (debug info files can be big) without changing XDG_CACHE_HOME. By adding `REVNG_CACHE_DIR` a user can put `export REVNG_CACHE_DIR=/external/directory/outside/of/home` in `.profile` and be done with it without having to move the entire `.cache` tree there (this is also useful with K8s).

>  Just set `XDG_CACHE_HOME=/tmp`

This is sort-of counter productive, because I (in the sense of hitman) might want to mount a semi-persistent directory exclusively for revng and this can only be accomplished by using `/tmp/revng` as the mountpoint and "knowing" that it's the correct one. Also setting it might mean that some other tool starts writing there, which might or might not cause problems.

aleclearmind: > revng.daemon (our asgi application) does not allow for command line arguments to be passed

OK, we can stick environment variables if that's the reason.

About `REVNG_CACHE_DIR`, your arguments are valid but not very very strong and I still lean towards having one less configuration option, do things in a standard way and let specific configurations deal with specific situations.
This said, I'll trust you on this one: let's keep it!

In exchange, let's drop the project ID variable `:)`.

mrjackv: Dropped `REVNG_PROJECT_ID`

PR #298, opened 2023-05-26
Add a way to represent user comments in the model and to emit them as ptml

Cvelth:


include/revng/Model/Register.h
```diff
@@ -509,8 +509,9 @@ inline llvm::StringRef getRegisterName(Values V) {
   llvm::StringRef FullName = getName(V);
   auto Architecture = getReferenceArchitecture(V);
   revng_assert(Architecture != model::Architecture::Invalid);
-  auto ArchitectureNameSize = model::Architecture::getName(Architecture).size();
-  return FullName.substr(0, FullName.size() - ArchitectureNameSize - 1);
+  llvm::StringRef Name = model::Architecture::getName(Architecture);
+  revng_assert(FullName.substr(FullName.size() - Name.size()) == Name);
```
aleclearmind, line 513: OK, but did this happen for some reason?

include/revng/Model/Helpers.h
```diff
@@ -12,23 +12,59 @@
...
+  if constexpr (EntityWithComment<LHS> && EntityWithComment<RHS>)
+    To.Comment() = From.Comment();
+
+  if constexpr (EntityWithReturnValueComment<LHS>
+                && EntityWithReturnValueComment<RHS>) {
+    To.ReturnValueComment() = From.ReturnValueComment();
+  }
 
   return To;
 }
```
aleclearmind, line 51: 1. Very enterprise.
2. `static_assert(EntityWithCustomName<LHS> == EntityWithCustomName<RHS>)` and friends.
3. I see why you're sad about `ReturnValueComment`. An alternative would be to have a trait `copyMetadata` that we can extend specifically for those with `ReturnValueComment`, but your solution works too.

include/revng/ABI/FunctionType/Layout.h
```diff
@@ -61,6 +61,9 @@ struct Layout {
   public:
     std::optional<StackSpan> Stack;
     ArgumentKind::Values Kind;
+
+    std::string Name;
+    std::string_view Comment;
```
aleclearmind, line 66: 1. Why `std::string` vs `std::string_view`?
2. I'd stick with `StringRef` over `std::string_view`, unless we make `std::string_view` preferable over `StringRef` project-wide.

lib/ABI/FunctionType/Layout.cpp
```diff
@@ -813,14 +823,22 @@ Layout::Layout(const model::RawFunctionType &Function) {
...
     revng_assert(StackArgType.is(model::TypeKind::StructType));
 
+    const model::Type *OriginalStackType = StackArgType.UnqualifiedType().get();
+
     auto &Argument = Arguments.emplace_back();
 
+    // (WIP): I have FUD about this, so I'm leaving it out for now, let's
+    //        discuss RFT's stack argument comment during the review process.
+    //
+    // Argument.Comment = StackArgType.UnqualifiedType().get()->Comment();
```
aleclearmind, line 833: Why not?

lib/ABI/FunctionType/Layout.cpp
```diff
@@ -813,14 +823,22 @@ Layout::Layout(const model::RawFunctionType &Function) {
...
+    // Argument.Name = "stack_args"; // revng-c also hard-codes this name and we
+    //                                  _really_ want to be consistent.
```
aleclearmind, line 835: No, we (almost) never commit generated stuff to the model.

include/revng/Yield/PTML.h
```diff
@@ -67,4 +68,41 @@ functionLink(std::string_view Location, const model::Binary &Binary);
...
+/// \param LineWidth the expected width of a line within the emitted comment.
+/// \param Indentation the number of spaces appended **before** the comment as
+///        an indentation token (see `ptml::tokens::Indentation`).
+///
+/// \returns a serialized PTML string containing the comment.
+std::string functionComment(const model::Function &Function,
+                            const model::Binary &Binary,
+                            std::string_view CommentIndicator,
+                            std::size_t Indentation = 0,
+                            std::size_t LineWidth = 80);
```
aleclearmind, line 111: I'd drop the default values for `CommentIndicator`, `Indentation` and `LineWidth`, let the caller be explicit.
If needed, declare a global variable `DefaultIndetnation`.

lib/Yield/PTML.cpp
```diff
@@ -701,3 +703,150 @@ std::string yield::ptml::shallowFunctionLink(std::string_view Location,
...
+    for (const auto &Argument : Layout.Arguments) {
+      if (!Argument.Comment.empty()) {
+        auto WithName = Argument.Name + ' ' + std::string(Argument.Comment);
+        ArgumentExplanation += Comment.multiline(WithName, "\\param ");
+      }
+    }
+
+    if (!Layout.ReturnValueComment.empty()) {
+      if (!ArgumentExplanation.empty())
+        ArgumentExplanation += Comment.line(); // newline
```
aleclearmind, line 847: No comments at the end of the line

lib/Yield/PTML.cpp
```diff
@@ -701,3 +703,150 @@ std::string yield::ptml::shallowFunctionLink(std::string_view Location,
...
+
+  std::string commentImpl(std::string &&Text) {
+    return ::ptml::Tag(tags::Span, std::move(Text))
+      .addAttribute(ptml::attributes::Token, ptml::tokens::Comment)
+      .serialize();
+  }
+
+  std::string commentImpl(std::string_view Text) {
+    return commentImpl(std::string(Text));
+  }
```
aleclearmind, line 815: Drop `Impl` suffix and make them verbs (e.g., `indent`).

lib/Yield/PTML.cpp
```diff
@@ -701,3 +703,150 @@ std::string yield::ptml::shallowFunctionLink(std::string_view Location,
...
+      Result = indentationImpl(Indentation) + commentImpl(Indicator)
+               + indentationImpl(InternalIndentation + 1)
+               + commentImpl(std::move(Text));
+    }
+
+    return ::ptml::Tag(tags::Div, std::move(Result)).serialize() += '\n';
+  }
+
+  std::string
+  multiline(std::string_view Text, std::string_view FirstLine = "") {
```
aleclearmind, line 741: This function probably makes sense as a standalone function we can reuse in the future in some public header.

include/revng/Yield/PTML.h
```diff
@@ -67,4 +68,41 @@ functionLink(std::string_view Location, const model::Binary &Binary);
...
+/// the model representation of both the function and its type.
+///
+/// \param Function the function to gather comment information from.
+/// \param Binary the model of the binary the function belongs to.
+/// \param CommentIndicator the indicator used at the start of commented lines.
+///        If no custom value is specified, the value obtained from
+///        `model::Architecture::getAssemblyCommentIndicator` is used instead,
+///        which is the preferred option for the assembly (see the inline
+///        overload calling this function).
+///        For c code, the recommended indicator is either `'///'` or `'  *'`.
```
aleclearmind, line 98: `s/\bc\b/C/`

lib/Yield/Pipes/AssemblyPipes.cpp
```diff
@@ -71,9 +71,11 @@ void YieldAssembly::run(pipeline::Context &Context,
     revng_assert(MaybeFunction && MaybeFunction->verify());
     revng_assert((*MaybeFunction)->Entry() == Address);
 
+    const model::Function &ModelFunction = Model->Functions().at(Address);
     Output.insert_or_assign((*MaybeFunction)->Entry(),
-                            yield::ptml::functionAssembly(**MaybeFunction,
-                                                          *Model));
+                            yield::ptml::functionComment(ModelFunction, *Model)
+                              + yield::ptml::functionAssembly(**MaybeFunction,
+                                                              *Model));
```
aleclearmind, line 78: Too dense, split in multiple statements

include/revng/Model/Register.h
Cvelth: Luckily not, just got a bit of FUD when looking through it.

include/revng/Model/Helpers.h
Cvelth: 1. Glad you like it.

2. What do you mean? It's intentional that these two have different types. For example, this is used for `MyRFT.CustomName() = MyCFT.CustomName()` and so on you, you get the point. The idea is, as long as BOTH sides have the field in question, copy/move the value.

3. Yeah, `ReturnValueComment` sticks as a sore thumb. But I can't think of anything better. To the point where I even abandoned the effort of encoding comments for each register separately in RFT and went with the same approach.

include/revng/ABI/FunctionType/Layout.h
Cvelth: 1. `std::string` for name because that's what `name()` returns - hence we, as the caller, can make no assumption about the lifetime of it, while for the comment we use the explicit accessor (`Comment()`), hence we know that the original object owns the string - hence only a view is enough.

2. Sounds good. The only qualm with `StringRef` I have is that it's `llvm`, so when some part of the codebase doesn't need to explicitly depend on it, I prefer `std::string_view`, but I guess we're way too deep inside the llvm ecosystem to worry about stuff like that 😅.

lib/ABI/FunctionType/Layout.cpp
Cvelth: Yep, but this is a layout object, it's always created anew on demand and is never serialized.

include/revng/Yield/PTML.h
Cvelth: Either way works for me.

lib/Yield/PTML.cpp
Cvelth: This is not a function, but an important part of this class's interface, so it'd be painful to extract just it. That being I'm Ok with disclosing the entire class if that's what you want.

include/revng/Model/Helpers.h
aleclearmind: 2. OK
3. >  I even abandoned the effort of encoding comments for each register separately in RFT and went with the same 

   What do you mean? How are we encoding comments of RFT args?

include/revng/ABI/FunctionType/Layout.h
aleclearmind: 1. Gotcha
2. Eh yes. We abandoned `llvm::Optional` in favor of `std::optional`, one day we'll do the same for `string_view`, but mixing is counterproductive. So I'd ban `string_view` until further notice.

lib/ABI/FunctionType/Layout.cpp
aleclearmind: You sure? Don't we use this for conversion purposes?

include/revng/Model/Helpers.h
Cvelth: 3. Sorry, I mean for return values. Arguments work as you would expect.

lib/ABI/FunctionType/Layout.cpp
Cvelth: Layouts are never used for conversion - they just share a bunch of code with `CFT -> RFT` path. This line is specifically inside the `Layout` constructor though.

lib/Yield/PTML.cpp
aleclearmind: OK, I was thinking a non-comments specific string primitive similar to [Python's `textwrap`](https://docs.python.org/3/library/textwrap.html#textwrap.indent). But if it's too much of an hassle, leave it.

lib/ABI/FunctionType/Layout.cpp
aleclearmind: Mmmmh, then why do we record this in `Layout` at all? Is there any real user?

Cvelth: To make the new comment emitting thing function-type agnostic. It only relies on the layout.

When implementing it initially, I noticed that I'm duplicating layout-like logic - so I just extended it with what I was missing instead.

include/revng/Model/Helpers.h
aleclearmind: Uhm, what's difficult about RFT return values? Can't we add `Comment` to `TypedRegister` and then concat them upon emission similarly to what we decided to do for `StackArgumentsType`?

Cvelth: Yeah, I abandoned that idea initially, but since we're now getting back to it, this might have some merit.

The main problem is how to encapsulate it into the type conversion. With stack it simple since we'll have a new argument for each field - so the comments are passed through in the same way as the rest of the metadata. In CFT it doesn't make sense to have more than a single comment for the return value. I guess we could make this "concatenation of multiple comments" thing generic enough to be used everywhere, but I have a lot of FUD about that.

lib/ABI/FunctionType/Layout.cpp
aleclearmind: For final emission we usually don't use layout, in revng-c for instance I think we have specific logic for both.

If it's not for conversion... I'd just let the emitter do ad-hoc work for `CABIFunctionType` and `RawFunctionType`. In this case, going through `Layout` doesn't seem to bring much benefits.

Cvelth: > in revng-c for instance I think we have specific logic for both

There's no need to integrate that deep. Since the comment is always before the function itself, it's as simple as calling `functionComment(ForThisFunction, InThisBinary)` and just concatentating resulting string with the function decompiler emitted.

Here's literally the entirety of code on that side that handles the functions: https://rev.ng/gitlab/revng-private/revng-c/-/merge_requests/151/diffs?commit_id=934b2ef51e4457f34495f7e67e62230fb953e242#be93403bff9d784fb72272719009d4716bd81c81_1632_1635

Cvelth: > In this case, going through `Layout` doesn't seem to bring much benefits.

I feel like it reduces code duplication a bit. But frankly, I'm fine with either way.

aleclearmind: You can factor common code, but don't go through `Layout`. `Layout` was not designed for this, let's leave it out of the equation entirely.

include/revng/Model/Helpers.h
aleclearmind: Concatenation is fine.

Just make sure you wrap each thing in a proper [PTML `data-model-edit-attribute` tag](https://pad.rev.ng/s/L3m9LemOh) so that users can edit comments directly from the UI.

Cvelth: Oh, right, thank you for the remainder!

Cvelth: > Concatenation is fine.

I have a feeling like we already had this conversation I'm **really** hesitant to use any non-reversible transformations when converting user input (which comments are). Doing something stupid like going `RFT->CFT->RFT->CFT->RFT` could get really messy if we have some non-trivial comment conversions in the mix, especially once we would have different versions of the same analysis in the wild, which introduces possibilities of some of these conversions done differently from the others.

lib/ABI/FunctionType/Layout.cpp
Cvelth: As per mattermost discussion, we decided to
1. decouple this from the layout,
2. concatenate **type** comments in a fashion similar to
      /// \param stack_args: field_at_offset_0: <comment here>
      ///                    field_at_offset_64: <another comment here>
      /// . . .
3. do the same for RFT return values as well

lib/Yield/PTML.cpp
Cvelth: This is gone.

lib/ABI/FunctionType/Layout.cpp
Cvelth: I dropped changes to the layout.

include/revng/ABI/FunctionType/Layout.h
Cvelth: This is gone.

Also, I replaced `std::string_view` with `llvm::StringRef` everywhere I noticed them.

lib/Yield/Pipes/AssemblyPipes.cpp
Cvelth: Done.

lib/ABI/FunctionType/Layout.cpp
Cvelth: All three are done :tada:

include/revng/Yield/PTML.h
Cvelth: Done.

lib/Yield/Doxygen.cpp
```diff
@@ -221,3 +223,247 @@ std::string yield::ptml::freeFormComment(const ::ptml::PTMLBuilder &PTML,
   auto Result = Builder.emit(Text.str());
   return Result.empty() ? Result : '\n' + Result;
 }
+
+using pipeline::serializedLocation;
+namespace ranks = revng::ranks;
+
+static llvm::SmallVector<DoxygenLine, 16>
```
aleclearmind, line 230: NOTETOSELF: review

lib/Yield/Doxygen.cpp
```diff
@@ -0,0 +1,223 @@
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+/// \file Doxygen.cpp
```
aleclearmind, line 5: The whole comment thing is not specific to `Yield`, we're putting way too much stuff in `Yield`.

Let's move this to `librevngPTML`, since it already depends on `librevngModel`.

aleclearmind (changes requested):

include/revng/ADT/STLExtras.h
```diff
@@ -392,6 +408,12 @@ constexpr bool constexprOr(CallableType &&Callable) {
                                  std::forward<CallableType>(Callable));
 }
 
+template<std::size_t IterationCount, typename CallableType>
+constexpr std::optional<std::size_t> constexprFind(CallableType &&Callable) {
+  return detail::constexprFindImpl(std::make_index_sequence<IterationCount>(),
+                                   std::forward<CallableType>(Callable));
+}
```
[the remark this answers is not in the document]
Cvelth: > What does this function do?

Verifies that when `Callable` is called for each number `[0, IterationCount)` it only returns true for either one number (in which case it returns that number) or for none at all (in which case it returns `std::nullopt`).

It's a pretty generic helper for some non-trivial template magic. I'm not opposed moving it closer to its user, but I feel like I might want to reuse it sooner or later.

> I'd avoid a property such as `constexpr` in the name

That's for consistency with its siblings already present in the file (like `constexprRepeat`, `constexprAnd`, and so on). I'm not Ok with renaming any one of them, but all for a better name for the whole group. Any ideas?

include/revng/Model/Helpers.h
```diff
@@ -186,6 +186,81 @@ inline std::string customName(const model::RawFunctionType &Function,
   return customName(Function, Argument.key());
 }
 
+template<EntityWithKey T, typename Root = model::Binary>
+inline std::string comment(const T &Value) {
+  static_assert(EntityWithComment<T>);
+  return detail::pathImpl<Root, T>(Value, "Comment");
```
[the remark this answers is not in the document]
Cvelth: 1. 👌.

> Why not simply accept `EntityWithComment`

`EntityWithComment` is not guaranteed to be an `EntityWithKey`, and this needs both to be true. I prefer this approach (one check for the major trait + one static assert for a more minor one) to a `requires` clause checking both concepts, but I'm not opposed to switching - if that's what you prefer.

> return `Value.Comment`

I have trouble understanding what you mean here. This function returns the path to the comment field for an arbitrary element of the binary (if it has one), how is returning value of said field related to anything?

lib/Yield/Doxygen.cpp
```diff
@@ -0,0 +1,223 @@
...
+/// \brief
```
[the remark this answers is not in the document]
Cvelth: I'm pretty sure I copied it from some other cpp file.

We need to get this straight. Please point me to a correct header so I can update all the "wrong" ones in #294.

lib/Yield/Doxygen.cpp
Cvelth: It starts to feel like `librevngPTML` is starting to become what `librevngYield` was intended to be, and I feel responsible because of putting to much stuff in `yield` temporarily 😅.

This comment of yours makes me think that we need to deal with [this thing](https://mattermost.rev.ng/revng/pl/qjsnb8z73if85kfpb7z447cety) sooner that I expected.

include/revng/ADT/STLExtras.h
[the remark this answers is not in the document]
aleclearmind: We can move them in a `constexpr` namespace?

Please add doc for each one of these with example, otherwise they're quite mysterious.

include/revng/Model/Helpers.h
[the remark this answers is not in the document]
aleclearmind: 2.2 Mh, that was not clear at all.

Why are we concatenating strings instead of return a `TupleTreePath` and then let the user serialize it?

This seems a bit misguided.

lib/Yield/Doxygen.cpp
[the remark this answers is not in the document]
aleclearmind:
````
/// \file ControlFlowEdgesGraph.cpp
/// \brief

//
//

```

lib/Yield/Doxygen.cpp
aleclearmind: Yeah, but in the meantime, let's move things there.

include/revng/ADT/STLExtras.h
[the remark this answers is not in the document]
Cvelth: Sounds good.

include/revng/Model/Helpers.h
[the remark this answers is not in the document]
Cvelth: Because of ease of implementation. I thought about doing it through `TupleTreePath`s, but that started to seem like an overkill really fast (since it's one more level of indirection between the information we see and what we want to end up with), but if you'd prefer that, I can take a deeper look.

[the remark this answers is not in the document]
aleclearmind: Mmmh, not now. But if we concat strings, this will break when we change the serialization format.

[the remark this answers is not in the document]
Cvelth: Consider that before I touched it, it was also serializing strings. So, at the very least, it's not  worse after compared to what it was before.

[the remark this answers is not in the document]
aleclearmind: OK, please leave a TODO about this and let's move on.

aleclearmind (changes requested):

lib/Yield/Doxygen.cpp
```diff
@@ -0,0 +1,469 @@
...
+static constexpr auto Identifier = "doxygen.identifier";
+
+} // namespace ptml::tokens
+
+struct Attribute {
+  std::string Name, Value;
+};
+
+struct DoxygenTag {
+  enum class Types { Untagged, Indentation, Keyword, Identifier };
```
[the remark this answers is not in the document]
Cvelth: 1-2. 👌.
3. `Untagged` is just a normal "comment" - something we have no additional information about. `Indentation` is just that - extra spaces somebody might not like and want to remove.

lib/Yield/Doxygen.cpp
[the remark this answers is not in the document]
Cvelth: Forgot to close this. It's done in #294.

include/revng/Model/Type.h
```diff
@@ -53,16 +56,12 @@ class model::Type : public model::generated::Type {
...
+  // Constructors that do not accept an ID are manually implemented because
+  // they need to generate a random one.
   Type();
   Type(TypeKind::Values TK);
-  Type(TypeKind::Values Kind, uint64_t ID) : Type(Kind, ID, Identifier(), "") {}
-  Type(TypeKind::Values Kind,
-       uint64_t ID,
-       Identifier CustomName,
-       std::string OriginalName) :
-    model::generated::Type(Kind, ID, CustomName, OriginalName) {}
```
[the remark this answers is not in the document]
Cvelth: Done, see `Model: clean type constructors up`.

include/revng/ADT/STLExtras.h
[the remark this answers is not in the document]
Cvelth: Done, see `CompilationTime.h` header.

include/revng/Model/Helpers.h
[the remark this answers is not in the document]
Cvelth: Done.

lib/Yield/Doxygen.cpp
Cvelth: Done.

lib/Yield/Doxygen.cpp
```diff
@@ -0,0 +1,469 @@
...
+
+  return Result;
+}
+
+std::string yield::ptml::functionComment(const ::ptml::PTMLBuilder &PTML,
+                                         const model::Function &Function,
+                                         const model::Binary &Binary,
+                                         llvm::StringRef CommentIndicator,
+                                         std::size_t Indentation,
+                                         std::size_t LineWidth) {
```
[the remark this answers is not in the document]
Cvelth: Done.

lib/Yield/Doxygen.cpp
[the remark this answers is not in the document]
Cvelth: 1-2. Done.

3. Not sure if you expect me to do anything about them 🤔.

lib/Yield/Doxygen.cpp
```diff
@@ -0,0 +1,469 @@
...
+    for (DoxygenLine &&Line : as_rvalue(Lines)) {
+      auto [ResultLine, CurrentSize] = line();
+
+      if (Line.Tags.empty())
+        Result += serialize(std::move(ResultLine));
+
+      bool WereTagsEmittedSinceLastBreak = false;
+      for (auto Iterator = Line->begin(); Iterator != Line->end(); ++Iterator) {
+        llvm::StringRef TagText = Iterator->Value;
+        while (!TagText.empty()) {
```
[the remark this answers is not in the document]
Cvelth: Done, I do have a bit of distaste for how many `inout` argument that function ended up with, but I guess it's still easier to read than before.

lib/Yield/Doxygen.cpp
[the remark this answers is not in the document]
Cvelth: 2. P. S. I went with `DoxygenToken` instead because it's a bit shorter, which resulted in better formatting.

aleclearmind:
Please rebase

Cvelth:
@aleclearmind rebased (sister PR too)

aleclearmind:
Please rebase again. Sorry but I'm merging lots of stuff today.

Cvelth:
@aleclearmind, rebased :)

aleclearmind:
I get check-conventions failures, can you please check?

aleclearmind:
Merged in 4daa8e6. Thanks!

PR #299, opened 2023-05-29
Propagate prototypes for wrappers

kolodziej:


lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -488,6 +494,130 @@ void DetectABI::interproceduralPropagation() {
...
+        //  - don't write to callee arguments
+        //  - every store instruction writes to registers (not memory)
+        auto *BB = GCBI.getBlockAt(Block.ID().start());
+
+        auto StackPointer = [this]() {
+          const auto &Architecture = Binary->Architecture();
+          auto SP = model::Architecture::getStackPointer(Architecture);
+          llvm::StringRef StackPointerName = model::Register::getCSVName(SP);
+          return M.getGlobalVariable(StackPointerName);
+        }();
```
aleclearmind, line 574: Nice, but I'd avoid a lambda for a function called just once.

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -488,6 +494,130 @@ void DetectABI::interproceduralPropagation() {
...
+
+        // Check if wrapper writes to Stack Pointer
+        const bool WriteSP = WrittenRegisters.count(StackPointer) > 0;
```
aleclearmind, line 577:
```suggestion
        const bool WritesSP = WrittenRegisters.count(StackPointer) > 0;
```

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -488,6 +494,130 @@ void DetectABI::interproceduralPropagation() {
...
+
+        using std::ranges::count_if;
+        auto IsWritingToCalleeArg = [this, &WrittenRegisters](auto &Argument) {
+          auto *CSV = M.getGlobalVariable(model::Register::getName(Argument));
+          return WrittenRegisters.count(CSV);
+        };
+        const auto &Arguments = CalleeLayout.argumentRegisters();
+        const bool WritesCalleeArgs = count_if(Arguments, IsWritingToCalleeArg);
```
aleclearmind, line 585: Nice

aleclearmind (changes requested):

lib/EarlyFunctionAnalysis/DetectABI.cpp
kolodziej: I only do this trick to not make mess with temporary variable in the scope. In output code it will be optimized out, so what why should we avoid it?

share/revng/test/tests/analysis/DetectABI/x86_64/propagate.S.yml.filecheck
```diff
@@ -0,0 +1,7 @@
+#
+# Copyright (c) rev.ng Labs Srl. See LICENSE.md for details.
+#
+
+CHECK:      Functions:
+CHECK:        - Entry: "0x400660:Code_x86_64"
+CHECK-NEXT:     Prototype: "/Types/CABIFunctionType-5831792892140344958"
```
aleclearmind, line 7: We can't hard code addresses and type IDs.

Consider using `revng model compare`, see the other usages, if they make sense to you or ping me for some support.
`revng model compare` basically enables you verify the "shape" of the model (e.g., two functions sharing the same prototype) without relying on hardcoding addresses or type IDs. Basically your ground truth should look something like this:

```
Functions:
- CustomName: caller
  Prototype: "/Types/CABIFunctionType-66666666666"
DynamicFunctions:
- OriginalName: puts
  Prototype: "/Types/CABIFunctionType-66666666666"
Types:
- Type: CABIFunctionType
  ID: 66666666666
```

`revng model compare` is smart enough to detect if the model you feed to it is compatible with the one above, even if the type ID is not 66666666666 and even if the model above is invalid (functions are missing the mandatory `Entry` attribute).

kolodziej: You explicitly asked for FileCheck tests, so that's why I decided to hardcode type id seed and compare specific values.

aleclearmind: We can probably make it work with `FileCheck` too, if you put a placeholder in place of the ID, but then the test case is not super robust w.r.t. to the order of the functions.

It's better to switch to `revng model compare` which enables you to write more "semantically rich" ground truth.

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -488,6 +494,100 @@ void DetectABI::interproceduralPropagation() {
   }
 }
 
+void DetectABI::propagatePrototypes() {
+  for (model::Function &Function : Binary->Functions()) {
+    const auto &Entry = Function.Entry();
```
aleclearmind, line 499: Please drastically reduce the usage of `auto` ([B.15](https://pad.rev.ng/s/SJNUet_ku#))

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -488,6 +494,130 @@ void DetectABI::interproceduralPropagation() {
...
+        } else if (auto ID = Successor->Destination(); ID.isValid()) {
+          using std::ranges::find_if;
+          auto ByEntryAddress =
+            [Address = ID.notInlinedAddress()](const model::Function &F) {
+              return Address == F.Entry();
+            };
+          auto Result = find_if(Binary->Functions(), ByEntryAddress);
+          revng_assert(Result != Binary->Functions().end());
+
+          NewPrototype = &(Result->Prototype());
```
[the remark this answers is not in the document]
kolodziej: Now there is `getPrototype` call for everything, so this part has been dropped.

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -488,6 +494,130 @@ void DetectABI::interproceduralPropagation() {
...
+  for (model::Function &Function : Binary->Functions()) {
+    const auto &Entry = Function.Entry();
+
+    auto &Summary = Oracle.getLocalFunction(Entry);
+    auto &ABI = Summary.ABIResults;
+    auto &CFG = Summary.CFG;
+    auto &WrittenRegisters = Summary.WrittenRegisters;
+
+    const bool HasNotArguments = ABI.ArgumentsRegisters.size() == 0;
+    const bool HasNotReturns = ABI.FinalReturnValuesRegisters.size() == 0;
```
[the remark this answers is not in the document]
aleclearmind: This does not seem to be fixed.

aleclearmind (changes requested):

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -488,6 +494,100 @@ void DetectABI::interproceduralPropagation() {
...
+                if (GV->getName() == "pc") {
+                  return false;
+                }
+
+                return Register == model::Register::Invalid;
+              }
+            }
+
+            return false;
+          };
```
aleclearmind: Can you make a regular function? Do you really need to capture `this`?

If you do, consider making a method.

aleclearmind (changes requested):

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -488,6 +494,99 @@ void DetectABI::interproceduralPropagation() {
...
+
+    if (HasNoReturns and HasNoArguments and IsSingleNode) {
+      auto &Block = *CFG.begin();
+
+      if (Block.Successors().size() == 1) {
+        auto &Successor = *Block.Successors().begin();
+
+        // Select new prototype for wrapper function
+        if (auto Call = dyn_cast<efa::CallEdge>(Successor.get())) {
+          auto Prototype = getPrototype(*Binary, Entry, Block, *Call);
```
aleclearmind, line 518: Turns this "precondition checks" into early loop `continue`.

```c++
if (not FirstCondition)
  continue;
if (not SecondCondition)
  continue;
if (not ThirdCondition)
  continue;

File: lib/EarlyFunctionAnalysis/DetectABI.cpp [hunk 10]
```diff
@@ -488,6 +494,99 @@ void DetectABI::interproceduralPropagation() {
...
+    const bool IsSingleNode = CFG.size() == 1;
+
+    if (HasNoReturns and HasNoArguments and IsSingleNode) {
+      auto &Block = *CFG.begin();
+
+      if (Block.Successors().size() == 1) {
+        auto &Successor = *Block.Successors().begin();
+
+        // Select new prototype for wrapper function
+        if (auto Call = dyn_cast<efa::CallEdge>(Successor.get())) {
```
aleclearmind, line 517: `auto &`?

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -488,6 +494,99 @@ void DetectABI::interproceduralPropagation() {
...
+          // Get layout of wrapped function
+          Layout CalleeLayout = Layout::make(Prototype);
+
+          // Verify that wrapper function:
+          //  - don't write stack pointer
+          //  - don't write to callee arguments
+          //  - every store instruction writes to registers (not memory)
+          auto *BB = GCBI.getBlockAt(Block.ID().start());
+
+          llvm::GlobalVariable *StackPointer = nullptr;
```
aleclearmind, line 530: `GCBI.spReg()`

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -488,6 +494,99 @@ void DetectABI::interproceduralPropagation() {
...
+          {
+            const auto &Architecture = Binary->Architecture();
+            auto SP = model::Architecture::getStackPointer(Architecture);
+            llvm::StringRef StackPointerName = model::Register::getCSVName(SP);
+            StackPointer = M.getGlobalVariable(StackPointerName);
+          }
+
+          // Check if wrapper writes to Stack Pointer
+          const bool WritesSP = WrittenRegisters.count(StackPointer) > 0;
```
aleclearmind, line 539: Can we use `.contains`?

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -488,6 +494,99 @@ void DetectABI::interproceduralPropagation() {
...
+
+          using std::ranges::count_if;
+          auto IsWritingToCalleeArg = [this,
```
aleclearmind, line 542: * `IsWritingToCalleeArg` -> `IsWrittenByCaller` or...
* Rewrite creating a set of `model::Register` and see if intersection is empty

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -488,6 +494,99 @@ void DetectABI::interproceduralPropagation() {
...
+                                       &WrittenRegisters](auto &Argument) {
+            auto *CSV = M.getGlobalVariable(model::Register::getName(Argument));
+            return WrittenRegisters.count(CSV);
+          };
+          const auto &Arguments = CalleeLayout.argumentRegisters();
+          const bool WritesCalleeArgs = count_if(Arguments,
+                                                 IsWritingToCalleeArg);
```
aleclearmind, line 549: `> 0`, elsewhere too.

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -488,6 +494,99 @@ void DetectABI::interproceduralPropagation() {
...
+                if (GV->getName() == "pc") {
+                  return false;
+                }
+
+                return Register == model::Register::Invalid;
+              }
+            }
+
+            return false;
+          };
```
aleclearmind: Didn't we say to use `isMemory`?

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -553,16 +553,7 @@ void DetectABI::propagatePrototypes() {
...
-                auto Register = model::Register::fromCSVName(GV->getName(),
-                                                             Architecture);
-
-                // If destination is not register variable, it's write to memory.
-                if (GV->getName() == "pc") {
-                  return false;
-                }
-
-                return Register == model::Register::Invalid;
+                return isMemory(GV);
```
aleclearmind: * Ah, no fixup commits.
* You have to use `isMemory` on `Destination`

share/revng/test/configuration/revng/for-propagated-prototypes.yml
```diff
@@ -0,0 +1,15 @@
...
+commands:
+  - type: revng.test-propagated-prototypes
+    from:
+      - type: revng-qa.compiled-with-debug-info
+        filter: for-propagated-prototypes
+    command: |-
+      RESUME=$$(temp -d);
+      revng analyze --model-type-id-seed=0 --resume "$$RESUME" ImportBinary "$INPUT" -o /dev/null;
+      revng analyze --model-type-id-seed=0 --resume "$$RESUME" AddPrimitiveTypes "$INPUT" -o /dev/null;
+      revng analyze --model-type-id-seed=0 --resume "$$RESUME" DetectABI "$INPUT" |
```
aleclearmind: Thanks to `revng model compare`, tests should pass even without `--model-type-id-seed=0`.

Drop it.

share/revng/test/tests/analysis/DetectABI/x86_64/propagate.S.model.yml
```diff
@@ -0,0 +1,24 @@
+#
+# This file is distributed under the MIT LIcense. See LICENSE.md for details.
+#
+
+---
+Functions:
+  - Entry: "0x400660:Code_x86_64"
```
aleclearmind, line 7: Can't hardcode addresses in ground truth.

Can you put `CustomName`?

share/revng/test/tests/analysis/DetectABI/x86_64/propagate.S.model.yml
```diff
@@ -0,0 +1,24 @@
...
+    Prototype: "/Types/CABIFunctionType-1"
+Types:
+  - Kind: CABIFunctionType
+    ID: 1
```
aleclearmind, line 11: While any ID works here, for now, ID below a certain number are reserved for `PrimitiveType`s.
This will change, but for now, it's better to here a random `uint64_t`.

share/revng/test/tests/analysis/DetectABI/x86_64/propagate.S.model.yml
```diff
@@ -0,0 +1,24 @@
...
+    ABI: SystemV_x86_64
+    ReturnType:
+      UnqualifiedType: "/Types/PrimitiveType-1540"
```
aleclearmind, line 14: Import the definition of the `PrimitiveType`s you refer to.

share/revng/test/tests/analysis/DetectABI/x86_64/propagate.S.add-puts.yml
```diff
@@ -0,0 +1,26 @@
+#
+# This file is distributed under the MIT LIcense. See LICENSE.md for details.
+#
+
+---
+Changes:
```
aleclearmind, line 6: Nice

share/revng/test/configuration/revng/for-propagated-prototypes.yml
```diff
@@ -11,5 +11,6 @@ commands:
       RESUME=$$(temp -d);
       revng analyze --model-type-id-seed=0 --resume "$$RESUME" ImportBinary "$INPUT" -o /dev/null;
       revng analyze --model-type-id-seed=0 --resume "$$RESUME" AddPrimitiveTypes "$INPUT" -o /dev/null;
+      revng model apply "$$RESUME/context/model.yml" "$SOURCE.add-puts.yml" -o "$$RESUME/context/model.yml";
```
aleclearmind, line 14: No fixup commits

aleclearmind (changes requested):

lib/EarlyFunctionAnalysis/DetectABI.cpp
kolodziej: Temporary, so `const auto &`

share/revng/test/tests/analysis/DetectABI/x86_64/propagate.S.model.yml
kolodziej: This doesn't contain `CustomName` in model received from `DetectABI`.

lib/EarlyFunctionAnalysis/DetectABI.cpp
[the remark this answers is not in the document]
kolodziej: It is fixed. Please take a look on commit history.

lib/EarlyFunctionAnalysis/DetectABI.cpp
kolodziej: I think we will still need to use some lambda to call method (`std::bind` is not generally recommended). In next change capturing `this` is removed, because most of this function is replaced with `isMemory` call. Should I squash _isMemory-commit_ to this?

aleclearmind: No need for a lambda if don't capure anything, we can just make a `static` function and call it a day.

> Should I squash isMemory-commit to this?

No fixup commits. All new code should not be amended by later commits.

lib/EarlyFunctionAnalysis/DetectABI.cpp
aleclearmind: You mean [this](https://github.com/revng/revng/pull/299#discussion_r1217965540)?

Outlining and having `return`s seems like a good thing to do.

Having a lot of nested `if`s is not great, a lot of `continue`s are sad too, outlining and using `return` is the best solution.

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -488,6 +507,81 @@ void DetectABI::interproceduralPropagation() {
...
+      Function.Prototype() = Prototype;
+    }
+  }
+}
+
+void DetectABI::propagatePrototypes() {
+  using std::ranges::for_each;
+  for_each(Binary->Functions(), [this](model::Function &Function) {
+    propagatePrototypesInFunction(Function);
+  });
```
aleclearmind, line 582: Use a regular range-based `for` loop

aleclearmind (changes requested):

kolodziej:
Related PR: https://github.com/revng/revng-qa/pull/19

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -720,6 +720,7 @@ void DetectABI::analyzeABI(llvm::BasicBlock *Entry) {
 
   // Commit ABI analysis results to the oracle
   Oracle.getLocalFunction(EntryAddress).ABIResults = ABIResults;
+  Oracle.getLocalFunction(EntryAddress).WrittenRegisters = WrittenRegisters;
```
aleclearmind, line 817:
```c++
... &LocalFunction = Oracle.getLocalFunction(EntryAddress);
LocalFunction.ABIResult = ...;
LocalFunction.WrittenRegisters = ...;
```

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -99,6 +101,18 @@ static pipeline::RegisterAnalysis<DetectABIAnalysis> A1;
 
 namespace efa {
 
+static bool isWritingToMemory(llvm::Instruction &I) {
+  if (auto *StoreInst = dyn_cast<llvm::StoreInst>(&I)) {
+    auto *Destination = StoreInst->getPointerOperand();
+
+    if (auto *GV = dyn_cast<llvm::GlobalVariable>(Destination)) {
```
aleclearmind, line 108: You need to just use `isMemory`, no need to test `GlobalVariable`, it doesn't make much sense, see the implementation of `isMemory`.

aleclearmind (changes requested):

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -488,6 +507,81 @@ void DetectABI::interproceduralPropagation() {
...
+                               << Function.Prototype().toString()
+                               << ") with wrapped function's prototype: "
+                               << Prototype.toString());
+      Function.Prototype() = Prototype;
+    }
+  }
+}
+
+void DetectABI::propagatePrototypes() {
+  using std::ranges::for_each;
```
aleclearmind, line 579: Drop

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -488,6 +507,81 @@ void DetectABI::interproceduralPropagation() {
...
+  const MetaAddress &Entry = Function.Entry();
+
+  FunctionSummary &Summary = Oracle.getLocalFunction(Entry);
+  ABIAnalyses::ABIAnalysesResults &ABI = Summary.ABIResults;
+  SortedVector<efa::BasicBlock> &CFG = Summary.CFG;
+  std::set<llvm::GlobalVariable *> &WrittenRegisters = Summary.WrittenRegisters;
+
+  const bool HasArguments = ABI.ArgumentsRegisters.size() > 0;
+  const bool HasReturns = ABI.FinalReturnValuesRegisters.size() > 0;
+  const bool IsSingleNode = CFG.size() == 1;
```
aleclearmind, line 520: Drop the `const`, one day `const` will be default, but for now the return on putting `const` like this is too little compared to the benefit.

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -488,6 +507,81 @@ void DetectABI::interproceduralPropagation() {
...
+
+  if (HasReturns or HasArguments or !IsSingleNode) {
```
aleclearmind, line 522: Either use trigraphs only or don't, we prefer trigraphs: `or not IsSingleNode`.

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -159,6 +176,8 @@ class DetectABI {
   void initializeInterproceduralQueue();
   void runInterproceduralAnalysis();
   void interproceduralPropagation();
+  void propagatePrototypesInFunction(model::Function &Function);
+  void propagatePrototypes();
```
aleclearmind, line 180: Add doxygen summarizing under what conditions we propagate.

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -488,6 +507,81 @@ void DetectABI::interproceduralPropagation() {
...
+  const bool HasSingleSuccessor = Block.Successors().size() == 1;
+  if (!HasSingleSuccessor) {
+    return;
+  }
+
+  auto &Successor = *Block.Successors().begin();
+
+  // Select new prototype for wrapper function
+  if (const auto &Call = dyn_cast<efa::CallEdge>(Successor.get())) {
+    auto Prototype = getPrototype(*Binary, Entry, Block, *Call);
```
aleclearmind, line 537: `auto` abuse

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -488,6 +507,81 @@ void DetectABI::interproceduralPropagation() {
...
+
+    using abi::FunctionType::Layout;
+    // Get layout of wrapped function
+    Layout CalleeLayout = Layout::make(Prototype);
+
+    // Verify that wrapper function:
+    //  - don't write stack pointer
+    //  - don't write to callee arguments
+    //  - every store instruction writes to registers (not memory)
+    llvm::BasicBlock *BB = GCBI.getBlockAt(Block.ID().start());
```
aleclearmind, line 547: We have `using namespace llvm`, drop all `llvm::`.

share/revng/test/configuration/revng/for-propagated-prototypes.yml
aleclearmind: Updates?

lib/EarlyFunctionAnalysis/DetectABI.cpp
kolodziej: The problem is that we also have `efa::BasicBlock` and skipping namespace here generates a conflict. In other places I'll remove namespace.

PR #300, opened 2023-06-19
Add pipeline invalidation

drblallo:
Replace the stub implementation of invalidation with the proper implementation. A ReadPathCache is added to each global so that it can keep tracks of what target are associated to which read paths.


include/revng/Pipeline/Committer.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+//
```
aleclearmind, line 2:
```suggestion

//
```

lib/Lift/LiftPipe.cpp
```diff
@@ -44,6 +45,8 @@ void Lift::run(Context &Ctx,
   PM.add(new LoadBinaryWrapperPass(Buffer->getBuffer()));
   PM.add(new LiftPass);
   PM.run(TargetsList.getModule());
+  // for (auto &Target : TargetsList.enumerate())
+  // Committer.commit(Target, TargetsList.name());
```
aleclearmind, line 49: Uh?

lib/Pipeline/Global.cpp
```diff
@@ -18,7 +18,34 @@ using namespace std;
...
+
+  PathCache = std::move(*MaybeCache);
+  return llvm::Error::success();
+}
+
+llvm::Error Global::serializeCache(llvm::raw_ostream &OS) const {
+  return getPathCache().serialize(OS);
+}
+
+Error Global::storeToDisk(StringRef Path, bool StoreCache) const {
```
aleclearmind, line 37: I'd make storing the cache mandatory. Let's not have potentially harmful configuration options unless we have a good reason for them.

include/revng/Pipeline/Context.h
```diff
@@ -120,11 +123,47 @@ class Context {
...
+  void collectReadFields(const LocatedTarget &Target) const {
+    Globals.collectReadFields(Target);
+  }
+
+  void clearReadFields() const { Globals.clearReadFields(); }
+  void pushReadFields() const { Globals.pushReadFields(); }
+  void popReadFields() const { Globals.popReadFields(); }
+  void stopTracking() const { Globals.stopTracking(); }
+
+  Committer makeCommitter() { return Committer(*this, CurrentStepName); }
```
aleclearmind, line 163: The whole `Committer` infrastructure is undocumented but it's very important for devs to understand it.

Please thoroughly document it, possibly with examples.

include/revng/Pipeline/Context.h
```diff
@@ -120,11 +123,47 @@ class Context {
 
 public:
   llvm::Error storeToDisk(llvm::StringRef Path) const {
-    return Globals.storeToDisk(Path);
+    if (auto Error = Globals.storeToDisk(Path); Error)
+      return Error;
+
+    return llvm::Error::success();
```
aleclearmind, line 129: Isn't this a more verbose version of what was there previously?

Same below.

include/revng/Pipeline/Context.h
```diff
@@ -120,11 +123,47 @@ class Context {
...
+  }
+
+public:
+  void getTargetsAssociatedToPaths(const TupleTreePath &Path,
+                                   InvalidationMap &Out) const {
+    for (auto &Name : Globals.getGlobalNames()) {
+      llvm::cantFail(Globals.get(Name))->getTargetsAssociatedToPaths(Path, Out);
+    }
   }
```
aleclearmind, line 145: I understand why you might want `Global::getTargetsAssociatedToPaths` to take an `InvalidationMap` as an out argument, as opposed to returning it, but I think `Context::getTargetsAssociatedToPaths` can return it.

include/revng/Pipeline/Context.h
```diff
@@ -120,11 +123,47 @@ class Context {
...
+      return Error;
+
+    return llvm::Error::success();
+  }
+
+public:
+  void getTargetsAssociatedToPaths(const TupleTreePath &Path,
+                                   InvalidationMap &Out) const {
+    for (auto &Name : Globals.getGlobalNames()) {
+      llvm::cantFail(Globals.get(Name))->getTargetsAssociatedToPaths(Path, Out);
```
aleclearmind, line 143: `llvm::cantFail(Globals.get(Name))` is weird. Can't we have `GlobalsMap` expose the list of globals without going through names?

include/revng/Pipeline/Context.h
```diff
@@ -120,11 +123,47 @@ class Context {
...
+  void clearReadFields() const { Globals.clearReadFields(); }
+  void pushReadFields() const { Globals.pushReadFields(); }
+  void popReadFields() const { Globals.popReadFields(); }
+  void stopTracking() const { Globals.stopTracking(); }
+
+  Committer makeCommitter() { return Committer(*this, CurrentStepName); }
+
+  void setCurrentStepName(llvm::StringRef NewName) {
+    CurrentStepName = NewName;
+  }
```
aleclearmind, line 167: A `Context` should not have a "current step", this seems like an hack. Maybe we can pass to `MyPipe::run` an instance of `Committer` so we give the right dignity to invalidation?

aleclearmind (changes requested):

share/revng/test/configuration/revng/revng.yml
```diff
@@ -11,11 +11,16 @@ commands:
...
     suffix: .bc
     command: |-
-      revng llvm pipeline "$INPUT" "$OUTPUT" Lift Lift ":Root" --analysis DetectABI;
-      revng llvm pipeline "$OUTPUT" "$OUTPUT" Lift EnforceABI "*:CSVsPromoted";
+      MODEL="$$(temp)";
+      TMP="$$(temp -d)";
+      revng model dump "$INPUT2" -o "$$MODEL";
+      revng analyze DetectABI "$INPUT1" --resume "$$TMP" -m "$$MODEL" -o /dev/null;
+      revng artifact EnforceABI "$INPUT1" --resume "$$TMP" -o /dev/null;
+      revng model inject "$$TMP"/context/model.yml "$$TMP"/EnforceABI/module.ll -o "$OUTPUT";
```
aleclearmind, line 23: Can we have a single commit for the commit dropping `revng llvm pipeline`?

python/revng/cli/_commands/llvm_pipeline.py
```diff
@@ -1,88 +0,0 @@
-#
```
aleclearmind, line 1: Goodbye!

lib/Pipes/RootKind.cpp
```diff
@@ -42,27 +42,6 @@ IsolatedRootKind::symbolToTarget(const llvm::Function &Symbol) const {
   return std::nullopt;
 }
 
-void RootKind::getInvalidations(const Context &Ctx,
```
aleclearmind, line 45: Goodbye

include/revng/Pipeline/ReadPathsCache.h
```diff
@@ -0,0 +1,215 @@
...
+
+class SerializableReadPathCache {
+public:
+  using ValueType = std::pair<pipeline::TargetReadPathsSerializable,
+                              pipeline::ParsableLocatedTarget>;
+  using Vector = std::vector<ValueType>;
+  Vector Data;
+};
+
+class ReadPathCache {
```
aleclearmind, line 123: This is a bit of a loaded name that implies its usages, not what it does "in the void".

Maybe we can call this `PathToTargetBimap` or something along that line?

include/revng/Pipeline/ReadPathsCache.h
```diff
@@ -0,0 +1,215 @@
...
+  static PathSerializationPrimitives make() {
+    PathSerializationPrimitives Primitives;
+    Primitives.Deserialize = &stringAsPath<T>;
+    Primitives.Serialize = &pathAsString<T>;
+    return Primitives;
+  }
+
+  std::optional<TupleTreePath> (*Deserialize)(llvm::StringRef);
+  std::optional<std::string> (*Serialize)(const TupleTreePath &);
+};
```
aleclearmind, line 86: * AFAIU this abstraction used to type-erase `ReadPathCache` from the type of the global variable. Is this correct?
* Can we use a more orthodox `virtual` class? I find this a bit "surprising"

include/revng/Pipeline/ReadPathsCache.h
```diff
@@ -0,0 +1,215 @@
...
+
+  template<TupleTreeCompatibleAndVerifiable T>
+  static ReadPathCache make() {
+    return ReadPathCache(PathSerializationPrimitives::make<T>());
+  }
+
+  void insert(const LocatedTarget &LocatedTarget,
+              const TargetReadPaths &TargetReadPaths) {
+    Map[TargetReadPaths].push_back(LocatedTarget);
+    ReverseMap[LocatedTarget].push_back(TargetReadPaths);
```
aleclearmind, line 145: Can you elaborate on why we need both directions for the map?
It's not obvious to me.

include/revng/Pipeline/ReadPathsCache.h
```diff
@@ -0,0 +1,215 @@
...
+class Context;
+
+class ParsableLocatedTarget {
+public:
+  std::string Content;
+  std::string StepName;
+  std::string ContainerName;
+};
+
+class LocatedTarget {
```
aleclearmind, line 26:
```suggestion
class TargetInStep {
```
?

"location" is a bit of an overloaded term.

include/revng/Pipeline/ReadPathsCache.h
```diff
@@ -0,0 +1,215 @@
...
+    return Primitives;
+  }
+
+  std::optional<TupleTreePath> (*Deserialize)(llvm::StringRef);
+  std::optional<std::string> (*Serialize)(const TupleTreePath &);
+};
+
+class TargetReadPaths {
+private:
+  TupleTreePath Path;
```
aleclearmind, line 90: 1. Why is this class called `TargetReadPaths` (plural) if it holds a single path?
2. Can we drop this class in favor of a plain `TupleTreePath` plus a pair of free functions to serialize/deserialize?
   Maybe instead of free functions these could be methods of `Context` or `Global`?
   There's quite a few classes with non-trivial names but very simple content that make decoding this file rather difficult.

aleclearmind (changes requested):

include/revng/Pipeline/Context.h
drblallo: eh, it is a hack, i guess one could extract the context from the committer? that would change all pipes

include/revng/Pipeline/ReadPathsCache.h
drblallo: we need the one from path to target so that we can figure out who is invalidated when the element pointed by that path changes, when we need the other to be able to compute the hash when we are doing the cold start, as well as to check if a target has been produced without committing it

include/revng/Pipeline/Context.h
drblallo: i would prefer not too because sometimes this is invoked very intensively for a lot of paths different paths, and the result are all accumulated into the same Out invalidation map

include/revng/Pipeline/ReadPathsCache.h
drblallo: i entirely removed the class

lib/Pipeline/Global.cpp
aleclearmind: Let's move to have cache serialized on disk on the side of containers.

include/revng/Pipeline/ReadPathsCache.h
aleclearmind: Please add a `TODO` stating that we can remove one of the two directions of the map once the tracking becomes mandatory ("strict").

For the cold start use case, we can revert the map on demand.

aleclearmind (changes requested):

include/revng/Pipeline/GlobalsMap.h
```diff
@@ -116,6 +131,15 @@ class GlobalsMap {
...
 
+  std::vector<llvm::StringRef> getGlobalNames() const {
+    std::vector<llvm::StringRef> ToReturn;
+
+    for (const auto &Global : Map)
+      ToReturn.push_back(Global.first);
+
+    return ToReturn;
+  }
+
```
aleclearmind, line 142: Now unused

lib/Pipes/TaggedFunctionKind.cpp
```diff
@@ -31,7 +31,7 @@ using namespace llvm;
 
 std::optional<pipeline::Target>
 TaggedFunctionKind::symbolToTarget(const llvm::Function &Symbol) const {
-  if (not Tag->isTagOf(&Symbol) or Symbol.isDeclaration())
+  if (not Tag->isExactTagOf(&Symbol) or Symbol.isDeclaration())
```
aleclearmind, line 34: Here we're fixing a long standing bug, but the commit does not mention it.

The commit message just says "Preliminaries" and includes a set of changes that do not seem strictly related to me. I think one commit can be the one fixing this bug and the other the `stopTracking` thing.

include/revng/Pipes/TaggedFunctionKind.h
```diff
@@ -20,6 +20,16 @@ class TaggedFunctionKind : public pipeline::LLVMKind {
...
 
+  // It is on the heap to avoid Initialization Order Fiasco
+  std::unique_ptr<llvm::SmallVector<TaggedFunctionKind *>> Children = nullptr;
+
+  void registerChild(TaggedFunctionKind *Child) {
+    if (Children == nullptr)
+      Children = std::make_unique<llvm::SmallVector<TaggedFunctionKind *>>();
+
+    Children->push_back(Child);
+  }
```
aleclearmind, line 31: * This commit introduces this function but does not use it.
* I'm confused why we need `children`. `Kind` can already enumerate its children.

include/revng/Model/CollectAccesses.h
```diff
@@ -0,0 +1,53 @@
+#pragma once
```
aleclearmind, line 1: 1. This file does not seem to be included by anyone.
2. Files should have the same name as the "main" thing in a class.
3. This file only contains stuff in the `detail` namespace.

include/revng/Pipeline/PathTargetBimap.h
```diff
@@ -0,0 +1,172 @@
...
+    Primitives.Deserialize = &stringAsPath<T>;
+    Primitives.Serialize = &pathAsString<T>;
+    return Primitives;
+  }
+
+  std::optional<TupleTreePath> (*Deserialize)(llvm::StringRef);
+  std::optional<std::string> (*Serialize)(const TupleTreePath &);
+};
+
+class SerializableReadPathCache {
```
aleclearmind, line 83: A cache is something one introduces for performance reasons containing information that could be otherwise recomputed.

This is not information we can recompute, I'd call it `ReadPathsSet` or if, in principle, it doesn't need to be used only for *read* data I'd just call it `TupleTreePathsSet` or `PathsSet`.

include/revng/Pipeline/PathTargetBimap.h
```diff
@@ -0,0 +1,172 @@
...
+namespace yaml {
+
+// YAML traits for TargetInStep
+template<>
+struct MappingTraits<pipeline::SerializableTargetInStep> {
+  static void
+  mapping(IO &IO, pipeline::SerializableTargetInStep &TargetInStep) {
+    IO.mapRequired("target", TargetInStep.Content);
+    IO.mapRequired("stepName", TargetInStep.StepName);
+    IO.mapRequired("containerName", TargetInStep.ContainerName);
```
aleclearmind, line 150: As a convention we preserve the casing of in-memory fields.

aleclearmind (changes requested):

pfez:
This MR depends on https://github.com/revng/revng/pull/310/commits

It should be merged only after https://github.com/revng/revng/pull/310/commits has been merged too

aleclearmind (changes requested):

include/revng/Pipeline/Committer.h
```diff
@@ -0,0 +1,61 @@
...
+// Example: void SomePipeProducingRootElement(const Context& Ctx, SomeContainer&
+// Container) {
+//   auto Committer = Ctx.getCommitter();
+//   auto RelevantInformations = readSomeFields(getModelFromContext(Ctx));
+//   produceRoot(RelevantInformations, Container);
+//   Ctx.commit(getGeneratedTarget(Container), Container);
+// }
+//
+class Committer {
```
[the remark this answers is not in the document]
drblallo: resolved with the new ExecutionCotext mechanism

include/revng/Pipes/TaggedFunctionKind.h
drblallo: the issue is that the childs in kind are of type Kind and here we need them of type TaggedFunctionKind, and i have not implemented the required stuff to make them upcastable with llvm::cast

aleclearmind (changes requested):

lib/Pipeline/ExecutionContext.cpp
```diff
@@ -0,0 +1,40 @@
...
+#include "revng/Pipeline/Step.h"
+#include "revng/Pipeline/Target.h"
+
+using namespace pipeline;
+
+ExecutionContext::ExecutionContext(Context &Ctx,
+                                   Step &Step,
+                                   PipeWrapper *Pipe) :
+  Ctx(&Ctx), CurrentStep(&Step), Pipe(Pipe) {
+  if (Pipe != nullptr)
```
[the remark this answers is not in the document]
drblallo: added a comment saying that it is null only when executing a analysis

include/revng/Pipeline/ExecutionContext.h
```diff
@@ -0,0 +1,68 @@
...
+//   auto RelevantInformations = readSomeFields(getModelFromContext(Ctx));
+//   produceRoot(RelevantInformations, Container);
+//   Ctx.commit(getGeneratedTarget(Container), Container);
+// }
+//
+class ExecutionContext {
+private:
+  Context *Ctx;
+  Step *CurrentStep;
+  PipeWrapper *Pipe;
```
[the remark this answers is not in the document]
drblallo: they were initialized in the only available constructor, anyway i initialized them inline too.

include/revng/Pipeline/GlobalTupleTreeDiff.h
```diff
@@ -101,7 +113,12 @@ class GlobalTupleTreeDiff {
     return &Casted->getDiff();
   }
 
+  llvm::SmallVector<const TupleTreePath *, 4> getPaths() const {
+    return Diff->getPaths();
+  }
+
```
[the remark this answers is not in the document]
drblallo: i have a forwarding method here because there is no getDiff method. i could drop the entire class, but it is useful to implement clone.

include/revng/Pipeline/Pipe.h
```diff
@@ -198,6 +198,75 @@ class PipeWrapperImpl : public PipeWrapperBase {
...
+  const PathTargetBimap &getPathCache(llvm::StringRef GlobalName) const {
+    revng_assert(PathCache.find(GlobalName) != PathCache.end());
+    return PathCache.find(GlobalName)->second;
+  }
+  PathTargetBimap &getPathCache(llvm::StringRef GlobalName) {
+    return PathCache[GlobalName];
+  }
+};
+
+struct PipeWrapper {
```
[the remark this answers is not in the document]
drblallo: unfortunately we have that strange inheritance structure that was needed to not duplicate code in analysis and pipes, and PipeWrapperBase is not accessible from outside.

include/revng/TupleTree/Tracking.h
```diff
@@ -183,6 +233,11 @@ struct Tracking {
   static void pop(const M &LHS) {
     visitTuple<M, PopVisitor>(LHS);
   }
+
+  template<typename M>
+  static void stop(const M &LHS) {
+    visitTuple<M, StopTrackingVisitor>(LHS);
```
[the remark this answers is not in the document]
drblallo: i not sure i understand the rationale behind this one. how does it differ from  my version?

lib/Lift/LiftPipe.cpp
```diff
@@ -44,6 +44,9 @@ void Lift::run(Context &Ctx,
   PM.add(new LoadBinaryWrapperPass(Buffer->getBuffer()));
   PM.add(new LiftPass);
   PM.run(TargetsList.getModule());
+
+  for (auto &Target : TargetsList.enumerate())
+    Ctx.commit(Target, TargetsList);
```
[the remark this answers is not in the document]
drblallo: the issue with this one is that i don't have a nice way to keep it working if the root target should ever change, say because it changed the name or the kind or something similar. We probably need to take a global decision about this since it will be relevant in other pipes too.

lib/Pipeline/ExecutionContext.cpp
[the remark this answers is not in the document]
aleclearmind: `Pipe` only seems to be used to check if `!= nullptr`. Can we turn this is in something like `bool EnableTracking`?

include/revng/TupleTree/Tracking.h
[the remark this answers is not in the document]
aleclearmind: You don't need a class, you can just use a lambda.

lib/Lift/LiftPipe.cpp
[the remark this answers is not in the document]
aleclearmind: I'm not sure I understand your answer.

aleclearmind (changes requested):

include/revng/Pipeline/Step.h
```diff
@@ -68,26 +68,120 @@ class Step {
...
+    for (auto &Pipe : Pipes) {
+      Pipe.ContainerWiseCache.getTargetsAssociatedToPaths(*Ctx,
+                                                          GlobalName,
+                                                          Path,
+                                                          Out[getName()]);
+    }
+  }
+
+  bool isTargetCached(llvm::StringRef GlobalName,
+                      const TargetInStep &Target) const {
```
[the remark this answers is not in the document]
drblallo: went for `Step::invalidationMetadataContain`, the intent of the method is to query if a existing target has paths associated to it or not, which would mean the user did not commited that target in a pipe

include/revng/TupleTree/Tracking.h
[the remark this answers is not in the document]
drblallo: i don't understand how this is supposed to work.
did you meant
```
visit([]<typename T>(const T &Obj){
   if contexper (IsSetOrKOC<T>) 
     Obj.stopTracking() 
});
```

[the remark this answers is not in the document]
aleclearmind: If that's necessary, yup. But in these cases I'd just use `auto` instead of `<typename T>`.
But yeah, whatever works with lambda is good.

include/revng/Pipeline/PathTargetBimap.h
drblallo: i went for ContainerInvalidationMetadata, as agreed

lib/Pipeline/ExecutionContext.cpp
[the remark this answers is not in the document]
drblallo: i can, but that would imply that you can run a pipe while not tracking, which is not intented. Do i do it anyway?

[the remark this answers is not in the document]
aleclearmind: > you can run a pipe while not tracking

I don't think it's a huge risk.

> Do i do it anyway?

Yes

include/revng/Pipeline/Pipe.h
```diff
@@ -198,6 +198,75 @@ class PipeWrapperImpl : public PipeWrapperBase {
 
 } // namespace detail
 
-using PipeWrapper = InvokableWrapper<detail::PipeWrapperBase>;
+class ContainerToCacheMap {
```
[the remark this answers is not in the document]
drblallo: went for InvalidationMetadata as agreed

include/revng/Pipeline/Global.h
```diff
@@ -149,6 +166,25 @@ class TupleTreeGlobal : public Global {
...
+  }
+
+  void collectReadFields(const TargetInStep &Target,
+                         PathTargetBimap &Out) override {
+    const auto &AsConst = Value;
+    auto Results = revng::Tracking::collect(AsConst);
+    for (auto &Result : Results.ExactVectors)
+      Out.insert(Target, Result);
+    for (auto &Result : Results.ExactVectors)
+      Out.insert(Target, Result);
```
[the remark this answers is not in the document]
drblallo: 1: ok
2: yep, it was a bug
3: we do not need to differentiate between them now, but we will need to differentiate between them when we introduce the hash discriminator when doing the cold start, because exact vectors will need to contribute to the hash in a different way.

lib/Pipeline/ExecutionContext.cpp
```diff
@@ -0,0 +1,40 @@
...
+    getContext().stopTracking();
+}
+
+void ExecutionContext::commit(const Target &Target,
+                              llvm::StringRef ContainerName) {
+  TargetInStep ToCollect(Target, ContainerName.str());
+  getContext().collectReadFields(ToCollect,
+                                 Pipe->ContainerWiseCache.getPathCache());
+
+  getContext().clearReadFields();
```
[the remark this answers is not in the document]
drblallo: i can, but it effectively eats away from the maximum number of possible elements in the stack. Not sure we care about that.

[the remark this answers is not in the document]
aleclearmind: > it effectively eats away from the maximum number of possible elements in the stack

Whta is "it"? push + pop eats only on element on the stack. `clearReadFields` acts only on the topmost one AFAIU.

Alternatively, we can rename `clearReadFields`, but I'm not sure it's easy to come up with a nice name saying that only the topmost layer is dropped.

include/revng/Pipeline/Global.h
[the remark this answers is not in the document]
aleclearmind: 3. Leave a comment about this.

aleclearmind (changes requested):

include/revng/Pipeline/ReadPathsCache.h
aleclearmind: https://github.com/revng/revng/pull/300#discussion_r1285786953

drblallo: i will close this one since i did the thing of dropping Primitives, and there is a equivalent issue down below

include/revng/Pipeline/GlobalTupleTreeDiff.h
```diff
@@ -12,14 +12,18 @@ namespace pipeline {
 class GlobalTupleTreeDiffBase {
 private:
   char *ID;
+  llvm::StringRef GlobalName;
```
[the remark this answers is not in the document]
drblallo: this should no longer be relevant since i rebased on a master which did had you commit

lib/Pipeline/ExecutionContext.cpp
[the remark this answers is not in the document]
drblallo: i considered to implement it with pop followed by push, but that is broken when the stack was empty, since you can't pop a empty stack. i could place a check and not do the pop if it is empty, but i would prefer to keep that as an assertion. i guess clearReadFields works as a name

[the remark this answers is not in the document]
drblallo: i went for clearTop

include/revng/Pipeline/Global.h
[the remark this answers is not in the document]
drblallo: i put the comment on the struct instead of that snippet.

include/revng/Pipeline/Step.h
```diff
@@ -68,26 +68,120 @@ class Step {
...
     Containers(std::move(Containers)),
     Pipes({ std::forward<PipeWrapperTypes>(PipeWrappers)... }),
-    PreviousStep(&PreviousStep) {}
+    PreviousStep(&PreviousStep),
+    Ctx(&Ctx) {}
+
+public:
+  void getTargetsAssociatedToPaths(llvm::StringRef GlobalName,
+                                   const TupleTreePath &Path,
+                                   InvalidationMap &Out) const {
```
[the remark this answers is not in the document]
drblallo: 1, 2, 3) went for registerTargetsDependingOn so it is clear they are registered in Out.

4) yeah, if we want we can try to wipe out all functions written in this style and replace them all.

include/revng/TupleTree/Tracking.h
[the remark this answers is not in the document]
drblallo: as we said this one it is possible to write the visitors with a inline lambda, but since we have a custom visitor here doing so would require to write a lambda that is longer and uglier than writing a the struct like they are now. in the end i would keep this one like this.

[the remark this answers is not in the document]
drblallo: previous message was incorrect. this is not writable with a lamda because one of the visitor methods has two template parameters and the other has 1.

include/revng/Pipeline/Pipe.h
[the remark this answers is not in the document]
aleclearmind: 1. If `PipeWrapperBase` is not for pipes only but also for analysis, rename it (or at least add a `TODO` about it).
2. Add a comment about the reason to be of this `class` as you just explained it to me.

include/revng/TupleTree/Tracking.h
```diff
@@ -23,6 +23,46 @@
 struct ReadFields {
   std::set<TupleTreePath> Read;
   std::set<TupleTreePath> ExactVectors;
+
+  template<typename M>
+  std::optional<llvm::MD5::MD5Result> getHash(const M &TupleTree) {
```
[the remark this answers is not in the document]
aleclearmind: This is still here, somehow

include/revng/Pipeline/GlobalTupleTreeDiff.h
```diff
@@ -55,6 +59,14 @@ class GlobalTupleTreeDiffImpl : public GlobalTupleTreeDiffBase {
 
   TupleTreeDiff<T> &getDiff() { return Diff; }
   const TupleTreeDiff<T> &getDiff() const { return Diff; }
+
+  llvm::SmallVector<const TupleTreePath *, 4> getPaths() const override {
+    llvm::SmallVector<const TupleTreePath *, 4> ToReturn;
+    for (const auto &Entry : Diff.Changes) {
```
aleclearmind, line 65: Drop `auto`

include/revng/Pipeline/GlobalTupleTreeDiff.h
```diff
@@ -12,14 +12,18 @@ namespace pipeline {
...
   virtual void serialize(llvm::raw_ostream &OS) const = 0;
   virtual ~GlobalTupleTreeDiffBase() = default;
   virtual std::unique_ptr<GlobalTupleTreeDiffBase> clone() const = 0;
   virtual bool isEmpty() const = 0;
-  GlobalTupleTreeDiffBase(char *ID) : ID(ID) {}
+  virtual llvm::SmallVector<const TupleTreePath *, 4> getPaths() const = 0;
+  GlobalTupleTreeDiffBase(char *ID, llvm::StringRef GlobalName) :
+    ID(ID), GlobalName(GlobalName) {}
   const char *getID() const { return ID; }
+  llvm::StringRef getGlobalName() const { return GlobalName; }
```
aleclearmind, line 26: Reorganize methods according to [B.16](https://pad.rev.ng/s/SJNUet_ku#B-CC)

include/revng/Pipeline/PathTargetBimap.h
```diff
@@ -0,0 +1,121 @@
...
+        Map.at(Path).erase(ToErase);
+      }
+
+      ReverseMap.erase(Iter);
+    }
+  }
+
+public:
+  void registerTargetsDependingOnPath(const TupleTreePath &Path,
+                                      ContainerToTargetsMap &Out) const;
```
aleclearmind, line 113: This doesn't seem very generic, it's probably better to inline it in its only caller.

include/revng/Pipeline/Pipe.h
```diff
@@ -54,6 +54,8 @@ constexpr bool checkPipe(auto (C::*)(First, Rest...))
 template<typename PipeType>
 class PipeWrapperImpl;
 
+// ToDo: Rename, there are 3 layers of wrappers around a pipe and they are
```
aleclearmind, line 57: `// TODO` all caps.
We never used `ToDo`, please fix everywhere.

include/revng/Pipeline/Step.h
```diff
@@ -92,7 +96,40 @@ class Step {
     Component(std::move(Component)),
     Containers(std::move(Containers)),
     Pipes({ std::forward<PipeWrapperTypes>(PipeWrappers)... }),
-    PreviousStep(&PreviousStep) {}
+    PreviousStep(&PreviousStep),
+    Ctx(&Ctx) {}
+
+public:
+  // ToDo, drop the Out parameter pattern if favour of coorutines in the whole
```
aleclearmind, line 103:
```
// TODO: blabla blabla
//       more blabla

File: lib/Pipeline/Step.cpp [hunk 42]
```diff
@@ -144,40 +330,144 @@ void Step::removeSatisfiedGoals(ContainerToTargetsMap &Targets,
...
+  if (not MaybeBool)
+    return MaybeBool.takeError();
+
+  if (not MaybeBool.get())
+    return llvm::Error::success();
+
+  if (auto Error = Containers.load(DirPath))
+    return Error;
+
+  return loadCacheFromDisk(DirPath);
```
aleclearmind, line 383: Feels like `loadCacheFromDisk` should be squashed into `Containers.load`, unless there's a reason to split them.

include/revng/Pipeline/Step.h
```diff
@@ -92,7 +96,40 @@ class Step {
...
+  }
+
+private:
+  llvm::Error loadCacheFromDiskImpl(const revng::DirectoryPath &Path,
+                                    ContainerSet::value_type &Pair);
+
+public:
+  llvm::Error loadCacheFromDisk(const revng::DirectoryPath &Path);
+
+  llvm::Error storeCacheToDisk(const revng::DirectoryPath &Path) const;
```
aleclearmind, line 132: 1. Drop "disk"
2. We said not to call it cache, please fix everywhere.

include/revng/Support/AccessTracker.h
```diff
@@ -17,14 +17,18 @@ namespace revng {
 class AccessTracker {
```
aleclearmind, line 17: AFAIU upon construction we're not tracking, when we call `.clear` we start tracking.
This is quite unexpected for a class called `AccessTracker`.
What's the ratio here?

include/revng/TupleTree/Tracking.h
```diff
@@ -73,7 +77,9 @@ struct Tracking {
     }
 
     template<typename Type, size_t FieldIndex>
-    static void visitTupleElement(const Type &CurrentItem) {}
+    static void visitTupleElement(const Type &CurrentItem) {
+      CurrentItem.template getTracker<FieldIndex>().stopTracking();
+    }
```
aleclearmind: Bugfix?

include/revng/TupleTree/TrackingContext.h
```diff
@@ -0,0 +1,46 @@
...
+//
+
+#include "revng/Pipeline/Target.h"
+#include "revng/TupleTree/Tracking.h"
+#include "revng/TupleTree/TupleTreeCompatible.h"
+
+namespace revng {
+
+template<TupleTreeCompatible T>
+class TrackingContext {
```
aleclearmind, line 14: Unused, untested, undocumented. Can we postpone its merge?

include/revng/TupleTree/TupleTree.h
```diff
@@ -25,11 +25,30 @@
...
+template<typename T>
+concept HasTracking = requires(T _) { T::HasTracking; };
+
+template<typename T>
+struct TrackGuard {
+  const T *TrackedObject;
+
+public:
+  TrackGuard(const T &TrackedObject) : TrackedObject(&TrackedObject) {
+    if constexpr (HasTracking<T>)
```
aleclearmind, line 43: Why would we want to `TrackGuard` something that doesn't have tracking?

lib/Lift/JumpTargetManager.cpp
```diff
@@ -358,7 +358,7 @@ MaterializedValue JumpTargetManager::readFromPointer(MetaAddress LoadAddress,
       uint64_t Addend = Relocation.Addend();
       auto RelocationSize = model::RelocationType::getSize(Relocation.Type());
       if (LoadAddress == Relocation.Address() and LoadSize == RelocationSize) {
-        revng_assert(not StringRef(Function.name()).contains('\0'));
+        revng_assert(not StringRef(Function.OriginalName()).contains('\0'));
```
aleclearmind, line 361: Add `// TODO: add this to model's verify function`

lib/Model/Binary.cpp
```diff
@@ -95,6 +100,8 @@ bool Binary::verifyTypes(VerifyHelper &VH) const {
 }
 
 void Binary::dump() const {
+  revng::Tracking::push(*this);
```
aleclearmind, line 155: I think we should use `TrackGuard` here. Just pushing is a bug.

lib/Pipeline/PathTargetBimap.cpp
```diff
@@ -0,0 +1,22 @@
+/// \file ReadPathsCache.cpp
```
aleclearmind, line 1: Outdated

aleclearmind:
1. "Preliminaries for pipeline invalidation": provide more details about what you did in the commit message.
2. "Add pipeline invalidation": commit message is outdated, it mentions `ReadPathCache`.

NOTETOSELF: resume review from "Add tracking debug infrastructure".

aleclearmind (changes requested):

include/revng/Model/VerifyHelper.h
```diff
@@ -41,6 +42,11 @@ class VerifyHelper {
 
   ~VerifyHelper() { revng_assert(InProgress.size() == 0); }
 
+public:
+  bool hasPushedTracking() const { return HasPushedTracking; }
+
+  void setPushedTracking() { HasPushedTracking = true; }
```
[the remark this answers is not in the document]
drblallo: 2) i fail to see how is it fundamentally different now 

```cpp
bool Binary::verifyTypes(VerifyHelper &VH) const {
  revng::TrackingPushingScope Dc(VH, *this);
```

i did it this way because if i return a lock from a function it is implementation defined if it pushes once or it pushes, pops, and pushes again, due to rvo.

3) i considered it, but it can't be done, the fundamentally try to solve different problems, one is trying shutdown the tracking of a whole structure and is intended to let you push multiple times, assuming that it will not happen too many times. The other is instead used to prevent mutually recursive heterogeneous functions from running out of available pushes when they don't care about tracking, which are just 7.

include/revng/Pipeline/Context.h
```diff
@@ -127,6 +128,16 @@ class Context {
...
+public:
+  void collectReadFields(const TargetInContainer &Target,
+                         llvm::StringMap<PathTargetBimap> &Out) const {
+    Globals.collectReadFields(Target, Out);
+  }
+
+  void clearTopReadFields() const { Globals.clearTopReadFields(); }
+  void pushReadFields() const { Globals.pushReadFields(); }
+  void popReadFields() const { Globals.popReadFields(); }
+  void stopTracking() const { Globals.stopTracking(); }
```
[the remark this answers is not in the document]
drblallo: we went for this design because we agreed that a temporary stop is represented by a push and a pop

lib/Pipeline/ExecutionContext.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+ExecutionContext::ExecutionContext(Context &Ctx,
+                                   Step &Step,
+                                   PipeWrapper *Pipe) :
+  TheContext(&Ctx),
+  CurrentStep(&Step),
+  Pipe(Pipe),
+  RunningOnPipe(Pipe != nullptr) {
+  // pipe is null when execution a analysis.
+  if (RunningOnPipe)
```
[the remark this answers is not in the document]
drblallo: i went for this solution because otherwise it would overcomplicate the invokable.h logic. It can be done, but there is no much point to do so, and surely at some point one will wish to get some run time dependent information from the execution context, such as which step is the current one.

lib/Lift/LiftPipe.cpp
```diff
@@ -26,7 +26,7 @@ using namespace llvm;
 using namespace pipeline;
 using namespace ::revng::pipes;
 
-void Lift::run(Context &Ctx,
+void Lift::run(ExecutionContext &Ctx,
                const BinaryFileContainer &SourceBinary,
                LLVMContainer &TargetsList) {
```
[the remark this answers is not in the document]
drblallo: you mean to rename all Ctx in all pipes to ExecutionCtx?

include/revng/Pipeline/PathTargetBimap.h
drblallo: i did not wanted to expose the find method to make this implementable from outside, maybe i can instead look for a better name than the current one for this method?

lib/Pipeline/Step.cpp
drblallo: i did it this way just because it easier to read.

include/revng/Support/AccessTracker.h
drblallo: we went for this solution because it class is embedded everywhere in the model, so by default we do not wish to track, and when starting to track we always want to clear, and when we clear we always want to track, so i saw no point in exposing two primitives. I can't really find a better name than this, except maybe `clearAndStartTracking`

include/revng/TupleTree/TupleTree.h
drblallo: because this mechanism ties to the optionality of emitting tracking information from jinja generated classes. i did not wanted to leak out from here this feature of the generation, so i hide it in here.

include/revng/Model/VerifyHelper.h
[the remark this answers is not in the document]
aleclearmind: 2. I think implementing the move constructor can solve the RVO problem you're mentioning.
3. `VH.disableTracking` should first check if it has already been disabled. If it has been disabled, it should return a no-op `TrackGuard` (as I suggested, we can have a `bool` in the constructor to make `TrackGuard` a no-op). Otherwise, it returns a regular `TrackGuard`.
   Basically I'm suggesting to move part of the implementation of the `TrackingPushingScope` constructor, into `VH.disableTracking`. I don't see why it can't be done.

include/revng/Pipeline/Context.h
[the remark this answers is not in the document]
aleclearmind: Unless I'm mistaken, the only usage of `stopTracking` is right before `collectReadFields`.

We should probably remove the `stopTracking` method and put its implementation into `collectReadFields`.
Also, we could assert that no one uses globals-related methods after `collectReadFields`.

lib/Pipeline/ExecutionContext.cpp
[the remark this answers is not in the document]
aleclearmind: Add a comment about the fact that, as of now, we could just use `Context`.

include/revng/TupleTree/TupleTree.h
aleclearmind: I'm not sure about the impact of this choice, but at a minimum add a comment explaining this.
It's very surprising.

lib/Model/Binary.cpp
aleclearmind: You have a well self-contained method (`dump`) that pushes stuff but does not pop, how is this correct? o_O

include/revng/Pipeline/ExecutionContext.h
```diff
@@ -0,0 +1,87 @@
...
+// Less formally, you must produce and commit a target at the time.
+//
+// Example:
+//
+// void SomePipeProducingRootElement(const ExecutionContext& Ctx, SomeContainer&
+// Container) {
+//   auto RelevantInformations = readSomeFields(getModelFromContext(Ctx));
+//   produceRoot(RelevantInformations, Container);
+//   Ctx.commit(getGeneratedTarget(Container), Container);
+// }
```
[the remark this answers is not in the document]
drblallo: i rewrote it like 

```
// void SomePipeProducingRootElement(const ExecutionContext& Ctx, SomeContainer&
// Container) {
//   auto GlobalRelevantInformations =
//   readSomeGlobalFields(getModelFromContext(Ctx));
//   for (auto& Function : Container)
//   {
//     Ctx.getContext().pushReadFields();
//     auto NonGlobalRelevantInformations =
//     readSomeFunctionSpecificFields(getModelFromContext(Ctx), Function);
//     transformCurrentFunnction(RelevantInformations,
//     NonGlobalRelevantInformations, Container); Ctx.commit(Container,
//     somehowGetGeneratedTargetOf(Function)); Ctx.getContext().popReadFields();
//   }
// }

```
does this works?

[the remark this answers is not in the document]
aleclearmind: 1. Fix indentation
2. I'd drop 1) the declaration of `GlobalRelevantInformations`, `NonGlobalRelevantInformations` and the call to `transformCurrentFunnction` (just leave some `...`).
4. Don't we have some helper that does the following part for us?
   ```c++
   Ctx.getContext().pushReadFields();
   // ...
   Ctx.commit(Container, somehowGetGeneratedTargetOf(Function)); 
   Ctx.getContext().popReadFields();
   ```

   Do we really want devs to do this each time? There's seems to be a low hanging fruit for a RAII object.
5. Instead of using `somehowGetGeneratedTargetOf`, use `Container.enumerate` which I suppose already returns a `Target`.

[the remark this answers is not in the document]
drblallo: 3) we said in another issue to pull out that helper and similar ones and delay them until a commit that actually uses them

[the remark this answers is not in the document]
aleclearmind: 3. OK. Let's not forget.

aleclearmind (changes requested):

include/revng/TupleTree/TupleTree.h
```diff
@@ -25,11 +25,102 @@
...
+      if (TrackedObject != nullptr) {
+        revng::Tracking::pop(*TrackedObject);
+      }
+    }
+    TrackedObject = nullptr;
+  }
+};
+
+template<typename T, typename TrackableTupleTree>
+class TrackingPushingScope {
```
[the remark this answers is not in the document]
drblallo: nono we agreed that the to classes had different semantics but this could be implemented using the other as primitive.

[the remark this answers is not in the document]
aleclearmind: Right.

However, you're expecting `T` to have `setPushedTracking`. This is very tightly coupled with `VerifyHelper`, I don't think this has a general use case, at least for now. I'd move it into `VerifyHelper.h`.
Also, I'd call this `DisableTrackingOnce` or `DisableTrackingIfEnabled`. `TrackingPushingScope` is confusing.

lib/Model/Binary.cpp
```diff
@@ -22,6 +22,58 @@
 
 using namespace llvm;
 
+static llvm::cl::opt<std::string> StructNameRegex("tracking-debug-struct-name",
```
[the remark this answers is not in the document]
aleclearmind: Nice

lib/Model/Binary.cpp
```diff
@@ -152,6 +161,8 @@ bool VerifyHelper::registerGlobalSymbol(const model::Identifier &Name,
 
 static bool verifyGlobalNamespace(VerifyHelper &VH,
                                   const model::Binary &Model) {
+
+  auto Guard = VH.stopTracking(Model);
```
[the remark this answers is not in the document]
drblallo: suspendTracking, and the class is VerifyHelper::TrackingSuspender works too

aleclearmind:
Merged in 840015f. Thanks.

PR #301, opened 2023-06-20
Add HexDump pipe

kolodziej:


lib/Yield/HexDump.cpp
```diff
@@ -0,0 +1,302 @@
...
+  using type = uint64_t; // assume we compare only .address() within same type,
+                         // epoch and address space
+};
+
+template<>
+struct size_type_of<MetaAddress> {
+  using type = uint64_t;
+};
+
+} // namespace boost::icl
```
aleclearmind, line 52: Do we need all of this stuff? For `interval_set` we don't need anything like this.

If we need this, move it to `include/revng/Support/MetaAddress/IntervalContainers.h`.

lib/Yield/HexDump.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+inline MetaAddress &operator++(MetaAddress &Address) {
+  return Address += 1u;
+}
+
+inline MetaAddress &operator--(MetaAddress &Address) {
+  return Address -= 1u;
+}
```
aleclearmind, line 60: Why this? We already have `MetaAddress:operator+`.

lib/Yield/HexDump.cpp
```diff
@@ -0,0 +1,302 @@
...
+template<>
+struct has_difference<MetaAddress> {
+  using type = has_difference;
+  BOOST_STATIC_CONSTANT(bool, value = true);
+};
+
+template<>
+struct difference_type_of<MetaAddress> {
+  using type = uint64_t; // assume we compare only .address() within same type,
+                         // epoch and address space
```
aleclearmind, line 44: 1. Capital letter
2. Most comments should take the whole line, no indented comments

lib/Yield/HexDump.cpp
```diff
@@ -0,0 +1,302 @@
...
+  return Address += 1u;
+}
+
+inline MetaAddress &operator--(MetaAddress &Address) {
+  return Address -= 1u;
+}
+
+namespace revng::pipes {
+
+static llvm::Error outputHexDump(const TupleTree<model::Binary> &Binary,
```
aleclearmind, line 64: `.cpp` files usually greatly benefit from `using namespace llvm`.

lib/Yield/HexDump.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+  std::error_code ErrorCode;
+  llvm::raw_fd_ostream Output(OutputPath,
+                              ErrorCode,
+                              llvm::sys::fs::CD_CreateAlways);
+  if (!!ErrorCode) {
+    return llvm::make_error<llvm::StringError>("Could not open file "
+                                                 + OutputPath.str(),
+                                               ErrorCode);
+  }
```
aleclearmind, line 80: 1. Drop `!!` if no necessary
2. Can you use `llvm::createStringError`?

lib/Yield/HexDump.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+  FunctionMetadataCache Cache;
```
aleclearmind, line 82: It's not a cache

aleclearmind (changes requested):

lib/Yield/HexDump.cpp
kolodziej: I took this `!!` from some other file in revng project :(

lib/Yield/HexDump.cpp
kolodziej: I think it was required by interval map from boost. I'll verify it again.

lib/Yield/HexDump.cpp
kolodziej: Unfortunately, without this it's not building. We have to keep this. I'll move that part to separate header.

lib/Yield/HexDump.cpp
kolodziej: Yes, we need it.

lib/Yield/HexDump.cpp
kolodziej: 1. boost expects `type` with lower case.

lib/Yield/HexDump.cpp
aleclearmind: 1. Move this inline in `MetaAddress.h`
2. Reuse the existing `operator+`, manipulating directly `Address` violates the semantics of `MetaAddress`

lib/Yield/HexDump.cpp
```diff
@@ -0,0 +1,280 @@
...
+        }
+      }
+    }
+  }
+
+  Output << "<div>\n";
+
+  auto FormatNumber = [](uint64_t Number, unsigned Width = 8) {
+    return FormattedNumber(Number, 0, Width, true, false, false);
+  };
```
aleclearmind, line 120: Push to `static` function.

lib/Yield/HexDump.cpp
```diff
@@ -0,0 +1,280 @@
...
+    uint64_t Offset = Segment.StartOffset();
+    MetaAddress Address = Segment.StartAddress();
+    uint64_t Size = Segment.VirtualSize();
+
+    CurrentAddress = Address;
+
+    std::optional<llvm::ArrayRef<uint8_t>>
+      SegmentBinary = BinaryView.getByAddress(Address, Size);
+
+    const size_t BytesInLine = 16;
```
aleclearmind, line 137: Hoist to the top of the file.

lib/Yield/HexDump.cpp
```diff
@@ -0,0 +1,280 @@
...
+  auto FormatNumber = [](uint64_t Number, unsigned Width = 8) {
+    return FormattedNumber(Number, 0, Width, true, false, false);
+  };
+
+  MetaAddress CurrentAddress;
+  Map::const_iterator Current = Instructions.begin();
+  Map::const_iterator Next = std::next(Current);
+  const Map::const_iterator End = Instructions.end();
+
+  for (const model::Segment &Segment : Binary->Segments()) {
```
aleclearmind, line 127: Use `BinyarView.segments()` and be happy

aleclearmind (changes requested):

lib/Yield/HexDump.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+    const size_t BytesInLine = 16;
+
+    size_t Counter = 0;
+
+    std::set<ptml::Tag> PreviousAddresses;
+    bool SpanIsOpen = false;
+    if (SegmentBinary.has_value()) {
+      SmallString<16> PrintableChars;
+      for (size_t Index = 0; Index < SegmentBinary->size(); ++Index) {
```
aleclearmind, line 145: Please thoroughly comment all of this loop, it's rather delicate and not obvious to follow.

lib/Yield/HexDump.cpp
```diff
@@ -0,0 +1,280 @@
...
+            Output << Tag.PTMLTag.close();
+          }
+        }
+
+        if (Counter == BytesInLine) {
+          // Output ASCII representation at the end of the line
+          std::string Temp;
+          raw_string_ostream Printable(Temp);
+          printHTMLEscaped(PrintableChars, Printable);
+          Output << "   | " << Printable.str() << " |\n";
```
aleclearmind, line 204: I never noticed `raw_string_ostream` has `.str`, I always ensure `raw_string_ostream` is destructed and then use the backing `std::string`. Nice!

lib/Yield/HexDump.cpp
```diff
@@ -0,0 +1,280 @@
+/// \file HexDump.cpp
+
```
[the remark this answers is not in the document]
kolodziej: I think `\brief` is forbidden (commit ce6d9a6dbf251f14398839aec3a657fb330dbd99)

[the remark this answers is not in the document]
kolodziej: I see only `\file` in most files

lib/Support/IRHelpers.cpp
```diff
@@ -336,27 +336,28 @@ void dumpUsers(llvm::Value *V) {
...
+    for (const BasicBlock *Predecessor : predecessors(BB)) {
       if (!Visited.contains(Predecessor))
         rc_recur findJumpTarget(Result, Predecessor, Visited);
+    }
   }
 
   rc_return;
 }
 
-llvm::BasicBlock *getJumpTargetBlock(llvm::BasicBlock *BB) {
```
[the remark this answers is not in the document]
kolodziej: I described only `getJumpTarget` and not other things, because `getJumpTarget` depends on others, so I had to add `const` in these places to make it compatible. I'll add this to description in commit message.

include/revng/Support/MetaAddress/IntervalContainers.h
```diff
@@ -0,0 +1,51 @@
...
+template<>
+struct has_difference<MetaAddress> {
+  using type = has_difference;
+  BOOST_STATIC_CONSTANT(bool, value = true);
+};
+
+template<>
+struct difference_type_of<MetaAddress> {
+  // assume we compare only .address() within same type, epoch and address space
+  using type = uint64_t;
```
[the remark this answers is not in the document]
kolodziej: Why is it returning `std::optional<uint64_t>`?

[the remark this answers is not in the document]
kolodziej: But it looks like for `MetaAddress.Address` always have some value. It's primitive type. So `operator-(MetaAddress)` will always have some value, because [here](https://github.com/revng/revng/blob/aa74e194bcdac0849fa47c4f9c4d8524c0781369/include/revng/Support/MetaAddress.h#L646C23-L646C23) we use `OverflowSafeInt` and it's `operator-` ([here](https://github.com/revng/revng/blob/aa74e194bcdac0849fa47c4f9c4d8524c0781369/include/revng/Support/OverflowSafeInt.h#L42) will always return optional with value). We construct `OverflowSafeInt` with primitive value, not optional, so it's `Value` is set for both addresses (MetaAddresses), so the result should be safely unwrapped from optional. Can't we add an assertion on `has_value` there and return bare `uint64_t`?

lib/Yield/HexDump.cpp
```diff
@@ -0,0 +1,280 @@
...
+  std::string LocationString;
+  ptml::Tag PTMLTag;
+
+  bool operator==(const AddressTag &B) const {
+    return LocationString == B.LocationString;
+  }
+
+  bool operator<(const AddressTag &B) const {
+    return LocationString < B.LocationString;
+  }
```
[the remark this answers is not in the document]
kolodziej: Is `ptml::Tag` comparable? I think we don't have operators for it and it contains field that are not comparable (`StringMap`). Am I right or I'm missing something?

include/revng/Support/MetaAddress/IntervalContainers.h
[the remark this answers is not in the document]
aleclearmind: If you do `MetaAddress(100) - MetaAddress(101)` it should return a nullopt.

lib/Yield/HexDump.cpp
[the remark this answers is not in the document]
aleclearmind: Ah, uh, new stuff, not used to this. OK then.

lib/Support/IRHelpers.cpp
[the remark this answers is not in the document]
aleclearmind: 👍🏿

lib/Yield/HexDump.cpp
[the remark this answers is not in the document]
kolodziej: Yes, I can postpone, but later I have to keep track of `ptml::Tag` instances attached to location string, although location string is already inside `ptml::Tag`. Without keeping `ptml::Tag` instances somewhere, I will not be able to call `ptml::Tag::close()`.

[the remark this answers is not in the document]
aleclearmind: You need the `interval_map` only while collecting the locations of each range. Once you're done with that you can turn the `interval_map` into some other data structure, e.g., `std::map<MetaAddress, ptml::Tag>`. At that point, the `MetaAddress` won't be ambiguous.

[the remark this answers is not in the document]
kolodziej: Then it should be either `std::multimap` or map should keep collection of `ptml::Tag`s starting at given address. Other aspect is where tag has to be closed. Should I wrap every byte separately?

include/revng/Support/MetaAddress/IntervalContainers.h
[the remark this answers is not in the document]
kolodziej: Correct, I see now that `OverflowSafeInt::sub` can return `std::nullopt`. Could you suggest any solution? Should I use some other key type for interval_map?

lib/Yield/HexDump.cpp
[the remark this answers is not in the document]
kolodziej: I think now it should be fine.

include/revng/Support/MetaAddress/IntervalContainers.h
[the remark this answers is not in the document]
aleclearmind: @kolodziej is this fixed?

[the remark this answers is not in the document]
kolodziej: Done

aleclearmind:
Merged in db6c6b1. Thanks!

PR #302, opened 2023-06-23
Improved Navigation & Look

mrjackv:


python/revng/api/manager.py
```diff
@@ -117,7 +117,13 @@ def _produce_target(
             # TODO: we really should be able to provide a detailed error here
             raise RevngException("Failed to produce targets")
 
-        return {t.serialize(): t.extract() for t in targets}
+        result = {}
+        for target in targets:
+            extracted_target = target.extract()
+            if extracted_target is None:
+                raise RevngException("Target extraction failed")
```
aleclearmind, line 124: Can we have a more informative error message?
Maybe add the target name.

What would help you if you were to read an error message about this?

typescript/tuple_tree.ts
```diff
@@ -39,6 +39,22 @@ export function safeParseInt(str?: string): number | undefined {
...
 
+/**
+ * Parses a bigint safely, returning undefined if the number cannot be represented as a decimal
+ * string
+ */
+export function safeParseBigInt(str?: string): bigint | undefined {
+    if (str !== undefined) {
+        try {
+            return BigInt(str);
+        } catch (err) {
```
aleclearmind, line 50: Can we limit what we catch here?

typescript/model.ts
```diff
@@ -29,33 +48,56 @@ const metaAddressRegex = new RegExp(
...
+            this.type = type || "Invalid";
+            this.address = stringOrAddress;
+            this.epoch = epoch || 0n;
+            this.addressSpace = addressSpace || 0n;
+        }
+    }
+
+    addOffset(offset: bigint): MetaAddress {
+        if (this.type === "Invalid") {
+            return new MetaAddress(":Invalid");
```
aleclearmind, line 84: Most safe to throw an exception

typescript/model.ts
```diff
@@ -29,33 +48,56 @@ const metaAddressRegex = new RegExp(
...
+        }
+        const newAddress = this.address + offset;
+        if (newAddress > maxValue(this.type)) {
+            return new MetaAddress(":Invalid");
```
aleclearmind, line 88: In `MetaAddress.h`, do we go to `Invalid` or wrap around?

typescript/model.ts
```diff
@@ -29,33 +48,56 @@ const metaAddressRegex = new RegExp(
...
+                this.type = "Invalid";
+            }
         } else {
-            this.type = "Invalid";
+            this.type = type || "Invalid";
+            this.address = stringOrAddress;
+            this.epoch = epoch || 0n;
+            this.addressSpace = addressSpace || 0n;
+        }
+    }
```
aleclearmind, line 80: Can we add validation of address, i.e., if it's too big or negative decay into `Invalid` (or throw an exception).

aleclearmind:
@drblallo can you review "PipelineManager: add produceTargets with checks "?

typescript/model.ts
mrjackv: That's not the behaviour in `MetaAddress.h`: https://github.com/revng/revng/blob/develop/include/revng/Support/MetaAddress.h#L627

typescript/tuple_tree.ts
```diff
@@ -39,6 +39,22 @@ export function safeParseInt(str?: string): number | undefined {
     }
 }
 
+/**
+ * Parses a bigint safely, returning undefined if the number cannot be represented as a decimal
+ * string
+ */
+export function safeParseBigInt(str?: string): bigint | undefined {
+    if (str !== undefined) {
```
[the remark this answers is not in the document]
mrjackv: It sort-of hand-tailored to the use in `model.ts` where we do:
```
this.epoch = safeParseBigInt(match.groups["Epoch"]) || 0n;
```

typescript/model.ts
aleclearmind: OK, then.

typescript/tuple_tree.ts
mrjackv: `BigInt` will throw `TypeError`, `RangeError` and `SyntaxError` which are 3 out of the 8 errors that Javascript natively supports. The other 5, in this context, cannot happen.

typescript/tuple_tree.ts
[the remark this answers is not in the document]
aleclearmind: Can't you push `|| "0"` in?

python/revng/api/manager.py
mrjackv: Added the serialized target to the exception message

typescript/tuple_tree.ts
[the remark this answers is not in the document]
mrjackv: Having looked a bit closer to the regex that matches the `MetaAddress`, I've concluded that `BigInt` can be used directly because the content of the regex is very strict. I've updated the commit accordingly.

typescript/tuple_tree.ts
mrjackv: Removed `safeParseBigInt`

typescript/model.ts
mrjackv: Added

lib/Pipes/PipelineManager.cpp
```diff
@@ -448,3 +448,35 @@ PipelineManager::invalidateFromDiff(const llvm::StringRef Name,
   // TODO: once invalidations are working, return `Map` instead of this
   return invalidateAllPossibleTargets();
 }
+
+llvm::Error PipelineManager::produceTargets(const llvm::StringRef StepName,
+                                            const Container &TheContainer,
+                                            const TargetsList &List,
+                                            llvm::raw_ostream &Output) {
+  if (CurrentState.count(StepName) == 0)
```
drblallo, line 456: CurrentState.count(StepName) == 0 implies that no target can be produced is the given Step, not that no such step exists. 

while the code in the commit kind of works, the error message will some time be confusing. Current state is a table that contains the list of produced targets, it does not wishes to properly keep track of whch steps exists (even tough maybe the current implementation may happen to guarantee it). You must interrogate the runner to check if StepName exists, not CurrentState.

lib/Pipes/PipelineManager.cpp
```diff
@@ -448,3 +448,35 @@ PipelineManager::invalidateFromDiff(const llvm::StringRef Name,
...
+    return llvm::createStringError(llvm::inconvertibleErrorCode(),
+                                   "Step not found");
+
+  if (!CurrentState[StepName].contains(TheContainer.first()))
```
drblallo, line 460: same issue as before, if no target can be produced for the given container, there is no guarantee the container will apear at all in the state.

lib/Pipes/PipelineManager.cpp
```diff
@@ -448,3 +448,35 @@ PipelineManager::invalidateFromDiff(const llvm::StringRef Name,
...
+  ContainerToTargetsMap Targets;
+  for (const pipeline::Target &Target : List)
+    Targets[TheContainer.second->name()].push_back(Target);
+
+  if (auto Error = getRunner().run(StepName, Targets); Error)
+    return Error;
+
+  const auto &ToFilter = Targets.at(TheContainer.second->name());
+  const auto &Cloned = TheContainer.second->cloneFiltered(ToFilter);
+  llvm::cantFail(Cloned->serialize(Output));
```
drblallo, line 480: i would keep this part of the function in the pipeline C, there is no reason why the production routine should have a opinion regarding how to return what has been requested, in particular if it returns them by serializing them into a stream

lib/Pipes/PipelineManager.cpp
mrjackv: Resolved as explained in the comment above

lib/Pipes/PipelineManager.cpp
mrjackv: As discussed IRL, split the functionality in `materializeTargets` and `produceTargets`

aleclearmind:
Merged in abff564. Thanks!

PR #305, opened 2023-07-11
Drop requirement of prefixing analyses list in CL

drblallo:


aleclearmind:
Merged in 6707a20. Thanks!

PR #309, opened 2023-07-19
Implement S3 storage backend

mrjackv:


include/revng/Pipes/FileContainer.h
```diff
@@ -116,32 +109,41 @@ class FileContainer
...
       return llvm::Error::success();
     }
+
     getOrCreatePath();
-    return llvm::errorCodeToError(llvm::sys::fs::copy_file(Path, this->Path));
+    auto MaybeClient = revng::StorageClient::fromURL(this->Path);
+    if (!MaybeClient)
+      return MaybeClient.takeError();
+
+    revng::Path DestinationPath(MaybeClient->get());
```
aleclearmind, line 145: 1. I'm confused, you're not using `this->Path` here.
2. While I like the cross-client copy infrastructure, I'm not sure it makes a lot of sense to create a `StorageClient` every time. Maybe it would be simpler to just have a `writeFromLocalFile` or `writeFromStream` method.

include/revng/Storage/Path.h
```diff
@@ -0,0 +1,94 @@
...
+      return MaybeWritableFile.takeError();
+
+    auto &ReadableFile = MaybeReadableFile.get();
+    auto &WritableFile = MaybeWritableFile.get();
+    WritableFile.getOS() << ReadableFile.getBuffer().getBuffer();
+
+    if (auto Error = WritableFile.close(); Error)
+      return Error;
+
+    return llvm::Error::success();
```
aleclearmind, line 82: Can you do the following?
```c++
return WritableFile.close();
```

include/revng/Pipes/PipelineManager.h
```diff
@@ -88,12 +88,12 @@ class PipelineManager {
...
   /// Triggers the full serialization of every step and every container to the
   /// the specified DirPath or the Execution directory if omitted.
-  llvm::Error storeToDisk(llvm::StringRef DirPath = llvm::StringRef());
+  llvm::Error storeToDisk(revng::Path *DirPath = nullptr);
 
   /// Trigger the serialization of a single Step to the specified DirPath or
   /// the Execution directory if omitted.
   llvm::Error storeStepToDisk(llvm::StringRef StepName,
-                              llvm::StringRef DirPath = llvm::StringRef());
+                              revng::Path *DirPath = nullptr);
```
aleclearmind, line 96: Does anyone actually use this argument? If not, drop it.

include/revng/Pipes/ToolCLOptions.h
```diff
@@ -48,7 +48,12 @@ class ToolCLOptions {
                             PipelineManager &Manager) {
     const auto &Name = ModelGlobalName;
     auto *Model(cantFail(Manager.context().getGlobal<ModelGlobal>(Name)));
-    return Model->loadFromDisk(ModelOverride);
+
+    revng_assert(!isURLS3(ModelOverride));
```
aleclearmind, line 52: 1. Why? I see many places using this function that should be storage-backend agnostic.
2. Rename to `isS3URL`.

aleclearmind (changes requested):

include/revng/Storage/Path.h
```diff
@@ -0,0 +1,94 @@
...
+
+namespace revng {
+
+class Path {
+private:
+  revng::StorageClient *Client;
+  std::string StoredPath;
+
+public:
+  Path(StorageClient *Client, std::string StoredPath = "") :
```
aleclearmind, line 21: Drop `= ""`

include/revng/Pipes/PipelineManager.h
mrjackv: It's used by `PipelineC.cpp` in `rp_manager_save`. This was added to support the `FIFO` synchronization. I'd drop it, but I'm double checking if you see some other use for it.

include/revng/Pipes/FileContainer.h
mrjackv: Fixed by the introduction of `Path::fromLocalStorage`

include/revng/Storage/Path.h
mrjackv: Done

include/revng/Pipes/PipelineManager.h
```diff
@@ -174,7 +174,7 @@ class PipelineManager {
   /// be produced by the pipeline in the current state
   void writeAllPossibleTargets(llvm::raw_ostream &OS) const;
 
-  llvm::StringRef executionDirectory() const { return ExecutionDirectory; }
+  revng::Path executionDirectory() const { return ExecutionDirectory; }
```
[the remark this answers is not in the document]
mrjackv: Done

include/revng/Pipes/ToolCLOptions.h
mrjackv: 1. It was to prevent "surprising" accidental usage by command-line tools, but with the introduction of `Path::fromLocalStorage` this is not needed anymore
2. Done

include/revng/Pipes/ToolCLOptions.h
```diff
@@ -48,7 +48,12 @@ class ToolCLOptions {
...
+    auto MaybeClient = revng::StorageClient::fromURL(ModelOverride);
+    if (!MaybeClient)
+      return MaybeClient.takeError();
+    return Model->loadFromDisk({ MaybeClient->get() });
```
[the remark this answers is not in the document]
mrjackv: 1. Done
2. Implemented with `Path::fromLocalStorage` as discussed

include/revng/Storage/Path.h
mrjackv: Done

include/revng/Storage/StorageClient.h
```diff
@@ -0,0 +1,114 @@
...
+  StorageClient(const StorageClient &Other) = delete;
+  StorageClient &operator=(const StorageClient &Other) = delete;
+  StorageClient(const StorageClient &&Other) = delete;
+  StorageClient &operator=(StorageClient &&Other) = delete;
+
+public:
+  static llvm::Expected<std::unique_ptr<StorageClient>>
+  fromURL(const llvm::StringRef URL);
+
+  virtual std::string toString() const debug_function = 0;
```
aleclearmind, line 92: What does `toString` return? The root URL from which we can rebuild this object using `fromURL`?

If not, it it's just for debug purpose, use `dump` and write to `dbg`.

lib/Storage/S3StorageClient.cpp
```diff
@@ -0,0 +1,379 @@
...
+    [this, TempFile, PathCopy, Encoding]() -> llvm::Error {
+      Aws::S3::Model::PutObjectRequest Request;
+      Request.SetBucket(Bucket);
+      std::string NewFilename = generateNewFilename(PathCopy);
+      Request.SetKey(resolvePath(SubPath, NewFilename));
+
+      if (Encoding == ContentEncoding::Gzip)
+        Request.SetContentEncoding("gzip");
+
+      auto File = Aws::MakeShared<Aws::FStream>("ALLOC_TAG",
```
aleclearmind, line 333: What is `ALLOC_TAG`?

lib/Storage/S3StorageClient.cpp
```diff
@@ -0,0 +1,379 @@
...
+  }
+};
+
+namespace revng::detail {
+
+namespace {
+using Aws::Utils::Logging::FormattedLogSystem;
+using Aws::Utils::Logging::LogLevel;
+
+static Logger<> AWSLogger("aws");
```
aleclearmind, line 50: In `.cpp` files we usually just go for `Log` if it's the default for the `.cpp` file.

lib/Storage/S3StorageClient.cpp
```diff
@@ -0,0 +1,379 @@
...
+public:
+  LoggerSystem(LogLevel LogLevel) : FormattedLogSystem(LogLevel) {}
+  ~LoggerSystem() override = default;
+  void Flush() override { AWSLogger.flush(); }
+
+  void ProcessFormattedStatement(Aws::String &&Statement) override {
+    // This function needs to be thread-safe, hence the lock_guard
+    std::lock_guard Guard(Mutex);
+    revng_log(AWSLogger, Statement);
+  }
```
aleclearmind, line 65: Despite the lock, this might still trash `stderr` due to other threads writing there freely.

lib/Storage/S3StorageClient.cpp
```diff
@@ -0,0 +1,379 @@
...
+};
+
+class SDKInitImpl {
+private:
+  Aws::SDKOptions Options;
+
+public:
+  SDKInitImpl() : Options() {
+    // TODO: unix-specific, add #ifdef _WIN32 when needed
+    setenv("AWS_EC2_METADATA_DISABLED", "true", 1);
```
aleclearmind, line 75: What is this? Do we really have to `setenv`?

lib/Storage/S3StorageClient.cpp
```diff
@@ -0,0 +1,379 @@
...
+    Options.loggingOptions.logLevel = Level;
+    Options.loggingOptions.logger_create_fn = [Level]() {
+      return Aws::MakeShared<LoggerSystem>("ALLOC_TAG", Level);
+    };
+    Aws::InitAPI(Options);
+    OnQuit->add([this] { Aws::ShutdownAPI(Options); });
+  }
+};
+
+static std::optional<SDKInitImpl> SDKInit;
```
aleclearmind, line 86: Don't use `static` in anonymous namespaces

lib/Storage/S3StorageClient.cpp
```diff
@@ -0,0 +1,379 @@
...
+  void Flush() override { AWSLogger.flush(); }
+
+  void ProcessFormattedStatement(Aws::String &&Statement) override {
+    // This function needs to be thread-safe, hence the lock_guard
+    std::lock_guard Guard(Mutex);
+    revng_log(AWSLogger, Statement);
+  }
+};
+
+class SDKInitImpl {
```
aleclearmind, line 68: Using a `class` for this kind of things makes sense if you use the destructor.

You could turn `std::optional<SDKInitImpl> SDKInit` into a simple `bool SDKIsInitialized = false` and copy the body of the constructor where you're currently checking the optional.

If you go this route, you also probably need to capture `Options` by copy or make a global variable out of it.

Give it a think.

lib/Storage/S3StorageClient.cpp
```diff
@@ -0,0 +1,379 @@
...
+llvm::Expected<bool> S3StorageClient::exists(const llvm::StringRef Path) {
+  return FilenameMap.count(Path) != 0;
+}
+
+llvm::Error S3StorageClient::createDirectory(const llvm::StringRef Path) {
+  return llvm::Error::success();
+}
+
+llvm::Error S3StorageClient::remove(const llvm::StringRef Path) {
+  FilenameMap.erase(Path);
```
aleclearmind, line 240: Assert it was there, IIRC `erase` fails silently.

include/revng/Storage/S3StorageClient.h
```diff
@@ -0,0 +1,51 @@
...
+#include "aws/s3/S3Client.h"
+
+namespace revng::detail {
+
+class S3StorageClient : public StorageClient {
+private:
+  Aws::S3::S3Client Client;
+  std::string Bucket;
+  std::string SubPath;
+  std::string SerializedString;
```
aleclearmind, line 21: What is this? The base URL?

lib/Storage/S3StorageClient.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+  llvm::SmallString<128> TempFile;
+  std::error_code EC = createTemporaryFile("revng-s3-storage-", "", TempFile);
+  if (EC)
+    return llvm::createStringError(EC, "Could not create temporary file");
+
+  std::ofstream OS(TempFile.str().str(), std::ios::binary);
+  if (OS.fail())
+    return llvm::createStringError(llvm::inconvertibleErrorCode(),
+                                   "Could not open temporary file");
```
aleclearmind, line 283: Use `{ }` for multi-line (even if single-statement) `if` bodies.

lib/Storage/S3StorageClient.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+  constexpr std::size_t BufSize = 4096;
+  char Buffer[BufSize];
```
aleclearmind, line 286: Don't put things so big on the stack. Use a `std::vector`.

include/revng/Storage/DummyStorageClient.h
```diff
@@ -0,0 +1,41 @@
...
+//
+
+#include "llvm/ADT/StringSet.h"
+
+#include "revng/Storage/StorageClient.h"
+#include "revng/Support/Debug.h"
+
+namespace revng::detail {
+
+class DummyStorageClient : public StorageClient {
```
aleclearmind, line 14: Docs, I expect it will not do much, but not much *how*?

include/revng/Storage/Path.h
```diff
@@ -0,0 +1,93 @@
...
+#include "llvm/Support/Path.h"
+
+#include "revng/Storage/StorageClient.h"
+#include "revng/Support/Debug.h"
+
+namespace revng {
+
+class Path {
+private:
+  revng::StorageClient *Client;
```
aleclearmind, line 17: As a policy, always initialize primitive types and pointers inline. Add `= nullptr`.

include/revng/Storage/Path.h
```diff
@@ -0,0 +1,93 @@
...
+  std::string StoredPath;
+
+public:
+  explicit Path(StorageClient *Client, const llvm::StringRef StoredPath) :
+    Client(Client), StoredPath(StoredPath.str()){};
+
+  static Path fromLocalStorage(llvm::StringRef Path);
+
+public:
+  std::string toString() const debug_function {
```
aleclearmind, line 27: This looks like `.dump`, not `toString`.

include/revng/Storage/StorageClient.h
```diff
@@ -0,0 +1,114 @@
...
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/MemoryBuffer.h"
+#include "llvm/Support/Path.h"
+
+#include "revng/Support/Assert.h"
+#include "revng/Support/Debug.h"
+
+namespace revng {
+
+class WritableFile {
```
aleclearmind, line 20: Given that you created `librevngStorage`, move `WritableFile` and `ReadableFile` to their own file.

aleclearmind (changes requested):

lib/Storage/S3StorageClient.cpp
mrjackv: > What is this?

When running the SDK inside an aws EC2 instance (a VM), there's a mechanism [described here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html) where the SDK, at init time, does some HTTP GETs to retrieve the credentials of the VM "user". This mechanism can only be disabled via the environment variable above

lib/Storage/S3StorageClient.cpp
mrjackv: OK, is there some reference on why this is bad practice?

include/revng/Storage/StorageClient.h
```diff
@@ -0,0 +1,114 @@
...
+  WritableFile(WritableFile &&Other) = default;
+  WritableFile &operator=(WritableFile &&Other) = default;
+  ~WritableFile() {
+    if (OS != nullptr)
+      Aborter();
+  }
+
+  llvm::raw_pwrite_stream &getOS() { return *OS; }
+
+  llvm::Error close() {
```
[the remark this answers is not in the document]
aleclearmind: `keep` is for when the file is finalized, I don't think anyone is used to write stuff after invoking `keep`.

If you prefer, we can call this guy and `StorageClient::finalizeWrites` `commit`, and in the doc we can say "remember to always call `commit` or it will be like you did nothing!" (even if that's not true for non-S3).

lib/Storage/StorageClient.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+#include "revng/Storage/DummyStorageClient.h"
+#include "revng/Storage/LocalStorageClient.h"
+#include "revng/Storage/S3StorageClient.h"
+#include "revng/Storage/StorageClient.h"
+
+llvm::Expected<std::unique_ptr<revng::StorageClient>>
+revng::StorageClient::fromURL(const llvm::StringRef URL) {
+  if (URL.empty()) {
+    return std::make_unique<revng::detail::DummyStorageClient>();
```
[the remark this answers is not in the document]
aleclearmind: Mmmh, `DummyStorageClient` should be used for testing purposes only.

Writing an LLVM IR module to `/dev/null` it's a huge waste of time, if we can avoid it.

lib/Storage/S3StorageClient.cpp
aleclearmind: By any chance, maybe the flush method is called from the main thread?

Alternatively, we could buffer them all and have some mechanism to flush them every time we're done with AWS objects, but that would be very sad.

I don't know, put a TODO and let's leave it like this, let's see what happens.

lib/Storage/S3StorageClient.cpp
aleclearmind: This seems more like an hitman/orchestra concern to me. Shall we force it in orchestra?

What about this instead?
```c
// TODO: unix-specific, add #ifdef _WIN32 when needed
```

include/revng/Storage/S3StorageClient.h
aleclearmind: `RedactedURL`?

lib/Storage/S3StorageClient.cpp
aleclearmind: Things in anonymous namespace are automatically `static`.

lib/Storage/S3StorageClient.cpp
aleclearmind: > what if the container on K8s doesn't set the env?

It will be set by the `revng` wrapper script.

It's not a concern of `revng/revng` if we're running in a AWS EC2 or not, it's not the right place to set it.

include/revng/Support/GzipTarFile.h
```diff
@@ -0,0 +1,71 @@
...
+  std::string Filename;
+  llvm::SmallVector<char> Data;
+};
+
+class GzipTarReader {
+private:
+  archive *Archive;
+
+public:
+  GzipTarReader(const char *Begin, size_t Size);
```
aleclearmind, line 57: Use `ArrayRef`

python/revng/daemon/event_manager.py
```diff
@@ -39,98 +26,27 @@ async def inner(*args, **kwargs):
...
-
-
-EventManager.add_class_handler(FIFOListener())
+        self.running = True
+        self.next_save: float | None = None
+
+    def run(self):
+        while self.running:
+            if self.next_save is not None and self.next_save < time.time():
+                self.manager.save()
```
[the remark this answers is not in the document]
mrjackv: > Aren't we saving every time we change something?

That would be ultra-lagging: it takes a good 20s to `storeToDisk` (on a local setup, I'd say 2x conservatively when on the cloud). As [described in the pad](https://pad.rev.ng/99Bh2R02RNuL8uNugMv3-Q) I'd tweak a few things in pipeline before saving on every change.

> Why is this required?

So that saves are done only when the daemon is idling, so that the user does not "feel the lag"

lib/Support/CMakeLists.txt
```diff
@@ -23,11 +23,21 @@ revng_add_library_internal(
...
+find_package(Boost REQUIRED COMPONENTS iostreams)
+include(FindLibArchive)
+
+if(NOT LibArchive_FOUND)
+  message(SEND_ERROR "libarchive not found")
+endif()
+
+target_include_directories(revngSupport PUBLIC ${Boost_INCLUDE_DIRS})
+target_link_libraries(revngSupport ${LibArchive_LIBRARIES} ${Boost_LIBRARIES}
+                      ${LLVM_LIBRARIES})
```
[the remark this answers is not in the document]
mrjackv: > Iostreams is header only.

```
$ l root/lib/libboost_iostreams.so.1.81.0 
-rwxr-xr-x 1 user user 111K Aug  1 10:28 root/lib/libboost_iostreams.so.1.81.0
```

> Also `target_include_directories` should not be strictly required

I'll check if it's needed

include/revng/Support/GzipTarFile.h
```diff
@@ -0,0 +1,71 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "llvm/Support/Error.h"
+#include "llvm/Support/MemoryBuffer.h"
+#include "llvm/Support/raw_ostream.h"
+
+#include "revng/Support/Assert.h"
+
+struct archive;
```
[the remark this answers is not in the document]
mrjackv: > libarchive is a C library?

Yes

> wrap C stuff in extern "C"

I didn't because [libarchive already does it](https://github.com/libarchive/libarchive/blob/master/libarchive/archive.h#L144)

python/revng/daemon/event_manager.py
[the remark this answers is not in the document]
mrjackv: > What about after an analysis?

Both `runAnalysis` and `runAnalysesList` save everything (model and `Lift/module.ll`) if the diff map is not empty

lib/Support/CMakeLists.txt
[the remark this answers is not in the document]
aleclearmind: Ah shit. I googled this but I misread.

I really don't want to link against boost.
We're currently building boost with exceptions and RTTI and that does not play well with our way of building things.

I think we don't really need a new dependency just for gzip.
We have libarchive, let's use that one.

Sorry, but I misread. If I read correctly, I would have refused iostreams in the first place.

include/revng/Support/GzipTarFile.h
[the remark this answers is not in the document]
aleclearmind: I know, I'm talking about this foward declaration.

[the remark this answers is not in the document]
mrjackv: Ah sorry, misread

lib/Support/CMakeLists.txt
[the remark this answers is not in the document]
aleclearmind: > We can't use that one, we need to have finer-grain control

Like what?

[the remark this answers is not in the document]
mrjackv: > Like what?

When writing the tar archive we need to:
* Write a compressed header
* Write the compressed file contents (in a separate gzip stream)
* Write the compressed trailing passing (also in a separate gzip stream)

libarchive does not offer this functionality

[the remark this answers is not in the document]
mrjackv: In theory we could, but then we'd have to convert thing into `libarchive`-specific data structures which isn't great ([example](https://github.com/libarchive/libarchive/blob/dff77a6783e5de548ceb265aad1f7a74261be390/libarchive/archive_read_support_filter_gzip.c#L104)) (they do weird vtable stuff in c)

I'm looking at what libarchive does for streaming compression, which seems to be to use `deflateInit2` from zlib. I think we can use that instead of boost

README.md
```diff
@@ -60,6 +60,7 @@ armv7a-hardfloat-linux-uclibceabi-gcc \
 
 # Translate
 ./bin/revng translate hello.arm
+chmod +x hello.arm.translated
```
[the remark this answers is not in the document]
mrjackv: Done

include/revng/Pipes/PipelineManager.h
mrjackv: Dropped it and the ancilliary functions in PipelineC

share/revng/test/tests/storage.sh
```diff
@@ -0,0 +1,23 @@
...
+  --resume="s3://S3RVER:S3RVER@region+127.0.0.1:$S3_PORT/test/project-test-dir" \
+  --analyses-list=revng-initial-auto-analysis \
+  -o /dev/null \
+  YieldAssembly "$1"
+
+INDEX_FILE="$2/test/project-test-dir/index.yml._S3rver_object"
+test -f "$INDEX_FILE"
+INPUT_ID="$(grep -Po '(?<=^begin/input:).*' "$INDEX_FILE" | tr -d ' ' | tr -d "'")"
+INPUT_FILE="$2/test/project-test-dir/$INPUT_ID._S3rver_object"
+[[ $(sha256sum "$INPUT_FILE" | cut -d' ' -f1) == "$(sha256sum "$1" | cut -d' ' -f1)" ]]
```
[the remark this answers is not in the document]
mrjackv: Done

include/revng/Storage/StorageClient.h
```diff
@@ -0,0 +1,114 @@
...
+
+  virtual llvm::Expected<WritableFile>
+  getWritableFile(const llvm::StringRef Path, ContentEncoding Encoding) = 0;
+
+  virtual llvm::Error finalizeWrites() = 0;
+};
+
+inline bool isS3URL(llvm::StringRef Input) {
+  return Input.starts_with("s3://") or Input.starts_with("s3s://");
+}
```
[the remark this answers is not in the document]
mrjackv: Done

include/revng/Storage/StorageClient.h
mrjackv: Renamed to `dump` and added the `dump(T &OS)` method as well (as specified in the pad)

lib/Storage/S3StorageClient.cpp
```diff
@@ -0,0 +1,379 @@
...
+      Request.SetBody(File);
+      Aws::S3::Model::PutObjectOutcome Result = Client.PutObject(Request);
+      if (not Result.IsSuccess())
+        return toError(Result);
+
+      FilenameMap[PathCopy] = NewFilename;
+      return llvm::Error::success();
+    },
+    [TempFile]() { llvm::sys::fs::remove(TempFile); }
+  };
```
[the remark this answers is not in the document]
mrjackv: Converted `WritableFile` and `ReadableFile` into abstract classes
Also tweaked `TemporaryFile` since it had a couple of shortcomings

include/revng/Storage/StorageClient.h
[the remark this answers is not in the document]
mrjackv: Changed to `WritableFile::os` and `WritableFile::commit` and `ReadableFile::buffer` (for consistency)
Also renamed `StorageClient::finalizeWrites` to `StorageClient::commit`

include/revng/Storage/Path.h
```diff
@@ -0,0 +1,93 @@
...
+    return WritableFile.close();
+  }
+
+  llvm::Error renameTo(const revng::Path &Destination) const {
+    if (auto Error = copyTo(Destination); Error)
+      return Error;
+    return remove();
+  }
+
+  llvm::Error finalizeWrites() const { return Client->finalizeWrites(); }
```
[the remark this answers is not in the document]
mrjackv: Dropped it, now `PipelineManager` calls `StorageClient::commit` directly

include/revng/Storage/StorageClient.h
```diff
@@ -0,0 +1,114 @@
...
+  virtual llvm::Error copy(const llvm::StringRef Source,
+                           const llvm::StringRef Destination) = 0;
+
+  virtual llvm::Expected<ReadableFile>
+  getReadableFile(const llvm::StringRef Path) = 0;
+
+  virtual llvm::Expected<WritableFile>
+  getWritableFile(const llvm::StringRef Path, ContentEncoding Encoding) = 0;
+
+  virtual llvm::Error finalizeWrites() = 0;
```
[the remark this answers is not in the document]
mrjackv: 1. Done
2. Also done

lib/Storage/S3StorageClient.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+#include "aws/core/Aws.h"
+#include "aws/core/auth/AWSCredentials.h"
+#include "aws/core/utils/logging/FormattedLogSystem.h"
+#include "aws/s3/S3Client.h"
+#include "aws/s3/model/CopyObjectRequest.h"
+#include "aws/s3/model/DeleteObjectRequest.h"
+#include "aws/s3/model/GetObjectRequest.h"
+#include "aws/s3/model/HeadObjectRequest.h"
+#include "aws/s3/model/PutObjectRequest.h"
```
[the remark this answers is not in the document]
mrjackv: Added `.clang-format` rule and reformatted

lib/Storage/S3StorageClient.cpp
mrjackv: Renamed to `Logger`, since `Log` is already taken by an exposed AWS SDK function

lib/Storage/S3StorageClient.cpp
mrjackv: > By any chance, maybe the flush method is called from the main thread?

Dunno, I haven't investigated the matter in depth

> I don't know, put a TODO and let's leave it like this, let's see what happens.

Done, we'll see if in the future we have to tinker with this

lib/Storage/S3StorageClient.cpp
mrjackv: Dropped

lib/Storage/S3StorageClient.cpp
mrjackv: Changed to:
```cpp
bool SDKIsInitialized = false;
void initializeSDK(); // asserts if called twice
```

lib/Storage/S3StorageClient.cpp
mrjackv: added

include/revng/Storage/S3StorageClient.h
mrjackv: Sounds good, renamed

lib/Storage/S3StorageClient.cpp
mrjackv: Double checked everywhere, I think I got all of them

lib/Storage/S3StorageClient.cpp
mrjackv: Switched to `llvm::SmallVector` since it has `resize_for_overwrite`

include/revng/Storage/Path.h
mrjackv: Added and double-checked everywhere else

include/revng/Storage/Path.h
mrjackv: renamed to `.dump`

include/revng/Support/GzipTarFile.h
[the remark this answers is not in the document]
mrjackv: Done

lib/Support/CMakeLists.txt
```diff
@@ -23,11 +23,21 @@ revng_add_library_internal(
...
+  GzipTarFile.cpp)
 
 llvm_map_components_to_libnames(LLVM_LIBRARIES Support Core Object)
 
-target_link_libraries(revngSupport ${LLVM_LIBRARIES})
+find_package(Boost REQUIRED COMPONENTS iostreams)
+include(FindLibArchive)
+
+if(NOT LibArchive_FOUND)
+  message(SEND_ERROR "libarchive not found")
```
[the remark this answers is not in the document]
mrjackv: Done

lib/Support/GzipTarFile.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+  revng_assert(Res == ARCHIVE_OK);
+  int64_t Size = archive_entry_size(Entry);
+  revng_assert(Size >= 0);
+
+  llvm::SmallVector<char, 0> Data;
+  if (Size > 0) {
+    Data.resize_for_overwrite(Size);
+    size_t SizeRead = archive_read_data(Archive, Data.data(), Size);
+    revng_assert(SizeRead == (size_t) Size);
```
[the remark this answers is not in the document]
mrjackv: Dropped

include/revng/Support/GzipTarFile.h
```diff
@@ -0,0 +1,71 @@
...
+    GzipTarReader(Buffer.getBufferStart(), Buffer.getBufferSize()){};
+  ~GzipTarReader();
+
+  GzipTarReader(const GzipTarReader &Other) = delete;
+  GzipTarReader &operator=(const GzipTarReader &Other) = delete;
+
+  GzipTarReader(GzipTarReader &&Other) = default;
+  GzipTarReader &operator=(GzipTarReader &&Other) = default;
+
+  std::optional<ArchiveEntry> nextElement();
```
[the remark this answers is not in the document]
mrjackv: 1. Done
2. Fixed

lib/Storage/StorageClient.cpp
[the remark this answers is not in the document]
mrjackv: After `perf`-ing a bit, I'm convinced that skipping serialization is worth it.
I've re-introduced some checks, except that now they rely on `StorageClient == nullptr`

lib/Storage/S3StorageClient.cpp
mrjackv: Re-checking again, it seems you need to use `Aws::MakeShared` only if you use a custom allocator (which we don't)
I've replaced all occurrencies with `std::make_shared`

include/revng/Storage/StorageClient.h
```diff
@@ -0,0 +1,114 @@
...
+
+  llvm::MemoryBuffer &getBuffer() { return *Buffer; }
+};
+
+enum ContentEncoding {
+  None,
+  Gzip
+};
+
+class StorageClient {
```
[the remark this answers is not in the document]
mrjackv: Done

include/revng/Storage/DummyStorageClient.h
mrjackv: Added docs

include/revng/Support/GzipTarFile.h
```diff
@@ -0,0 +1,71 @@
...
+  ~GzipTarWriter() { revng_assert(OS == nullptr); }
+
+  GzipTarWriter(const GzipTarWriter &Other) = delete;
+  GzipTarWriter &operator=(const GzipTarWriter &Other) = delete;
+
+  GzipTarWriter(GzipTarWriter &&Other) = default;
+  GzipTarWriter &operator=(GzipTarWriter &&Other) = default;
+
+  OffsetDescriptor append(const llvm::StringRef Name,
+                          const llvm::StringRef Data);
```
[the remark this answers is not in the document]
mrjackv: Added

lib/Storage/LocalStorageClient.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+public:
+  LocalWritableFile(std::unique_ptr<llvm::raw_fd_ostream> &&OS) :
+    OS(std::move(OS)) {}
+  ~LocalWritableFile() override = default;
+  llvm::raw_pwrite_stream &os() override { return *OS; }
+  llvm::Error commit() override { return llvm::Error::success(); }
+};
+
+std::string LocalStorageClient::resolvePath(const llvm::StringRef Path) {
```
aleclearmind, line 37: Please review all the `StorageClient`s to prevent path traversals.

A `StorageClient` should act as a sort of sandbox. It should not be possible to write out of `Root` using it.

lib/Storage/Path.cpp
```diff
@@ -0,0 +1,32 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "revng/Storage/LocalStorageClient.h"
+#include "revng/Storage/Path.h"
+#include "revng/Storage/StorageClient.h"
+
+namespace revng {
+
+namespace {
```
aleclearmind, line 14: Uhm, I'd prefer anonymous namespaces to be top-level namespaces.

lib/Storage/S3StorageClient.cpp
```diff
@@ -0,0 +1,423 @@
...
+};
+
+namespace revng::detail {
+
+namespace {
+
+using Aws::Utils::Logging::FormattedLogSystem;
+using Aws::Utils::Logging::LogLevel;
+
+Logger<> Logger("aws");
```
aleclearmind, line 52:
```suggestion
Logger<> Logger("s3-storage");
```

include/revng/Storage/S3StorageClient.h
```diff
@@ -0,0 +1,59 @@
...
+
+#include "llvm/ADT/StringRef.h"
+
+#include "revng/Storage/StorageClient.h"
+
+namespace revng::detail {
+
+class S3WritableFile;
+
+class S3StorageClient : public StorageClient {
```
aleclearmind, line 17: Can we push all the `StorageClient`s to be private headers?

include/revng/Support/TemporaryFile.h
```diff
@@ -14,20 +14,47 @@ class TemporaryFile {
 private:
   llvm::SmallString<32> Path;
 
+  TemporaryFile(const llvm::StringRef Path, int /* dummy */) :
+    Path(Path.str()) {
+    llvm::sys::RemoveFileOnSignal(Path);
+  }
```
aleclearmind, line 20: I see some non-trivial changes to `TemporaryFile` in a commit that's not strictly related.
What's going on here?

include/revng/Support/TemporaryFile.h
```diff
@@ -14,20 +14,47 @@ class TemporaryFile {
...
+
 public:
-  TemporaryFile(const llvm::Twine &Prefix, llvm::StringRef Suffix = "") {
+  TemporaryFile(const llvm::Twine &Prefix, const llvm::StringRef Suffix = "") {
```
aleclearmind, line 23: `const` on non-reference arguments doesn't make much sense. Maybe you meant `const &` but `StringRef`s are meant to be passed by copy (like pointers).

lib/Pipes/PipelineManager.cpp
```diff
@@ -97,12 +96,12 @@ static llvm::Expected<Runner>
 setUpPipeline(pipeline::Context &PipelineContext,
               Loader &Loader,
               llvm::ArrayRef<std::string> TextPipelines,
-              llvm::StringRef ExecutionDirectory) {
+              revng::Path ExecutionDirectory) {
```
aleclearmind, line 99: Do we want to pass them by copy or `const &`? Let's be coherent.

lib/Storage/S3StorageClient.cpp
```diff
@@ -0,0 +1,423 @@
...
+                               const llvm::StringRef Path) {
+  if (Root.empty())
+    return Path.str();
+  else if (Path.empty())
+    return Root.str();
+  else
+    return joinPath(Root, Path);
+}
+
+static std::string generateNewFilename(const llvm::StringRef Path) {
```
aleclearmind, line 116: Doc the pattern of the generated file name

lib/Storage/S3StorageClient.cpp
```diff
@@ -0,0 +1,423 @@
...
+
+  llvm::raw_pwrite_stream &os() override { return *OS; }
+  llvm::Error commit() override {
+    OS->flush();
+
+    Aws::S3::Model::PutObjectRequest Request;
+    Request.SetBucket(Client.Bucket);
+
+    if (Encoding == ContentEncoding::Gzip)
+      Request.SetContentEncoding("gzip");
```
aleclearmind, line 183: This only has the effect of letting the S3 server known that has to serve this file with the proper content-encoding, correct? Maybe add a comment.

share/revng/test/tests/storage.sh
```diff
@@ -0,0 +1,23 @@
...
+  --resume="s3://S3RVER:S3RVER@region+127.0.0.1:$S3_PORT/test/project-test-dir" \
+  --analyses-list=revng-initial-auto-analysis \
+  -o /dev/null \
+  YieldAssembly "$1"
+
+INDEX_FILE="$2/test/project-test-dir/index.yml._S3rver_object"
+test -f "$INDEX_FILE"
+INPUT_ID="$(grep -Po '(?<=^begin/input:).*' "$INDEX_FILE" | tr -d ' ' | tr -d "'")"
+INPUT_FILE="$2/test/project-test-dir/$INPUT_ID._S3rver_object"
+sha256sum --quiet --check <(sha256sum <"$INPUT_FILE") <"$1"
```
aleclearmind, line 23: Comment

tests/unit/GzipTarFile.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+static void checkOffset(llvm::SmallVector<char> &Buffer,
+                        size_t Start,
+                        size_t Size,
+                        const llvm::StringRef ExpectedValue) {
+  std::string Result = gzipDecompress({ Buffer.data() + Start, Size });
+  BOOST_TEST((Result == ExpectedValue));
+}
+
+BOOST_AUTO_TEST_SUITE(GzipTarFileTestSuite)
```
aleclearmind, line 31: Drop?

tests/unit/CMakeLists.txt
```diff
@@ -426,3 +426,14 @@ target_link_libraries(test_PipelineCTracing revngPipelineC revngSupport
                       Boost::unit_test_framework ${LLVM_LIBRARIES})
 revng_add_test(NAME test_PipelineCTracing COMMAND test_PipelineCTracing)
 set_tests_properties(test_PipelineCTracing PROPERTIES LABELS "unit")
+
+#
```
aleclearmind, line 430: Can we have a bash test that checks files can be uncompressed with `tar`?

include/revng/Support/GzipStream.h
```diff
@@ -0,0 +1,27 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/Support/raw_ostream.h"
+
+void gzipCompress(llvm::raw_ostream &OS, const llvm::ArrayRef<uint8_t> Buffer);
+
+inline void gzipCompress(llvm::raw_ostream &OS,
+                         const llvm::ArrayRef<char> Buffer) {
```
aleclearmind, line 13: Drop `const`. This is the second instance of this pattern, please check elsewhere too.

lib/Support/GzipStream.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include "revng/Support/Assert.h"
+#include "revng/Support/GzipStream.h"
+
+#include "zlib.h"
+
+constexpr size_t OutputBufferSize = 4 * 1024;
+static_assert(OutputBufferSize <= UINT_MAX);
+
+constexpr int windowBits = 15 /* 2**15 bytes (32k) of window */
+                           + 16 /* Magic offset for gzip */;
```
aleclearmind, line 19: 1. `WindowBits`
2. No C-style comments, use full line comments. Please check everywhere.

lib/Support/GzipStream.cpp
```diff
@@ -0,0 +1,99 @@
...
+static void zlibCopyStream(z_stream &Stream,
+                           llvm::raw_ostream &OS,
+                           const llvm::ArrayRef<uint8_t> Buffer) {
+  llvm::SmallVector<uint8_t> OutBuffer;
+  OutBuffer.resize_for_overwrite(OutputBufferSize);
+  const char *OutBufferPtr = reinterpret_cast<const char *>(OutBuffer.data());
+
+  size_t RemainingInput = Buffer.size();
+
+  Stream.next_in = const_cast<uint8_t *>(Buffer.data());
```
aleclearmind, line 31: `const_cast` are absolutely illegal, please check everywhere.
Use `MutableArrayRef`.

include/revng/Support/GzipStream.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/Support/raw_ostream.h"
+
+void gzipCompress(llvm::raw_ostream &OS, const llvm::ArrayRef<uint8_t> Buffer);
```
aleclearmind, line 10: Doc and rename arguments to explicitly reflect who's the input and who's the output.

include/revng/Support/GzipTarFile.h
```diff
@@ -0,0 +1,101 @@
...
+  size_t DataStart;
+  size_t PaddingStart;
+  size_t End;
+
+  size_t headerSize() { return DataStart - Start; }
+  size_t dataSize() { return PaddingStart - DataStart; }
+  size_t paddingSize() { return End - PaddingStart; }
+};
+
+/**
```
aleclearmind, line 32: No C-style comments, check everywhere.

aleclearmind (changes requested):

lib/Support/GzipStream.cpp
mrjackv: That's not the intention, it's the `libz` api's fault that it only accepts `byte *` but **does not change the contents**.

include/revng/Support/TemporaryFile.h
mrjackv: 1. Added `TemporaryFile::make` for the cases where you want to catch the `error_code` if it happens (with the weird dummy constructor to avoid ambiguity)
2. Fixed the move assignment, since it would leak the original file if it was present
3. Fixed the destructor, otherwise it would be called even on a moved object

lib/Pipeline/ContainerSet.cpp
```diff
@@ -127,26 +129,33 @@ ContainerToTargetsMap ContainerSet::enumerate() const {
...
-    return llvm::createStringError(EC,
-                                   "could not write file at %s",
-                                   Path.str().c_str());
+llvm::Error ContainerBase::storeToDisk(const revng::Path &Path) const {
+  auto MaybeWritableFile = Path.getWritableFile();
+  if (not MaybeWritableFile) {
+    return MaybeWritableFile.takeError();
+  }
+
+  if (auto Error = serialize(MaybeWritableFile.get()->os()); Error)
```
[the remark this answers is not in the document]
mrjackv: It's a `llvm::Expected<std::unique_ptr<WritableFile>>`, the `.get()` is needed

```
ContainerSet.cpp:138:49: error: no member named 'os' in 'std::unique_ptr<revng::WritableFile>'
  if (auto Error = serialize(MaybeWritableFile->os()); Error)
                             ~~~~~~~~~~~~~~~~~~~^
1 error generated.
```

lib/Pipes/PipelineManager.cpp
```diff
@@ -132,46 +125,40 @@ PipelineManager::create(llvm::ArrayRef<std::string> Pipelines,
...
 PipelineManager::createFromMemory(llvm::ArrayRef<std::string> PipelineContent,
                                   llvm::ArrayRef<std::string> EnablingFlags,
                                   llvm::StringRef ExecutionDirectory) {
+  std::unique_ptr<revng::StorageClient> Client;
+  if (not ExecutionDirectory.empty()) {
+    auto MaybeClient = revng::StorageClient::fromURL(ExecutionDirectory);
+    if (!MaybeClient)
+      return MaybeClient.takeError();
+    Client = std::move(MaybeClient.get());
+  }
```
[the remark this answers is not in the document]
mrjackv: PipelineManager is the god object of the entire pipeline, there's nothing above. Also this needs to somehow work with python. Moving it besides like `rp_make_storage_client` to then just pass it to the manager seems a bit redundant IMHO

lib/Storage/LocalStorageClient.cpp
```diff
@@ -0,0 +1,133 @@
...
+/// \brief Implementation of StorageClient operations on the local filesystem
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "revng/Storage/LocalStorageClient.h"
+#include "revng/Support/PathList.h"
+
+namespace revng::detail {
```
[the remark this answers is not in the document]
mrjackv: I used `detail` because I wanted to stress the fact that a "user" (in the sense of, anyone not touching the internals) should not explicitly instantiate any of the implementation classes

lib/Storage/LocalStorageClient.cpp
```diff
@@ -0,0 +1,133 @@
...
+public:
+  LocalWritableFile(std::unique_ptr<llvm::raw_fd_ostream> &&OS) :
+    OS(std::move(OS)) {}
+  ~LocalWritableFile() override = default;
+  llvm::raw_pwrite_stream &os() override { return *OS; }
+  llvm::Error commit() override { return llvm::Error::success(); }
+};
+
+std::string LocalStorageClient::resolvePath(const llvm::StringRef Path) {
+  if (Path == "-")
```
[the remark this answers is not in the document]
mrjackv: Dropping `-` basically breaks all our tests and command-line tools and is not as portable as `/dev/stdin` (since that device node can be missing)

lib/Pipeline/ContainerSet.cpp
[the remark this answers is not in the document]
aleclearmind: OK, this is why I hope that was not necessary:
    https://stackoverflow.com/questions/10460662/recursive-application-of-operator

But I got confused.

lib/Pipes/PipelineManager.cpp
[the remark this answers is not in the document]
aleclearmind: > PipelineManager is the god object of the entire pipeline, there's nothing above

There are `main`s, which is the appropriate place where these things need to be handled.

> rp_make_storage_client

The C API can do it. I'm just suggesting to enforce the invariant that there no paths/URLs anywhere from `PipelineManager` below, just `revng::Path`.

lib/Storage/LocalStorageClient.cpp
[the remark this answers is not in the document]
aleclearmind: Drop `detail`, move to private header as suggested elsewhere.

lib/Pipes/PipelineManager.cpp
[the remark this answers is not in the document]
mrjackv: > just revng::Path

But `PipelineManager` has to `StorageClient::commit` (unless mains suddenly decide, but I think is poor coupling), I think having this level of indirection is unnecessary

include/revng/Support/TemporaryFile.h
aleclearmind: Gotcha, a commit would be good.

lib/Support/GzipStream.cpp
aleclearmind: OK, then add a comment about this.

We should put `const_cast` in check conventions and use `NONEXTLINELINT` in such cases.

lib/Pipes/PipelineManager.cpp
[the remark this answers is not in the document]
aleclearmind: I'm suggesting to:
1. Have `PipelineManager` get a `std::move`'d `StorageClient` in the constructor for the execution directory. Then, it can call `commit` whenever appropriate.
2. Any other method that does not implicitly use the execution directory, accept a `revng::Path`.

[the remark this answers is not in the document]
mrjackv: Doing (1) just means duplicating a lot of code around to call `StorageClient::fromURL`, I think it worsens the API
For (2) that's already happening, the only places where there's `StringRef` for paths is `PipelineManager::create*` and `Runner::storeToDiskDebug`

include/revng/Support/OnQuit.h
```diff
@@ -0,0 +1,29 @@
...
+  /// Registers an object for having its onQuit method called upon program
+  /// termination
+  void add(std::function<void()> &&Handler) {
+    Register.push_back(std::move(Handler));
+  }
+
+  void quit();
+
+private:
+  std::vector<std::function<void()>> Register;
```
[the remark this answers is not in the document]
mrjackv: Done

lib/Storage/Path.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+// TODO: unix-specific, will not work on Windows
+std::unique_ptr<StorageClient>
+  LocalClient = std::make_unique<detail::LocalStorageClient>("/");
```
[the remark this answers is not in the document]
mrjackv: Dropped

lib/Pipeline/Runner.cpp
```diff
@@ -146,20 +146,24 @@ Error Runner::invalidate(const Target &Target) {
...
+  if (ToParse.count(':') < 1) {
     auto *Message = "could not parse %s into two parts "
                     "file_path:step/container";
     return createStringError(inconvertibleErrorCode(),
                              Message,
                              ToParse.str().c_str());
   }
 
-  auto [StepName, ContainerName] = Parts.back().split('/');
+  auto Parts = ToParse.rsplit(':');
```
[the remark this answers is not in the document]
mrjackv: Done

include/revng/Storage/StorageClient.h
```diff
@@ -0,0 +1,109 @@
...
+#include "revng/Support/Debug.h"
+
+namespace revng {
+
+enum ContentEncoding {
+  None,
+  Gzip
+};
+
+/**
```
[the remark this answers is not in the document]
mrjackv: 1. Dropped all of them
2. Thanks

lib/Pipeline/Runner.cpp
```diff
@@ -171,68 +175,70 @@ Error PipelineFileMapping::loadFromDisk(Runner &LoadInto) const {
...
     return llvm::createStringError(llvm::inconvertibleErrorCode(),
                                    "No known container " + Container);
+  }
+
+  if (auto Error = LoadInto[Step].containers()[Container].storeToDisk(Path);
+      Error)
+    return Error;
 
-  return LoadInto[Step].containers()[Container].storeToDisk(InputFile);
+  return Client->commit();
```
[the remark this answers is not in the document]
mrjackv: Dropped the `StorageClient` in `PipelineFileMapping`, switched to using `fromLocalStorage`

lib/Storage/Path.cpp
mrjackv: Moved

include/revng/Support/TemporaryFile.h
mrjackv: Split off to a separate commit

include/revng/Support/TemporaryFile.h
mrjackv: Dropped all `const llvm::(Array|String)Ref`s

lib/Storage/S3StorageClient.cpp
mrjackv: > correct?

Yes

> Maybe add a comment

Will do

tests/unit/GzipTarFile.cpp
mrjackv: Dropped

include/revng/Support/GzipStream.h
mrjackv: Dropped all occurences

lib/Support/GzipStream.cpp
mrjackv: 1. Done
2. Done

lib/Support/GzipStream.cpp
mrjackv: Added comment, noted adding the rule to check-conventions in TODO

include/revng/Support/GzipTarFile.h
mrjackv: Dropped

lib/Storage/LocalStorageClient.cpp
```diff
@@ -0,0 +1,133 @@
...
+    return Path.str();
+  else if (Root.empty())
+    return joinPath(".", Path);
```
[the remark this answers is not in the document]
mrjackv: Added assertion, removed the `.`

include/revng/Support/GzipStream.h
mrjackv: Done

lib/Storage/LocalStorageClient.cpp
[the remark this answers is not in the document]
mrjackv: Dropped

lib/Storage/S3StorageClient.cpp
mrjackv: Added

share/revng/test/tests/storage.sh
mrjackv: Commented

python/revng/daemon/event_manager.py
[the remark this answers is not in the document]
mrjackv: Added TODO

lib/Pipes/PipelineManager.cpp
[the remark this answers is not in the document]
aleclearmind: Strings should not be the only way to instantiate `PipelineManager`, otherwise we're coupling `PipelineManager` to the currently known list of `StorageClient`s. One day in testing we might have some custom `StorageClient`.

You can either make the constructor public or add another factory method that takes `StorageClient &&`.

Maybe `storeToDiskDebug` should be rename to `dump`.

lib/Pipeline/ContainerSet.cpp
```diff
@@ -88,11 +87,14 @@ llvm::Error ContainerSet::storeToDisk(StringRef Directory) const {
   return Error::success();
 }
 
-llvm::Error ContainerSet::loadFromDisk(StringRef Directory) {
+llvm::Error ContainerSet::loadFromDisk(const revng::Path &Directory) {
```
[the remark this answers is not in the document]
mrjackv: Implemented

lib/Pipes/PipelineManager.cpp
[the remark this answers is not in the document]
mrjackv: Added an overload to `createFromMemory` that accepts a `StorageClient&&`
Renamed `storeToDiskDebug` to `dump`

lib/Storage/LocalStorageClient.cpp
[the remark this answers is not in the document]
mrjackv: Changes made:
* Added `StdinStorageClient` and `StdoutStorageClient` and `FilePath::std(in|out)`
* Added `CLRevngPath.h` that implements `(Input|Output)Path` that can be used instead of string e.g. `opt<string> Filename -> opt<InputFile> Filename`
* Dropped special-handling of `-` from `LocalStorageClient`

lib/Pipes/PipelineManager.cpp
```diff
@@ -249,26 +236,23 @@ void PipelineManager::writeAllPossibleTargets(llvm::raw_ostream &OS) const {
...
-
-  if (ExecutionDirectory.empty())
+llvm::Error PipelineManager::storeToDisk() {
+  if (StorageClient == nullptr)
     return llvm::Error::success();
 
-  return Runner->storeToDisk(ExecutionDirectory);
+  if (auto Error = Runner->storeToDisk(ExecutionDirectory); Error)
+    return Error;
+  return StorageClient->commit();
```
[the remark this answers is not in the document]
mrjackv: Done

lib/Storage/LocalStorageClient.cpp
mrjackv: Added a `normalizePath` function that removes the possibility of path traversal altogether

lib/Storage/S3StorageClient.cpp
mrjackv: Added comments

tests/unit/CMakeLists.txt
mrjackv: Added

include/revng/Pipes/ToolCLOptions.h
```diff
@@ -48,7 +48,7 @@ class ToolCLOptions {
                             PipelineManager &Manager) {
     const auto &Name = ModelGlobalName;
     auto *Model(cantFail(Manager.context().getGlobal<ModelGlobal>(Name)));
-    return Model->loadFromDisk(ModelOverride);
+    return Model->loadFromDisk(revng::Path::fromLocalStorage(ModelOverride));
```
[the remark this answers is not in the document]
mrjackv: Done

include/revng/Storage/StorageClient.h
```diff
@@ -0,0 +1,114 @@
...
+public:
+  virtual ~StorageClient() = default;
+  StorageClient(const StorageClient &Other) = delete;
+  StorageClient &operator=(const StorageClient &Other) = delete;
+  StorageClient(const StorageClient &&Other) = delete;
+  StorageClient &operator=(StorageClient &&Other) = delete;
+
+public:
+  static llvm::Expected<std::unique_ptr<StorageClient>>
+  fromURL(const llvm::StringRef URL);
```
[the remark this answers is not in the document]
aleclearmind: Can we have two distinct methods?
If not, I'm OK with `fromPathOrURL`.

lib/Storage/Utils.h
```diff
@@ -0,0 +1,30 @@
...
+#include "revng/Support/PathList.h"
+
+// Utility function that removes any attempt to introduce parent notation in a
+// path. This is used in the StorageClient implementations to avoid escaping
+// outside the "root" directory. This function will produce the same output
+// ('foo/bar') given any of these:
+// * foo/bar
+// * ./foo/../foo/bar
+// * /foo/bar
+// * ../../foo/bar
```
aleclearmind, line 18: These two are weird, I'd either: resolve the path and ensure it's below the root or simply abort if the path starts with `/`, ends with `/..` or contains `../`.

I'd prefer the latter, unless we use this for command line arguments too. Then the user might actually go with `..`, then we can go with the former option.

aleclearmind (changes requested):

include/revng/Storage/StorageClient.h
```diff
@@ -0,0 +1,113 @@
...
+  fromURL(llvm::StringRef URL);
+
+  void dump() const debug_function { dump(dbg); }
+
+  template<typename T>
+  void dump(T &OS) const debug_function {
+    OS << dumpString();
+  }
+
+  virtual llvm::Expected<ExistType> exists(llvm::StringRef Path) = 0;
```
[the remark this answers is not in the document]
mrjackv: 1. Ok
2. That's only on the `StorageClient` side, if you look at `FilePath`/`DirectoryPath` the `exists` function (which is the "public API") returns an `Expected<bool>`

include/revng/Storage/Path.h
```diff
@@ -0,0 +1,142 @@
...
+#include "llvm/ADT/SmallString.h"
+#include "llvm/Support/Path.h"
+
+#include "revng/Storage/StorageClient.h"
+#include "revng/Support/Debug.h"
+#include "revng/Support/PathList.h"
+
+namespace revng {
+
+class BasePath {
```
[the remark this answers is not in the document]
mrjackv: I've added the `Base` to stress that it's the parent of `FilePath` and `DirectoryPath` and to avoid confusion like "Ah it's a `Path`, I can use it to do file operations" whereas `BasePath` is more explicit

include/revng/Storage/Path.h
```diff
@@ -0,0 +1,142 @@
...
+protected:
+  revng::StorageClient *Client = nullptr;
+  std::string StoredPath;
```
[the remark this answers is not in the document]
mrjackv: Would `SubPath` be better?

lib/Storage/Utils.h
mrjackv: > unless we use this for command line arguments too

we do

> resolve the path and ensure it's below the root

That's by design of the `normalizePath` function: the output path is guaranteed to be below the root

include/revng/Storage/StorageClient.h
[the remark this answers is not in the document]
mrjackv: Renamed to `fromPathOrURL`

include/revng/Storage/Path.h
```diff
@@ -0,0 +1,142 @@
...
+    Client(Client), StoredPath(StoredPath.str()){};
+
+public:
+  void dump() const debug_function { dump(dbg); }
+
+  template<typename T>
+  void dump(T &OS) const debug_function {
+    OS << "<Client: ";
+    Client->dump(OS);
+    OS << " Path: " + StoredPath + ">";
```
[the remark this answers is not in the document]
mrjackv: Done

include/revng/Storage/Path.h
```diff
@@ -0,0 +1,142 @@
...
+    WritableFile->os() << ReadableFile->buffer().getBuffer();
+
+    return WritableFile->commit();
+  }
+
+  llvm::Error renameTo(const FilePath &Destination) const {
+    if (auto Error = copyTo(Destination); Error)
+      return Error;
+    return remove();
+  }
```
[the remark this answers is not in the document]
mrjackv: Dropped

include/revng/Storage/CLPathOpt.h
```diff
@@ -0,0 +1,85 @@
+#pragma once
```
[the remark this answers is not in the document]
mrjackv: Added comments explaining what's going on

lib/Storage/Path.cpp
```diff
@@ -0,0 +1,47 @@
+/// \file S3StorageClient.cpp
+/// \brief Implementation of StorageClient operations in S3
```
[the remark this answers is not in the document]
mrjackv: 1. Fixed
2. Dropped

include/revng/Storage/Path.h
```diff
@@ -0,0 +1,142 @@
...
+
+  bool isValid() const { return Client != nullptr; }
+};
+
+class FilePath : public BasePath {
+public:
+  explicit FilePath(StorageClient *Client, llvm::StringRef StoredPath) :
+    BasePath(Client, StoredPath){};
+
+  // Utility function to get a FilePath that can read stdin
```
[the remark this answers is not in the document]
mrjackv: Fixed

include/revng/Storage/Path.h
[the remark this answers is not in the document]
aleclearmind: Right, I though it was a abstract class, but it's not. Change to `PathBase`.

include/revng/Storage/Path.h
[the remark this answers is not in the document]
aleclearmind: I'd prefer just `Path`, but `SubPath` works too.

include/revng/Storage/StorageClient.h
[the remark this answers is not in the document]
aleclearmind: 2. OK, still. Either find a better name or split up.

lib/Storage/Utils.h
aleclearmind: Yes, but handling `../foo` as `foo` is weird.
Let's make the path absolute, check if it's below the root and if it's not fail instead of succeeding in unexpected ways.

include/revng/Storage/StorageClient.h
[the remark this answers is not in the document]
mrjackv: `pathType`? `stat`?

include/revng/Storage/Path.h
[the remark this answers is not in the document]
mrjackv: Renamed to `SubPath`

include/revng/Storage/StorageClient.h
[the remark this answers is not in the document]
aleclearmind: Just `type` might work

include/revng/Storage/Path.h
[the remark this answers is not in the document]
mrjackv: Renamed

include/revng/Storage/StorageClient.h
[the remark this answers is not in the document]
mrjackv: Renamed to `type`

lib/Storage/Utils.h
mrjackv: Changed how things work:
* `fromLocalStorage` will compute the absolute path before creating the `{File|Directory}Path`, so on that side it's guaranteed that we don't get `../`
* Changed the logic to do `revng_assert(checkPath(Path))` where checkPath returns false when it detects `../`, `./`, `//` or leading `/`

aleclearmind:
Merged in 9c37b70. Thanks.

PR #310, opened 2023-07-20
Make dynamic functions `nomerge`

pfez:


mrjackv (approved):
Looks good to me

aleclearmind:
In revng-c, you copy the attributes instead of forcing `nomerge`, which is nice.

Can you do the same in `EnforceABI.cpp` so we have only One True Place where we set `nomerge` (i.e., Isolate)?

```patch
diff --git a/lib/FunctionIsolation/EnforceABI.cpp b/lib/FunctionIsolation/EnforceABI.cpp
index f9ccee673..d0e3501cc 100644
--- a/lib/FunctionIsolation/EnforceABI.cpp
+++ b/lib/FunctionIsolation/EnforceABI.cpp
@@ -395,6 +395,7 @@ void EnforceABIImpl::handleRegularFunctionCall(CallInst *Call) {
                                    *CallerBlock,
                                    *CallSite);
   NewCall->copyMetadata(*Call);
+  NewCall->setAttributes(Call->getAttributes());
 
   // Set PC to the expected value
   GCBI.programCounterHandler()->setPC(Builder, CallerBlock->ID().start());
@@ -461,7 +462,6 @@ CallInst *EnforceABIImpl::generateCall(IRBuilder<> &Builder,
   // Produce the call
   //
   auto *Result = Builder.CreateCall(Callee, Arguments);
-  Result->addFnAttr(Attribute::NoMerge);
   revng_assert(Result->getDebugLoc());
 
   if (ReturnCSVs.size() != 1) {
```

pfez:
I pushed a new commit with the change to EnforceABI that you requested.

I'm running tests.

pfez:
The tests pass. @aleclearmind it's good to merge for me.

aleclearmind:
Merged in cbddc85. Thanks!

PR #312, opened 2023-07-25
Avoid hangs when running `revng daemon` tests

mrjackv:


share/revng/test/tests/daemon/test.py
```diff
@@ -86,20 +87,32 @@ async def client(pytestconfig: Config, request) -> AsyncGenerator[AsyncClientSes
...
         env=new_env,
     )
 
+    def stop_daemon():
+        if process.returncode is not None:
+            return process.returncode
+
+        process.send_signal(signal.SIGINT)
+        try:
+            return process.wait(30.0)
```
aleclearmind, line 101: Will I have to wait 30 extra seconds before getting a crash?
These tests are already quite slow, waiting 30 more seconds would make me very sad.

You mentioned `.communicate` as the proper solution, is there anything stopping us from implementing the proper solution?

mrjackv: > You mentioned .communicate as the proper solution, is there anything stopping us from implementing the proper solution?

This *is* a proper solution, we're using `TemporaryFile`, but it works well just as well. The problem with using `.communicate` is that it forces collection of stdout when terminating rather than allowing `termination` and `reading stdout` to happen at different times (which would look awkward and ugly in the code).

> Will I have to wait 30 extra seconds before getting a crash?

No, that's just a safeguard in case a similar situation shows up, basically:
* If the test runs smoothly -> instant termination
* If the test crashes -> instant termination
* If the test hangs at shutdown (such as with PIPE) -> termination happens 30 seconds late with `KILL` and the test fails because the exit code is `-9` as opposed to `0`

share/revng/test/tests/daemon/test.py
```diff
@@ -77,6 +77,7 @@ async def response_trace(session, trace_config_ctx, params: TraceRequestEndParam
 @fixture
 async def client(pytestconfig: Config, request) -> AsyncGenerator[AsyncClientSession, None]:
     temp_dir = TemporaryDirectory()
```
aleclearmind, line 79: Can we have a name for this too?

share/revng/test/tests/daemon/test.py
```diff
@@ -77,6 +77,7 @@ async def response_trace(session, trace_config_ctx, params: TraceRequestEndParam
...
+    log_file = TemporaryFile("wb+", suffix="revng-daemon-test-log-")
```
aleclearmind, line 80: Why not use a `using` statement?

share/revng/test/tests/daemon/test.py
aleclearmind: Gotcha

aleclearmind (changes requested):

share/revng/test/tests/daemon/test.py
mrjackv: * Everything under should be indented in the `with`
* `Temporary*` classes delete the resource either via `with` but also on destruction, so once `log_file` goes out of scope it will be deleted

aleclearmind:
Merged in 878f9ef. Thanks!

PR #313, opened 2023-07-25
Drop hardcoded analyses

mrjackv:


python/revng/cli/_commands/graphql_upload.py
```diff
@@ -0,0 +1,65 @@
...
+from typing import List
+
+from aiohttp import TCPConnector
+from gql import Client, gql
+from gql.transport.aiohttp import AIOHTTPTransport
+
+from revng.cli.commands_registry import Command, CommandsRegistry, Options
+
+
+class GraphQLUploadCommand(Command):
```
aleclearmind, line 17: 1. What is this command for? If it's never tested (and used) let's not merge it.
2. The name is misleading, this also performs analyses.

share/revng/components/revng.yml
```diff
@@ -0,0 +1,24 @@
...
+  - lib/revng/analyses/librevngBasicAnalyses.so
+  - lib/revng/analyses/librevngABIAnalyses.so
+  - lib/revng/analyses/librevngEarlyFunctionAnalysis.so
+  - lib/revng/analyses/librevngFunctionIsolation.so
+  - lib/revng/analyses/librevngModel.so
+  - lib/revng/analyses/librevngModelImporter.so
+  - lib/revng/analyses/librevngModelImporterBinary.so
+  - lib/revng/analyses/librevngRecompile.so
+  - lib/revng/analyses/librevngTypeShrinking.so
+  - lib/revng/analyses/librevngYieldPipes.so
```
aleclearmind, line 24: This will go out of sync. We either:
1. Need to generate this in CMake, or...
2. We can we can put in each `share/revng/pipelines/*.yml` file a field to specify the "component", expose it through GraphQL and then let `revng-daemon-self-test` select only the artifacts with the `revng` component. This sounds better, it's more informative for other use cases and does not increase the `dlopen` trickery.

aleclearmind (changes requested):

python/revng/cli/_commands/graphql_upload.py
mrjackv: 1. The files:
   ```
   python/revng/daemon/demo_webpage.py
   python/revng/daemon/templates/base.html
   python/revng/daemon/templates/index.html
   ```
   That were dropped as part of this PR were part of the demo page cremo made way back, they were completely unused expect by me for a specific purpose: they easily allowed loading a binary in daemon and running analyses *without* using revng-ui, which is very handy when testing some things. Since those needed analyses to be hardcoded, I've dropped them but added this command to preserve the functionality (which might be useful to others, not just me)

share/revng/components/revng.yml
mrjackv: 1. I'll check if this can be done without too much magic
2. This can work for pipelines, but for libraries it's non-trivial: unless we do `$unusual_things` to them (e.g. stuffing the component in `.note` section)

Overall if (1) can be done in a simple way, it's preferable to (2)

python/revng/cli/_commands/graphql_upload.py
aleclearmind: We need to test it. I propose to:
1. Drop `revng daemon-self-test` in favor of this tool plus a `--all-artifacts` switch.
2. Rename this tool into `revng graphql` and see where we go from there.

share/revng/components/revng.yml
aleclearmind: We can keep loading the libraries, we just need to prevent it from running revng-c analyses and producing revng-c artifacts.
Simply loading a library won't lead to non-determinism in most cases, I'm fine with that.
Let's strive for option 2.

mrjackv: Ok for (2), I've noted down in my pad other options if we want to go back to (1)
Regarding the field: should it be part of the schema or e.g. embedded in a comment?

aleclearmind: A proper field

mrjackv: Done

python/revng/api/_capi.py
```diff
@@ -122,8 +123,16 @@ def __setattr__(self, name, value):
             raise NotImplementedError()
 
 
+def get_components():
+    if "REVNG_COMPONENTS" in os.environ:
+        return os.environ["REVNG_COMPONENTS"].split(",")
+    else:
+        return None
+
+
```
aleclearmind, line 132: Drop, we already have `REVNG_OPTIONS`

python/revng/cli/commands_registry.py
```diff
@@ -68,6 +68,11 @@ def __init__(self):
         parser.add_argument(
             "--prefix", action="append", metavar="PREFIX", help="Additional search prefix."
         )
+        parser.add_argument(
+            "--components",
+            type=lambda x: x.split(","),
+            help="Limit using the specified comma-separated components",
+        )
```
aleclearmind, line 75: `--component` + `action="append"`

python/revng/support/collect.py
```diff
@@ -51,3 +53,19 @@ def collect_libraries(search_prefixes: AnyPaths) -> Tuple[List[str], Set[str]]:
...
+    if components is None:
+        return pipelines
+
+    result: List[str] = []
+    for pipeline in pipelines:
+        with open(pipeline) as f:
+            data = yaml.safe_load(f)
+            assert "Component" in data, "Pipeline file does not specify its component"
+            if data["Component"] in components:
+                result.append(pipeline)
```
aleclearmind, line 69: Python never inspected the pipelines files before this. Unnecessary coupling.

My original idea was to 1) always load all the pipelines, 2) expose for each artifact what was its component and let `revng graphql --all-artifacts` filter based on that.

If you think that's too annoying, at least push handling of `--component` into C++, ignoring pipelines that to not match one of the whitelisted components.

aleclearmind (changes requested):

python/revng/api/_capi.py
mrjackv: Dropped

python/revng/cli/commands_registry.py
mrjackv: Done

python/revng/support/collect.py
mrjackv: Moved the logic on the pipeline side

include/revng/Pipeline/Step.h
```diff
@@ -137,6 +138,12 @@ class Step {
     return llvm::Error::success();
   }
 
+  void setComponent(const llvm::StringRef Component) {
+    this->Component = Component.str();
+  }
```
aleclearmind, line 143: Can we drop this?

aleclearmind (changes requested):

python/revng/api/_capi.py
```diff
@@ -163,12 +169,18 @@ def initialize(
...
+        if not revng_options_has("load"):
+            libraries, _ = collect_libraries(ROOT)
+        else:
+            libraries = []
     else:
         libraries = to_iterable(libraries)
 
     if pipelines is None:
-        pipelines = collect_files(ROOT, ["share", "revng", "pipelines"], "*.yml")
+        if not revng_options_has("pipeline-path"):
```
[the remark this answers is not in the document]
mrjackv: Otherwise `initialize` would be passing files that (possibly the user) does not expect
e.g.
```
REVNG_OPTIONS="-pipeline-path=foo.yml" osh revng daemon
```

Should not call `collect_pipelines`

include/revng/Pipeline/Step.h
mrjackv: Do you mean integrating it in the constructor? Dunno, the component is not an *essential* part of a `Step` (e.g. `begin` does not belong to any component)

python/revng/api/_capi.py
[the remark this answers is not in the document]
aleclearmind: So you're saying that if in `REVNG_OPTIONS` there's `-pipeline-path` we do not autocollect them?

I think we should either have the C API ignore `REVNG_OPTIONS` entirely or have an argument to enable/disable auto-collection of things. Parsing `REVNG_OPTIONS` like this is sad.

include/revng/Pipeline/Step.h
aleclearmind: Empty string?

mrjackv: > Empty string in such case?

That's the way it is currently

aleclearmind: Then drop `setComponent` and let's go through the constructor only?

python/revng/api/_capi.py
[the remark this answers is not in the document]
aleclearmind: What specific situation are we trying to handle here?
Under which circumstances do we have `REVNG_OPTIONS` with `pipeline-path` and we'd like not to load any other pipeline?

include/revng/Pipeline/Step.h
mrjackv: Done

python/revng/api/_capi.py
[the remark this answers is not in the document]
mrjackv: Added code that removes `REVNG_OPTIONS` when using `revng.api._capi`

aleclearmind:
Merged in e88985e. Thanks.

PR #322, opened 2023-09-04
Add S3 proxy

mrjackv:


include/revng/Pipeline/Container.h
aleclearmind, line 1: Nice

lib/Storage/S3StorageClient.h
```diff
@@ -49,6 +51,7 @@ class S3StorageClient : public StorageClient {
   getWritableFile(llvm::StringRef Path, ContentEncoding Encoding) override;
 
   llvm::Error commit() override;
+  llvm::Error setCredentials(llvm::StringRef Credentials) override;
```
aleclearmind, line 54: The credentials string is opaque and storage-backend specific.
Document its format here.

python/revng/daemon/__init__.py
```diff
@@ -34,7 +36,23 @@
...
 
-def get_middlewares() -> List[Middleware]:
+class ManagerCredentialsMiddleware:
+    def __init__(self, app: ASGIApp, event_manager: EventManager):
+        self.app = app
+        self.event_manager = event_manager
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] != "http":
+            return await self.app(scope, receive, send)
```
aleclearmind, line 46: Can you make an example of a non-`http` request type that we expect?

python/revng/daemon/__init__.py
```diff
@@ -34,7 +36,23 @@
...
+
+        credentials = Headers(scope=scope).get("x-revng-set-credentials")
```
aleclearmind, line 48: What's the convention about custom headers naming?
I'd expect `X-revng-Set-Credentials` or something like that. What's the best practice here?

aleclearmind (changes requested):

python/revng/daemon/__init__.py
mrjackv: `lifetime` and `websocket`, per [ASGI Spec](https://asgi.readthedocs.io/en/latest/specs/main.html#protocol-specifications)

Starlette guarantees it's one of those 3 [here](https://github.com/encode/starlette/blob/master/starlette/routing.py#L729)

include/revng/Pipeline/Container.h
mrjackv: Note that this change is here because `revng:feature/revng-s3` does not include the last commit of `feature/revng-s3` for some reason

python/revng/daemon/__init__.py
mrjackv: Headers are case-insensitive ([Reference](https://datatracker.ietf.org/doc/html/rfc9112#name-field-syntax))

Most implementations that I know of use `.lower` when parsing headers

lib/Storage/S3StorageClient.h
mrjackv: Documented

python/revng/daemon/__init__.py
aleclearmind: I guess we'll have to do something about websockets one day, but not today.

python/revng/daemon/__init__.py
aleclearmind: Yeah, it was more for aesthetic purposes, no biggie.

aleclearmind:
Merged in e2504dd. Thanks.

PR #324, opened 2023-09-08
GraphAlgorithms: fix `nodesBetween` implementation

andrealinux1:
During the DFS exploration performed in the `nodesBetweenImpl` implementation, when we encounter a node that has been already visited (and therefore inserted in `VisitStack`), we save all the nodes on the current exploration stack, so that later, during the post-processing of the identified reachable nodes, we are able to insert them as reachable nodes if the encountered already visited node, will eventually became part of the reachable nodes.

This widen the previous criterion, which perfomed this operation only for nodes that were on the exploration stack when encountered.

Unit test to check this new behavior have been added.

### DIFF
--- a/include/revng/Support/GraphAlgorithms.h
+++ b/include/revng/Support/GraphAlgorithms.h
@@ -202,7 +202,7 @@ public:
     return InternalMap.insert(std::make_pair(Block, OnStack));
   }
 
-  // Return true if b is currently on the active stack of visit
+  // Return true if Block is currently on the active stack of visit
   bool onStack(NodeT Block) const {
     auto Iter = InternalMap.find(Block);
     return Iter != InternalMap.end() && Iter->second;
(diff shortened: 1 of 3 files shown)


aleclearmind:
Can you make a graphviz example and produce a list of visited nodes that would lead to an incorrect result in the old implementation?

include/revng/Support/GraphAlgorithms.h
```diff
@@ -202,12 +202,19 @@ class DFStack {
...
   bool onStack(NodeT Block) const {
     auto Iter = InternalMap.find(Block);
     return Iter != InternalMap.end() && Iter->second;
   }
 
+  // Return true if Block has been already visited, both if it is or not on the
+  // stack of visit
+  bool alreadyVisited(NodeT Block) const {
+    auto Iter = InternalMap.find(Block);
+    return Iter != InternalMap.end();
```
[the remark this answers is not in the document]
andrealinux1: b59a8ed15a71b7ccbb2ebc2f3b1fedec0dc00671 now reflects this change

andrealinux1:
A couple of comments regarding the new implementation, in addition to what we already discussed:

1) There are a couple of corner cases we didn't consider, they are minor stuff, and there is a dedicated part in the code after the double DFS dedicated to handle them.

2) There is a situation where we invoke `nodesBetween` with a `Target` equivalent to `nullptr`. This is a use case we need in some situations, and the intended semantics on the callsite is to collect all the nodes between the `Source` and any exit node in an acyclic graph.

I made the `nodesBetween` implementation behave in the following way, when invoked with the `Target == nullptr`: just perform the forward DFS from the `Source` node, and collect all the reachable nodes.

This, in general, is not equivalent to the semantics cited above (in particular, if we have an edge looping on a node _above_ the `Source`, we would also collect all the nodes on that path, even if they don't lead to an exit), but the fact that we invoke the function with the guarantee of an acyclic graph, makes the two semantics coincide. In this way, the behavior of the `nodesBetween` routine invoked with `nullptr` is clear and simple, and the callsite is in charge of making sure that the results respects the intended semantics.

include/revng/Support/GraphAlgorithms.h
```diff
@@ -406,138 +312,117 @@ getBackedgesBlackList(GraphT Block,
...
-  State.insertInMap(Target, false);
+  // Perform the forward DFS visit, stopping at the `Target` node
+  using ExtType = typename llvm::df_iterator_default_set<NodeRef>;
+  ExtType ForwardExtSet;
+  ForwardExtSet.insert(Target);
 
-  using nbdf_iterator = llvm::df_iterator<GraphT, StateType, true, GT>;
-  auto Begin = nbdf_iterator::begin(Source, State);
-  auto End = nbdf_iterator::end(Source, State);
+  using fdf_iterator = llvm::df_iterator<GraphT, ExtType, true, FGT>;
```
aleclearmind, line 349: Why do we set `ExtStorage = true`? What does that do?

include/revng/Support/GraphAlgorithms.h
```diff
@@ -406,138 +312,117 @@ getBackedgesBlackList(GraphT Block,
...
+  auto ForwardBegin = fdf_iterator::begin(Source, ForwardExtSet);
+  auto ForwardEnd = fdf_iterator::end(Source, ForwardExtSet);
 
-  for (NodeRef Block : llvm::make_range(Begin, End)) {
-    (void) Block;
+  for (NodeRef Block : llvm::make_range(ForwardBegin, ForwardEnd)) {
+    ForwardDFSNodes.insert(Block);
```
aleclearmind, line 354: Any reason you need `ForwardDFSNodes` instead of using `ForwardExtSet` directly?

include/revng/Support/GraphAlgorithms.h
```diff
@@ -406,138 +312,117 @@ getBackedgesBlackList(GraphT Block,
...
+    Result.insert(Target);
+    Result.insert(Source);
+  } else if (Source == Target) {
+
+    // If `Source` is equal to `Target`, due to fact that `Target` (which is `==
+    // Source`) is added to the ext set, (and vice-versa), we will not find any
+    // candidate note in our exploration, therefore we need to manually add it
+    // to the `Result` set
+    Result.insert(Source);
+  }
```
aleclearmind, line 406: Seems very nice!

aleclearmind (changes requested):

include/revng/Support/GraphAlgorithms.h
andrealinux1: You are right, my bad, in the current usage of the `df_iterator` we do not make use of the `ExtStorage = true`, I changed it to `false` in 9b5a0ed0e76593d28bf4787c768a11739a7f2469.

include/revng/Support/GraphAlgorithms.h
andrealinux1: This is a very good question, we discussed a bit over it, and the final decision was to leave it as it is 9b5a0ed0e76593d28bf4787c768a11739a7f2469

include/revng/Support/GraphAlgorithms.h
andrealinux1: Eh, unfortunately due to the fact that the intersection will not contain the `Source` and the `Target` (they are mutually excluded by the fact they are inserted in the ext set, therefore not in the intersection), we need to manually add them afterwards (even in the corner case of when `Source == Target`.

andrealinux1:
I also added the `findReachableNodes` primitive that is able to perform a standard DFS without stopping, when the `Stop` parameter is `nullptr`.

PR #328, opened 2023-09-21
Context global index

mrjackv:
Sister PRs:
* [revng-ui](https://rev.ng/gitlab/revng-private/revng-ui/-/merge_requests/19)


include/revng/Pipeline/CLOption.h
```diff
@@ -6,31 +6,43 @@
...
+#include "llvm/Support/Casting.h"
 #include "llvm/Support/CommandLine.h"
 
+#include "revng/ADT/Concepts.h"
 #include "revng/Support/Assert.h"
 
 namespace pipeline {
 
+template<typename T>
+concept PipelineOptionType = anyOf<T, bool, int, uint64_t, std::string>();
```
aleclearmind, line 18: I'd put as a first line of the commit message something like "Analyses options: restrict to bool, int and strings"

include/revng/Pipeline/Invokable.h
```diff
@@ -197,12 +207,27 @@ void createAndAppend(std::vector<std::unique_ptr<CLOptionBase>> &Out,
...
                     const std::integer_sequence<size_t, S...> &,
                     llvm::cl::OptionCategory *Cat = nullptr) {
   using cat = llvm::cl::cat;
   (createAndAppend<T, S>(Out, T::Name, getOptionName<T, S>(), cat(*Cat)), ...);
+  (createAndAppendPathSwitch<T, S>(Out,
+                                   T::Name,
+                                   getOptionName<T, S>().str() + "-is-path",
+                                   cat(*Cat),
+                                   llvm::cl::init(false)),
+   ...);
```
aleclearmind, line 230: I understand why you might want this, but I really don't want to have to use two options.

```
revng analyze --ConvertToCABIFunctionType-confidence=/lel --ConvertToCABIFunctionType-confidence-is-path
```

It's not the most ergonomic thing.

I'd go for something like:
```
revng analyze --ConvertToCABIFunctionType-confidence-path=/lel
```

include/revng/Pipes/GlobalsAnalyses.h
```diff
@@ -0,0 +1,71 @@
...
+
+constexpr inline std::tuple DiffOptions = { Option("global-name", ""),
+                                            Option("diff-content", "") };
+
+constexpr inline std::tuple SetOptions = { Option("global-name", ""),
+                                           Option("global-content", "") };
+
+} // namespace options
+
+struct ApplyDiffAnalysis {
```
aleclearmind, line 27: Pretty nice, just add some docs.

lib/PipelineC/PipelineC.cpp
```diff
@@ -254,8 +285,9 @@ _rp_manager_run_analysis(rp_manager *manager,
                          const char *step_name,
                          const char *analysis_name,
                          const rp_container_targets_map *target_map,
+                         const rp_string_map *options,
                          rp_invalidations *invalidations,
-                         const rp_string_map *options) {
+                         rp_error *error) {
```
aleclearmind, line 290: What was happening before this? Just returning a `nullptr`?

python/revng/api/manager.py
```diff
@@ -116,40 +116,13 @@ def deserialize_target(self, serialized_target: str) -> Optional[Target]:
...
-        return result
-
     def produce_target(
         self,
         step_name: str,
         target: None | str | List[str],
         container_name: Optional[str] = None,
         only_if_ready=False,
-    ) -> Dict[str, str | bytes]:
+    ) -> Dict[str, str | bytes] | Error:
```
aleclearmind, line 125: What's the ratio around returning `Error` as opposed to throwing an exception?

python/revng/daemon/graphql.py
```diff
@@ -41,10 +43,30 @@ def run_in_executor(function: Callable[P, T], *args: P.args, **kwargs: P.kwargs)
...
 
 
+@dataclass
+class Diff:
+    diff: str
+
+
+@dataclass
+class Produced:
+    result: str
```
aleclearmind, line 60: What's the reason for these guys to exist?

python/revng/daemon/schema.graphql
```diff
@@ -19,8 +25,30 @@ type Target {
 type Mutation {
     uploadB64(input: String!, container: String!): Boolean!
     uploadFile(file: Upload, container: String!): Boolean!
-    runAnalysis(step: String!, analysis: String!, containerToTargets: String, options: String): String!
-    runAnalysesList(name: String!, options: String): String!
+    runAnalysis(step: String!, analysis: String!, containerToTargets: String, options: String): AnalysisResult!
+    runAnalysesList(name: String!, options: String): AnalysisResult!
+}
+
+union AnalysisResult = Diff | SimpleError | DocumentError
```
aleclearmind: What's the ratio for handling errors in-band?

GraphQL seems to have [its own error reporting](https://learning.postman.com/open-technologies/blog/graphql-error-handling/). Have you considered using it?

include/revng/PipelineC/Prototypes.h
```diff
@@ -364,7 +364,8 @@ bool rp_manager_container_deserialize(rp_manager *manager,
                                       rp_step *step,
                                       const char *container_name,
                                       const char *content,
-                                      uint64_t size);
+                                      uint64_t size,
+                                      rp_invalidations *invalidations);
```
aleclearmind, line 373: We should really get rid of this primitive. It's... impure.
One day we'll have hashes in the model and we'll be able to get rid of it.

No action needed.

aleclearmind (changes requested):

aleclearmind:
NOTETOSELF: resume from "Add Context Global Index".

lib/Storage/Path.cpp
```diff
@@ -17,15 +17,20 @@ revng::LocalStorageClient LocalClient("/");
 revng::StdinStorageClient StdinClient;
 revng::StdoutStorageClient StdoutClient;
 
+std::string normalizeLocalPath(llvm::StringRef Path) {
+  llvm::SmallString<256> PathCopy(Path);
+  revng_assert(!llvm::sys::fs::make_absolute(PathCopy));
+  llvm::sys::path::remove_dots(PathCopy, true);
+  llvm::StringRef Result(PathCopy.substr(1)); // Remove leading '/'
```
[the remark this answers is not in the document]
mrjackv: `llvm::sys::fs::make_absolute` guarantees that the path begins with `/`, we can check but it seems a bit redundant.

include/revng/Pipeline/CLOption.h
```diff
@@ -40,8 +52,17 @@ class CLOptionBase {
   }
 };
 
-template<typename T>
-class CLOptionWrapper : public CLOptionBase {
+template<PipelineOptionType T>
+class CLOptionWrapper final : public CLOptionBase {
+private:
+  std::string FullName;
+  llvm::cl::opt<T> Option;
```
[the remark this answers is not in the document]
mrjackv: Noted

lib/PipelineC/PipelineC.cpp
mrjackv: Yes

include/revng/Pipeline/Invokable.h
mrjackv: Ah! I misunderstood what you meant the other day, I'll fix it.

python/revng/daemon/graphql.py
mrjackv: graphql `union`s can only be done between structs and not basic types (e.g. I cannot write `AnalysisResult = string | SimpleError | DocumentError`) so those classes are needed as a dummy struct to package the string.

python/revng/daemon/schema.graphql
mrjackv: Yes, however it does not provide strong typing for the errors at the schema level, so it would have been considerably worse. The current implementation:
* Makes it clear which operations are failable right on the schema
* Makes it clear what the structure of the errors is, providing all the GraphQL goodies such as aliases etc.

python/revng/api/manager.py
mrjackv: Related to [this](https://github.com/revng/revng/pull/328#discussion_r1347297699) but basically it makes it clear at the API level that an error might be returned, so that it can't be ignored by callers and we notice it via `mypy`

include/revng/PipelineC/Prototypes.h
mrjackv: I agree, this change is kind of in that direction because it introduces the same invalidation mechanism as analyses/diffs, so that once it does become one it will be transparent w.r.t. revng-ui.

include/revng/Pipeline/Invokable.h
mrjackv: Fixed

include/revng/Pipes/GlobalsAnalyses.h
mrjackv: Added

lib/Storage/Path.cpp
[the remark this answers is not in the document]
aleclearmind: OK, but take `make_absolute` out of the assertion, they are stripped in release mode.

python/revng/daemon/schema.graphql
aleclearmind: OK, add a comment about this.

include/revng/Pipeline/Context.h
```diff
@@ -36,6 +36,7 @@ extern Logger<> CommandLogger;
 class Context {
 private:
   GlobalsMap Globals;
+  uint64_t GlobalIndex = 0;
```
aleclearmind, line 39: Calling this feature "global index" is very vague.
What about "commit index" or "epoch"?

lib/Pipeline/Context.cpp
```diff
@@ -20,3 +20,42 @@ Logger<> pipeline::CommandLogger("commands");
...
+llvm::Error Context::store(const revng::DirectoryPath &Path) const {
+  if (auto Error = Globals.store(Path); Error)
+    return Error;
+
+  revng::FilePath IndexPath = Path.getFile(".index");
+  auto MaybeWritableFile = IndexPath.getWritableFile();
+  if (not MaybeWritableFile)
+    return MaybeWritableFile.takeError();
+
+  MaybeWritableFile->get()->os() << GlobalIndex;
```
aleclearmind, line 33: Newline?

lib/Pipeline/Context.cpp
```diff
@@ -20,3 +20,42 @@ Logger<> pipeline::CommandLogger("commands");
...
+    return llvm::Error::success();
+
+  auto MaybeReadableFile = IndexPath.getReadableFile();
+  if (not MaybeReadableFile)
+    return MaybeReadableFile.takeError();
+
+  llvm::StringRef Buffer = MaybeReadableFile->get()->buffer().getBuffer();
+  if (Buffer.getAsInteger(10, GlobalIndex)) {
+    return llvm::createStringError(llvm::inconvertibleErrorCode(),
+                                   "Malformed Index");
```
aleclearmind, line 57: Malformed `.index` file

python/revng/daemon/schema.graphql
```diff
@@ -53,6 +53,11 @@ type DocumentErrorReason {
 }
 
 type Subscription {
+    invalidations: Invalidation!
+}
+
+type Invalidation {
+    index: BigInt!
```
aleclearmind, line 64: Index is to vague. We need a name that makes what it is obvious.

python/revng/daemon/__init__.py
```diff
@@ -131,6 +132,7 @@ def generate_context(request: Request):
         return {
             "manager": manager,
             "event_manager": event_manager,
+            "index_lock": asyncio.Lock(),
```
aleclearmind, line 135: Why do we need a lock?
Docs.

python/revng/daemon/graphql.py
```diff
@@ -60,6 +60,11 @@ class Produced:
     result: str
 
 
+@dataclass
+class IndexError_:
```
aleclearmind, line 64: Why the trailing `_`?

aleclearmind (changes requested):

python/revng/daemon/graphql.py
mrjackv: `IndexError` is a builtin python exception when you do an out-of-bounds array access

include/revng/Pipeline/Context.h
mrjackv: [The pad](https://pad.rev.ng/NSpY-VvKTnG23vjTDJO70g) explicitly forbid using 'epoch'. I guess 'commit index' works

python/revng/daemon/schema.graphql
mrjackv: IMHO in the context of an invalidation it's obvious what the index is about, in the same way as the `index` parameter of `runAnalysis`

python/revng/daemon/__init__.py
mrjackv: Without the lock, there's a TOCTOU between the check of the index and the actual analysis is being run
ACK on adding docs

lib/Storage/Path.cpp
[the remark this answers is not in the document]
mrjackv: Done

lib/Pipeline/Context.cpp
mrjackv: It can be added, but then it requires `.trim`-ing at deserialization

aleclearmind: OK. I don't like text files that do not end with `\n` `:D`

python/revng/daemon/schema.graphql
aleclearmind: I'm not sure I agree, it might be the index of the invalidation.

In `runAnalysis` it's the index we *expect* to be in remote, not just any index.

python/revng/daemon/graphql.py
aleclearmind: OK, I guess that calling it `CommitIndex` imprves the situation.

python/revng/daemon/schema.graphql
mrjackv: > I'm not sure I agree, it might be the index of the invalidation.

That's the thing, invalidations are (currently) tied to the index, so `1 index bump -> 1 invalidation`. There isn't a separate ordering.

aleclearmind: I see, but the "index of invalidation events" is a confusing concept. I'd expect the "index of an event" to be zero at the first event, but this is not the case (unless you just opened the project).

I don't see much benefit in having two similar concept with the same name with slightly different semantics.
Let's stick to one concept only: the index of the commit on globals, so we're always sure what we're talking about.

lib/Pipeline/Context.cpp
```diff
@@ -20,3 +20,42 @@ Logger<> pipeline::CommandLogger("commands");
 
 Context::Context() : TheKindRegistry(Registry::registerAllKinds()) {
 }
+
+llvm::Error Context::store(const revng::DirectoryPath &Path) const {
+  if (auto Error = Globals.store(Path); Error)
+    return Error;
+
+  revng::FilePath IndexPath = Path.getFile(".index");
```
aleclearmind, line 28: The resume directory is already private, let's not use hidden files.

In general, let's never use hidden files unless there's a very good reason to.

python/revng/daemon/schema.graphql
mrjackv: I think there might be a misunderstanding: (in my understanding of the mechanism) there was never, and there will never be a separate index for invalidations. Invalidations happen because the context index has increased, period. The payload of an invalidation basically says to (e.g. `revng-ui`): "We're at global index $index, if you were at $index-1 here's the $invalidation list of target you need to invalidate to stay in sync"

lib/Pipeline/Context.cpp
mrjackv: Done

python/revng/daemon/__init__.py
mrjackv: Added docs

lib/Pipeline/Context.cpp
mrjackv: Fixed

lib/Pipeline/Context.cpp
mrjackv: Done

python/revng/daemon/schema.graphql
mrjackv: Done

include/revng/Pipeline/Context.h
aleclearmind: Yup

python/revng/daemon/schema.graphql
aleclearmind: I'm just saying, let's never talk about "invalidation event index", let's just talk about "globals commit index".

I understand they would go in parallel, but what's would be weird to me is that on the client the "invalidation index" on the third run of `revng-daemon` will not be `0` (which would make sense being the first "invalidation event" I get) but some other number.

include/revng/Pipeline/Context.h
mrjackv: Done

python/revng/daemon/schema.graphql
mrjackv: Ok, I've renamed the `index` field of the invalidation to `commitIndex` just to be extra-clear

aleclearmind:
Merged in 70344e0. Thanks.

PR #330, opened 2023-10-09
Tracing improvements

mrjackv:


include/revng/PipelineC/Tracing/Trace.h
```diff
@@ -124,6 +124,9 @@ struct RunTraceOptions {
   std::set<uint64_t> BreakAt;
   // If set, all temporary directories will be created within this directory
   std::string TemporaryRoot;
+  // Instead of using a temporary directory, the first invocations will use
+  // these directories instead
+  std::vector<std::string> ResumeDirectories;
```
aleclearmind, line 129: The fact that you can specify multiple resume directories (one for the first manager being created, one for the second) seems quite subtle.

I don't think anyone will use it, in practice. I'd go for the simpler "override all resume directories option", or we can assert if you specified the override directory and more than one manager is created.

mrjackv: The "override all resume directories option" could be very dangerous (think 2 managers running simultaneously)
I'll implement the second variant where it'll `revng_abort` if a second manager is created

mrjackv:
@aleclearmind I've noticed there was a change missing in this PR. It's contained in the last commit, named `Tracing/Runner: improve pointer vector handling`, which needs review.

lib/PipelineC/Tracing/Runner.cpp
```diff
@@ -32,13 +32,48 @@ static Logger<> TraceRunnerLogger("trace-runner");
 namespace utils {
 // Utility functions for decoding arguments
 
-// Put a nullptr at the end of the array and poison it, this will make it so
-// out-of-bounds errors are caught quickly
+// Utility class to store pointer arrays. When `.data()` is called, a nullptr is
+// added at the end of the array and it's poisoned by ASAN. This allows to spot
+// out of bounds accesses. After `.data()` has been called the internal vector
```
aleclearmind, line 37: ASan can already spot this kind of basic out of bounds accesses. [Godbolt](https://godbolt.org/z/98dqcnrYc)

mrjackv: Hmmm, the general idea is that:
* Without ASAN, we put `nullptr` so that it's extra obvious if there is an OOB
* With ASAN, unless it's poisoned it would mask off-by-ones

Alternatively, I can:
* revert this and add a `#if __has_feature(address_sanitizer)` that omits adding the nullptr so that ASAN catches it
* drop the nullptr business entirely, but (IMHO) it might make those bugs harder to spot

aleclearmind: Ah, it's for when we *don't* have ASan, got it.

Can you persuade me that this particular piece of code is more worthy of the extra love compared to, say, any `std::vector<T *>` in the rest of the codebase?
Is this FUD or did you have a real scenario where this bit you?

mrjackv: The origin of this implementation originates from the C implementation where padding the array with a `NULL` at the end was necessary to properly handle 0-length arrays. In this implementation it's recycled for 2 purposes:
* Have an obvious deference crash for out of bounds accesses
* Guarantee that even 0-sized arrays have a valid `.data()` pointer (in cppreference it says that it might be `nullptr` or it might be junk)

>  Is this FUD or did you have a real scenario where this bit you?

It didn't happen to me yet, but the spirit of `revng trace run` (in part) is to add extra checks so that easy-to-spot errors bubble up quickly (e.g. checking if the value of returns matches)

aleclearmind: OK

PR #331, opened 2023-10-17
`TypeSystemPrinter`: allow prototype-less functions

pfez:
Before this commit, the `TypeSystemPrinter` implicitly assumed that `model::Function` and `model::DynamicFunction` always had a prototype.
However, this is not guaranteed by the model schema.
This commit extends the `TypeSystemPrinter` so that it doesn't make that wrong assumption anymore, and can successfully print any `model::Function` and `model::DynamicFunction`, even when a prototype is not present.


pfez:
Merged in 43d6a2c5c5ec705b381bf6c0050ead3e540a046f !

PR #333, opened 2023-11-02
Convert all nullable type refs in model to `TupleTreeReference`

pfez:


pfez:
The branch passes `orc check-branch` and tests locally. The CI is running.

pfez:
The CI passes!

lib/Model/Binary.cpp
```diff
@@ -396,19 +396,35 @@ bool Function::verify(VerifyHelper &VH) const {
...
+  if (not StackFrameType().empty()) {
+
+    if (not StackFrameType().isValid())
+      return VH.fail("Invalid stack frame type", *this);
+
+    // The stack frame has a type
+
+    if (not isa<model::StructType>(StackFrameType().get())) {
+      return VH.fail("The stack frame type is not a StructType", *this);
+    }
```
aleclearmind, line 447: Drop `{ ... }`

lib/Model/Binary.cpp
```diff
@@ -366,6 +366,28 @@ bool Segment::verify(VerifyHelper &VH) const {
...
+
+    auto *Struct = dyn_cast<model::StructType>(Type().get());
+    if (not Struct)
+      return VH.fail("The segment type is not a StructType", *this);
+
+    if (VirtualSize() != Struct->Size()) {
+      return VH.fail(Twine("VirtualSize: ") + Twine(VirtualSize())
+                       + Twine(" != StackFrameType->Size: ")
+                       + Twine(Struct->Size()),
+                     *this);
```
aleclearmind, line 384: These are errors we expose to end users, let's use sentences, not C-like expressions.

lib/Model/TypeSystemPrinter.cpp
```diff
@@ -591,6 +592,64 @@ void TypeSystemPrinter::print(const model::DynamicFunction &F) {
...
 }
 
+void TypeSystemPrinter::dumpSegmentNode(const model::Segment &S, int NodeID) {
+  const model::Type *T = S.Type().empty() ? nullptr : S.Type().getConst();
+
+  // Print the name of the node
+  Out << "node_" << to_string(NodeID) << "[";
+
+  // Choose the node's border color
+  auto Color = Orange;
```
aleclearmind, line 602: [B.15](https://pad.rev.ng/s/SJNUet_ku#). Below too.

lib/Model/Pass/PurgeUnnamedAndUnreachableTypes.cpp
```diff
@@ -76,14 +76,26 @@ void model::purgeTypesImpl(TupleTree<model::Binary> &Model) {
...
+    for (const auto &Segment : Model->Segments())
+      if (not Segment.Type().empty())
+        ToKeep.insert(const_cast<Type *>(Segment.Type().get()));
+
     for (UpcastablePointer<model::Type> &T : Model->Types()) {
       TypeToNode[T.get()] = TypeGraph.addNode(NodeData{ T.get() });
     }
+
+    if (not Model->DefaultPrototype().empty())
+      ToKeep.insert(const_cast<Type *>(Model->DefaultPrototype().get()));
```
aleclearmind, line 98: I think the whole semantics of the `PruneAllUnusedTypes == true` code paths is dubious and should be dropped or reviewed.

In any case, this change is not probably what we want. It tries to mirror (by hand) what we already do (automatically) below in the `!PruneAllUnusedTypes` code path (the `visitTupleExcept` business).

If you want this branch to go in, leave this commit out.

aleclearmind (changes requested):

lib/Model/Pass/PurgeUnnamedAndUnreachableTypes.cpp
pfez: Eh without this commit the tests don't pass, because types are purged too aggressively and we're left with a model that doesn't verify with the new stricter version of `verify` that this patchset introduces.

I realize the proposed change is not correct, but at the moment I'm very overcommitted with other more pressing stuff. Could you please try and remove the commit and take a look yourself when you have time?

Otherwise I'm gonna have to leave this sit there for a while before I can look at it.

Thanks

lib/Model/TypeSystemPrinter.cpp
pfez: Right. Looking around I've seen that this file was pretty `auto`-heavy, and I've converted a bunch of them to using explicit types.

lib/Model/Pass/PurgeUnnamedAndUnreachableTypes.cpp
pfez: I see that you've pushed a commit to fix this. I'm closing the conversation.

aleclearmind:
Reminder that there are minor issues here that prevent merging.

pfez:
> Reminder that there are minor issues here that prevent merging.

I fixed them. The only thing left is the `auto`-heaviness, which I fixed but I'd like you to take a look at the fix like I asked in the conversation.

Anything else?

PR #334, opened 2023-11-03
OpaqueFunctionsPool: Remove the assert

dwarfyresearch:


pfez:
I'm closing this.

This patch was broken and its only purpose was to hide a bug in another Pull Request in `revng-c`. I'm reviewing that one and I have fixed the actual problem.

PR #335, opened 2023-11-06
Fix: isolated functions targets overappresented

drblallo:
Isolated functions no longer show up in the content of the pipeline multiple times equal to the number of tags.


aleclearmind:
Merged in b0614c3. Thanks.

PR #336, opened 2023-11-08
Hub artifacts

mrjackv:


python/revng/daemon/__init__.py
aleclearmind, line 1: Very nice

python/CMakeLists.txt
```diff
@@ -169,3 +182,25 @@ python_module(
...
+#
+configure_file(
+  pyproject.toml "${CMAKE_BINARY_DIR}/${PYTHON_INSTALL_PATH}/pyproject.toml"
+  COPYONLY)
+
+add_custom_command(
+  OUTPUT "${CMAKE_BINARY_DIR}/revng-1.0.0-py3-none-any.whl"
+  COMMAND python -m build --no-isolation --wheel --outdir "${CMAKE_BINARY_DIR}"
+          &> /dev/null
+  DEPENDS "${CMAKE_BINARY_DIR}/${PYTHON_INSTALL_PATH}/pyproject.toml"
```
aleclearmind, line 197: As discussed, let's have two wheels. One internal and the other one public.

aleclearmind (changes requested):

python/CMakeLists.txt
mrjackv: Done

python/CMakeLists.txt
```diff
@@ -18,118 +23,185 @@ function(python_module)
...
-      FILES "${CMAKE_BINARY_DIR}/${PYTHON_INSTALL_PATH}/${MODULE_FILE}"
-      DESTINATION
-        "${CMAKE_INSTALL_PREFIX}/${PYTHON_INSTALL_PATH}/${MODULE_FILE_DIR}")
+  foreach(MODULE_FILE ${PYTHON_MODULE_MODULE_GENERATED_FILES})
+    add_custom_command(
+      OUTPUT
+        "${CMAKE_BINARY_DIR}/build-wheel/${PYTHON_MODULE_WHEEL}/${MODULE_FILE}"
+      DEPENDS "${CMAKE_BINARY_DIR}/${PYTHON_INSTALL_PATH}/${MODULE_FILE}"
+      COMMAND
+        ${CMAKE_COMMAND} -E copy
```
aleclearmind, line 41: I would just use `cp`. We don't really support anything non-Linux for building as of now and we're unlikely to ever support building on Windows.

python/CMakeLists.txt
```diff
@@ -146,26 +218,39 @@ endforeach()
...
 #
-set(REVNG_DAEMON_MODULE_FILES
-    revng/daemon/__init__.py revng/daemon/multiqueue.py
-    revng/daemon/event_manager.py revng/daemon/schema.graphql
-    revng/daemon/graphql.py revng/daemon/util.py)
-python_module(TARGET_NAME revng-python-daemon MODULE_FILES
+file(
+  GLOB REVNG_DAEMON_MODULE_FILES
+  RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}"
+  revng/internal/daemon/*.py revng/internal/daemon/*.graphql)
```
aleclearmind, line 224: Never use `GLOB`.
Developers must know that they have to list files if they want them to be picked up.
If you use `GLOB` you don't need to *but* you need to re-run configure which is very confusing.
Also, `GLOB` picks up scratch files you might have around.

No `GLOB`.

python/wheel_install.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
aleclearmind, line 1: Use `-` (`wheel-install.sh`), unless there's a reason not to.

python/pyproject_revng_internal.toml
```diff
@@ -0,0 +1,46 @@
+# This file is distributed under the MIT License. See LICENSE.md for details.
```
aleclearmind, line 1: What's the `pyproject_` prefix for?

python/CMakeLists.txt
```diff
@@ -18,118 +23,185 @@ function(python_module)
...
+      COMMAND
+        ${CMAKE_COMMAND} -E copy
+        "${CMAKE_BINARY_DIR}/${PYTHON_INSTALL_PATH}/${MODULE_FILE}"
+        "${CMAKE_BINARY_DIR}/build-wheel/${PYTHON_MODULE_WHEEL}/${MODULE_FILE}")
+    list(
+      APPEND "WHEEL_FILES-${PYTHON_MODULE_WHEEL}"
+      "${CMAKE_BINARY_DIR}/build-wheel/${PYTHON_MODULE_WHEEL}/${MODULE_FILE}")
   endforeach()
+
+  set("WHEEL_FILES-${PYTHON_MODULE_WHEEL}"
```
aleclearmind, line 49: Do not use `-` in variable names, use `_`.

python/CMakeLists.txt
```diff
@@ -2,14 +2,19 @@
 # This file is distributed under the MIT License. See LICENSE.md for details.
 #
 
-# Declares a python module as a target Named arguments: - TARGET_NAME: the name
-# of the CMake target to be created - MODULE_FILES: list of relative paths of
-# the python module sources - MODULE_GENERATED_FILES: list of relative paths of
-# the python module generated sources The module sources will be copied to the
-# build directory (in the lib/python folder) and installed in /lib/python.
+# ~~~
```
aleclearmind, line 5: What are these? I'd leave an empty comment line if you want to better isolate this part.

aleclearmind (changes requested):

python/CMakeLists.txt
mrjackv: That's a fence for `cmakeformat` (See [README](https://github.com/cheshirekow/cmake_format), search for 'fences:'). It avoids the whitespace being collapsed, leading to the previous state of the comments where the bullet point list was squashed

python/pyproject_revng_internal.toml
mrjackv: The "correct" name of the file is `pyproject.toml`. Since we need the `revng` and `revng` internal version, those have been used as a postfix.

python/CMakeLists.txt
aleclearmind: Right right, I forgot.

python/pyproject_revng_internal.toml
mrjackv: That's not the structure that python expects. for both packages the structure before running `python -m build` should be:
```
drwxrwxr-x python
-rw-r--r-- pyproject.toml
```

aleclearmind: Ah OK, the conventions are important for the CMake magic to happen, correct?
CMake looks up `pyproject_$X.toml` where `$X` is the (normalized) name of the package?

mrjackv: Correct on both

python/CMakeLists.txt
mrjackv: Replaced with `cp`

python/CMakeLists.txt
mrjackv: Dropped `GLOB`

python/wheel_install.sh
mrjackv: Done

python/CMakeLists.txt
mrjackv: Done

aleclearmind:
Merged in c895d53. Thanks.

PR #337, opened 2023-11-09
UI QA Fixes

mrjackv:


include/revng/Model/RawFunctionType.h
```diff
@@ -21,7 +21,7 @@ inherits: Type
   - name: ReturnValues
     sequence:
       type: SortedVector
-      elementType: TypedRegister
+      elementType: NamedTypedRegister
```
aleclearmind, line 24: I assume in revng-c we have the corresponding stuff to 1) emit the artificial struct with the proper names and 2) have a way to rename it from the UI.

Correct?

mrjackv: 1. Yes
2. Yes

aleclearmind:
Merged in 9fe7b86. Thanks.

PR #338, opened 2023-11-10
Add msvc support to the abi testing setup

Cvelth:
A sister PR for https://github.com/revng/revng-qa/pull/20


tests/abi/tools/revng-ensure-rft-equivalence/Main.cpp
```diff
@@ -216,7 +216,7 @@ int main(int Argc, char *Argv[]) {
   LeftModel.replaceReferences(LeftReplacements);
   RightModel.replaceReferences(RightReplacements);
 
-  // Remove all the dynamic functions so that they don't interfere
+  // Remove all the dynamic functions so that they don't interfere.
```
aleclearmind, line 242: Ehr, we try to go the other way around. Trailing `.` in single-line comments should be banned.

lib/ABI/Analyses/ConvertToCABIFunctionType.cpp
```diff
@@ -126,10 +126,10 @@ checkVectorRegisterSupport(model::VerifyHelper &VH, const model::Type &Type) {
     revng_abort("Unknown type.");
   }
 
-  if (Result)
+  if (Result) {
     VH.setVerified(&Type);
-
-  VH.verificationCompleted(&Type);
+    VH.verificationCompleted(&Type);
+  }
```
aleclearmind, line 132: What's going on? The commit is not very descriptive.

tests/abi/tools/revng-ensure-rft-equivalence/Main.cpp
```diff
@@ -177,6 +177,28 @@ int main(int Argc, char *Argv[]) {
     Iterator->second.Right = Right->key();
   }
 
+  // Deduplicate the list of IDs to replace.
```
aleclearmind, line 180: Why do we need to do this?

tests/abi/tools/revng-ensure-rft-equivalence/Main.cpp
```diff
@@ -243,6 +243,13 @@ int main(int Argc, char *Argv[]) {
   LeftModel->ImportedDynamicFunctions().clear();
   RightModel->ImportedDynamicFunctions().clear();
 
+  // Remove all the functions starting with `__` as well.
+  auto StartsWithDoubleUnderscore = [](const model::Function &Function) {
+    return llvm::StringRef(Function.OriginalName()).substr(0, 2) == "__";
+  };
+  llvm::erase_if(LeftModel->Functions(), StartsWithDoubleUnderscore);
+  llvm::erase_if(RightModel->Functions(), StartsWithDoubleUnderscore);
+
```
aleclearmind, line 252: Can't we convert all the types that are directly referenced by a `model::Function` instead?

share/revng/test/tests/abi/verify-abi-test-binaries.sh
```diff
@@ -28,6 +28,7 @@ revng \
     analyze \
     ImportBinary \
     "$BINARY" \
+    --use-pdb="${BINARY}.pdb" \
```
aleclearmind, line 31: What's the effect of this?
What was the problem before?

aleclearmind (changes requested):

lib/ABI/Analyses/ConvertToCABIFunctionType.cpp
Cvelth: The same thing was silently changed in the `model::Type::size` this is modeled after some time in the past, so this coroutine was silently broken.

tests/abi/tools/revng-ensure-rft-equivalence/Main.cpp
Cvelth: Because after importing PDBs we end up with the same prototype (function type) used for multiple functions. And since this is gathered function-wise, we end up with weirdness because it tries to replace the same multiple times.

share/revng/test/tests/abi/verify-abi-test-binaries.sh
Cvelth: > What's the effect of this?

This flag explicitly says "if you need a `pdb` when importing a PE binary, instead of the encoded path (build directory of `test/revng-qa`) look wherever the binary is - the `pdb` is named the same but with a suffix".

> What was the problem before?

Since this was only ever working on `elf` binaries, `pdb`s were not relevant.

Cvelth:
> Last commit is somehow empty

cc: @aleclearmind

I intentionally made it so. This way I won't forget about it. I didn't want to update docs before the review in case we end up changing something major or whatever. but it doesn't look like there's anything big in here, so it can be filled in now.

tests/abi/tools/revng-ensure-rft-equivalence/Main.cpp
aleclearmind: Why? Add a comment at least.

Cvelth: > Why?

I don't remember specifically. But I feel like at the time this code was written, `model diff` was the hip new thing so we were kind of pushing for it wherever applicable.

That being said, I'd say it fits here pretty well, aside from having to deal with a couple of weird cases we don't care about like this one, it does indeed do all the heavy lifting (and produces pretty nice error message as a bonus).

tests/abi/tools/revng-ensure-rft-equivalence/Main.cpp
aleclearmind: Comment

tests/abi/tools/revng-ensure-rft-equivalence/Main.cpp
aleclearmind: OK, but we can't leave `llvm::erase_if(LeftModel->Functions(), StartsWithDoubleUnderscore)` without an explanation on why we're dropping those.

Cvelth: Yeah, my bad on the explanation.

tests/abi/tools/revng-ensure-rft-equivalence/Main.cpp
Cvelth: (I just realized I forgot to close a couple of threads on Friday)

This one is done.

aleclearmind (commented):
NOTETOSELF: resume from "ABI: introduce a way to specify shadow word count ".

include/revng/ABI/Definition.h
```diff
@@ -152,6 +152,16 @@ type: struct
       ARM ABI), the value of `StackAlignment` should be equal to 4.
     type: uint64_t
 
+  - name: StackShadowWordCount
+    doc: |
+      States the number of words reserved for the callee to be able to "push"
+      register arguments in.
```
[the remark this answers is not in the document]
Cvelth: I guess "push" is a wrong word to use here.

The idea is that all the register arguments are also _technically_ passed on the stack too. As in, the space for them is allocated, but the values are not put there - the callee is free to do that if it needs to.

So, in short, these 4 words "shift" all the other stack arguments.

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
```diff
@@ -446,21 +446,16 @@ getPrototypeLayout(const model::Function &Function, model::ABI::Values ABI) {
...
-
-    if (ABI != PrototypeCopy.ABI()) {
-      // Workaround for dwarf sometimes misdetecting a very specific ABI,
-      // for example, it labels `SystemV_x86_regparam_N` as just `SystemV_x86`
-      // despite them being incompatible.
-      std::array AllowedABIs = { model::ABI::SystemV_x86_regparm_3,
-                                 model::ABI::SystemV_x86_regparm_2,
-                                 model::ABI::SystemV_x86_regparm_1 };
-      revng_assert(llvm::is_contained(AllowedABIs, ABI));
-      PrototypeCopy.ABI() = ABI;
```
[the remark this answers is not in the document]
Cvelth: Yeah, this is not good news, just my bad. Last time when we discussed this and came up with a [cleaner solution](https://github.com/revng/revng/blob/develop/share/revng/test/tests/abi/replace-abi.py), I somehow missed this and it also got merged. Practically, this was a no-op all along (since all the cases it is designed to handle were already handled by the python script).

include/revng/ABI/FunctionType/Support.h
```diff
@@ -39,6 +39,9 @@ inline constexpr uint64_t paddedSizeOnStack(uint64_t RealSize,
                                             uint64_t RegisterSize) {
   revng_assert(llvm::isPowerOf2_64(RegisterSize));
 
+  if (RealSize == 0)
+    return RealSize;
```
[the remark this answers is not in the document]
Cvelth: I added this check because I was definitely hitting this branch somewhere.

But, thinking on it now, I was probably due to `void` arguments added by PE importer. Now that @dwarfyresearch fixed those, this could probably became an assert.

lib/ABI/Definition.cpp
```diff
@@ -374,6 +374,14 @@ std::optional<uint64_t> Definition::alignment(const model::QualifiedType &QT,
...
     return std::nullopt;
 }
+std::optional<uint64_t> Definition::alignment(const model::Type &Type,
+                                              AlignmentCache &Cache) const {
+  std::optional<AlignmentInfo> Result = naturalAlignment(*this, Type, Cache);
+  if (Result.has_value() && Result->Value != 0)
+    return Result->IsNatural ? Result->Value : 1;
+  else
+    return std::nullopt;
+}
```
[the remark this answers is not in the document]
Cvelth: f... Looks like a rebase went bad at some point again. I'll redo this commit.

share/revng/test/tests/model/import/pe/main.c.model.yml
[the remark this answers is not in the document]
Cvelth: Before now they were only used for 32-bit. Since this seemed like the way to make it work (tm) again with minimal effort. If you have something better in mind, me (and I'm sure @dwarfyresearch too) are open to hear it.

include/revng/ABI/Definition.h
```diff
@@ -152,6 +152,16 @@ type: struct
...
+
+      In out context, this serves as the starting offset for the first stack
```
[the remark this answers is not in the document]
Cvelth: Done.

lib/Model/Importer/DebugInfo/PDBImporter.cpp
```diff
@@ -289,6 +289,7 @@ void PDBImporterImpl::run(NativeSession &Session) {
   deduplicateEquivalentTypes(Importer.getModel());
   promoteOriginalName(Importer.getModel());
   purgeUnnamedAndUnreachableTypes(Importer.getModel());
+  fixModel(Importer.getModel());
```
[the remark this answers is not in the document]
Cvelth: cc: @dwarfyresearch

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
```diff
@@ -213,22 +213,36 @@ VH::LeftToVerify VH::verifyAnArgument(const runtime_test::State &State,
...
+        if (Remaining.Stack.size() < ABI.getPointerSize())
+          Remaining.Stack = {};
+        else
+          Remaining.Stack = Remaining.Stack.drop_front(ABI.getPointerSize());
+      }
 
-    if (tryToVerifyStack(Remaining.Stack, ArgumentBytes))
-      break; // Stack matches after accounting for alignment.
+      if (tryToVerifyStack(Remaining.Stack, ArgumentBytes))
+        break; // Stack matches after accounting for alignment.
```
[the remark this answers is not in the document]
Cvelth: Done.

lib/ABI/Definition.cpp
```diff
@@ -384,5 +392,14 @@ Definition::hasNaturalAlignment(const model::QualifiedType &QT,
...
 }
+std::optional<bool>
+Definition::hasNaturalAlignment(const model::Type &Type,
+                                AlignmentCache &Cache) const {
+  std::optional<AlignmentInfo> Result = naturalAlignment(*this, Type, Cache);
+  if (Result.has_value() && Result->Value != 0)
+    return Result->IsNatural;
+  else
+    return std::nullopt;
+}
```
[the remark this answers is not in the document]
Cvelth: Done.

lib/ABI/Definition.cpp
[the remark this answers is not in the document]
Cvelth: Should be fixed, but please double-check.

lib/Model/Importer/DebugInfo/PDBImporter.cpp
```diff
@@ -172,12 +180,39 @@ void PDBImporterImpl::populateTypes() {
...
+      auto OrignalModelType = ProcessedTypes[OriginalType];
+
+      auto CheckKey =
+        [&ForwardModelTypeKey](const model::TypePath &Reference) -> bool {
+        if (Reference.empty())
+          return false;
+
+        return ForwardModelTypeKey == Reference.getConst()->key();
+      };
+      Importer.getModel().replaceReferencesIf(OrignalModelType, CheckKey);
```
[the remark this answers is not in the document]
dwarfyresearch: Well, it does make sense! I will try to do that way.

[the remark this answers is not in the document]
dwarfyresearch: >You create a type for A (not sure which one).

I use an ugly hack for it: As a place holder type, I create a pointer to `void`, but the size of the pointer is the `sizeof` original type - and it can be an arbitrary `sizeof(a struct)`. It is important that I have the correct size for it, since I can still manually calculate the size of structs for the Model if it depends on the forward reference, and that is why I invoke `fixModel` and friends to clean up it later (since it is obviously wrong to have a pointer of size `X` in the Model).

lib/Model/Importer/DebugInfo/PDBImporter.cpp
```diff
@@ -668,6 +718,8 @@ Error PDBImporterTypeVisitor::visitKnownRecord(CVType &Record,
         revng_log(DILogger,
                   "LF_STRUCTURE: Unknown field type "
                     << Field.getType().getIndex());
+        // Avoid incomplete struct types.
+        return Error::success();
```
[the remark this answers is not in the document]
dwarfyresearch: Hmm, I will check, but not sure how is this useful

lib/Model/Importer/DebugInfo/PDBImporter.cpp
```diff
@@ -172,12 +180,39 @@ void PDBImporterImpl::populateTypes() {
...
     consumeError(std::move(Err));
+  } else {
+    // Post process the forward references.
+    for (auto &TypeInfo : ForwardReferencedTypes) {
+      auto ForwardDeclaration = TypeInfo.first;
+      auto OriginalType = TypeInfo.second;
+
+      if (ProcessedTypes.count(OriginalType) == 0
+          or ProcessedTypes.count(ForwardDeclaration) == 0)
+        continue;
```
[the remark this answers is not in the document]
dwarfyresearch: At least, it should `revng_log`.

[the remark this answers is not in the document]
aleclearmind: Let me rephrase it: is there a reason why this is not an assertion (other than FUD)?

lib/Model/Importer/DebugInfo/PDBImporter.cpp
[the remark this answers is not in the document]
aleclearmind: > size of the pointer is the sizeof original type

Mmh, this is not OK. Please see if what I suggested above can save us from doing this.
If not, let's figure out a proper solution.

In any case, when you leave such hacks, please leave a thorough comment explaining what's going on and why.

lib/Model/Importer/DebugInfo/PDBImporter.cpp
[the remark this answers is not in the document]
aleclearmind: It's useful if in the original file you had a incomplete type.

Consider the following code:

```c
typedef void FILE;
// Or simply
// struct FILE;

void my_function(FILE *);
```

We should be able to handle this, it's legitimate code that can lead to legitimate debug information.

share/revng/test/tests/model/import/pe/main.c.model.yml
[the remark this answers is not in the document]
aleclearmind: It works for me, just let's not let the reader figure it out and let's leave an explicit note.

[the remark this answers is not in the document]
Cvelth: Sounds good.

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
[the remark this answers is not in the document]
aleclearmind: Uhm OK.

tests/abi/tools/revng-ensure-rft-equivalence/Main.cpp
aleclearmind: Ping?

tests/abi/tools/revng-ensure-rft-equivalence/Main.cpp
aleclearmind: Ping?

include/revng/ABI/Definition.h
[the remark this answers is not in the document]
Cvelth: From the [docs]():

> The x64 Application Binary Interface (ABI) uses a four-register fast-call calling convention by default. Space is allocated on the call stack as a shadow store for callees to save those registers.

...

> Any parameters beyond the first four must be stored on the stack after the shadow store before the call

---

In practice (from the outputs of our testing suite), these are no different from usual stack aruguments, except we're not interested in their values (which look like ~~garbage~~ whatever was previously written there)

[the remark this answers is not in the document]
Cvelth: TL;DR: this parameter is literally "Ignore the first X words of the stack argument struct - they contain garbage".

Luckily, it's always [four](https://devblogs.microsoft.com/oldnewthing/20160623-00/?p=93735), which makes our (and apparently their, we're not the only ones who have trouble with `vararg` functions) lives _so_ much easier!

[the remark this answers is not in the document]
aleclearmind: Let's make this more explicit in the comment and then OKGO.

lib/Model/Importer/DebugInfo/PDBImporter.cpp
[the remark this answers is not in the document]
aleclearmind: Uhm OK.

include/revng/ABI/Definition.h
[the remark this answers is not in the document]
Cvelth: Sounds good.

lib/Model/Importer/DebugInfo/PDBImporter.cpp
```diff
@@ -172,12 +180,39 @@ void PDBImporterImpl::populateTypes() {
...
-                                     ProcessedTypes);
+                                     ProcessedTypes,
+                                     ForwardReferencedTypes,
+                                     *StreamTpiOrErr);
   if (auto Err = visitTypeStream(InputFile->types(), TypeVisitor)) {
     revng_log(DILogger, "Error during visiting types: " << Err);
     consumeError(std::move(Err));
+  } else {
+    // Post process the forward references.
+    for (auto &TypeInfo : ForwardReferencedTypes) {
```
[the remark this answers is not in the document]
dwarfyresearch: I think I wont need this anymore.

lib/Model/Importer/DebugInfo/PDBImporter.cpp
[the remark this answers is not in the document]
dwarfyresearch: I think I wont need this anymore.

lib/Model/Importer/DebugInfo/PDBImporter.cpp
```diff
@@ -172,12 +180,39 @@ void PDBImporterImpl::populateTypes() {
...
+      auto ForwardDeclaration = TypeInfo.first;
+      auto OriginalType = TypeInfo.second;
+
+      if (ProcessedTypes.count(OriginalType) == 0
+          or ProcessedTypes.count(ForwardDeclaration) == 0)
+        continue;
+
+      auto ForwardModelType = ProcessedTypes[ForwardDeclaration];
+      auto ForwardModelTypeKey = ForwardModelType.get()->key();
+      auto OrignalModelType = ProcessedTypes[OriginalType];
```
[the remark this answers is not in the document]
dwarfyresearch: I think I wont need this anymore.

share/revng/test/tests/model/import/pe/main.c.model.yml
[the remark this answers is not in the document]
Cvelth: Done:
```yaml
#
# NOTE: this file is stripped of any architecture-specific information (things
#       like ABI, pointer sizes, and so on): it reflects all the _common_
#       expectations for a specific source file (see revng-qa-side importPE
#       sources) after being compiled and lifted by a variety of different
#       msvc compiler versions.
#
```

tests/abi/tools/revng-ensure-rft-equivalence/Main.cpp
Cvelth: Done:
```cpp
  // We have to do this because in case a single prototype is used for multiple
  // functions (common in binaries lifted from PE/COFF) we still only want to
  // "replace" it once.
```

tests/abi/tools/revng-ensure-rft-equivalence/Main.cpp
Cvelth: Since then, I ran into similar trouble again, so it's now even stricter:
```cpp

  // Remove all the functions that are not strictly related to the test.
  auto ShouldRemove = [](const model::Function &Function) {
    llvm::StringRef Name = Function.OriginalName();
    if (Name.take_front(5) == "test_")
      return false;
    if (Name.take_front(6) == "setup_")
      return false;
    if (Name == "main")
      return false;
    return true;
  };
  llvm::erase_if(LeftModel->Functions(), ShouldRemove);
  llvm::erase_if(RightModel->Functions(), ShouldRemove);
```

Do you feel like this version also needs an even more explicit comment? (because I feel like lambda being right there makes it explicit enough)

include/revng/ABI/Definition.h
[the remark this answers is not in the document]
Cvelth: I went with
```
      States the number of words reserved for the callee to be able to "mirror"
      register arguments on the stack. This value should be set to `0` for any
      ABI that does not such space.

      In our context, this serves as the starting offset for the first stack
      argument. As in, it specifies the number of words on the top of the stack
      struct that should be ignored.
```

tests/abi/tools/revng-ensure-rft-equivalence/Main.cpp
aleclearmind: Works

share/revng/test/tests/model/import/pe/main.c.model.yml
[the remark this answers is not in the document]
aleclearmind: Works

include/revng/ABI/Definition.h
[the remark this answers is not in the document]
aleclearmind: Works

lib/Model/Importer/DebugInfo/PDBImporter.cpp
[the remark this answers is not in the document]
Cvelth: @aleclearmind, I just pushed a new patch from @dwarfyresearch. Please take a look when you have a moment.

(`PDBImport: Handle forward referenced types` commit)

lib/ABI/FunctionType/ValueDistributor.h
```diff
@@ -20,16 +20,30 @@ struct DistributedValue {
...
-  /// The total size of the argument (including padding if necessary) in bytes.
+  /// The total size of the argument in bytes WITHOUT any padding.
   uint64_t Size = 0;
 
+  /// The total size of padding that has to be added to the previous argument
+  /// because of alignment specifics.
+  uint64_t PrePaddingSize = 0;
+
+  /// The total size of padding that has to be added to the current argument.
+  uint64_t PostPaddingSize = 0;
```
aleclearmind, line 31: These fields are set but never read. What's their purpose?

lib/ABI/FunctionType/ValueDistributor.cpp
```diff
@@ -232,6 +232,11 @@ DistributedValues ArgumentDistributor::positionBased(bool IsFloat,
 
   DistributedValue Result;
   Result.Size = Size;
+  if (Result.Size > ABI.getPointerSize()) {
+    Result.Size = ABI.getPointerSize();
+    Result.UsesPointerToCopy = true;
+    IsFloat = false; // Pointers never use vector registers.
```
aleclearmind, line 238: Comments go on their own line

include/revng/ABI/Definition.h
```diff
@@ -466,10 +466,12 @@ class Definition : public generated::Definition {
...
     for (auto Register : GPRs)
       if (Lookup.contains(Register))
-        Result.emplace_back(Register);
+        if (!llvm::is_contained(Result, Register))
+          Result.emplace_back(Register);
     for (auto Register : VectorArgumentRegisters())
       if (Lookup.contains(Register))
-        Result.emplace_back(Register);
+        if (!llvm::is_contained(Result, Register))
+          Result.emplace_back(Register);
```
aleclearmind, line 472: The message commit is quite vague, what faulty situation are we trying to prevent?
Can you make an example?

lib/ABI/Analyses/ConvertToCABIFunctionType.cpp
```diff
@@ -189,16 +189,36 @@ class ConvertToCABIFunctionType {
...
+        // model::ABI::AAPCS64,
+
+        // There are known issues
+        // model::ABI::SystemV_MIPS_o32,
+        // model::ABI::SystemV_MIPSEL_o32
+
+        // Unable to reliably test: no easy access to a compiler
+        // model::ABI::Microsoft_x86_64_clrcall,
+        // model::ABI::Microsoft_x86_clrcall,
+        // model::ABI::Pascal_x86,
```
aleclearmind, line 218: We don't have a compiler for PASCAL?

lib/ABI/Analyses/ConvertToCABIFunctionType.cpp
```diff
@@ -189,16 +189,36 @@ class ConvertToCABIFunctionType {
...
+        // There are known issues
+        // model::ABI::AAPCS64,
+
+        // There are known issues
+        // model::ABI::SystemV_MIPS_o32,
+        // model::ABI::SystemV_MIPSEL_o32
+
+        // Unable to reliably test: no easy access to a compiler
+        // model::ABI::Microsoft_x86_64_clrcall,
+        // model::ABI::Microsoft_x86_clrcall,
```
aleclearmind, line 217: Please let's purge untested things, I don't think we want to support these for now.

share/revng/rcc-config.yml
```diff
@@ -131,6 +130,7 @@ read_passes:
               - "<="
               - '//\s*WIP'
               - '#if\s*[01]'
+              - "#pragma clang optimize off"
```
aleclearmind, line 133: Why do we need this?

lib/ABI/Definition.cpp
```diff
@@ -366,40 +366,53 @@ naturalAlignment(const abi::Definition &ABI,
   rc_return rc_recur naturalAlignment(ABI, *QT.UnqualifiedType().get(), Cache);
 }
 
+template<typename T>
+void abortIfFalse(bool Success, const T &Type) {
```
aleclearmind, line 370: `s/abortIfFalse/assertOnType/`?

include/revng/ABI/Definition.h
```diff
@@ -141,6 +141,11 @@ type: struct
       call. If equal to `true`, it's the callee, otherwise it the caller.
     type: bool
 
+  - name: FloatsAlwaysUseGenericRegisters
+    doc: |
+      Setting this to true disables usage of vector registers entirely.
+    type: bool
```
aleclearmind, line 136: This sounds like a property of the architecture, not of the ABI.
We should also check if this is true for other parts of `Definition.h`.

tests/unit/Alignment.cpp
```diff
@@ -43,6 +43,18 @@ void testAlignment(const model::QualifiedType &Type,
...
+bool ABIhasNeededScalars(const abi::Definition &ABI, Types... Values) {
+  return (llvm::is_contained(ABI.ScalarTypes(), Values) && ...);
+}
+
+template<typename... Types>
+  requires(std::convertible_to<Types, uint64_t> && ...)
+bool ABIhasNeededFloats(const abi::Definition &ABI, Types... Values) {
+  return (llvm::is_contained(ABI.FloatingPointScalarTypes(), Values) && ...);
+}
+
```
aleclearmind, line 57: Please describe what's going on in this commit, I'm confused.

aleclearmind (changes requested):

aleclearmind:
* `ToRaw: take advantage of the updated distributor`: what changes specifically? Consider squashing commits.ù
* `ABI: improve mips ABIs`: MIPS is upper case.
*  ABI-tests: check equivalence after compatibility: commit message `s/chance/change/`.

tests/unit/Alignment.cpp
Cvelth: If you remember, scalar alignment requirements are now a lot stricter. Which lead this test to fail.

This is pretty much some extra machinery to be able to express "if ABI has no such prototype, skip a test".

include/revng/ABI/Definition.h
Cvelth: It is a property of the ABI. MIPS spec clearly describes the vector register usage. Except GCC doesn't care (I guess their MIPS ABI implementation is about as good as their Microsoft ones is), so this is pretty much a workaround to tell "hey, ignore the vector data". I think it's viable to add this as an ABI configuration option though. We might also want to reuse for stuff like "soft-float" ARM or some vector-less i386 ABIs. You get the point.

lib/ABI/Analyses/ConvertToCABIFunctionType.cpp
Cvelth: I just listed them here to keep track of what we are testing and what we don't. I think there's a value in explicitly specifying "do not rely to much on this if you use X ABI" in the body of the analysis.

lib/ABI/Analyses/ConvertToCABIFunctionType.cpp
Cvelth: Wait, we do? Microsoft one dropped the support (maybe we could try to use some old one, but I have doubts).

And, in my experience, gcc's support of all the "foreign" ABIs is _really_ bad.

include/revng/ABI/Definition.h
Cvelth: The case where the same register is mentioned more than once. I found out about this limitation when trying to encode `gcc's mips` ABI before adding `FloatsAlwaysUseGenericRegisters` flag.

But also, consider, that this is a register-_sorting_ method. I don't see a reason for it to complain about duplication. Let ABI definition specify whichever registers it wants - all this cares about is their _order_.

include/revng/Model/ABI.h
```diff
@@ -295,15 +311,24 @@ inline constexpr llvm::StringRef getDescription(model::ABI::Values V) {
     return "32-bit Microsoft x86 abi that was intended to "
            "mimic 32-bit SystemV x86 abi but has minor "
            "differences";
+  case model::ABI::Microsoft_x86_cdecl_gcc:
+    return "Turns out gcc's version of `cdecl` is subtly different from "
+           "the original";
```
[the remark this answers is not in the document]
Cvelth: 1. This is a comment that is specifically used by the CLI tools (the ones that list all the ABIs we support), it makes a lot more sense in context - where it comes up right after the relevant msvc abi.

```yaml
Tool options:

  --abi=<value>                  - Specifies the default ABI of the binary.
    =Invalid                     -   Unknown and/or unsupported ABI
    # . . .
    =Microsoft_x86_cdecl         -   32-bit Microsoft x86 abi that was intended to mimic 32-bit SystemV x86 abi but has minor differences
    =Microsoft_x86_cdecl_gcc     -   Turns out gcc's version of `cdecl` is subtly different from the original
```

2. This predates YAML. But, IMO, since we plan to get rid of this entire enum eventually, the problem with duplicated descriptions will go away on its own.

include/revng/Model/ABI.h
```diff
@@ -70,6 +70,10 @@ type: enum
       differences. The documentation can be found
       \sa https://docs.microsoft.com/en-us/cpp/cpp/cdecl
 
+  - name: Microsoft_x86_cdecl_gcc
+    doc: |
+      Turns out gcc's version of `cdecl` is subtly different from the original
```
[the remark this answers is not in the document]
Cvelth: Wait, how is mingw even related? This is specifically the result of adding `__attribute__((ms_abi,cdecl))` (and friends) to the function prototype.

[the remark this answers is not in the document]
Cvelth: > To be frank, this was a non-goal that has been pulled in without agreement.

This wasn't pulled in. This is the remainder from the previous PR where these were the only microsoft ABIs we _were_ testing.

2. Sadly, the real ones are too different, so it's impossible to keep them with the same name - hence "_gcc" suffix.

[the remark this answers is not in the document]
Cvelth: > how much does this slow down the testing pipeline

3. There are 3 of these. Out of 16 total (so 3/13 (20%)) slowdown _worst_ case (because x86 gcc ones are generally pretty fast, compared to stuff like ARM or, worse, msvc).

include/revng/ABI/Definition.h
```diff
@@ -152,6 +152,18 @@ type: struct
       ARM ABI), the value of `StackAlignment` should be equal to 4.
     type: uint64_t
 
+  - name: StackShadowWordCount
```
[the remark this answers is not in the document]
Cvelth: [MSVC](https://learn.microsoft.com/en-us/cpp/build/x64-calling-convention?view=msvc-170#calling-convention-defaults) describes them as:

> The x64 Application Binary Interface (ABI) uses a four-register fast-call calling convention by default. _Space_ is allocated on the call stack _as a shadow store_ for callees to save those registers.

> ...

> Any parameters beyond the first four must be stored on the stack after _the shadow store_ before the call

As for MIPS, it's mentioned as:

> Each stack frame has sufficient space allocated for:

> ...

> * function call argument area. In a non-leaf function the maximum number of bytes of arguments (ivan's comment: including register ones) used to call other functions from the non-leaf function must be allocated. However, _at least four words_ (16 bytes) must always be reserved, even if the maximum number of arguments to any called function is fewer than four words.

(emphasis mine)

So, apart from the `shadow` "word" nothing is really used to describe them.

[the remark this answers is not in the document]
Cvelth: Something like `StackBytesAllocatedForRegisterArguments` sounds good for me, to be honest, what do you think?

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
```diff
@@ -461,7 +463,28 @@ getPrototypeLayout(const model::Function &Function, model::ABI::Values ABI) {
...
+    if (ABI.StackShadowWordCount() != 0 && !Raw->StackArgumentsType().empty()) {
+      uint64_t SWSize = ABI.StackShadowWordCount() * ABI.getPointerSize();
+
+      const model::TypePath &StackArgTypeRef = Raw->StackArgumentsType();
+      auto &StackStruct = *llvm::cast<model::StructType>(StackArgTypeRef.get());
+      if (!StackStruct.Fields().empty())
+        if (StackStruct.Fields().begin()->Offset() < SWSize)
+          revng_abort("Stack arguments in the shadow space?");
+
+      auto &LastArgument = *std::prev(Result.Arguments.end());
```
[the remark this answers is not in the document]
Cvelth: It's a "tracking" `SortedVector`, as in, a container that mimics a `set` interface. And `set` has no "back".

tests/abi/tools/revng-check-compatibility-with-abi/Verify.cpp
```diff
@@ -461,7 +463,28 @@ getPrototypeLayout(const model::Function &Function, model::ABI::Values ABI) {
...
+      revng_assert(LastArgument.Stack && LastArgument.Stack->Offset == 0);
+      revng_assert(LastArgument.Stack->Size > SWSize);
+      LastArgument.Stack->Offset = SWSize;
+      LastArgument.Stack->Size -= SWSize;
+    }
```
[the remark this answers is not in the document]
Cvelth: 1. Yep.
2. Yep.
3. Yep.
4. Yep.
5. It knows the ABI, so it knows what that space is _supposed_ to be. As of now, it just silently skips it, but we can add some extra "sanity" checks in there (for example, that no fields in that area have any names/comments, you get the point).

> to be explicitly represented in `RawFunctionType`

I thought about it. But how do you detect it? The only places we can reliably tell what that space is those that _explicitly_ know what the ABI of the function is.

lib/ABI/FunctionType/Layout.cpp
```diff
@@ -437,32 +428,29 @@ Layout::Layout(const model::CABIFunctionType &Function) {
...
-                                               ArgumentType);
-        Current.Stack->Offset = CurrentStackOffset;
+        Current.Stack = Layout::Argument::StackSpan{
+          .Offset = Distributed.OffsetOnStack, .Size = Distributed.SizeOnStack
+        };
 
-        // And carry the size over unchanged.
-        Current.Stack->Size = Distributed.SizeOnStack;
-        CurrentStackOffset += Current.Stack->Size;
+        StackStructSize = Distributed.OffsetOnStack + Distributed.SizeOnStack;
```
Cvelth, line 454: > `ToRaw: take advantage of the updated distributor`: what changes specifically? Consider squashing commits.

@aleclearmind, specifically, it drops all the alignment-related computations that used to be done "at the leaves" and replaces them with data that's newly embedded into the `ArgumentDistributor` in of the previous commits.

include/revng/Model/ABI.h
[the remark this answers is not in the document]
aleclearmind: > This wasn't pulled in. This is the remainder from the previous PR where these were the only microsoft ABIs we were testing.

Right, but I thought we'd drop the old stuff.

In my mind, the GCC toolchain for Windows is MinGW, but yeah, technically it's a GCC feature, even if probably the mingw guys implemented it.

Anyway, let's keep it and enable testing.

Btw, can you make an example of a prototype that would not be compatible between GCC and MSVC and why?

include/revng/Model/ABI.h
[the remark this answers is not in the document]
aleclearmind: I'd go for something like `32-bit Microsoft x86 abi as implemented in GCC (subtly different from the MSVC one)`.

lib/ABI/FunctionType/ValueDistributor.h
aleclearmind: OK, can you add a comment stating that as of now these are mainly for debugging purposes?
If this change, we can drop it.

include/revng/ABI/Definition.h
aleclearmind: I'd prefer to assert and force callers to dedup things. If have 5 registers and I ask someone to sort them and I get back 4, I'd like an abort.

lib/ABI/Analyses/ConvertToCABIFunctionType.cpp
aleclearmind: I'd prefer to drop `clrcall` ABIs entirely, having them there without ever testing them is mostly noise.

include/revng/Model/ABI.h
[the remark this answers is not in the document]
Cvelth: > MS breaks ABI at every major release

I don't think it's the same _ABI_, that's talked about.

This means the binary-interface side of things. Stuff like gcc/clang's `std::regex` or `std::function` stuck in an inferior state because changing some fields would mean binaries compiled with an older version is no longer compatible.

We, on the other hand, only think about the "basic" ABI level, as in which registers are used, how is stack space allocated for arguments, how they aligned, etc. I can't imagine this changing much, if ever.

> I wonder if GCC is maybe implementing and older MSVC ABI version

I severely doubt it. GCC's implementations _looks_ very similar to those of `regparm`s, and in general, they are very close to natural GCC ABIs. I'd imagine that's because the implementation is very crude, but still works for basic use cases (like when you need to call a function or two passing a couple of ints or something - without anything non-trivial), so that's why it's kept. But yeah, that's just my speculation.

share/revng/rcc-config.yml
aleclearmind: This might get committed :-1:.
I know it's not the best but you can have a command that does `orc shell -c revng ctest -j16 && orc install test/revng` if you want to bypass `check-conventions`. Usually I don't run `ctest` an `test/revng` though, I'm usually interested only in one of the two.

include/revng/ABI/Definition.h
aleclearmind: 1. When do we use "vector" and when "float"? Are we being coherent?
2. Can we call this `PassFloatThroughGPRs`? "Generic registers" is not a term we use AFAIK.

include/revng/ABI/Definition.h
Cvelth: Actually, wait, yeah. That's my bad on the implementation (I just sticked "if contained" thing in without thinking). This needs a bit of a rewrite. The intention is to deduplicate the { `ReturnValueLocationRegister` + `GeneralPurposeArgumentRegisters` + `VectorArgumentRegisters` } set, not the callee's input.

Thanks for catching this!

lib/ABI/Analyses/ConvertToCABIFunctionType.cpp
Cvelth: Ah, you were talking about `clrcall`, not about the comments in itself. My bad then.

No complaints on dropping `clrcall`.

share/revng/rcc-config.yml
Cvelth: Gotcha.

include/revng/ABI/Definition.h
Cvelth: 1. `vector` is only ever applied to registers. It just happens that on all the major architectures "float"s re-use "vector" infrastructure. In this case we're talking about _float_ values using _generic_ registers, as opposed to _vector_ ones, which is why I explicitly mentioned _float_ which I only do when forced to.

> `PassFloatThroughGPRs`

2. My only complaint the `pass` term. Since it also applies to return values.

`AlwaysUseGPRsForFloats`?

tests/unit/Alignment.cpp
aleclearmind: 1. Use `SortedVector::contains` instead of `llvm::is_contained`.
2. Avoid templates, too big
3. Wouldn't a more appropriate name be `hasFloatOfSize(...)`? `ABIhasNeededFloats` does not resonate a lot with me.

Cvelth: 1. 👌
2. It's the usual dilemma of "do you want callee to be ugly? Or the caller?" and I prefer the callee. This let's caller just say
```cpp
if (ABIhasFloatsOfSizes(Size1, Size2, Size3))
  do_something(Size1, Size2, Size3);
```

instead of doing ugly stuff like
```cpp
if (ABIhasFloatOfSize(Size1))
  if (ABIhasFloatOfSize(Size2))
    if (ABIhasFloatOfSize(Size3))
      do_something(Size1, Size2, Size3);
```

3. `ABIhasFloatsOfSizes` works (without plurality if we drop the multi-check version :sadparrot:).

include/revng/ABI/Definition.h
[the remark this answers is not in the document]
aleclearmind: Sounds good! I'd speak in terms of bytes.

include/revng/Model/ABI.h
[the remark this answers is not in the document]
aleclearmind: > I don't think it's the same ABI, that's talked about.

I think I agree... but it'd be nice to make a quick check, if we can, with some other MSVCs that we have available.
Which version(s) of MSVC are we using right now?

Also:

> Can you make an example of a prototype that would not be compatible between GCC and MSVC and why?

include/revng/ABI/Definition.h
aleclearmind: Works!

include/revng/Model/ABI.h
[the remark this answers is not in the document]
Cvelth: > Which version(s) of MSVC are we using right now?

```
toolchain/win32-vc19/vc
```

> Can you make an example of a prototype that would not be compatible between GCC and MSVC and why?

for `cdecl`/`stdcall` the easiest one is return value:
`msvc` returns small structs (<=8 bytes) in registers (rdx+rax) which `gcc` always uses SPTAR (this one would be believable if it was a recent-ish change and old `msvc` also used SPTAR)

for `fastcall` (also `thiscall`, but that's way to broken), `gcc` uses `rcx` for for SPTAR, which `msvc` - stack. (this one I would not believe to be version related - there's no reason to "degrade" the SPTAR performance by suddenly starting to use stack).

include/revng/ABI/Definition.h
Cvelth: On the second thought, just `UseGPRsForFloats` sounds better (and is more in line with your original suggestion), let's go with that instead.

Ah, ok, but `imperative` clashes with all the adjacent "bool"s. Let's go with `FloatsUseGPRs` after all - for consistency with the other variables (If you want a wider naming overhaul, let's do it in a separate PR, this one is already way too big).

include/revng/Model/ABI.h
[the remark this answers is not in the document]
aleclearmind: OK, we also have `vc{12,13,16}`, correct?
Do you think you can give a quick shot at some of them to see what happens?

I don't want an in-depth analysis, but at least understand if the situations is 1) they are all identical, 2) they 16 and 19 are almost the same, earlier are different, 3) they all are wildly different.

Maybe there's some reason due to which this is not easy to test, let me know.

include/revng/ABI/Definition.h
aleclearmind: Works.

tests/unit/Alignment.cpp
aleclearmind: 2. I'd just take an `ArrayRef<uint64_t>` and use `std::any_of` or something similar

Cvelth:
```cpp
if (ABIhasFloatsOfSizes(ABI, { Size1, Size2, Size3 }))
  do_something(ABI, Size1, Size2, Size3);
```

is still pretty ugly, but I guess I can live with it.

include/revng/ABI/Definition.h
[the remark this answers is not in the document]
Cvelth: Done.

include/revng/Model/ABI.h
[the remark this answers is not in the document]
Cvelth: Done.

lib/ABI/FunctionType/ValueDistributor.h
Cvelth: Done

I also used that chance to update the log, please take a look at the new `ValueDistributor: greatly improve success log message` commit.

lib/ABI/FunctionType/ValueDistributor.cpp
Cvelth: Fixed.

include/revng/ABI/Definition.h
Cvelth: Fixed.

I also took this opportunity to modernize this code.
Please review the new `ABI: redesign register sorting from the ground up` commit.

lib/ABI/Analyses/ConvertToCABIFunctionType.cpp
Cvelth: Dropped. See the new `ABI: drop incomplete `clrcall` support` commit.

share/revng/rcc-config.yml
Cvelth: Dropped.

lib/ABI/Definition.cpp
Cvelth: I went with `checkIfComputationFailed`, but also pulled the `Value == 0` check in as well.

Please take a look at the updated `ABI: make alignment-computation fails stricter` commit.

include/revng/ABI/Definition.h
Cvelth: Done.

tests/unit/Alignment.cpp
Cvelth: Done (I went with the `std::initializer_list`)

Also, while at it, I renamed `ABIhasScalarsOfSizes` into `ABIhasIntsOfSizes` because, technically `Floats` are also `Scalars`.

include/revng/Model/ABI.h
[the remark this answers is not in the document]
Cvelth: `vc12` also uses registers:
```cpp
// source
struct two_words {
  int a, b;
};

struct two_words global;
struct two_words whatever(void) { return global; }

// vc12
  401000: a1 28 54 40 00               	movl	0x405428, %eax
  401005: 8b 15 2c 54 40 00            	movl	0x40542c, %edx
// vc19
  401010: a1 00 53 41 00               	movl	0x415300, %eax
  401015: 8b 15 04 53 41 00            	movl	0x415304, %edx
```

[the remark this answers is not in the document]
Cvelth: `gcc` 9.2.0 from our out toolchain with `__attribute((ms_abi, cdecl))`:
```cpp
010001a0 <whatever>:
 10001a0: 8b 44 24 04                  	movl	0x4(%esp), %eax
 10001a4: 8b 15 78 31 00 01            	movl	0x1003178, %edx
 10001aa: 8b 0d 7c 31 00 01            	movl	0x100317c, %ecx
 10001b0: 89 10                        	movl	%edx, (%eax)
 10001b2: 89 48 04                     	movl	%ecx, 0x4(%eax)
 10001b5: c3                           	retl
```

lib/Model/Importer/DebugInfo/PDBImporter.cpp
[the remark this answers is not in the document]
dwarfyresearch: Resolved. We do not have this code any more.

lib/Model/Importer/DebugInfo/PDBImporter.cpp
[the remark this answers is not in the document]
dwarfyresearch: Resolved. We do not have this code any more.

lib/Model/Importer/DebugInfo/PDBImporter.cpp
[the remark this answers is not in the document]
dwarfyresearch: Resolved. We do not have this code any more.

lib/Model/Importer/DebugInfo/PDBImporter.cpp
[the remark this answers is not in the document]
dwarfyresearch: Done.

lib/Model/Importer/DebugInfo/PDBImporter.cpp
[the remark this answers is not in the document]
dwarfyresearch: Done.

include/revng/Model/ABI.h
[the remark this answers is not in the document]
aleclearmind: I was not just talking about this specific situation, I was curious to see if the test suite for MSVC ABIs with binaries produced by older MSVC versions would still work.

But maybe it's better to back log this.

lib/ABI/Definition.cpp
aleclearmind: If a function aborts unconditionally, it should contain "abort" in the name.
If a function aborts depending on a `bool` argument, it should contain "assert" in the name.

include/revng/Model/ABI.h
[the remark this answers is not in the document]
Cvelth: I don't remember why specifically (or even if it's still in there - switching to event handlers turned everything upside down), but we explicitly depended on c11. (and `vc16` shipped with `visual studio 2010`, which cannot have c11 support by definition).

lib/ABI/Definition.cpp
Cvelth: To be fair, this is not really a function, as in, it's not a part of any interface. The only reason it exists is to unload repetitive checks from the functions that come next.

Either way, it looks like this now:
```cpp
template<typename T>
bool checkIfComputationFailed(std::optional<AlignmentInfo> ComputationResult,
                              const T &ThingToDumpOnFailure) {
  if (!ComputationResult) {
    std::string Error = "Unable to compute the alignment of "
                        + serializeToString(ThingToDumpOnFailure);
    revng_abort(Error.c_str());
  }

  return ComputationResult->Value == 0;
}

// users
  std::optional<AlignmentInfo> Result = naturalAlignment(*this, QT, Cache);
  if (checkIfComputationFailed(Result, QT))
    return std::nullopt;
```

What do you think the name should be?

lib/Model/Importer/DebugInfo/PDBImporter.cpp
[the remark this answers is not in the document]
dwarfyresearch: No need for this.
Done.

lib/Model/Importer/DebugInfo/PDBImporter.cpp
```diff
@@ -804,7 +804,8 @@ Error PDBImporterTypeVisitor::visitKnownRecord(CVType &Record,
       // as `static` or `friend`.
       if (Function.getMethodKind() != MethodKind::Static
           and Function.getMethodKind() != MethodKind::Friend
-          and ProcessedTypes.count(CurrentTypeIndex) != 0) {
+          and ProcessedTypes.count(CurrentTypeIndex) != 0
+          and ProcessedTypes[CurrentTypeIndex].get() != nullptr) {
         auto MaybeSize = ProcessedTypes[CurrentTypeIndex].get()->size();
         if (MaybeSize and *MaybeSize != 0) {
```
Cvelth, line 810: @dwarfyresearch, ah, I almost forgot, I had to apply something like this because `model::TypePath{}.get()->size()` was failing. Not sure if this fix is sufficient though.

When you have a moment, could you, please, take a look into where these empty references come from?

include/revng/Model/ABI.h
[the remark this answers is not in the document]
aleclearmind: OK, please backlog to check if we can relax constraints on the compiler and test older MSVC, not super urgent.

lib/ABI/Analyses/ConvertToCABIFunctionType.cpp
aleclearmind: The compiler we have (`vc19`) doesn't support PASCAL?

aleclearmind:
NOTETOSELF: Review the following commits:
* ABI: drop incomplete 'clrcall' support
* ABIhasScalarsOfSizes into ABIhasIntsOfSizes
* ABI: redesign register sorting from the ground up
* ValueDistributor: greatly improve success log message
* ToCABI: always preserve stack types with metadata

lib/ABI/Definition.cpp
aleclearmind: It's a bit weird how the check is organized. I'd make "can't fail" version of `naturalAlignment` and then instead of

```c++
  if (checkIfComputationFailed(Result, QT))
    return std::nullopt;
```

I'd put `if (xxx == 0) return std::nullopt;`.

lib/ABI/FunctionType/Layout.cpp
aleclearmind: OK, can you elaborate on this a bit in the commit message?

lib/ABI/Definition.cpp
Cvelth: > I'd make "can't fail" version of `naturalAlignment`

I thought about it, but frankly, we already have way too many versions. I'd rather we avoid adding any more.

All I want is for all these practically identical bodies to be as short as possible so that they are at least somewhat reasonable to deal with.

To be honest, I this 🤏 close to just saying "screw it" and making a heavily templated `alignmentPITA_IMPL`-like function that turns all of these into one-liners like `auto whatever(/* args */) { return alignmentPITA_IMPL</* params */>(/* args */); }`

lib/ABI/FunctionType/Conversion.cpp
```diff
@@ -409,120 +409,123 @@ TCC::tryConvertingStackArguments(model::TypePath StackArgumentTypes,
...
-
-    if (!canBeNext(Distributor,
-                   CurrentArgument.Type(),
-                   CurrentArgument.Offset(),
-                   TheNextOne.Offset(),
-                   NextAlignment)) {
-      // We met a argument we cannot just "add" as is.
-      NaiveConversionFailed = true;
-      break;
+  if (!model::hasMetadata(Stack)) {
```
aleclearmind, line 424: Add a comment about the ratio behind this

lib/ABI/Definition.cpp
aleclearmind:
```c++
  AlignmentInfo Result = myCantFail(*this, naturalAlignment(*this, QT, Cache));
  if (ComputationResult->Value == 0)
    return std::nullopt;
```

lib/ABI/FunctionType/Conversion.cpp
Cvelth: Wanting to not drop types with metadata (if I remember correctly, now it's just custom names and comments) sounds pretty logical to me.

The rationale behind this change is slightly more pragmatic though: tests were failing if we were to unroll some of the structs. So ensuring they have a custom name was an easy and non-intrusive way to preserve them.

TL;DR: if converter ever meets a struct with name or comment or whatever new metadata we add in the future, it ensures it stays as a struct no matter what.

aleclearmind: I guess we should do a recursive check.

Anyway, add a comment.

Cvelth: > I guess we should do a recursive check.

No need, we only ever unroll the `StackArgumentsType` struct itself. Everything nested any deeper is only ever treated as a struct.

In practice this only ever affects functions with a single struct stack argument, when the entire `StackArgumentsType` is replaced with the said struct.

aleclearmind: OK, resolve once you added a comment

Cvelth: Done:
```cpp
    // NOTE: Only proceed with trying to "split" the stack struct up if it has
    //       no metadata attached: as in, it doesn't have a user-defined
    //       `CustomName` or a `Comment` or any other fields we'll mark as
    //       metadata in the future.
```

lib/ABI/FunctionType/Layout.cpp
Cvelth: Done:
```
ToRaw: take advantage of the updated distributor

This lets us drop the repeated alignment calculations at
the "leaves" of the conversion. Since new distributor already
provides the necessary information, we can just grab it from
there.
```

lib/ABI/Definition.cpp
Cvelth: Done (`assertOnFailure`).

lib/Model/Importer/DebugInfo/PDBImporter.cpp
Cvelth: Fixed.

aleclearmind:
Merged in b21ff0b. Thanks.

PR #341, opened 2023-12-22
Implement type editing in the UI

mrjackv:


python/revng/internal/daemon/__init__.py
```diff
@@ -135,6 +135,7 @@ def make_startlette() -> Starlette:
 
     if DEBUG:
         print(f"Manager workdir is: {manager.workdir}")
```
aleclearmind, line 137: Debug log should always go to `stderr`

python/revng/internal/daemon/__init__.py
```diff
@@ -135,6 +135,7 @@ def make_startlette() -> Starlette:
...
+        signal.signal(signal.SIGUSR2, lambda s, f: event_manager.save())
```
aleclearmind, line 138: 1. Why do we need this?
2. What do we use `SIGUSR1` for?

lib/Yield/Pipes/AssemblyPipes.cpp
```diff
@@ -75,9 +75,10 @@ void YieldAssembly::run(pipeline::ExecutionContext &Context,
   for (auto [Address, S] : Input) {
     auto MaybeFunction = TupleTree<yield::Function>::deserialize(S);
     revng_assert(MaybeFunction && MaybeFunction->verify());
-    revng_assert((*MaybeFunction)->Entry() == Address);
+    revng_assert((*MaybeFunction)->Entry() == std::get<0>(Address));
```
aleclearmind, line 78: Mmmh, why do we need `std::get<0>` here? Is this a single-element tuple?

include/revng/Pipes/StringMap.h
```diff
@@ -31,7 +31,8 @@ struct DataOffset {
   size_t End;
 };
 
-using OffsetMap = std::map<MetaAddress, DataOffset>;
+template<typename T>
+using OffsetMap = std::map<T, DataOffset>;
```
aleclearmind, line 36: Nice

include/revng/Pipes/TargetListContainer.h
aleclearmind, line 1: What's the purpose of this?
Can't we deduce this information from the model?

typescript/tuple_tree.ts
```diff
@@ -47,8 +47,8 @@ export class Reference<T, M> {
         }
     }
 
-    resolve(root: M): T {
-        return _getElementByPath<T>(this.reference, root)!;
+    resolve(root: M): T | undefined {
+        return _getElementByPath<T>(this.reference, root);
     }
```
aleclearmind, line 52: What's the use case here?

lib/Yield/Pipes/AssemblyPipes.cpp
mrjackv: Instead of `MetaAddress` we how get `std::tuple<MetaAddress>` (because it's uniform w.r.t. model keys)

include/revng/Pipes/TargetListContainer.h
aleclearmind: Ah I got it. Please thoroughly document the reason for this to be, what limitation it addresses and what needs to be done for this hack to go away.

I'd appreciate if you could document hacks on your own, I always have to ask you `:)`.

python/revng/internal/daemon/__init__.py
mrjackv: 1. For debugging purposes, sometimes I want to test things that happen during save (e.g. S3, `store` function, etc.) and I'd like to avoid shutting down the server or waiting 5 minutes of inactivity
2. To reload opened files, references: [1](https://github.com/encode/uvicorn/blob/3dd0d44445cc2d96c77037ba1f7bfefd288f370e/uvicorn/workers.py#L75) [2](https://github.com/benoitc/gunicorn/blob/88fc4a43152039c28096c8ba3eeadb3fbaa4aff9/gunicorn/workers/base.py#L187)

typescript/tuple_tree.ts
mrjackv: Basically, in TS when traversing the model, you get to a reference object, let's call `ref`, if you want to get the pointed object you would do `ref.resolve(binary)`, however it's not guaranteed that the pointed paths exists, so to make sure the caller of `resolve` handles that case, we return `undefined` in case the path does not exist.

lib/Yield/Pipes/AssemblyPipes.cpp
aleclearmind: Sad but OK

python/revng/internal/daemon/__init__.py
aleclearmind: Gotcha, maybe we can merge SIGUSR2 into SIGUSR1, it doesn't sound harmful.

mrjackv: Don't get me wrong, we can, but nesting signal handlers isn't the nicest thing and `USR2` is unused

aleclearmind: Nesting? Just let the SIGUSR1 signal handler do both things. Can we?

mrjackv: Not directly, you need to do something like:
```python
old_handler = signal.getsignal(signal.SIGUSR1)  # get the old handler
def new_handler(sig, frame):
  save_stuff()
  old_handler(sig, frame)
signal.signal(signal.SIGUSR1, new_handler)
```

python/revng/internal/daemon/__init__.py
mrjackv: Done

aleclearmind:
Merged in 32d7976. Thanks.

PR #343, opened 2024-01-10
Container cache invalidation

mrjackv:
First five commits are shared with #341

Sister PRs:
* [revng-c](https://rev.ng/gitlab/revng-private/revng-c/-/merge_requests/187)
* [hitman](https://rev.ng/gitlab/revng-private/hitman/-/merge_requests/31)


include/revng/Support/InitRevng.h
```diff
@@ -35,12 +36,17 @@ class InitRevng : public llvm::InitLLVM {
       llvm::cl::HideUnrelatedOptions(CategoriesToHide);
     }
 
+    llvm::BumpPtrAllocator A;
+    llvm::StringSaver Saver(A);
+    llvm::SmallVector<const char *, 0> Arguments(Argv, Argv + Argc);
+    if (auto EnvValue = llvm::sys::Process::GetEnv("REVNG_OPTIONS")) {
+      llvm::cl::TokenizeGNUCommandLine(*EnvValue, Saver, Arguments);
+    }
```
aleclearmind, line 44: Add a comment explaining why we no longer use `ParseCommandLineOptions` for this.

include/revng/ADT/STLExtras.h
```diff
@@ -286,10 +286,9 @@ inline llvm::ArrayRef<uint8_t> toArrayRef(llvm::StringRef Data) {
...
 //
 template<std::ranges::sized_range FromType, std::ranges::sized_range ToType>
-auto append(FromType &&From, ToType &To) {
-  size_t ExistingElementCount = To.size();
-  To.resize(ExistingElementCount + From.size());
-  return llvm::copy(From, std::next(To.begin(), ExistingElementCount));
+void append(FromType &&From, ToType &To) {
+  To.reserve(To.size() + From.size());
+  llvm::copy(From, std::back_inserter(To));
 }
```
aleclearmind, line 292: @Cvelth do you approve?

CMakeLists.txt
```diff
@@ -114,6 +114,17 @@ file(RELATIVE_PATH RELATIVE_LLVM_TOOLS_BINARY_DIR "${CMAKE_INSTALL_PREFIX}"
...
+#
+# Component hash
+#
+file(MAKE_DIRECTORY "${CMAKE_BINARY_DIR}/share/revng/component-hashes")
+add_custom_target(
+  compute_sha1 ALL
+  COMMAND git -C "${CMAKE_SOURCE_DIR}" rev-parse HEAD >
+          "${CMAKE_BINARY_DIR}/share/revng/component-hashes/revng")
+install(FILES "${CMAKE_BINARY_DIR}/share/revng/component-hashes/revng"
+        DESTINATION "${CMAKE_INSTALL_DIR}/share/revng/component-hashes/")
```
aleclearmind, line 126: 1. Use a more descriptive target name. `ninja compute_sha1` is a bit obscure.
2. Are we already assume we are in a git repo in other parts of the codebase? If not, let's handle the situation in which we're not in a git repo.

lib/Support/ResourceFinder.cpp
```diff
@@ -17,4 +21,20 @@ PathList ResourceFinder({
 
 });
 
+std::string getComponentsHash() {
```
aleclearmind, line 24: This function does not return an hash. Can we get a better name? `getComponentIDs`? `getComponentHashes`?

lib/Pipeline/Runner.cpp
```diff
@@ -236,6 +236,24 @@ Error Runner::load(const revng::DirectoryPath &DirPath) {
   return Error::success();
 }
 
+std::vector<revng::FilePath>
+Runner::getWrittenFiles(const revng::DirectoryPath &DirPath) const {
+  std::vector<revng::FilePath> Result;
+
+  for (auto &StepEntry : Steps) {
```
aleclearmind, line 243: `auto` abuse

include/revng/Pipeline/ContainerFactory.h
```diff
@@ -22,11 +22,19 @@ class ContainerFactoryBase {
 
   virtual llvm::StringRef mimeType() const = 0;
 
+  virtual std::vector<revng::FilePath>
+  getWrittenFiles(const revng::FilePath &Path) const = 0;
```
aleclearmind, line 26: Maybe this is too powerful.
Maybe we should return `{ "", ".idx" }` so we're never tempted by doing anything else other than appending things to the path.

lib/Pipes/PipelineManager.cpp
```diff
@@ -92,6 +98,60 @@ PipelineManager::overrideContainer(llvm::StringRef PipelineFileMapping) {
   return MaybeMapping->load(*Runner);
 }
 
+static llvm::Error
+checkComponentsVersion(const revng::DirectoryPath &ExecutionDirectory,
```
aleclearmind, line 102: Comment this function

include/revng/ADT/STLExtras.h
Cvelth: If we go this route, it's probably a good idea to actually enforce that fact that `To` must be a vector-like: https://godbolt.org/z/MnxnYY4vc

I guess the old approach also assumed that (since it used `resize`). So my bad on not constraining the interface correctly.

But, while on topic, let's make it right (tm).
Here's what I'd suggest: https://godbolt.org/z/neYsYbjoq

What do you, guys, think?

Cvelth: We could also do something cool like
```cpp
template <typename T>
concept HasReserve = requires(T &&V) { { V.reserve(5) }; };
// . . .
  if constexpr(HasReserve<ToType>)
    To.reserve(To.size() + From.size());
```

But even I think that might be an overkill.

mrjackv: Wouldn't `std::ranges::move` imply that `From` is always an rvalue?
Wouldn't this be better?

```cpp
template<std::ranges::sized_range FromType, std::ranges::sized_range ToType>
auto append(FromType &&From, ToType &To) {
  if constexpr (std::is_lvalue_reference_v<FromType>)
    return std::ranges::copy(From, std::inserter(To, To.end()));
  else
    return std::ranges::move(From, std::inserter(To, To.end()));
}
```

Cvelth: Wait, yeah, I wasn't paying attention, thanks for catching that.

lib/Support/ResourceFinder.cpp
mrjackv: > This function does not return an hash.

It returns the concatenation of all components' hashes, so in a certain way, it's an hash
Both name propositions have plurality, but in this case the return is a singular string which says `if this equals to the one saved, the versions are the same` (just as a "normal" hash)

CMakeLists.txt
mrjackv: 1. Is `compute_component_hash` ok?
2. What would you do? `dd if=/dev/random`?

aleclearmind: 1. Yup
2. I'd put the project version according to CMake (`PROJECT_VERSION`)

lib/Support/ResourceFinder.cpp
aleclearmind: An hash certain properties that a string that's a concatenation of hashes does not have, so this is an hash if you consider the term in a very widened meaning, but OK.

include/revng/Pipeline/ContainerFactory.h
mrjackv: They only meaningful thing one can do with a `revng::FilePath` is to do `addExtension` (other than `getReadableFile`, but that's beside the point). I'm ok to return `std::vector<std::string>`, but that would be a bit less uniform w.r.t. the other methods (also, it might be better to be renamed to `getWrittenExtensions` or something similar)

aleclearmind: I'll leave this to you. Both solutions sound reasonable to me.

include/revng/Support/InitRevng.h
mrjackv: Done

CMakeLists.txt
mrjackv: Both done

lib/Pipes/PipelineManager.cpp
mrjackv: Added comments

include/revng/ADT/STLExtras.h
mrjackv: I've done some testing and using `reserve` cuts times considerably, so I've added the version above with `HasReserve`

lib/Pipeline/Runner.cpp
mrjackv: Fixed

include/revng/Pipeline/ContainerFactory.h
mrjackv: I've decided to keep the `revng::FilePath` version for consistency's sake with other functions.

PR #346, opened 2024-01-18
JumpTargetManager: improve `rebuildDispatcher`

andrealinux1:
Improve the way in which we connect the non reachable `JumpTargets` when rebuilding the dispatcher for non `SemanticsPreserving` `CFGForm`s.

When connecting group of jump targets that are currently not reachable from the entry dispatcher, we elect the jump target with the lowest program counter value, as the one to be connected to the dispatcher. We also mark the jump targets now transitively reachable from the elected one, as reachables, in order to avoid adding other unnecessary edges from the dispatcher.


aleclearmind:
Please always use `feature/` as a branch prefix.

aleclearmind:
Merged in b2ef969. Thanks.

PR #349, opened 2024-01-23
Reduce layout usage in early parts of the pipeline to aid invalidation

Cvelth:


include/revng/ABI/FunctionType/Layout.h
```diff
@@ -181,4 +182,27 @@ struct Layout {
   void dump() const debug_function;
 };
 
+inline std::span<const model::Register::Values>
+calleeSavedRegisters(const model::CABIFunctionType &Function) {
+  return abi::Definition::get(Function.ABI()).CalleeSavedRegisters();
+}
+inline std::span<const model::Register::Values>
```
aleclearmind, line 189: Leave a blank line between each function. If you really want to group them, use doxygen facilities for grouping.

lib/ABI/FunctionType/Layout.cpp
```diff
@@ -633,6 +633,30 @@ Layout::returnValueRegisters() const {
...
+      if (!Distributed.RepresentsPadding) {
+        if (Distributed.SizeOnStack != 0) {
+          uint64_t CurrentSize = Distributed.OffsetOnStack
+                                 + Distributed.SizeOnStack;
+          revng_assert(CurrentSize >= CombinedStackArgumentSize);
+          CombinedStackArgumentSize = CurrentSize;
+        }
+      }
+    }
+  }
```
aleclearmind, line 655: Can't we factor this code out with some other piece of `librevngABI`?

lib/ABI/FunctionType/ValueDistributor.h
```diff
@@ -194,6 +194,24 @@ class ArgumentDistributor : public ValueDistributor {
     }
   }
 
+  bool canNextArgumentUseRegisters() const {
```
aleclearmind, line 197: I'd prefer this commit to be merge in the commit where it's used.

include/revng/ABI/FunctionType/Layout.h
```diff
@@ -222,4 +222,30 @@ inline uint64_t finalStackOffset(const model::TypePath &Function) {
   return finalStackOffset(*Function.get());
 }
 
+struct UsedRegisters {
+  llvm::SmallVector<model::Register::Values, 8> Arguments;
+  llvm::SmallVector<model::Register::Values, 8> ReturnValues;
```
aleclearmind, line 227: `DenseSet`?

include/revng/ABI/FunctionType/Layout.h
```diff
@@ -222,4 +222,30 @@ inline uint64_t finalStackOffset(const model::TypePath &Function) {
...
+    return usedRegisters(*CABI);
+  else if (auto *Raw = llvm::dyn_cast<model::RawFunctionType>(&Function))
+    return usedRegisters(*Raw);
+  else
+    revng_abort("Layouts of non-function types are not supported.");
+}
+inline UsedRegisters usedRegisters(const model::TypePath &Function) {
+  revng_assert(Function.isValid());
+  return usedRegisters(*Function.get());
+}
```
aleclearmind, line 249: The lack of space between functions makes them very hard to read.
Always leave a blank line.

lib/ABI/FunctionType/Layout.cpp
```diff
@@ -657,6 +657,43 @@ uint64_t finalStackOffset(const model::CABIFunctionType &Function) {
   return ToRawConverter(ABI).finalStackOffset(CombinedStackArgumentSize);
 }
 
+UsedRegisters usedRegisters(const model::CABIFunctionType &Function) {
```
aleclearmind, line 660: I see a quite a bit of newly added code, I was hoping that this changeset would just outline pieces of things that were buried in `Layout` and the other classes...

lib/EarlyFunctionAnalysis/FunctionSummaryOracle.cpp
```diff
@@ -25,6 +25,19 @@ struct PrototypeImporter {
...
+      if (llvm::GlobalVariable *CSV = M.getGlobalVariable(Name, true))
+        PreservedRegisters.insert(CSV);
+    }
+
+    std::erase_if(Summary.ClobberedRegisters, [&](const auto &E) {
+      return PreservedRegisters.contains(E);
+    });
+
+    Summary.ElectedFSO = abi::FunctionType::finalStackOffset(Prototype);
+
```
aleclearmind: Why? Comment in the commit message

lib/EarlyFunctionAnalysis/FunctionSummaryOracle.cpp
```diff
@@ -25,13 +25,18 @@ struct PrototypeImporter {
...
+      return M.getGlobalVariable(model::Register::getCSVName(Register), true);
+    };
+    auto NullptrFilter = [](const auto *Pointer) -> bool {
+      return Pointer != nullptr;
+    };
+
+    auto PreservedRegisters = abi::FunctionType::calleeSavedRegisters(Prototype)
+                              | std::views::transform(RegisterToCSV)
+                              | std::views::filter(NullptrFilter)
+                              | revng::to<std::set<llvm::GlobalVariable *>>();
```
aleclearmind, line 46: `:O`

What do you think of:

```c++
    auto ToCSV = std::views::transform([this](Register Register) -> llvm::GlobalVariable * {
      return M.getGlobalVariable(model::Register::getCSVName(Register), true);
    });
    auto RemoveNullptr = std::views::filter([](const auto *Pointer) -> bool {
      return Pointer != nullptr;
    });

    auto PreservedRegisters = abi::FunctionType::calleeSavedRegisters(Prototype)
                              | ToCSV
                              | RemoveNullptr
                              | revng::to<std::set<llvm::GlobalVariable *>>();
```

I think this kind of paradigm is very effective when you first prepare all the components with very explicative, obvious, high level names and then you combine them without noise.

lib/EarlyFunctionAnalysis/FunctionSummaryOracle.cpp
```diff
@@ -150,11 +150,14 @@ void importModel(llvm::Module &M,
   revng_log(Log, "Importing from model");
   LoggerIndent Indent(Log);
 
-  std::set<llvm::GlobalVariable *> ABICSVs;
-  for (llvm::GlobalVariable *CSV : GCBI.abiRegisters())
-    if (CSV != nullptr && !(GCBI.isSPReg(CSV)))
-      ABICSVs.emplace(CSV);
-  PrototypeImporter Importer{ .M = M, .ABICSVs = ABICSVs };
+  auto IsAnABICSV = [SP = GCBI.spReg()](llvm::GlobalVariable *CSV) -> bool {
```
aleclearmind, line 173: As before, I'd prefer something like `| keepABICSVsOnly` or something like this.

lib/EarlyFunctionAnalysis/FunctionSummaryOracle.cpp
```diff
@@ -12,6 +12,13 @@ static Logger<> Log("efa-import-model");
 
 namespace efa {
 
+enum class PrototypeImportLevel {
+  Full,
+  Partial,
+  None
+};
```
aleclearmind, line 19: Docs

include/revng/EarlyFunctionAnalysis/FunctionSummaryOracle.h
```diff
@@ -131,6 +120,20 @@ class FunctionSummaryOracle {
...
+                       GeneratedCodeBasicInfo &GCBI,
+                       const model::Binary &Binary);
+  static FunctionSummaryOracle
+  importBasicPrototypeData(llvm::Module &M,
+                           GeneratedCodeBasicInfo &GCBI,
+                           const model::Binary &Binary);
+  static FunctionSummaryOracle
+  importWithoutPrototypes(llvm::Module &M,
+                          GeneratedCodeBasicInfo &GCBI,
+                          const model::Binary &Binary);
```
aleclearmind, line 135: Docs

include/revng/ABI/FunctionType/Layout.h
Cvelth: Wait, no, it has to be a vector - order matters.

Now it returns them ordered the same way `sortArguments`/`sortReturnValues` returns (as in, order depending on the ABI definition), switching to a set would mean ordering them the same way `model::Register::Values` does.

lib/ABI/FunctionType/Layout.cpp
Cvelth: I tried that edition first. But that made layout itself a bit more complex.

Consider that practically all the complexity here is because we want to handle SPTAR gracefully.

Without it, it's just
```
UsedRegisters usedRegisters(const model::CABIFunctionType &Function) {
  auto Result.RVs = move_return_value_registers_in();
  
  handle_SPTAR_gracefully();
  
  for (const auto *Argument = Function.Arguments().tryGet(Index++);
     Argument != nullptr;
     Argument = Function.Arguments().tryGet(Index++)) {
     // (if it's this complicated looking loop you don't like, we could use some range-magic and hide it under something like `TrackingContainer::iterate`)
     // The only reason I didn't is because it's only ever usable for CFT-Arguments (because it's just a vector disguised as a `SortedVector`,
     // So, there's not many points where we could reuse it, AND most of those are already happy relying on `usedRegisters` as is - so here feels general enough.
             std::ranges::move(distributeArgument(...).Registers,
                          std::back_inserter(Result.Arguments));
  }
}
```

include/revng/ABI/FunctionType/Layout.h
Cvelth: Done.

include/revng/ABI/FunctionType/Layout.h
Cvelth: Fixed.

lib/EarlyFunctionAnalysis/FunctionSummaryOracle.cpp
Cvelth: Done:
```
EFA: reorder prototype-based summary importer

This lets up cleanly "bail out" after only some of the computation
is done. This way we avoid reading information we don't really use
from the model based on who is the specific user of the oracle.
```

lib/EarlyFunctionAnalysis/FunctionSummaryOracle.cpp
Cvelth: Done.

(I went with `TransformToCSV` and `IgnoreNullptr`)

the only problem is that I don't really like how clang-format reformatted it into
```cpp
    auto Preserved = abi::FunctionType::calleeSavedRegisters(Prototype)
                     | TransformToCSV | IgnoreNullptr
                     | revng::to<std::set<llvm::GlobalVariable *>>();
```
But I guess I can live with that.

lib/EarlyFunctionAnalysis/FunctionSummaryOracle.cpp
Cvelth: Done, except I went with `RegisterFilter`.
```
  PrototypeImporter Importer{ .M = M,
                              .ABICSVs = GCBI.abiRegisters() | RegisterFilter
                                         | revng::to<std::set<GV *>>() };
```

lib/EarlyFunctionAnalysis/FunctionSummaryOracle.cpp
Cvelth: Done:
```cpp
enum class PrototypeImportLevel {
  // Do not import any prototype-related information except for
  // the callee saved register list (which is ABI-defined for CFTs).
  None,

  // Only import final stack offset on top of the callee saved register list.
  Partial,

  // Import everything present in the prototype.
  Full
};
```

include/revng/EarlyFunctionAnalysis/FunctionSummaryOracle.h
Cvelth: Done:
```cpp
  /// Create an oracle based on the binary, but importing all the function
  /// related information available in it.
  static FunctionSummaryOracle
  importFullPrototypes(llvm::Module &M,
                       GeneratedCodeBasicInfo &GCBI,
                       const model::Binary &Binary);

  /// Create an oracle based on the binary, but when importing functions, only
  /// pull preserved registers and FSO from the prototype.
  static FunctionSummaryOracle
  importBasicPrototypeData(llvm::Module &M,
                           GeneratedCodeBasicInfo &GCBI,
                           const model::Binary &Binary);

  /// Create an oracle based on the binary, but when importing functions, only
  /// pull preserved registers from the prototype.
  static FunctionSummaryOracle
  importWithoutPrototypes(llvm::Module &M,
                          GeneratedCodeBasicInfo &GCBI,
                          const model::Binary &Binary);
```

lib/ABI/FunctionType/Layout.cpp
```diff
@@ -342,6 +346,19 @@ ToRawConverter::finalStackOffset(uint64_t SizeOfArgumentsOnStack) const {
   return Result;
 }
 
+uint64_t ToRawConverter::combinedStackArgumentSize(CFT Function) const {
```
aleclearmind, line 349: I'd prefer to have `using TRC = ToRawConverter` rather than `using CFT =`.

lib/ABI/FunctionType/Layout.cpp
aleclearmind: OK, comment the function about what each stage is doing.

> this complicated looking loop you don't like...

Why can't you just iterate over the arguments? What situation are you trying to handle?
This `SortedVector` should have keys from `0` to its size `- 1`. If this is not checked in `::verify`, we should check it there and assume it here.

aleclearmind (changes requested):

lib/ABI/FunctionType/Layout.cpp
Cvelth: Done.

lib/ABI/FunctionType/Layout.cpp
Cvelth: 1. Done.

2. We don't need to read _every_ argument in every case. For some ABIs even as little as one argument is enough, hence why it's easier to use index - this lets to not call `end` and otherwise rely on the number of arguments in any way.

aleclearmind: 2. Ah, that's a pretty obscure feature to have without documentation. In any case, let's not push optimizations this far, unless we have proof that's beneficial. We're reading one argument, I'm OK with reading them all. Let's favor simplicity.

Cvelth: I went along and "hid" the complexity within the `TrackingContainer`.
This function now looks like
```cpp
UsedRegisters usedRegisters(const model::CABIFunctionType &Function) {
  UsedRegisters Result;

  // Ready the return value register data.
  const abi::Definition &ABI = abi::Definition::get(Function.ABI());
  auto RV = ToRawConverter(ABI).distributeReturnValue(Function.ReturnType());
  std::ranges::move(RV.Registers, std::back_inserter(Result.ReturnValues));

  // Handle shadow pointer return value gracefully.
  ArgumentDistributor Distributor(ABI);
  if (RV.SizeOnStack != 0) {
    Distributor.addShadowPointerReturnValueLocationArgument();

    revng_assert(Result.ReturnValues.empty());
    const auto &GPRs = ABI.GeneralPurposeReturnValueRegisters();
    revng_assert(!GPRs.empty());
    Result.ReturnValues.emplace_back(GPRs[0]);

    if (ABI.ReturnValueLocationRegister() != model::Register::Invalid)
      Result.Arguments.emplace_back(ABI.ReturnValueLocationRegister());
  }

  // Iterate over arguments until we sure no further argument can use
  // any registers.
  for (const model::Argument &Argument : Function.Arguments().asVector()) {
    auto Distributed = Distributor.nextArgument(Argument.Type());
    for (DistributedValue SingleEntry : Distributed)
      if (!SingleEntry.RepresentsPadding)
        std::ranges::move(SingleEntry.Registers,
                          std::back_inserter(Result.Arguments));

    if (!Distributor.canNextArgumentUseRegisters())
      break;
  }

  return Result;
}
```

aleclearmind:
Merged in cf3072d. Thanks.

PR #350, opened 2024-01-30
Allow multiple LLVM lit configuration files

drblallo:
revng check branch fails if there are multiple python modules with the same name. This commit explicitly turns off checks for modules with the same name as the configuration files of LLVM lit so that more than one can exists.


pfez:
Merged in 95aa074e4c0b707d79957a3cdc53701149b2ca70

Thanks!

PR #355, opened 2024-02-09
Use runtime ABI test binaries on the rest of the pipeline

Cvelth:


lib/ValueMaterializer/AdvancedValueInfo.cpp
```diff
@@ -244,10 +244,19 @@ runAVI(const DataFlowGraph &DFG,
...
 
-  revng_assert(InitialNodes.size() > 0);
+  if (InitialNodes.empty()) {
+    return {
+      std::map<llvm::Instruction *, ConstantRangeSet>{},
+      ControlFlowEdgesGraph(),
+      map<const ForwardNode<ControlFlowEdgesNode> *,
+          MFP::MFPResult<map<llvm::Instruction *, ConstantRangeSet>>>{}
+    };
+  }
```
aleclearmind, line 259: Can you ping me on mattermost on this?

share/revng/test/configuration/revng/runtime-abi-tests.yml
```diff
@@ -22,8 +22,10 @@ commands:
     from:
       - type: revng.runtime-abi-tests
       - type: revng-qa.runtime-abi-tests
+    suffix: /
```
aleclearmind, line 25: This is fine, but what's the reason? Are you planning to have downstream tests build on this?

share/revng/test/tests/abi/prepare-tested-model.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python3
+
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+import argparse
```
aleclearmind, line 7: What does this script do? Add brief summary on top.

aleclearmind (changes requested):

share/revng/test/configuration/revng/runtime-abi-tests.yml
Cvelth: Currently it uses a temporary directory, which is cleaned as soon as the test is done.

this will make it easier to report errors:

- Hey, @ivan, I see this ABI/segregation test failing.
- I see, can you please send me the contents of directory X?

share/revng/test/tests/abi/prepare-tested-model.py
Cvelth: It's just `replace-abi` script with a bit of extra filtering on top.

aleclearmind: OK, but add a brief summary.

Cvelth: Done:
```
# This script is used for model preparation for the tests using runtime-abi
# artifacts. It removes all the function not explicitly relevant to the test
# so only `test_X` function plus a couple of helpers (like `main` and `memcpy`)
# remain. After which, it ensures that the prototypes of all the remaining
# functions have their ABI set correctly to work around the limitations of
# ABI detection when importing from dwarf.
```

lib/ValueMaterializer/AdvancedValueInfo.cpp
Cvelth: Your suggested fix was added as the last commit.

lib/ABI/FunctionType/Conversion.cpp
```diff
@@ -405,13 +405,16 @@ TCC::tryConvertingStackArguments(model::TypePath StackArgumentTypes,
...
-                                                              .get());
+  auto &Stack = *llvm::cast<model::StructType>(StackArgumentTypes.get());
 
-  // Compute the full alignment.
-  model::QualifiedType StackStruct = { StackArgumentTypes, {} };
-  uint64_t StackAlignment = *ABI.alignment(StackStruct);
+  // As a workaround for the cases where expected alignment is missing at
+  // the very end of the RFT-style stack argument struct, use adjusted
+  // size and alignment values instead of the real ones.
+  uint64_t StackAlignment = *ABI.alignment(Stack);
```
Cvelth, line 413: @aleclearmind, please review the new `ToCFT: fix last stack argument verification` commit, it doesn't have anything to do with this PR, but it feels like this is the best place for it to get a piggy-back ride.

aleclearmind:
Merged in e990d49. Thanks.

PR #358, opened 2024-02-15
Introduce basic disassembly configuration

Cvelth:


lib/Yield/PTML.cpp
```diff
@@ -535,19 +543,71 @@ static std::string taggedText(const PTMLBuilder &ThePTMLBuilder,
...
+                  .serialize();
+    } else {
+      // TODO: handle column overflows more carefully.
+    }
+  }
+
+  if (options::RawByteColumnWidth) {
+    std::string SerializedBytes;
+    for (uint8_t Byte : Instruction.RawBytes()) {
+      std::string HexByte = Byte ? llvm::utohexstr(Byte, true, 2) : "00";
```
aleclearmind, line 540: Why is `Byte ? ... : "00"` necessary?

lib/Yield/PTML.cpp
```diff
@@ -535,19 +543,71 @@ static std::string taggedText(const PTMLBuilder &ThePTMLBuilder,
   return Result;
 }
 
+static std::string instructionPrefix(const PTMLBuilder &ThePTMLBuilder,
+                                     const yield::Instruction &Instruction) {
+  std::string Result;
+
+  if (options::InstructionAddressColumnWidth) {
+    std::string SerializedAddress = Instruction.Address().toString();
```
aleclearmind, line 519: I'd just dump the address, not the whole `MetaAddress`. Too much noise for little return.

A basic block can only be of one type (can't mix Thumb and regular ARM in a single block), but even a function mixing ARM and Thumb is rare enough that I don't think it's worth all the noise.

lib/Yield/PTML.cpp
```diff
@@ -535,19 +543,71 @@ static std::string taggedText(const PTMLBuilder &ThePTMLBuilder,
...
+
+    if (SerializedBytes.size() < options::RawByteColumnWidth) {
+      uint64_t Column = options::RawByteColumnWidth;
+      Result += ThePTMLBuilder
+                  .getTag(tags::Span,
+                          std::string(Column - SerializedBytes.size(), ' '))
+                  .addAttribute(attributes::Token, ptml::tokens::Indentation)
+                  .serialize();
+    } else {
+      // TODO: handle column overflows more carefully.
```
aleclearmind, line 559: So we don't emit anything in case there's too many bytes?

lib/Yield/PTML.cpp
```diff
@@ -535,19 +543,71 @@ static std::string taggedText(const PTMLBuilder &ThePTMLBuilder,
...
+  if (options::InstructionAddressColumnWidth) {
+    std::string SerializedAddress = Instruction.Address().toString();
+    Result += ThePTMLBuilder.getTag(tags::Span, SerializedAddress)
+                .addAttribute(attributes::Token, tokenTypes::InstructionAddress)
+                .serialize()
+              + ' ';
+
+    if (SerializedAddress.size() < options::InstructionAddressColumnWidth) {
+      uint64_t Column = options::InstructionAddressColumnWidth;
+      Result += ThePTMLBuilder
```
aleclearmind, line 527: I'd be really happy if I could have a commit that `s/ThePTMLBuilder/B/g` in the whole codebase.

scripts/tuple_tree_generator/tuple_tree_generator/generators/typescript.py
```diff
@@ -174,7 +174,7 @@ def get_default_value(self, field: StructField):
                 if isinstance(field.resolved_type, EnumDefinition):
                     return '"Invalid"'
                 else:
-                    return f"new {field.type}()"
+                    return f"new {field.type}({{}} as I{field.type})"
```
aleclearmind, line 177: @mrjackv do you approve?

include/revng/Model/DisassemblyConfiguration.h
```diff
@@ -0,0 +1,35 @@
...
+      Set this to 0 (default) to disable the raw byte emission.
+      Set it to 1 for non-column view (since each instruction is guaranteed
+      to have at least one byte).
+    type: uint64_t
+    optional: true
+  - name: InstructionAddressColumnWidth
+    doc: |
+      Set this to 0 (default) to disable the instruction address emission.
+      Set it to 1 for non-column view (since each instruction is guaranteed
+      to have an address).
```
aleclearmind, line 25: I'm not sure what "non-column view" means and what happens if it's greater than `1`. Do we wrap the address?

include/revng/Model/DisassemblyConfiguration.h
```diff
@@ -30,6 +34,10 @@ class model::DisassemblyConfiguration
   : public model::generated::DisassemblyConfiguration {
 public:
   using generated::DisassemblyConfiguration::DisassemblyConfiguration;
+  DisassemblyConfiguration() : generated::DisassemblyConfiguration() {
+    // TODO: introduce a better way of specifying default values.
+    UseIntelSyntax() = true;
+  }
```
aleclearmind: Mmmmh, I think this will create problems with all the other wrappers (Python and TypeScript).

Can we have `UseATTSyntax`? I know it's not ideal, but...

aleclearmind (changes requested):

lib/Yield/PTML.cpp
Cvelth: `llvm::utohexstr` was printing `0` for some reason for 0s. So instead of trying to figure out what's wrong with it, I just manually handled the problem case and called it a day.

lib/Yield/PTML.cpp
Cvelth: If there are too many bytes, they "overflow", as in, instruction itself gets further to the right, breaking tabulation.

Cvelth: But I'm fine with either option, we can emit nothing OR `byte byte byte...` or anything else. I don't have strong preferences.

include/revng/Model/DisassemblyConfiguration.h
Cvelth: `non-column view` means `no tabulation`, as in, the instruction just starts after a space.

include/revng/Model/DisassemblyConfiguration.h
Cvelth: Done, `UseIntelSyntax` -> `UseATTSyntax`. As for others, I added
```cpp
if (SelectedValue == SomeEnum::Invalid)
  SelectedValue = SomeEnum::Default;
```

which should do for now.

lib/Yield/PTML.cpp
aleclearmind: Ah this part emits spaces *after* the bytes. OK, maybe add a comment. I've to say that PTMLBuilder code is verbose and kinda hard to read.

In general I'd do what `objdump` does, but leaving this for later at this stage it's OK.

include/revng/Model/DisassemblyConfiguration.h
aleclearmind: Wait, are we emitting `\t`? Those are evil since you have somehow to assume the tabulation size.
If not, then please make a specific example of each option because it's unclear to me.

scripts/tuple_tree_generator/tuple_tree_generator/generators/typescript.py
mrjackv: Removing it still builds fine, I'd drop it

lib/Yield/PTML.cpp
Cvelth: Yeah, imitating `objdump` is non-trivial, since we'd need a way to introduce a "dummy" instructions that only have bytes, or fake it some other way. And then we need graph side to cooperate with that. And graph-side code is in a disarray already.

👌 on a comment.

include/revng/Model/DisassemblyConfiguration.h
Cvelth: :_t_burn:!!

We imitate it with spaces.

> example

So, if `RawByteColumnWidth` is set to `10`, then we would get something like:
```cpp
01        one_byte_instrunction
01 02     two_byte_instrunction
01 02 03 04 four_byte_instrunction
01 02 03 04 05 06 07 08 eight_byte_instrunction
```

Setting it to `20` leads to
```cpp
01                 one_byte_instrunction
01 02              two_byte_instrunction
01 02 03 04        four_byte_instrunction
01 02 03 04 05 06 07 08 eight_byte_instrunction
```

And so on.

scripts/tuple_tree_generator/tuple_tree_generator/generators/typescript.py
Cvelth: > Removing it still builds fine, I'd drop it

Did you try this branch? Or develop with this patch?

This matters in a case where there's a TTG-struct field inside another one.

include/revng/Model/DisassemblyConfiguration.h
Cvelth: The behave exactly the same.

scripts/tuple_tree_generator/tuple_tree_generator/generators/typescript.py
mrjackv: > Did you try this branch? Or develop with this patch?

This branch, with commit `b529e01f`, commit `76851834` gives me a C++ build error

> This matters in a case where there's a TTG-struct field inside a different struct.

Can you give me an example? Again as far as I remember, if all fields of a struct are optional then the constructor can be invoked with no arguments. Doing `{{}} as I{field.type}` would inevitably lead to bugs down the line.

include/revng/Model/DisassemblyConfiguration.h
aleclearmind: Can we have no configuration and simply right align addresses? Handling 32-bit and 64-bits as appropriate

32-bits:
```
    0x1000 lol
0x10001000 lel
```
64 bits:

```
        0x10001000 lol
0x1000100010001000 lel
```

Alternatively we can precompute the maximum size of an address in the current function.

Left aligning numbers is very sad.

Cvelth: Done.

scripts/tuple_tree_generator/tuple_tree_generator/generators/typescript.py
Cvelth: Dropped.

lib/Yield/PTML.cpp
Cvelth: Let's move this to the model PR instead.

lib/Yield/PTML.cpp
Cvelth: Since this got reworked, no reason to leave the comment anymore.

lib/Yield/PTML.cpp
Cvelth: Done.

aleclearmind:
Merged in 5cf07b5. Thanks.

PR #364, opened 2024-03-13
Miscellaneous fixes for Tier 2

pfez:
This patchset is needed for merging the latest `revng-c` fixes for tier 2.

It includes:
- addition of a `SetIntersectionLattice` to the new `MFP` implementation
- removal of old `MonotoneFramework` code, whose last users were in `revng-c`, and are now gone
- enforce determinism on CSV handling, which was causing further non-deterministic behaviour down the pipeline in `revng-c`


include/revng/Support/IRHelpers.h
```diff
@@ -943,11 +943,26 @@ inline std::vector<llvm::GlobalVariable *> extractCSVs(llvm::Instruction *Call,
   return Result;
 }
 
+inline bool CompareByName(const llvm::GlobalVariable *LHS,
+                          const llvm::GlobalVariable *RHS) {
+  revng_assert(LHS->hasName() and RHS->hasName());
+  return LHS->getName() < RHS->getName();
+};
+
+template<typename RangeOfPointersToGlobal>
```
aleclearmind, line 952: Can we have a (simple) concept constraint? cc/@Cvelth

lib/FunctionIsolation/PromoteCSVs.cpp
```diff
@@ -64,11 +66,13 @@ class PromoteCSVs {
   struct WrapperKey {
   public:
     Function *Helper;
-    std::set<GlobalVariable *> Read;
-    std::set<GlobalVariable *> Written;
+    std::vector<GlobalVariable *> ReadSortedByName;
+    std::vector<GlobalVariable *> WrittenSortedByName;
```
aleclearmind, line 70: I'd leave the old name and add doxygen about the fact that they are sorted.

include/revng/Support/IRHelpers.h
```diff
@@ -943,11 +943,26 @@ inline std::vector<llvm::GlobalVariable *> extractCSVs(llvm::Instruction *Call,
   return Result;
```
aleclearmind, line 943: Commit message says "In particulare".

aleclearmind (changes requested):

include/revng/Support/IRHelpers.h
Cvelth: We already have `range_with_value_type` just for this.

https://github.com/revng/revng/blob/3fb9e58b0a0826b52d9ae97c0e6cbb9920f71247/include/revng/ADT/Concepts.h#L143

lib/FunctionIsolation/PromoteCSVs.cpp
pfez: How about now?

include/revng/Support/IRHelpers.h
pfez: :facepalm: fixed

include/revng/Support/IRHelpers.h
pfez: Eh that doesn't always work, because it assumes `R` always have `value_type`, which is not the case if e.g. `R` is an iterator range. And that's one of the ways I'm using it in `PromoteGlobalToLocalVars.cpp` (with `llvm::make_first_range`).

I changed the concept not to use `value_type` and to use the typed obtained dereferencing `R.begin()` instead. How does that look now?

aleclearmind:
Merged in cdfd91c. Thanks.

PR #386, opened 2024-05-08
Fix license header

JackCrumpLeys:
The license header on some files refers to a non-existent `LICENSE.md`. Based on context I assume the license that should be linked is the one held at `LICENSE.mit`.


JackCrumpLeys:
unrelated note: letting any random pr trigger a CI run is inviting people to run evil code on your CI server.

aleclearmind:
> unrelated note: letting any random pr trigger a CI run is inviting people to run evil code on your CI server.

Yeah, we're going to disable it, but anon users shouldn't have access to any non-public information.
Still, they could start doing bad things from our CI's IP.

aleclearmind:
We'll do this internally, thanks for reporting this though.

Issue #367, opened 2024-03-29
Support AArch64 macOS ABI


trufae:
Steps to reproduce:

attached file [demo.zip](https://github.com/revng/revng/files/14806456/demo.zip)

$ cat revng.sh

```bash
#!/bin/bash
export BASH=/usr/bin/bash
cat > a.out
. /revng/environment
revng artifact --analyze --progress decompile-to-single-file $@ a.out | revng ptml --color
```

$ cat Dockerfile
```dockerfile
FROM ubuntu:22.04

# Avoid prompts from apt
ENV DEBIAN_FRONTEND=noninteractive

# Update and install basic utilities (optional)
RUN apt-get update && apt-get install -y \
    curl \
    git \
    xz-utils \
    && rm -rf /var/lib/apt/lists/*

# Your command here
RUN curl -L -s https://rev.ng/downloads/revng-distributable/master/install.sh | bash

COPY revng.sh /

ENTRYPOINT /revng.sh
```

$ cat Makefile

```Makefile
all:
	docker build --platform linux/amd64 -t revng:0.1 .

demo: all
	gcc -o demo demo.c
	docker run --rm --platform linux/amd64 -i revng:0.1 /revng.sh < demo

shell:
	docker run --platform linux/amd64 -it revng:0.1 /usr/bin/bash

.PHONY: demo run all
```

* then just run `make demo`

```
gcc -o demo demo.c
docker run --rm --platform linux/amd64 -i revng:0.1 /revng.sh < demo

[=======================================] 100%  0.93s Run module pass managers (1): MPM index: 0
[=======>                               ]  20%  0.95s Harvesting (10): Advanced Value Info
[>                                      ]   0% 12.29s Lifting: Peek new address
[=========>                             ]  25% 12.60s Translation (12): Lifting code
[===============================>       ]  81% 13.01s Lift pass (4): Translate
[===================>                   ]  50% 13.01s Run module passes (2): Lift Pass
[>                                      ]   0% 13.01s Run module pass managers (1): MPM index: 0
[>                                      ]   0% 13.03s Step lift (3): lift
[=============>                         ]  33% 13.03s Run step (3): Run the step
[==========================>            ]  66% 13.03s Produce steps required up to lift (2): lift
[========>                              ]  22% 13.03s Analysis execution (3): Produce step lift
[=========================>             ]  64% 13.12s Analysis list revng-initial-auto-analysis (5): detect-abi
[...]

PR #387, opened 2024-05-10
Improve arm64 ABI handling across the board

Cvelth:
Sister PRs:
- https://github.com/revng/orchestra/pull/150
- https://github.com/revng/revng-qa/pull/26


Cvelth:
This resolves #367

include/revng/ABI/Definition.h
```diff
@@ -284,6 +284,16 @@ type: struct
     type: bool
     optional: true
 
+  - name: ReturnValueLocationIsReturned
+    doc: |
+      Specifies whether functions accepting a return value location should
+      return it back.
```
aleclearmind, line 305: Specifies whether functions whose return value is saved in a memory area pointed by an argument passed by the caller (SPTAR) should also return such pointer on `ReturnValueLocationRegister`/`ReturnValueLocationOnStack`.

include/revng/ABI/Definition.h
```diff
@@ -145,6 +145,13 @@ type: struct
       ARM ABI), the value of `StackAlignment` should be equal to 4.
     type: uint64_t
 
+  - name: MinimumStackWordSize
```
aleclearmind, line 156: I'd avoid the "word" nomenclature, it's quite overloaded.

include/revng/ABI/Definition.h
```diff
@@ -145,6 +145,13 @@ type: struct
...
+    doc: |
+      States the minimum possible stack argument size in bytes.
+      When the factual type of an argument is below this value, it should be
```
aleclearmind, line 159: factual?

include/revng/ABI/Definition.h
```diff
@@ -102,6 +102,14 @@ type: struct
       (if `ArgumentsArePositionBased` is `false`).
     type: bool
 
+  - name: BigArgumentsUsePointersToCopy
+    doc: |
+      This allows a non-position-based ABI to take advantage of one of the main
+      traits of position-based ABI: ability to use pointers to copy to pass
+      big aggregates.
```
aleclearmind, line 109: These descriptions are very difficult to understand.

Can you make an example of the difference between an ABI using `true` and one using `false`?

include/revng/Model/Segment.h
```diff
@@ -79,7 +79,7 @@ class model::Segment : public model::generated::Segment {
 
 public:
   bool contains(MetaAddress Address) const {
-    auto EndAddress = StartAddress() + VirtualSize();
+    auto EndAddress = StartAddress().toGeneric() + VirtualSize();
```
aleclearmind, line 82: Wait segments should already be `Generic`

aleclearmind (changes requested):

include/revng/ABI/Definition.h
Cvelth: Was supposed to be `actual`, but I suppose I was thinking about some kind of `facts` at the time 😆

include/revng/Model/Segment.h
Cvelth: I saw `AArch64` segments failing because of alignment. Hence these changes. I guess PDB importer did something weird? 🤔

Cvelth: Do we want to have an assert about segments being generic instead?

aleclearmind: It should be in `verify`

include/revng/Model/ABI.h
```diff
@@ -235,28 +253,47 @@ inline constexpr model::ABI::Values getDefault(model::Architecture::Values V) {
...
+
+inline constexpr model::ABI::Values
+getDefaultForMachO(model::Architecture::Values V) {
+  switch (V) {
+  case model::Architecture::aarch64:
+    return model::ABI::Apple_AAPCS64;
+  default:
+    return getDefaultFallback(V);
   }
 }
```
aleclearmind, line 298: `getDefaultFallback` handles x86-64 ABIs, correct?

Cvelth: `getDefaultFallback` is the "current default" in a way. This lets you only specify "overrides" for the architectures that needs something special to limit duplication.

Cvelth: I guess we could be even stricter with it and explicitly disallow some combinations (for example MIPS in a Mach-O binary), but for now I opted to preserve the current behaviour.

include/revng/ABI/Definition.h
Cvelth: I went with
```cpp
      Specifies whether functions whose return value is saved in a memory area
      pointed by an argument passed by the caller (SPTAR) should also return
      such pointer using the first return value register.
```

include/revng/ABI/Definition.h
Cvelth: I went with `MinimumStackArgumentSize`.

include/revng/ABI/Definition.h
Cvelth: Done:
```
      This allows a non-position-based ABI to take advantage of one of the main
      traits of position-based ABI: ability to use pointers to copy to pass
      big aggregates.

      This is one of the defining features of arm64 ABI, which allows it to get
      the best of both approaches - an ability to pass aggregates by copy in
      registers from the non-position-based approach while also being able to
      sometimes avoid a copy of a bigger aggregate from the position-based
      approach.
```

include/revng/Model/Segment.h
Cvelth: Done: see `Model: make sure segments use generic addresses` for details.

lib/ABI/FunctionType/ValueDistributor.cpp
```diff
@@ -128,7 +128,16 @@ ValueDistributor::distribute(uint64_t Size,
     DA.OffsetOnStack = UsedStackOffset;
   } else {
     // This is a stack-only argument.
-    uint64_t PrePaddedOffset = ABI.alignedOffset(UsedStackOffset, Alignment);
+    if (not HasNaturalAlignment)
+      Alignment = ABI.MinimumStackArgumentSize();
```
aleclearmind, line 132: Suspicious `if`. Add `{ ... }`.

lib/ABI/FunctionType/Conversion.cpp
```diff
@@ -282,6 +299,10 @@ static bool verifyAlignment(const abi::Definition &ABI,
   if (*Offset == NextOffset) {
     revng_log(Log, "Argument slots in perfectly.");
     return true;
+  } else if (NextAlignment == 0) {
+    // This is the last argument.
+    // TODO: is there anything more we can check in here?
+    return true;
```
aleclearmind, line 305: What's special about the last argument?

include/revng/ABI/Definition.h
```diff
@@ -147,6 +147,13 @@ type: struct
       Setting this to true disables usage of vector registers entirely.
     type: bool
 
+  - name: AllowPuttingSimilarlyAlignedStackArgumentsTogether
```
aleclearmind, line 150: 1. This field is use before being introduced. Does this pass `check-branch`?
2. What does "similarly" mean?

include/revng/ABI/Definition.h
```diff
@@ -147,6 +147,13 @@ type: struct
...
+    doc: |
+      Setting this to true allows stack arguments to use struct-like alignment
+      rules instead of the common word-extension system. A notable ABI using
```
aleclearmind, line 153: I've no idea what the "word-extension system" is.
Please make examples on different behavior. Without examples, all of this documentation is rather obscure.

include/revng/ABI/Definition.h
```diff
@@ -154,6 +154,13 @@ type: struct
       this is apple flavor of the arm64 ABI.
     type: bool
 
+  - name: UseStrictAggregateAlignmentRules
+    doc: |
+      Treat all the structs passed using the stack as if they were aligned
+      unnaturally.
```
aleclearmind, line 160: 1. Can you define "unnatural alignment"?
2. Can you make an example where this makes a difference?

lib/ABI/FunctionType/Conversion.cpp
Cvelth: The padding after it seems to be somewhat inconsistent in some cases.

I still have some doubts about this (hence the TODO), but TL;DR of this change is: if we know this is the last argument, we don't really care about anything (as in, padding... or lack thereof) that comes after it.

include/revng/ABI/Definition.h
Cvelth: 1. Ah, thanks for noticing, something slipped through then. My bad.

2. Yeah, on the fresh look, the name is awful, what do you think about something like `StackArgumentsUseNormalStructAlignmentRules`?

include/revng/ABI/Definition.h
Cvelth: In practically all the ABIs, no argument can be smaller than a word, as in, they always get extended to a word (no matter the size, be it 1 byte or 2) before doing any allocation. Setting this to `true`, disables the behavior - making them more struct-like.

include/revng/ABI/Definition.h
Cvelth: 1. A struct that breaks natural alignment rules - as in contains at least one field for which the offset is not a multiple of its natural alignment.

2. Let me go into a bit of a tangent for context sake then. Technically both of these new options can be merged together since are only ever set to `true` for the apple arm64 abi. But I feel like they are distinct enough to be separate. Hopefully, for future reusal. the TL;DR is that `AllowPuttingSimilarlyAlignedStackArgumentsTogether` states that argument use relaxed alignment rules (as in they were normal fields of a struct). While `UseStrictAggregateAlignmentRules` states that "all the aggregates cannot be trusted, tip-toe around them and don't make any hard assumption", as in, let them behave the same way other unnatural struct behave, like, say
```cpp
struct __attribute__((packed)) {
  uint8_t field_at_offset_0;
  uint64_t unnaturally_aligned_field_at_offset_1;
};
```

lib/ABI/FunctionType/Conversion.cpp
aleclearmind: OK, elaborate the comment a bit.

include/revng/ABI/Definition.h
aleclearmind: Works, but I'd use "Regular" instead of "Normal".

include/revng/ABI/Definition.h
aleclearmind: OK, add what you said in the doc.

include/revng/ABI/Definition.h
aleclearmind: 1. Can we put a micro glossary at the top of the file?
2. So AFAIU we're following rules that are more typical for laying out fields in a `struct` rather than laying out arguments? If so, maybe we should have a single option that reflects this?

aleclearmind: I think this too complex for me too understand without a specific example. Please make a specific example explaining the different behavior.
Each configuration option should come with an example, otherwise they are inaccessible.

lib/ABI/FunctionType/ValueDistributor.cpp
Cvelth: Done.

lib/ABI/FunctionType/Conversion.cpp
Cvelth: Done.

```cpp
    // This is the last argument.
    //
    // Since we often find trailing padding to be inconsistent, it would be
    // hard to enforce it, as such, just assume that if we know this argument
    // is the last one, it's always right.
```

include/revng/ABI/Definition.h
Cvelth: Done.

include/revng/ABI/Definition.h
Cvelth: Done.

```
Setting this to true allows stack arguments to use slots smaller than a word (makes them behave like a single struct instead of a bunch of arguments): so they don't get extended unnecessarily. A notable ABI using this is apple flavor of the aarch64.

For example

void function(/* enough arguments of all the registers */, uint8_t, uint8_t);

When this is set to true, both `uint8_t` will get placed as if they were a struct, that is without any padding in-between. When this is false, both of them will get extended to register size before getting placed, leading to padding.
```

include/revng/ABI/Definition.h
Cvelth: While at it I also moved the documentation out of that crowded `markdown` within `yaml` within `c++ header` structure into it's own file. Please give it a read when you have a moment. It's in `references/abi-definition.md`. Might also be a good time to elaborate on some of them that feel unclear to you.

aleclearmind: Oh God this is so good. Let me read it and leave some comment there. Thanks thanks thanks.

share/doc/revng/references/abi-definition.md
```diff
@@ -0,0 +1,213 @@
...
+```cpp
+void function(uint32_t, Big);
+```
+the `uint32_t` argument would be passed in the first GPR (`r0`).
+
+But that would also mean that the remaining three available GPRs are not enough to fit the entirety of the `Big` object, meaning it needs to either be split between the registers and the memory, or passed using the stack. That's exactly what this option states.
+
+## `BigArgumentsUsePointersToCopy` (`bool`)
+
+This allows a non-position-based ABI to take advantage of one of the main traits of position-based ABI: ability to use pointers to copy to pass big aggregates.
```
aleclearmind, line 61: > ability to use pointers to copy to pass big aggregates

"to copy to pass" reads weird.

share/doc/revng/references/abi-definition.md
```diff
@@ -0,0 +1,213 @@
...
+
+This is one of the defining features of arm64 ABI, which allows it to get the best of both approaches - an ability to pass aggregates by copy in registers from the non-position-based approach while also being able to sometimes avoid a copy of a bigger aggregate from the position-based approach.
```
aleclearmind, line 63: > also being able to sometimes avoid a copy of a bigger aggregate from the position-based approach

You seem to imply this saves some copying, but in practice this will still be a pointer to a copy, I think, since the callee can change it.

share/doc/revng/references/abi-definition.md
```diff
@@ -0,0 +1,213 @@
...
+
+## `NoRegisterArgumentsCanComeAfterStackOnes` (`bool`)
+
+States whether ABI allows a stack argument (mainly if it's too big to be placed in the registers) to precede other register arguments.
+
+For example, if there is a `Big` struct that has to use the stack, and
+a function like
+```cpp
+void function(Big, uint32_t);
+```
```
aleclearmind, line 73: Leave blank lines around code snippets or GitHub syntax highlighting breaks.

image

share/doc/revng/references/abi-definition.md
```diff
@@ -0,0 +1,213 @@
...
+if this value is set to true, both argument will be passed on stack, otherwise, only the struct will.
+
+## `AllowUnnaturallyAlignedTypesInRegisters` (`bool`)
```
aleclearmind, line 76: Mention two notable examples (one for `true`, one for `false`).

Same for all other options.

share/doc/revng/references/abi-definition.md
```diff
@@ -0,0 +1,213 @@
...
+
+Specifies who is responsible for cleaning the stack after the function call. If equal to `true`, it's the callee, otherwise it the caller.
+
+## `FloatsUseGPRs` (`bool`)
+
+Setting this to true disables usage of vector registers entirely.
+
+## `StackArgumentsUseRegularStructAlignmentRules` (`bool`)
+
+Setting this to true allows stack arguments to use slots smaller than a word (makes them behave like a single struct instead of a bunch of arguments): so they don't get extended unnecessarily. A notable ABI using this is apple flavor of the aarch64.
```
aleclearmind, line 90: The concept of "slot" has not been defined, but seems rather important.

share/doc/revng/references/abi-definition.md
```diff
@@ -0,0 +1,213 @@
...
+```cpp
+void function(/* enough arguments of all the registers */, uint8_t, uint8_t);
+```
+When this is set to true, both `uint8_t` will get placed as if they were a struct, that is without any padding in-between. When this is false, both of them will get extended to register size before getting placed, leading to padding.
+
+## `UseStrictAggregateAlignmentRules` (`bool`)
+
+Treat all the structs passed using the stack as if they were aligned unnaturally[^1].
+
+This means that they are going to handled with extreme care (which usually just means extra padding).
```
aleclearmind, line 102: Obscure, I don't get what this means.

share/doc/revng/references/abi-definition.md
```diff
@@ -0,0 +1,213 @@
...
+
+## `StackAlignment` (`unsigned`)
+
+States the required alignment of the stack at the point of a function call in bytes.
+
+> **_Note:_** states minimum value for ABIs supporting multiple different stack alignment values, for example, if the ABI requires the stack to be aligned on 4 bytes for internal calls but on 8 bytes for interfaces (like 32-bit ARM ABI), the value of `StackAlignment` should be equal to 4.
```
aleclearmind, line 108: > the ABI requires the stack to be aligned on 4 bytes for internal calls but on 8 bytes for interfaces

WHAT?
What is an "internal call"?
ABI are evil.

share/doc/revng/references/abi-definition.md
```diff
@@ -0,0 +1,213 @@
...
+
+
+## `MinimumStackArgumentSize` (`unsigned`)
+
+States the minimum possible stack argument size in bytes. When the actual type of an argument is below this value, some padding with be inserted.
```
aleclearmind, line 113: I wonder how this plays with `StackArgumentsUseRegularStructAlignmentRules`.

share/doc/revng/references/abi-definition.md
```diff
@@ -0,0 +1,213 @@
...
+
+## `StackBytesAllocatedForRegisterArguments` (`unsigned`)
```
aleclearmind, line 115: AFAIU these are bytes *before* or *after* the stack arguments?

```
lol(1,2,3);
```
Is it
```
push 1
push 2
push 3
push --
push --
push --
call
```
or
```
push --
push --
push --
push 1
push 2
push 3
call
```

share/doc/revng/references/abi-definition.md
```diff
@@ -0,0 +1,213 @@
...
+
+In our context, this serves as the starting offset for the first stack argument. As in, it specifies the number of byte on the top of the stack struct that should be ignored.
+
+## `MaximumGPRsPerAggregateArgument` (`unsigned`)
+
+States the maximum number of GPRs available to pass a single aggregate (a struct, a union, etc.) argument, meaning that it can only be passed in the GPRs if `MaximumGPRsPerAggregateArgument` is less than or equal to the number of the registers required to fit the object including padding.
+
+> **_Note:_** If `MaximumGPRsPerAggregateArgument` is equal to 0, it means that the ABI does not allow aggregate arguments to use GPRs.
+
+> **_Note:_** If an argument doesn't fit into the specified registers or uses irregular padding, the registers are not used and the object is passed using the memory (stack, pointer-to-copy, etc.).
```
aleclearmind, line 127: 1. > irregular padding

   What is "irregular padding"? Is it the same as an "unnaturally aligned" `struct`?
2. Is there an ABI where this is not either "all argument registers or none"? That is, could this be a `bool`?
   Same question for the two next options.

share/doc/revng/references/abi-definition.md
```diff
@@ -0,0 +1,213 @@
...
+
+## `CalleeSavedRegisters` (`list of registers`)
+
+Specifies the list of registers for which the ABI requires the callee to preserve the value, meaning that when the callee returns, the value of those registers must be the same as it was when the function was called.
+
+## `ReturnValueLocationRegister` (`register`)
+
+Specifies a register to be used for returning (or even passing, depending on ABI) the pointer to the memory used for returning copies of big aggregate objects.
+
+Must be set to `model::Register::Invalid` for ABIs that do not support returning values by 'pointer-to-copy' or use stack for it.
```
aleclearmind, line 177: 1. Use `"` for `pointer-to-copy`.
2. I don't get if, in SystemV x86-64, this is `rdi` or `rax`. The "for returning (or even passing, ...)" part confuses me.
   If it's `rdi` I wouldn't say it's "the register to *return* the pointer to the memory", it's the register used to *pass* it.
3. I wonder if the situation with "return value locations" could be better represented with an `enum` instead of using multiple `bool`s.

share/doc/revng/references/abi-definition.md
```diff
@@ -0,0 +1,213 @@
...
+
+This is only relevant if `ReturnValueLocationRegister` is set to `Invalid`.
+
+## `ReturnValueLocationIsReturned` (`bool`)
+
+Specifies whether functions whose return value is saved in a memory area pointed by an argument passed by the caller (SPTAR) should also return such pointer using the first return value register.
+
+This is only relevant if either `ReturnValueLocationRegister` or `ReturnValueLocationOnStack` is enabled.
+
+## `ScalarTypes` (`list of scalar types`)
```
aleclearmind, line 191: I think there's some notable example why we need this. A 32-bit integer with 64-bit alignment or viceversa, please report it.

aleclearmind (changes requested):

share/doc/revng/references/abi-definition.md
Cvelth:
```diff
-use pointers to copy to pass
+use pointers-to-copy to pass
```
?

share/doc/revng/references/abi-definition.md
Cvelth: Well, that's usually the point. If it can prove that the callee never touches it, that copy is the very first thing to go away.

share/doc/revng/references/abi-definition.md
aleclearmind: Ah, that makes sense.

share/doc/revng/references/abi-definition.md
Cvelth: Actually, I yoinked that terminology from the arm doc and have no idea how to even express it otherwise.

Looking at their doc, it doesn't look like they explain it anywhere either. They just use it, like:

> When an argument is assigned to a stack slot any unused padding bytes have unspecified value.

or

> `ldr x2, [x0, #__stack] // get stack slot pointer`

share/doc/revng/references/abi-definition.md
Cvelth: When about something like
```
This means that options that assume regular alignment rules (like `StackArgumentsUseRegularStructAlignmentRules`) stop applying to them, while others (such as `AllowUnnaturallyAlignedTypesInRegisters`) start.
```
?

share/doc/revng/references/abi-definition.md
Cvelth: Yeah, this is courtesy of arms. They call it "public interface"s and "private interface"s though. Maybe we should switch to the same terminology 🤔

share/doc/revng/references/abi-definition.md
Cvelth: `StackArgumentsUseRegularStructAlignmentRules` is specifically added to ignore this size.

share/doc/revng/references/abi-definition.md
Cvelth: They are the first four arguments. All the ABIs we support that use these provide as many as there are register arguments. As far as I understand, they were originally untended to mirror register values, but that is not guaranteed anymore.

share/doc/revng/references/abi-definition.md
aleclearmind: By "never touches it" I guess you mean doesn't write to it.

You're right, but I'm not sure that was the original design goal, it's a rather advanced compiler optimization that requires view on the callee, seems weird for an ABI to be designed for such a thing.

It'd be interesting to find a counterexample on godbolt, but it's not too important.
I'll leave this to you.

share/doc/revng/references/abi-definition.md
Cvelth: 1. Yeah, as far as I remember, this comment predates our support of unnatural alignment, hence the weird terminology. Thanks for noticing!

2. aarch64 has `2`. There, if it's under 2 words - pass it as is, otherwise, use pointer-to-copy.

share/doc/revng/references/abi-definition.md
aleclearmind: Wait but is this private vs public something under the control of the dev?
Is this related to how far the target function is?
Or maybe whether is static or not?
Or maybe its visibility?

I need some reference.

share/doc/revng/references/abi-definition.md
aleclearmind: I would then add a preamble such as "Certain ABIs (for example XXX) typically extend each stack argument to a slot (i.e., to a size of a GPR). This option...".

AFAIU this is an option that it specific  to Apple flavor of ARM's ABI and has no real general way to be described.
I think pretending this is not the case becomes counterproductive. I think the docs should be explicit and just say "this is an ad-hoc thing for...". Then, if in the future we can subsume this option in something more general we can get rid of that part, but as of now, pretending this is not ad-hoc, just makes the docs very mysterious.

share/doc/revng/references/abi-definition.md
aleclearmind: Can't we merge the two options somehow? Maybe with a special value for `MinimumStackArgumentSize`?

> StackArgumentsUseRegularStructAlignmentRules is specifically added to ignore this size.

Since that's the case, state it in the docs.

I'm starting to thing that some grouping of these options (in the docs) would be helpful. Maybe we can have a section about stack arguments, a section about return values and things like this.

share/doc/revng/references/abi-definition.md
aleclearmind: 1. OK, so these are "shadow arguments" and where they are on the stack depends on whether arguments are pushed LTR or RTL. Add a note about this.
2. Now that I think about it, we don't have an option to specify if stack arguments are pushed LTR or RTL. Shouldn't we have such a things?

share/doc/revng/references/abi-definition.md
Cvelth: > By "never touches it" I guess you mean doesn't write to it.

Yeah, sorry.

------------

I looked around but couldn't find source I originally referred too. So I went and played around in godbolt.

I couldn't make it happen under msvc at all. It kept trying to inline the callee. And explicitly forbidding it from doing so seems to disable optimizations around the call in the first place (it's probably similar to gcc's `weak` attribute) - so not much success there.

I did manage to reproduce the behavior with [clang](https://godbolt.org/z/EETvKEKGK) using only `-O1`, but again, who knows how much of that is just clang being awesome.

That being said, I think you're right and we should just drop this from docs.

Cvelth:
```
This is one of the defining features of arm64 ABI, which allows it to get the best of both approaches - an ability to pass aggregates by copy in registers from the non-position-based approach while also being able to avoid "wasting" registers on passing big aggregates from the position-based approach.
```

?

share/doc/revng/references/abi-definition.md
Cvelth:
```
Most ABIs extend each stack argument to a GPR-sized slot. Setting this option to true, allows them occupy less space by behaving more like fields in a regular struct would.

This option is here exclusively for the apple flavor of the aarch64, since nothing else we support does anything similar.
```
?

share/doc/revng/references/abi-definition.md
aleclearmind: Works.

share/doc/revng/references/abi-definition.md
aleclearmind: Works.

share/doc/revng/references/abi-definition.md
Cvelth: Ok, digging a bit deeper into it, it doesn't seem as bad.

This is what their docs have to say on the matter:
```
This definition of conformance gives maximum freedom to implementers. For example, if it is
known that both sides of an externally visible interface will be compiled by the same compiler,
and that the interface will not be publicly visible, the AAPCS64 permits the use of private
arrangements across the interface such as using additional argument registers or passing data
in non-standard formats. Stack invariants must, nevertheless, be preserved because an
AAPCS64-conforming routine elsewhere in the call chain might otherwise fail. Rules for use of
IP0 and IP1 must be obeyed or a static linker might generate a non-functioning executable
program.
Conformance at a publicly visible interface does not depend on what happens behind that
interface. Thus, for example, a tree of non-public, non-conforming calls can conform because
the root of the tree offers a publicly visible, conforming interface and the other constraints are satisfied.
```

To me it just sounds like "hey, you can do what you want with static (or maybe even partially inlined?) calls, but the rules for public interfaces are as follows".

aleclearmind: Please use quote (`> some text here`) for long pasting text. Using code snippets disable word-wrapping and makes it hard to read.

> To me it just sounds like "hey, you can do what you want with static (or maybe even partially inlined?) calls, but the rules for public interfaces are as follows".

Same. So... can we drop this option?

share/doc/revng/references/abi-definition.md
Cvelth: > Can't we merge the two options somehow?

Would be pretty awkward, I think. Since it's not like `StackArgumentsUseRegularStructAlignmentRules` overrides `MinimumStackArgumentSize` per se - it's still there, kind of. It just allows multiple small arguments to get implicitly "struct"-ified.

> StackArgumentsUseRegularStructAlignmentRules is specifically added to ignore this size.

`*` in certain cases

(my bad on wording)

> grouping

Yeah, that was the original intention, but we only had a handful after the first iteration, so it got postponed. Eventually we'll have to get to it :D

As of now they are de-facto grouped like:
1. Name.
2. All the weird bools.
3. Register stuff.
4. Scalar type stuff.

share/doc/revng/references/abi-definition.md
Cvelth: 0. Ok, I was wrong about the intention. Digging for the pointers-to-copy, I also ran into this.

They are intended to just "save" volatile registers, but more efficiently (?), since you don't need to move stack pointer - you already have space ready for them.

As far as I understand there are no limitation on how that extra space is used though.

1. 👌.

2. I think we had it at some point, but you didn't like the fact that it was unused, so we got rid of it, or something like that.

share/doc/revng/references/abi-definition.md
Cvelth: 1. 👌 
2. Looks like a leftover from an old design, my bad. Yeah, it's only about passing the pointer in. Returning (when it happens) uses the first GPR return value register from the usual list.
3. It's only two bools though, so no idea if enum is worth the overhead. Don't forget that this first one is a register - and enum is not going to help us get rid of it.

share/doc/revng/references/abi-definition.md
Cvelth: Option, no - we still need to know alingment 😆.

Note - probably 🤔.

Cvelth: Ah, but wait... It's currently set to the less restrictive value 🤔.

This might be tricky. Because I don't think I would have set to the more obscure value without a good reason. Some static function somewhere was probably interfering. And I'm not sure we're ready to split `arm-aapcs-static` ABI from `arm-aapcs` 😭.

share/doc/revng/references/abi-definition.md
Cvelth: Nothing that drastic.

There are two factors that influenced this kind of list:

1. We wanted a very explicit list of supported scalars. So everything outside the list is out by definition.

2. The alignment is unpredictable, some ABI just say "everything is aligned at most at register size". Others are like "scalar size is the alignment no matter what", and all the steps in-between.

We could have probably got away with an enum / a couple of bools for the second point, but IMO, structuring it like this makes it a lot easier to read.

share/doc/revng/references/abi-definition.md
aleclearmind: I'm confused, the whole ABI project should entirely ignore optimizations that the compiler can do for `static` functions. This is very important: handling what the compilers do for `static` functions is a non-goal that should not be pulled in. Maybe one day we'll add `-static` ABIs, but not today.

We should entirely ignore the whole paragraph you posted. Compilers are already doing whatever they want in that case, there's no need for permission by the ABI.

share/doc/revng/references/abi-definition.md
aleclearmind: Can you reorganize the document in sections?

share/doc/revng/references/abi-definition.md
aleclearmind: > I think we had it at some point, but you didn't like the fact that it was unused, so we got rid of it, or something like that.

Mmmh, we support both x86 `cdecl` and `PASCAL` which have different ways to push arguments: https://en.wikipedia.org/wiki/X86_calling_conventions#List_of_x86_calling_conventions

share/doc/revng/references/abi-definition.md
aleclearmind: 2. OK, make sure we clarify that and make the SystemV x86-64 example.
3. Ah, right. Grouping them together and (if it makes sense) cross-referencing each other will probably make the situation clearer.

Cvelth: 2.
> Specifies which register is used for passing the pointer to the memory used for returning big aggregate objects.

> For example, for `SystemV_x86_64` ABI it is set to `rdi` (the first argument register), but for `arm64` ABI it is set to a dedicated `x8` register (hence the need for a separate option).

> It must be set to `model::Register::Invalid` for ABIs that do not support returning values by "pointer-to-copy" or use stack for it.

share/doc/revng/references/abi-definition.md
Cvelth: 1. Actually, on a fresh look. We should probably just drop the unnatural alignment part of the note to begin with. Since, unlike when it was written, we now have a dedicated option for that.

Cvelth: Or even the entire note? 🤔

It sounds so basic that I'm not sure if it is worth keeping.

share/doc/revng/references/abi-definition.md
Cvelth: > Compilers are already doing whatever they want in that case, there's no need for permission by the ABI.

Well, turns out, they do in fact have a permission! Who would have guessed.

> This is very important: handling what the compilers do for `static` functions is a non-goal that should not be pulled in.

But here the catch: we can't reliably tell if what we are looking at is a static function or not. So we still need at least some basic handling.

Let's look at this specific case. Arm abi requires that after every instruction, the stack is aligned at 4 bytes. But, on a "public interface" instruction, it requires 8 byte alignment.

By encoding less strict one in our system, we prevent horrible breakage in the case we run into static function (memory overlap), while only sacrificing trailing padding, which, let's be honest, we don't even care about that much - we already take some liberties around last argument as is.

share/doc/revng/references/abi-definition.md
aleclearmind: OK

share/doc/revng/references/abi-definition.md
aleclearmind: Perfect.

share/doc/revng/references/abi-definition.md
Cvelth: Wait-wait-wait. Don't forget that the lowest we ever go is RFT. Isn't RFT's stack struct already ordered the right way?

If not, then the current pascal implementation is broken. On top of that, it's probably not trivial to fix it, because we can't just "reverse" a struct. The endianness starts coming into the play, and we managed to somehow avoid worrying about it until now.

aleclearmind: Let's take a step back.
The ABI stuff is used for two purposes: 1) going from a `struct` representing stack arguments attached to an RFT to plain arguments in CABI and 2) given some CABI stack arguments get their offset on the stack.

In both cases, the order in which arguments are pushed on the stack is very relevant.

If you an RFT with stack arguments type `struct { uint32_t a; uint32_t b; }` you need to know if it's `void myfunc(uint32_t a, uint32_t b)` or `void myfunc(uint32_t b, uint32_t a)`.

In the same way, for use case 2), given `void myfunc(uint32_t a, uint32_t b)` you need to know if `a` is at offset -4 and `b` at offset `-8` or viceversa.

> we can't just "reverse" a struct

I don't think we need to do that. For 1), we can keep doing what we're doing but simply reverse stack arguments and for 2) we need walk CABI stack arguments backwards, I guess.
But this is a bug if we're not doing this.

share/doc/revng/references/abi-definition.md
Cvelth: Done.

share/doc/revng/references/abi-definition.md
Cvelth: Done.

share/doc/revng/references/abi-definition.md
Cvelth: Done.

share/doc/revng/references/abi-definition.md
Cvelth: Done.

share/doc/revng/references/abi-definition.md
Cvelth: Done.

share/doc/revng/references/abi-definition.md
Cvelth: Please take a look (it's in `squash 2`)

share/doc/revng/references/abi-definition.md
Cvelth: Gone.

share/doc/revng/references/abi-definition.md
Cvelth: Done.

share/doc/revng/references/abi-definition.md
aleclearmind: We're trying to guess what the compiler does in case of `static` functions, which is a losing battle.

> we prevent horrible breakage in the case we run into static function

Can you make an example of the consequence of encoding StackAlignment to 8 in such a situation? It's not clear to me.
The fact that this actually happens and has catastrophic consequence that we cannot prevent needs some significant backing evidence. Otherwise, I really want to strictly follow the ABI rules, and the ABI says 8.
If we were a compiler and we used this information to emit a call, we'd lose the alignment constraint and emit invalid code.

But in any case, if something horrible happens when we someone does not respect the ABI, that's fine. The problem is that conversion from RFT to CABI should have failed. If we don't have enough information to make it fail, it might be reasonable to add to RFT a new field to detect such situations. For instance, we could add `StackAlignmentUponCall` and have `detect-abi` try to detect it (maybe).

Cvelth: > Can you make an example of the consequence of encoding `StackAlignment` to 8 in such a situation?

It leads to the layout reporting arguments as too big. As in we would grab some memory that is not an argument, and label it as such. I'd say it's pretty likely that there could be a local variable or something in there - so now we observe extra reads and writes into such an "argument".

Given, we'd needs to investigate to reliably tell what'd happen in our case because of such misreporting, depends on how much we rely on the layout information. That being said, consider the opposite error: we report too little padding, end up with some hole in memory (that was never accessed to begin with), which will probably just disappear at some point (when we are deciding what to promote into an variable?), nobody is hurt.

> If we were a compiler

If we were a compiler, we'd never be able to get away with a handful of options! We'd need something akin to the headache clang is encoding in their :cursed_parrot: code gen tables.

I get what you're trying to say, but also, I'm not sure what's more important: being closer to the source of truth (even though we're never matching it) or going up in flames as rarely as possible.

> The problem is that conversion from RFT to CABI should have failed.

Well, yeah, if we could do that, the world would be a much nicer place (I would even let you ride my unicorn! /j). The problem is that everything related to alignment we have is pretty much guesswork. There's usually no way to tell padding apart for unused words. And it sometimes even depends on types alone - and we obviously can't rely on the fact that we recovered those correctly.

aleclearmind: I still do not understand the extent of the problem. I need a specific example, in a specific scenario (e.g., RFT->CABI or CABI->Layout), with a specific prototype where you can say "if `StackAlignment` is 8 this happens, if it's 4 this other thing happens".

share/doc/revng/references/abi-definition.md
aleclearmind: Works, but drop the index, it will go out of sync in no time.

Cvelth: It's generated by a vscode extension, so it's not going to get out of sync if I'm the one changing the options.

The question is how often do you expect other people adding removing options from here? Because if not very, I'd rather we keep the index.

aleclearmind: Frankly I see no reason why this shouldn't end up in docs.rev.ng. I'd add a "Developer's manual" section to `mkdocs.yml` and move it to `share/doc/revng/developer-manual/xxx.md`. It has been a while that I wanted to move stuff there.

This way we have no ToC issues, mkdocs will create it for us.

What do you think?

Cvelth: Perfect! I didn't do it because, hey, this PR is already stretching the scope to its limits, and documentation structuring is something we'll definitely have to take a closer look at once we have more of these random writeups piled up.

I didn't want to add anything more to that already crowded `markdown` in `yaml` in `c++ header` monstrosity we had, but giving this its place is even better than that.

That being said, I don't think this should fall under the manual category 🤔.

`developer-reference/abi-definition.md`? `developer-manual/reference/abi-definition.md`?

aleclearmind: Mmh, maybe for now put it under `references`. Then, once we'll create `developer-manual/` we'll split the references under  `user-manual/` and `developer-manual/`.

Also, one day this stuff might become part of the model, so not too bad to put it a more user facing area.
Let's put this in `references/`.

share/doc/revng/references/abi-definition.md
Cvelth: This is so hard to get right. I feel like the default behavior you're talking about is somewhat implicit, and beside the point of this option: it's about which _other options_ should apply to 
this type, not about the alignment specifics.

Cvelth: Either way, I added an extra note:

> Practically this option is mostly used to conditionally ensure [`StackArgumentsUseRegularStructAlignmentRules`](#stackargumentsuseregularstructalignmentrules-bool) does not apply to structs (see apple aarch64 abifor reference).

Do you still feel like this is not enough?

share/doc/revng/references/abi-definition.md
Cvelth: Done, as per the mattermost discussion.

share/doc/revng/references/abi-definition.md
Cvelth: Done.

share/doc/revng/references/abi-definition.md
```diff
@@ -0,0 +1,244 @@
...
+
+When this is set to true, both `uint8_t` will get placed as if they were a struct, that is without any padding in-between. When this is false, both of them will get extended to register size before getting placed, leading to padding.
+
+## `UseStrictAggregateAlignmentRules` (`bool`)
+
+Treat all the structs passed using the stack as if they were aligned unnaturally[^1].
+
+This means that options that assume regular alignment rules (like `StackArgumentsUseRegularStructAlignmentRules`) stop applying to them, while others (such as `AllowUnnaturallyAlignedTypesInRegisters`) start.
+
+Practically this option is mostly used to conditionally ensure [`StackArgumentsUseRegularStructAlignmentRules`](#stackargumentsuseregularstructalignmentrules-bool) does not apply to structs (see apple aarch64 abifor reference).
```
aleclearmind, line 132: > see apple aarch64 abifor reference

Typos/case

share/doc/revng/references/abi-definition.md
aleclearmind: > > It would be nice to have some section describing how naturally aligned structs (and maybe unnaturally aligned structs) are "normally" handled and then this option can read more easily as an exception to what's normal.
>
> the default behavior you're talking about is somewhat implicit, and beside the point of this option

What I meant here is a section on the top of the document, not in the description of this specific option.

I'm still confused about this set of options:
1. Maybe we should replace the term "unnaturally aligned" with "packed" all over the place. It's much more intuitive to me.
2. About `StackArgumentsUseRegularStructAlignmentRules`: can we negate this option and rename this to `PackStackArguments`? I feel like "stack arguments use regular struct alignment rules" states what's normal while here we want to specify an exceptional behavior. Alternatively we could call this `AlignStackArguments`, but I feel `PackStackArguments"`makes you immediately think "no padding", which is what happens in practice.
3. `UseStrictAggregateAlignmentRules` description mentions that `AllowUnnaturallyAlignedTypesInRegisters` "starts applying", which applies to register arguments, but `UseStrictAggregateAlignmentRules` seems to apply to stack arguments only. I'm confused.
   Does `UseStrictAggregateAlignmentRules` apply to register arguments? If not, don't mention `AllowUnnaturallyAlignedTypesInRegisters`.
4. AFAIU, `StackArgumentsUseRegularStructAlignmentRules` means "pack together stack arguments".
   `UseStrictAggregateAlignmentRules` means "behave as if the struct passed on the stack were packed".
   Is this correct?
   We need to document somewhere (at the top?) what's the "standard" behavior for non-packed (naturally aligned) and packed (unnaturally aligned) structs on the stack. We can't customize something that has not been defined.

Cvelth: 1. A single struct can be both packed and naturally aligned, they are not exclusive.
2. Technically we could, but reversing it would mean using the non-default value for every ABI but one, which is sad. Maybe we should just leave a TODO and do that once "default value" in TTG is implemented?
2.1. That being said, I do agree that `PackStackArguments` fits it better! Thanks for the idea!
3. `UseStrictAggregateAlignmentRules` (was named something along the lines of `treat all structs as if they were unnaturally aligned`), hence why all the "unnatural alignment facing" options start to apply. That being said, I see why it would be confusing for those who are not aware of that history. Should we rename it again?
4. I'll see what I can do!

Cvelth: 2. Actually, on the second thought

> I feel like "stack arguments use regular struct alignment rules" states what's normal while here we want to specify an exceptional behavior.

What is says is that ABI uses the normal behavior for _struct fields_ instead of the normal behavior for _stack arguments_. But yeah, good catch, word `pack` fits it really well.

aleclearmind: 1. If a `struct` is packed but happens to be naturally aligned, I would define it as "not packed" anyway. We can give such definition at the top.
   I understand that this might be confusing, since we unconditionally emit all of our structs as `packed`, but if a C programmer uses `packed`, it's because he wants them to ignore alignment rules, otherwise it would just leave them out.
   I find "packed" so much more intuitive than "unnaturaly aligned" (which makes me think of some mutatant monster).
   One day we might even decide to remove `packed` from some of our `struct`s in the decompiled C code, more readable at the cost of less portable C, even if in the end the C code we emit is already tied to an ABI. But let's stay focused on the original point.
2. > using the non-default value for every ABI but one

   That's fine with me.
3. Let's ignore history entirely and find a suitable name. What do you propose?
4. :+1:

Cvelth: 1. I still have some doubts, but if you're believe in it so strongly, sure.

3. The point of the option is kind of straightforward: "hey, struct handling in this ABI sucks, treat _them all_ as carefully as you possibly can". But expressing is shorter than that is kind of hard. And, even after your argument for packed in `1` which I can kind of get behind, it doesn't really hold in here, since all of our structs are _packed_, that would probably just end up confusing people. And if we are dropping `unnaturally-aligned` thing elsewhere, it's definitely not worth keeping it only here - for consistency sake if for nothing else. And those two were my best bets 😅.

I'm starting to doubt whether dropping `unnaturally-aligned` term is a good idea 🤔.

aleclearmind: 2. 1. I've absolutely no idea what "treating arguments *carefully*" means... Like, are we being sloppy with others? You have this concept of "careful" (I've seen such comments in the code too) which don't bind to anything for me. It seems like you've been working so much on this that certain things are obvious to you, but they are not obvious to others.
   2. As for the confusion on the term `packed`, I really think we should stop emit all structs as packed, or at least have an option to emit them non-`packed` if the chosen ABI for the backend allows that.
   Apart from this, as long as we put a warning at the top, I think we're fine.
   Consider `StackArgumentsUseRegularStructAlignmentRules`, would you prefer to hear "you have to pretend those two `uint8_t` are in a `struct` and that `struct` respects the standard alignment conventions (which you don't remember)" or just "those two `uint8_t` are packed together".

Cvelth: > It seems like you've been working so much on this that certain things are obvious to you, but they are not obvious to others.

I wouldn't be surprised to be honest.

That being said, by "carefully", I mean taking all the precautions: as in full alignment everywhere around it (padding, yay) as well as not using registers (aside from some ABIs, like 32-bit, ARM that specifically allow it) for any it.

> an option to emit them non-`packed` if the chosen ABI for the backend allows that.

Honestly, I wouldn't mind adding a new `StructABI` field to the binary and then, as long every field of said struct makes sense in terms of the abi (we already do very similar checks when converting stack struct into separate arguments, shouldn't be hard to extrapolate it out).

At the very least we should do
```cpp
for (auto Field : Struct.Fields())
  if (Field.Offset() % ABI.alignment(Field.Type()) == 0 && ABI.isNaturallyAligned(Field.Type()))
    assume_struct_is_not_packed();
```
but I'd go full route of "deconstructing" and then "reconstructing" struct just to be sure. Since we rarely get more than handful of fields, it shouldn't be computationally expensive and I would be able to sleep better at night :D.

Going to those length would let us enable "non-packed struct" by default without many downsides. And it shouldn't be hard either, since, again, all the work is already done, it's the question of factoring the relevant piece of code out and making it callable from outside.

But, yeah, if you are on board with something like this, it should be its own PR.

> Consider `StackArgumentsUseRegularStructAlignmentRules`, would you prefer to hear "you have to pretend those two `uint8_t` are in a `struct` and that `struct` respects the standard alignment conventions (which you don't remember)" or just "those two `uint8_t` are packed together".

Honestly, agreed. I think the reason I preferred that way of phrasing it is because of sensible defaults: I'd rather it inherited the struct behavior from the same ABI in case we ever introduce a new one, rather than trying to define an extra concept of "packing" which ABI currently is not aware of.

That being said, in the hindsight, I was definitely overthinking it. Let's go with packed terminology ~~for now~~ until we have a reason not to.

Cvelth: > At the very least we should do

Actually, now that I think about it, we might be able to get away with just `ABI.isNaturallyAligned(Struct)` 🤔. But that would need extensive testing.

aleclearmind: :+1:

aleclearmind:
Merged in 5993cee. Thanks.

PR #388, opened 2024-05-13
Introduce `Undirected` `GraphTraits` for `GenericGraph`

andrealinux1:
Introduce the `Undirected` GraphTraits, which treat a `Bidirectional` `GenericGraph` as an undirected graph.

### DIFF
--- a/include/revng/ADT/GenericGraph.h
+++ b/include/revng/ADT/GenericGraph.h
@@ -15,6 +15,7 @@
 
 #include "revng/ADT/STLExtras.h"
 #include "revng/Support/Debug.h"
+#include "revng/Support/Undirected.h"
 
 /// GenericGraph is our implementation of a universal graph architecture.
 /// First and foremost, it's tailored for the revng codebase, but maybe you
@@ -996,6 +997,46 @@ public:
     return llvm::make_range(predecessors_begin(), predecessors_end());
   }
 
+public:
+  // TODO: The following methods could, in principle, be implemented using just
+  //       the `GraphTraits`, and become available for any implementer of
+  //       `GraphTraits` and not just for `GenericGraph`
+  auto undirected_children() {
+    using ResultIterator = UndirectedChildIterator<SuccessorIterator,
+                                                   PredecessorIterator,
+                                                   EdgeView>;
+    return llvm::make_range(ResultIterator::begin(successors(), predecessors()),
+                            ResultIterator::end(successors(), predecessors()));
+  }
+
+  auto undirected_children() const {
+    using ResultIterator = UndirectedChildIterator<ConstSuccessorIterator,
+                                                   ConstPredecessorIterator,
+                                                   EdgeView>;
+    return llvm::make_range(ResultIterator::begin(successors(), predecessors()),
+                            ResultIterator::end(successors(), predecessors()));
+  }
+
+  auto undirected_child_edges() {
+    using ResultIterator = UndirectedChildIterator<SuccessorEdgeIterator,
+                                                   PredecessorEdgeIterator,
+                                                   EdgeView>;
+    return llvm::make_range(ResultIterator::begin(successor_edges(),
+                                                  predecessor_edges()),
+                            ResultIterator::end(successor_edges(),
+                                                predecessor_edges()));
+  }
+
+  auto undirected_child_edges() const {
+    using ResultIterator = UndirectedChildIterator<ConstSuccessorEdgeIterator,
+                                                   ConstPredecessorEdgeIterator,
+                                                   EdgeView>;
+    return llvm::make_range(ResultIterator::begin(successor_edges(),
+                                                  predecessor_edges()),
+                            ResultIterator::end(successor_edges(),
+                                                predecessor_edges()));
+  }
+
 private:
   template<typename IteratorType>
   static auto findImpl(DerivedType const *N,
@@ -1577,6 +1618,41 @@ public:
 
   static T *edge_dest(EdgeRef Edge) { return &Edge.Neighbor; }
   static T *getEntryNode(llvm::Inverse<T *> N) { return N.Graph; };
+};
+
+/// Specializes GraphTraits<llvm::Undirected<MutableEdgeNode<...> *>>
+template<StrictSpecializationOfMutableEdgeNode T>
+struct GraphTraits<llvm::Undirected<T *>> {
+public:
+  using NodeRef = T *;
+  using EdgeRef = typename T::EdgeView;
+
+public:
+  static auto child_begin(NodeRef N) {
+    return N->undirected_children().begin();
+  }
+
+  static auto child_end(NodeRef N) { return N->undirected_children().end(); }
+
+  static auto child_edge_begin(NodeRef N) {
+    return N->undirected_child_edges().begin();
+  }
+
+  static auto child_edge_end(NodeRef N) {
+    return N->undirected_child_edges().end();
+  }
+
+  static NodeRef edge_dest(EdgeRef Edge) { return Edge.Neighbor; }
+
+  static NodeRef getEntryNode(llvm::Undirected<NodeRef> N) { return N.Graph; }
+
+public:
+  // We infer the `ChildIteratorType` directly from the `child_begin` method
+  using ChildIteratorType = decltype(child_begin(NodeRef{ nullptr }));
+
+  // We infer the `ChildEdgeIteratorType` directly from the `child_edge_begin`
+  // method
+  using ChildEdgeIteratorType = decltype(child_edge_begin(NodeRef{ nullptr }));
 };
 
 /// Specializes GraphTraits<GenericGraph<...> *>>
--- a/include/revng/ADT/GenericGraph.h
+++ b/include/revng/ADT/GenericGraph.h
@@ -1460,10 +1460,10 @@ public:
 
   using EdgeRef = typename T::Edge &;
   template<typename Ty, typename True, typename False>
-  using if_const = std::conditional_t<std::is_const_v<Ty>, True, False>;
-  using ChildEdgeIteratorType = if_const<T,
-                                         typename T::const_edge_iterator,
-                                         typename T::edge_iterator>;
+  using if_const_t = std::conditional_t<std::is_const_v<Ty>, True, False>;
+  using ChildEdgeIteratorType = if_const_t<T,
+                                           typename T::const_edge_iterator,
+                                           typename T::edge_iterator>;
 
 public:
   static ChildIteratorType child_begin(NodeRef N) {
@@ -1496,6 +1496,13 @@ public:
                                                typename T::const_child_iterator,
                                                typename T::child_iterator>;
 
+  using EdgeRef = typename T::Edge &;
+  template<typename Ty, typename True, typename False>
+  using if_const_t = std::conditional_t<std::is_const_v<Ty>, True, False>;
+  using ChildEdgeIteratorType = if_const_t<T,
+                                           typename T::const_edge_iterator,
+                                           typename T::edge_iterator>;
+
 public:
   static ChildIteratorType child_begin(NodeRef N) {
     return N->predecessors().begin();
@@ -1504,6 +1511,16 @@ public:
   static ChildIteratorType child_end(NodeRef N) {
     return N->predecessors().end();
   }
+
+  static ChildEdgeIteratorType child_edge_begin(NodeRef N) {
+    return N->predecessor_edges().begin();
+  }
+
+  static ChildEdgeIteratorType child_edge_end(NodeRef N) {
+    return N->predecessor_edges().end();
+  }
+
+  static NodeRef edge_dest(EdgeRef Edge) { return Edge.Neighbor; }
 
   static NodeRef getEntryNode(llvm::Inverse<NodeRef> N) { return N.Graph; };
 };
--- a/include/revng/Support/Undirected.h
+++ b/include/revng/Support/Undirected.h
@@ -1,0 +1,147 @@
+#pragma once
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "llvm/ADT/iterator.h"
+
+#include "revng/Support/Assert.h"
+
+namespace llvm {
+
+/// This class is used as a marker class to tell the graph iterator to treat the
+/// underlying graph as an undirected one
+template<class GraphType>
+struct Undirected {
+  const GraphType &Graph;
+
+  inline Undirected(const GraphType &G) : Graph(G) {}
+};
+
+} // namespace llvm
+
+/// Custom child iterator which iterates over the concatenation of the
+/// successors and predecessors of a node. This is used for the `Undirected`
+/// `GraphTraits` specialization. This is done taking inspiration from the
+/// internals of `llvm::concat` implementation, but overcomes one of its
+/// limitations. The limitation is caused by the fact that internally, the `*`
+/// operator of `llvm::concat` (and more specifically the `getHelper`), first
+/// dereferences the internal iterator that is being concatenated, and then
+/// takes and returns its pointer (`&*It`). If the iterator used internally,
+/// when dereferenced, returns a temporary object with a limited lifetime, which
+/// goes out of scope before its used by the user of `llvm::concat`, we have a
+/// problem. Instead here, we artificially extend the lifetime of the object
+/// pointed to by the internal iterator, wrapping it into the `Proxy` struct.
+/// When dereferenciation happens, no problems of lifetime happen.
+// TODO: Drop this once we adopt C++26
+//       (https://en.cppreference.com/w/cpp/ranges/concat_view) or if we adopt
+//       range-v3
+//       (https://ericniebler.github.io/range-v3/structranges_1_1views_1_1concat
+//        __fn.html).
+//       In such case, we can drop the below custom iterator, and use the
+//       `concat`  directly in the body of the `struct
+//       GraphTraits<llvm::Undirected<T *>>` implementation.
+template<typename SuccessorIterator,
+         typename PredecessorIterator,
+         typename ValueType>
+class UndirectedChildIterator : public llvm::iterator_facade_base<
+                                  UndirectedChildIterator<SuccessorIterator,
+                                                          PredecessorIterator,
+                                                          ValueType>,
+                                  std::forward_iterator_tag,
+                                  ValueType> {
+private:
+  SuccessorIterator SBegin;
+  SuccessorIterator SEnd;
+  PredecessorIterator PBegin;
+  PredecessorIterator PEnd;
+
+  explicit UndirectedChildIterator(SuccessorIterator SBegin,
+                                   SuccessorIterator SEnd,
+                                   PredecessorIterator PBegin,
+                                   PredecessorIterator PEnd) :
+    SBegin(SBegin), SEnd(SEnd), PBegin(PBegin), PEnd(PEnd) {}
+
+public:
+  static UndirectedChildIterator<SuccessorIterator,
+                                 PredecessorIterator,
+                                 ValueType>
+  begin(llvm::iterator_range<SuccessorIterator> Successors,
+        llvm::iterator_range<PredecessorIterator> Predecessors) {
+    return UndirectedChildIterator<SuccessorIterator,
+                                   PredecessorIterator,
+                                   ValueType>(std::begin(Successors),
+                                              std::end(Successors),
+                                              std::begin(Predecessors),
+                                              std::end(Predecessors));
+  }
+  static UndirectedChildIterator<SuccessorIterator,
+                                 PredecessorIterator,
+                                 ValueType>
+  end(llvm::iterator_range<SuccessorIterator> Successors,
+      llvm::iterator_range<PredecessorIterator> Predecessors) {
+    return UndirectedChildIterator<SuccessorIterator,
+                                   PredecessorIterator,
+                                   ValueType>(std::end(Successors),
+                                              std::end(Successors),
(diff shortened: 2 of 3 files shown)


include/revng/ADT/GenericGraph.h
```diff
@@ -1498,6 +1498,13 @@ struct GraphTraits<llvm::Inverse<T *>> {
                                                typename T::const_child_iterator,
                                                typename T::child_iterator>;
 
+  using EdgeRef = typename T::Edge &;
+  template<typename Ty, typename True, typename False>
+  using if_const = std::conditional_t<std::is_const_v<Ty>, True, False>;
```
aleclearmind, line 1503: Use `IfConst` or `if_const_t`

include/revng/ADT/GenericGraph.h
```diff
@@ -1527,6 +1527,37 @@ struct GraphTraits<llvm::Inverse<T *>> {
   static NodeRef getEntryNode(llvm::Inverse<NodeRef> N) { return N.Graph; };
 };
 
+// Undirected - This class is used as a marker class to tell the graph iterator
```
aleclearmind, line 1530: Drop "Undirected - "

include/revng/ADT/GenericGraph.h
```diff
@@ -1527,6 +1527,37 @@ struct GraphTraits<llvm::Inverse<T *>> {
...
+// to treat the underlying graph as an undirected one
+template<class GraphType>
+struct Undirected {
```
aleclearmind, line 1533: Add a `TODO` about moving this to `Undirected.h` and offering a default implementation of `GraphTraits<Undirected<T>>` for any `T` for which you have both `GraphTraits<T>` and `GraphTraits<Inverse<T>>` (what I was suggesting during the rev.ng hour).

include/revng/ADT/GenericGraph.h
```diff
@@ -1541,6 +1541,7 @@ template<StrictSpecializationOfBidirectionalNode T>
 struct GraphTraits<llvm::Undirected<T *>> {
 public:
   using NodeRef = T *;
+  using EdgeRef = typename T::Edge;
```
aleclearmind, line 1544: Squash `GenericGraph: edge traits for llvm::Undirected` into the previous commit, no need to add the feature in a second step.

aleclearmind (changes requested):

include/revng/ADT/GenericGraph.h
andrealinux1: I changed this to `if_const_t` [here](https://github.com/revng/revng/blob/d25a140596ffda169b22a0334654fc7bb27b028d/include/revng/ADT/GenericGraph.h#L1504). While I was at it, I also made the instances a couple of lines (not part of this changeset originally) above coherent with this [here](https://github.com/revng/revng/blob/d25a140596ffda169b22a0334654fc7bb27b028d/include/revng/ADT/GenericGraph.h#L1466)

include/revng/ADT/GenericGraph.h
andrealinux1: I added a TODO explaining the situation [here](https://github.com/revng/revng/blob/d25a140596ffda169b22a0334654fc7bb27b028d/include/revng/ADT/GenericGraph.h#L1530). Do you want me to implement this change right away?

include/revng/ADT/GenericGraph.h
andrealinux1: Right, I squashed the mentioned commit into the previous one in a single step

include/revng/ADT/GenericGraph.h
aleclearmind: No, thanks!

aleclearmind:
This could go in, but since the branch has the same name as another WIP PR for revng-c, I'm not merging.
Next time, if you have self-contained changes that you'd like to go in earlier, use a different branch name. But I guess there's no urgency.

pfez (approved):

include/revng/ADT/GenericGraph.h
```diff
@@ -996,6 +996,124 @@ class MutableEdgeNode
...
+  /// causing by the return `EdgeView` going of out scope.
+  template<typename SuccessorIterator, typename PredecessorIterator>
+  class UndirectedChildIterator
+    : public llvm::iterator_facade_base<
+        UndirectedChildIterator<SuccessorIterator, PredecessorIterator>,
+        std::bidirectional_iterator_tag,
+        EdgeView> {
+  private:
+    SuccessorIterator SBegin, SEnd;
+    PredecessorIterator PBegin, PEnd;
```
aleclearmind, line 1013: Split in 4 fields

include/revng/ADT/GenericGraph.h
```diff
@@ -996,6 +996,124 @@ class MutableEdgeNode
     return llvm::make_range(predecessors_begin(), predecessors_end());
   }
 
+private:
+  /// Custom child iterator which iterates over the concatenation of the
+  /// successors and predecessors of a node. This is used for the `Undirected`
+  /// `GraphTraits` specialization. This is done taking inspiration from the
+  /// internals of `llvm::concat` implementation, but overcoming the limitation
+  /// causing by the return `EdgeView` going of out scope.
```
aleclearmind, line 1004: 1. > overcoming the limitation causing by the return `EdgeView` going of out scope

   This does not provide me with enough information to understand why all of this is necessary.
   Can you add an example of how you'd use `llvm::concat` highlighting what the problem would be?

 2. How does `Proxy` solves the problem?
 3. Is there a way to have a `revng::concat` that does not suffers from `llvm::concat` limitation? If so, leave a TODO about this.

include/revng/ADT/GenericGraph.h
```diff
@@ -996,6 +996,124 @@ class MutableEdgeNode
...
+        return *SBegin;
+
+      if (PBegin != PEnd)
+        return *PBegin;
+
+      revng_abort("Attempted to dereference an end iterator!");
+    }
+
+    // Proxy object which extends the lifetime of the `EdgeView` object returned
+    // by the underlying `map_iterator`
```
aleclearmind, line 1079: I'm not sure what `map_iterator` we're talking about here. This piece of code seems to assume some specific usage.

include/revng/ADT/GenericGraph.h
```diff
@@ -996,6 +996,124 @@ class MutableEdgeNode
...
+    struct Proxy {
+      Proxy(EdgeView Value) : Temporary(std::move(Value)) {}
+      EdgeView *const operator->() { return &Temporary; }
+      EdgeView const *const operator->() const { return &Temporary; }
+
+    private:
+      EdgeView Temporary;
+    };
+    Proxy operator->() { return operator*(); }
```
aleclearmind, line 1088: Add blank line

include/revng/ADT/GenericGraph.h
```diff
@@ -996,6 +996,124 @@ class MutableEdgeNode
...
+  };
+
+public:
+  auto undirected_children() {
+    using ResultIterator = UndirectedChildIterator<SuccessorIterator,
+                                                   PredecessorIterator>;
+    return llvm::make_range(ResultIterator::begin(successors(), predecessors()),
+                            ResultIterator::end(successors(), predecessors()));
+  }
+  auto undirected_child_edges() {
```
aleclearmind, line 1108: Add blank line

include/revng/ADT/GenericGraph.h
```diff
@@ -996,6 +996,124 @@ class MutableEdgeNode
...
+    using ResultIterator = UndirectedChildIterator<SuccessorEdgeIterator,
+                                                   PredecessorEdgeIterator>;
+    return llvm::make_range(ResultIterator::begin(successor_edges(),
+                                                  predecessor_edges()),
+                            ResultIterator::end(successor_edges(),
+                                                predecessor_edges()));
+  }
```
aleclearmind, line 1115: I guess these methods could be in principle implemented just using `GraphTraits` and become available for any implementor of `GraphTraits` and not just `GenericGraph`, correct?

I assume this is reflected by the comment below:
```
// TODO: We should move the `Undirected` definition to a separate header, and
//       offer a default implementation for `GraphTraits<Undirected<T>>` for
//       any `T` for which you have a definition of both `GraphTraits<T>` and
//       `GraphTraits<Inverse<T>>`
```
If this is the case, this is fine.

include/revng/ADT/GenericGraph.h
```diff
@@ -1562,6 +1697,54 @@ struct GraphTraits<llvm::Inverse<T *>> {
...
+//       any `T` for which you have a definition of both `GraphTraits<T>` and
+//       `GraphTraits<Inverse<T>>`
+/// This class is used as a marker class to tell the graph iterator to treat the
+/// underlying graph as an undirected one
+template<class GraphType>
+struct Undirected {
+  const GraphType &Graph;
+
+  inline Undirected(const GraphType &G) : Graph(G) {}
+};
```
aleclearmind, line 1711: I don't want to raise the ambitions of this PR too much, it's fine not to do this tied to `GenericGraph`, but could we at least move `llvm::Undirected` and `MutableEdgeNode::UndirectedChildIterator` to `revng/Support/Undirected.h`.

`MutableEdgeNode` is already a quite big `class` and `UndirectedChildIterator` doesn't seem to be tied to it in any way.

aleclearmind:
Can you add a couple of tests to `tests/unit/GraphAlgorithms.cpp`?

include/revng/ADT/GenericGraph.h
andrealinux1: Yes, your observation is correct, even thought the comment below is indeed a bit outdated and refers to the previous implementation, so I'm updating this

include/revng/ADT/GenericGraph.h
andrealinux1: Done in https://github.com/revng/revng/blob/f14ffdc32258b98198148f45162302e6cbecc7c1/include/revng/Support/Undirected.h#L51

include/revng/ADT/GenericGraph.h
andrealinux1: I agree the explanation was not well laid out for one that did not see the problem in its entirety. I rewrite and centralized the explanation here https://github.com/revng/revng/blob/f14ffdc32258b98198148f45162302e6cbecc7c1/include/revng/Support/Undirected.h#L28

include/revng/ADT/GenericGraph.h
andrealinux1: Yeah, this was more confusing than anything, I centralized the explanation for having this at the beginning of the class, as stated in the comment above https://github.com/revng/revng/pull/388#discussion_r1705687147

include/revng/ADT/GenericGraph.h
andrealinux1: As you suggested, I extracted the `llvm::Undirected` and `MutableEdgeNode::UndirectedChildIterator` in the new `Undirected.h` header. The only downside of this is the fact that the `GraphTraits` specialization for `Undirected` is still in `GenericGraph`, so things are bit more scattered in different points, but whatever

include/revng/ADT/GenericGraph.h
andrealinux1: I added a specific, and more clear comment highlighting your observation with a TODO https://github.com/revng/revng/blob/f14ffdc32258b98198148f45162302e6cbecc7c1/include/revng/ADT/GenericGraph.h#L1001

andrealinux1:
> Can you add a couple of tests to `tests/unit/GraphAlgorithms.cpp`?

I added some unit tests https://github.com/revng/revng/blob/f14ffdc32258b98198148f45162302e6cbecc7c1/tests/unit/GenericGraph.cpp#L621

I asked Pietro to take a double look on this, since he knows how this works and may spot more easily problems on these tests.

include/revng/Support/Undirected.h
```diff
@@ -0,0 +1,144 @@
...
+
+} // namespace llvm
+
+/// Custom child iterator which iterates over the concatenation of the
+/// successors and predecessors of a node. This is used for the `Undirected`
+/// `GraphTraits` specialization. This is done taking inspiration from the
+/// internals of `llvm::concat` implementation, but overcomes one of its
+/// limitations. The limitation is caused by the fact that internally, the `*`
+/// operator of `llvm::concat` (and more specifically the `getHelper`), first
+/// deferences the internal iterator that is being concatenated, and then takes
```
aleclearmind, line 34: > deferences

typo

include/revng/Support/Undirected.h
```diff
@@ -0,0 +1,144 @@
...
+/// and returns its pointer (`&*It`). If the iterator used internally, when
+/// deferenced, returns a temporary object with a limited lifetime, which goes
+/// out of scope before its used by the user of `llvm::concat`, we have a
+/// problem.
```
aleclearmind, line 38: OK, now the problem is clear but can you add a line like "instead here we...".

include/revng/Support/Undirected.h
```diff
@@ -0,0 +1,144 @@
...
+#include "llvm/ADT/iterator.h"
+
+#include "revng/Support/Assert.h"
+
+namespace llvm {
+
+// TODO: Without the `llvm::concat` implementation limitation described below,
+//       we could offer a default implementation for
+//       `GraphTraits<Undirected<T>>` for any `T` for which you have a
+//       definition of both `GraphTraits<T>` and `GraphTraits<Inverse<T>>`
```
aleclearmind, line 16: I think we could still implement `GraphTraits<Undirected<T>>` using `UndirectedChildIterator`. The comment seems to imply this is not possible.

tests/unit/GenericGraph.cpp
```diff
@@ -617,3 +617,234 @@ BOOST_AUTO_TEST_CASE(MutableEdgeNodeRemovalTest) {
   revng_check(C->successorCount() == 0);
   revng_check(C->predecessorCount() == 1);
 }
+
+// Tests for the UndirectedChildIterator implementation
+BOOST_AUTO_TEST_CASE(MutableEdgeNodeUndirectedIteratorTest) {
```
aleclearmind, line 622: Nice!

include/revng/Support/Undirected.h
```diff
@@ -0,0 +1,144 @@
...
+/// and returns its pointer (`&*It`). If the iterator used internally, when
+/// deferenced, returns a temporary object with a limited lifetime, which goes
+/// out of scope before its used by the user of `llvm::concat`, we have a
+/// problem.
+// TODO: An alternative to implementing our custom iterator, would be to provide
+//       our own implementation of `revng::concat`, which does not have the
+//       limitation described above.
+template<typename SuccessorIterator,
+         typename PredecessorIterator,
+         typename ProxyObjectType>
```
aleclearmind, line 44: Can we rename this template argument? Iterators have the concept of "value type", i.e., what you get when you `*`. Maybe we can rename this to `ValueType`?

include/revng/Support/Undirected.h
andrealinux1: Right, fixed https://github.com/revng/revng/blob/542a271c138b3255d8c126850db12d2ae3e761d5/include/revng/Support/Undirected.h#L30

include/revng/Support/Undirected.h
andrealinux1: Oh right, added a comment https://github.com/revng/revng/blob/542a271c138b3255d8c126850db12d2ae3e761d5/include/revng/Support/Undirected.h#L34

include/revng/Support/Undirected.h
andrealinux1: Yeah, sound a lot better https://github.com/revng/revng/blob/542a271c138b3255d8c126850db12d2ae3e761d5/include/revng/Support/Undirected.h#L42

include/revng/Support/Undirected.h
andrealinux1: Yeah, that comment was a bit confusing from the start, but I left there because now we have two orthogonals thing to improve the codebase, but that can be combined in improving the `Undirected` trait specialization.

The two things are:
1) Use the `GraphTraits` to provide the `UndirectedChildIterator` not only for `GenericGraph`, but using the traits. But this is explained in this TODO indeed https://github.com/revng/revng/blob/542a271c138b3255d8c126850db12d2ae3e761d5/include/revng/ADT/GenericGraph.h#L1001.
2) Implement our own `revng::concat` in place of `llvm::concat`, and this is commented with a TODO here https://github.com/revng/revng/blob/542a271c138b3255d8c126850db12d2ae3e761d5/include/revng/Support/Undirected.h#.

If 2) is done, we can drop the `UndirectedChildIterator` and re-implement the `GraphTraits<Undirected<T>>` using the `revng::concat`. Hence the comment above. But I also think that it was more confusing than anything else, so I dropped it, and the two `TODO`s combined should lead us to conclude what I said above.

tests/unit/GenericGraph.cpp
```diff
@@ -617,3 +617,234 @@ BOOST_AUTO_TEST_CASE(MutableEdgeNodeRemovalTest) {
...
+  DEdgeView = *DUndirectedIt;
+  revng_check(DEdgeView.Neighbor == B);
+  revng_check(*DEdgeView.Label == 3);
+  DUndirectedIt++;
+  DEdgeView = *DUndirectedIt;
+  revng_check(DEdgeView.Neighbor == C);
+  revng_check(*DEdgeView.Label == 4);
+  DUndirectedIt++;
+  revng_check(DUndirectedIt == DUndirectedEdges.end());
+}
```
pfez, line 804: I think we're missing tests for weirder graphs.
- multiple edges between a given pair of node A->B
- loops and self-loops in the underlying directed graph

tests/unit/GenericGraph.cpp
```diff
@@ -617,3 +617,234 @@ BOOST_AUTO_TEST_CASE(MutableEdgeNodeRemovalTest) {
...
+  auto AUndirectedIt = AUndirectedEdges.begin();
+  AEdgeView = *AUndirectedIt;
+  revng_check(AEdgeView.Neighbor == B);
+  revng_check(*AEdgeView.Label == 1);
+  AUndirectedIt++;
+  AEdgeView = *AUndirectedIt;
+  revng_check(AEdgeView.Neighbor == C);
+  revng_check(*AEdgeView.Label == 2);
+  AUndirectedIt++;
+  revng_check(AUndirectedIt == AUndirectedEdges.end());
```
pfez, line 683: You do the same tests for all the nodes A B C and D.

Please refactor this in a helper function that takes a single node as argument, and the constants you want to test (e.g. 1 and 2 for this case).

tests/unit/GenericGraph.cpp
pfez: And I would like to test the visits for them too.

pfez (changes requested):

include/revng/Support/Undirected.h
```diff
@@ -0,0 +1,142 @@
...
+/// dereferences the internal iterator that is being concatenated, and then
+/// takes and returns its pointer (`&*It`). If the iterator used internally,
+/// when dereferenced, returns a temporary object with a limited lifetime, which
+/// goes out of scope before its used by the user of `llvm::concat`, we have a
+/// problem. Instead here, we artificially extend the lifetime of the object
+/// pointed to by the internal iterator, wrapping it into the `Proxy` struct.
+/// When dereferenciation happens, no problems of lifetime happen.
+// TODO: An alternative to implementing our custom iterator, would be to provide
+//       our own implementation of `revng::concat`, which does not have the
+//       limitation described above.
```
aleclearmind, line 39: After today's talk, please replace this comment with "drop this once we adopt C++34 or we could adopt rangesv3" and link the specific relevant file.

andrealinux1: I added a comment linking the page on `cppreference` for C++26 and the documentation for range-v3

https://github.com/revng/revng/blob/bf979cc12411a1e94654b1f77e6cfb0054b6daed/include/revng/Support/Undirected.h#L37

tests/unit/GenericGraph.cpp
andrealinux1: I added a test with a graph which has multiple edges between the same node pair https://github.com/revng/revng/blob/ff2b7e90cbf0ff106a7853aa468bfff9920cc855/tests/unit/GenericGraph.cpp#L728

and a graph with a main loop and a smaller self loop https://github.com/revng/revng/blob/ff2b7e90cbf0ff106a7853aa468bfff9920cc855/tests/unit/GenericGraph.cpp#L749

tests/unit/GenericGraph.cpp
andrealinux1: I factored all this code in an helper https://github.com/revng/revng/blob/ff2b7e90cbf0ff106a7853aa468bfff9920cc855/tests/unit/GenericGraph.cpp#L621

and I also factored the code to check the DFS order visits https://github.com/revng/revng/blob/ff2b7e90cbf0ff106a7853aa468bfff9920cc855/tests/unit/GenericGraph.cpp#L684 https://github.com/revng/revng/blob/ff2b7e90cbf0ff106a7853aa468bfff9920cc855/tests/unit/GenericGraph.cpp#L695

include/revng/Support/Undirected.h
pfez: C++34 :heart: :rofl:

tests/unit/GenericGraph.cpp
pfez: :+1:

tests/unit/GenericGraph.cpp
pfez: :+1:

PR #390, opened 2024-06-06
revng: allow passing additional args to wrappers

mrjackv:


aleclearmind:
Can we have `--gdb="--extra-gdb-arg"` instead of a distinct option?

mrjackv:
> Can we have `--gdb="--extra-gdb-arg"` instead of a distinct option?

I would have loved that, however we cannot do that since:
* we have positional arguments
* the python parser is not non-deterministic

Easy example:
```
revng --gdb artifact --analyze ...
```

Here the `artifact` part would be mis-identified as an argument of `--gdb` rather than the positional argument for the subcommand, so it would fail with `revng: error: argument command_name0: invalid choice: '--analyze'`

aleclearmind:
I'm not very persuaded. Maybe we should just have a `--wrapper` and let people do
```
revng --wrapper='gdb --mycustom-arg --args' artifact --analyze decompile-to-single-file calc
```

This PR introduces 7 new arguments just so the user can do:

```
revng --rr --rr-args='--lol' ...
```

Instead of

```
revng --wrapper='rr --lol' ...
```

Which is even shorter.

I think that we should either stick with the presets or we let the user take over entirely.

mrjackv:
> Maybe we should just have a `--wrapper`

That sounds neat and flexible, I'd go with that. ACK?

aleclearmind:
Sure, go ahead. Just to clarify: my proposal is about *adding* `--wrapper` and keep `--gdb` but *not adding* `--gdb-args`.

mrjackv:
Done

aleclearmind:
Merged in ba0f374. Thanks.

PR #393, opened 2024-07-09
New Mass Testing

mrjackv:


mrjackv:
> Why?

Basically, to allow a reliable way for the "main" process to be terminated by a signal of our choice. When testing the `revng-artifact` executable is wrapped by 3-4 other executables (`time`, `revng` cli, there might be `--heaptrack` & friends). This makes it difficult for the `test-harness` to kill the right process and allow a stacktrace to be produced without any "weird" heuristics.

PR #396, opened 2024-07-15
New mass testing

mrjackv:
Siter PRs:
* [orchestra](https://github.com/revng/orchestra/pull/155)
* [revng-qa](https://github.com/revng/revng-qa/pull/29)
* [revng-c](https://github.com/revng/revng-c/pull/32)
* [llvm-project](https://github.com/revng/llvm-project/pull/34)


CMakeLists.txt
```diff
@@ -143,12 +143,21 @@ install(FILES "${COMPONENT_HASH_PATH}"
...
 #
-add_custom_target(copy_share ALL COMMAND cp -Tar "${CMAKE_SOURCE_DIR}/share/"
-                                         "${CMAKE_BINARY_DIR}/share/")
+add_custom_target(
+  copy_share ALL
+  COMMAND mkdir -p "${CMAKE_BINARY_DIR}/share/"
+  COMMAND cp -Tar "${CMAKE_SOURCE_DIR}/share/doc"
+          "${CMAKE_BINARY_DIR}/share/doc"
+  COMMAND cp -Tar "${CMAKE_SOURCE_DIR}/share/revng"
+          "${CMAKE_BINARY_DIR}/share/revng")
```
aleclearmind, line 152: Why? The whole `share/` directory is designed to be copied as is.
Please reorganize things so that things stay as they are, or come and we can discuss.

libexec/mass-testing/readobj-json
```diff
@@ -0,0 +1,52 @@
...
+    proc = run(
+        ["llvm-readobj", "--elf-output-style=JSON", "--sections", filepath],
+        stdout=PIPE,
+        check=True,
+        text=True,
+    )
+    data = yaml.safe_load(proc.stdout)
+
+    result = {}
+    if len(data) == 1:
```
aleclearmind, line 25: 1. There has to be a better way
2. Are the ELF and PE outputs similar?

libexec/mass-testing/readobj-json
aleclearmind, line 1: 1. I'd rename this to `dump-sections`
2. Why do we have `libexec/mass-testing`? There needs to be a quite strong reason not to put this under `libexec/revng`.

share/mass-testing-report/.gitignore
aleclearmind, line 1: Mmmh, no. We don't encourage creating any non-tracked file in the source directory.

libexec/mass-testing/revng-test
```diff
@@ -0,0 +1,13 @@
...
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+set -euo pipefail
+SCRIPT_DIR=$(realpath "$(dirname "${BASH_SOURCE[0]}")")
+
+SECTIONS_JSON="$TEST_OUTPUT_DIR/sections.json"
+
+"$SCRIPT_DIR/readobj-json" "$TEST_INPUT" > "$SECTIONS_JSON"
+echo "text_size: $("$SCRIPT_DIR/text-size" "$SECTIONS_JSON")" >> "$TEST_OUTPUT_DIR/meta.yml"
```
aleclearmind, line 11: We usually prefer kebab-case

libexec/mass-testing/symbolizer_wrapper
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python3
```
aleclearmind, line 1: 1. Rename all the scripts to use kebab-case, unless they are modules that need to be imported. Then we can use snake case.
2. Add a comment on the top of each script describing the purpose of the script.

libexec/mass-testing/symbolizer_wrapper
aleclearmind, line 1: Rename all the scripts to use kebab-case, unless they are modules that need to be imported. Then we can use snake case.

libexec/mass-testing/test-harness
```diff
@@ -0,0 +1,235 @@
...
+
+    try:
+        process.wait(args.timeout)
+        timed_out = False
+    except TimeoutExpired:
+        handle_timeout(process)
+        timed_out = True
+
+    # Wait for things to settle and then take care of children
+    sleep(10)
```
aleclearmind, line 215: Do we really really really need this?
10 seconds is a long time.

libexec/mass-testing/test-harness
```diff
@@ -0,0 +1,235 @@
...
+            result.append("revng")
+        else:
+            result.append(part)
+    return result
+
+
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--input", help="Input binary")
+    parser.add_argument("--output", required=True, help="Output directory")
```
aleclearmind, line 149: As a policy, we use required arguments as positional arguments.

libexec/mass-testing/test-harness
```diff
@@ -0,0 +1,235 @@
...
+        pass
+
+
+def patch_cmd(original_cmd: List[str], input_: str | None) -> List[str]:
+    result = []
+    for part in original_cmd:
+        if part == input_:
+            result.append("$INPUT")
+        elif Path(part).resolve() == (SCRIPT_DIR / "revng-test"):
+            result.append("revng")
```
aleclearmind, line 140: This script must not assume anything `revng` specific. Please double check.

libexec/mass-testing/test-harness
```diff
@@ -0,0 +1,235 @@
...
+    except psutil.ChildProcessError:
+        # Process died before we could `wait`
+        pass
+
+
+def patch_cmd(original_cmd: List[str], input_: str | None) -> List[str]:
+    result = []
+    for part in original_cmd:
+        if part == input_:
+            result.append("$INPUT")
```
aleclearmind, line 138: This is brittle and upon failure fails silently because `input_` is optional.
We can:
1. Make `--input` mandatory (making it positional) and have users do
   ```
   test-harness my-input-file -- my-command --lol %INPUT%
   ```
   Alternatively, we can drop `%INPUT%` and assume that the command you give accepts the input file as the last parameter.
2. Push the symlinking logic in `revng-test`, but I think that's not viable since the report generation code (which can only assume to know about this script basically) assumes the symlink is there.

Let me know if you have better ideas.

libexec/mass-testing/test-harness
```diff
@@ -0,0 +1,235 @@
...
+    '"avg_shared_text_size": %X, "avg_unshared_data_size": %D, "avg_stack_size": %p, '
+    '"avg_total_size": %K, "max_resident_set_size": %M, "avg_resident_set_size": %t, '
+    '"major_io_page_faults": %F, "minor_reclaim_page_faults": %R, '
+    '"voluntary_context_switches": %w, "involuntary_context_switches": %c, "swaps": %W, '
+    '"fs_inputs": %I, "fs_outputs": %O, "socket_msgs_sent": %s, "socket_msgs_recvd": %r, '
+    '"signals_delivered": %k, "page_size": %Z, "exit_code": %x}'
+)
+
+
+def set_child_subreaper():
```
aleclearmind, line 37: Docs

libexec/mass-testing/test-harness
```diff
@@ -0,0 +1,235 @@
...
+                kill(process.pid, signal.SIGABRT)
+                process_killed = True
+                # Wait a bit between kills to allow for things to settle
+                sleep(2)
+            elif ppid == os.getpid():
+                os.waitpid(process.pid, 0)
+    return process_killed
+
+
+def pid_has_signal(pid: int, signalno: int) -> bool:
```
aleclearmind, line 102: Docs

libexec/mass-testing/test-harness
```diff
@@ -0,0 +1,235 @@
...
+    libc = cdll.LoadLibrary(libc_name)
+    libc.prctl(PR_SET_CHILD_SUBREAPER := 36, 1)  # noqa: F841,N806
+
+
+def kill(pid: int, signal: int):
+    with suppress(ProcessLookupError, PermissionError):
+        os.kill(pid, signal)
+
+
+def walk_process_tree(pid: int, include_root=True) -> Generator[psutil.Process, None, None]:
```
aleclearmind, line 49: Docs: does this yield root or leaves first?

libexec/mass-testing/text-size
aleclearmind, line 1: What kind of input does this script expect?

python/CMakeLists.txt
aleclearmind, line 1: I think that, as long as this package stays in this repo, despite its ambitions to be revng agnostic, should stay under the revng namespace.
The idea is that we can eventually easily move it out, but I would take another whole root namespace just for this.

What do you think?

python/revng/internal/cli/_commands/mass_testing/configure.py
```diff
@@ -0,0 +1,57 @@
...
+
+class MassTestingConfigureCommand(Command):
+    def __init__(self):
+        super().__init__(
+            ("mass-testing", "configure"), "Collect binary files and run test-configure"
+        )
+
+    def register_arguments(self, parser: argparse.ArgumentParser):
+        parser.description = "Collect binary files and run test-configure"
+        parser.add_argument("input", help="Input directory")
```
aleclearmind, line 25: This script collects all the file, no filtering takes place, correct?

python/revng/internal/cli/_commands/mass_testing/generate_report/stacktrace_transform.py
```diff
@@ -0,0 +1,99 @@
...
+    path: str
+    symbol: str
+
+    def to_string(self) -> str:
+        return f"{self.symbol} at {self.path}".replace(";", "")
+
+
+def normalize_path(path: str):
+    path = os.path.normpath(path)
+    # TODO: un-orchestra this
```
aleclearmind, line 30: Use `WIP` for comments that should not be merged

python/revng/internal/cli/_commands/mass_testing/generate_report/stacktrace_transform.py
```diff
@@ -0,0 +1,99 @@
...
+                lines += get_perf_line(f)
+            else:
+                lines += "no stack trace 1\n"
+
+    opts = ["--color=chain", "--width=1920", "--minwidth=0", "--countname=run(s)"]
+    if inverted:
+        opts.append("--inverted")
+        opts.append("--title=Stack traces - Top down - Thread entry point is at the bottom")
+    else:
+        opts.append("--title=Stack traces - Bottom up - Crash location is at the top")
```
aleclearmind, line 86: I'd consider having these arguments passed in from the user and forward them to `flamegraph.pl`.

For instance, this script is used with non-crashes too (while we hardcode "crash location").

python/revng/internal/cli/_commands/mass_testing/run.py
```diff
@@ -0,0 +1,34 @@
...
+    def run(self, options: Options):
+        args = options.parsed_args
+
+        path = os.environ["PATH"]
+        libexec_path = get_root() / "libexec/mass-testing"
+
+        new_env = os.environ.copy()
+        new_env["PATH"] = f"{libexec_path!s}:{path}"
+
+        # TODO: drop
```
aleclearmind, line 31: Use `WIP`

python/revng/internal/cli/_commands/mass_testing/generate_report/test_directory.py
```diff
@@ -0,0 +1,84 @@
...
+from shutil import copy2
+from typing import Generator
+
+import yaml
+
+REQUIRED_FILES = ("input", "output.log", "test-harness.json")
+COPY_EXCLUDED_FILES = ("meta.yml",)
+
+
+class TestDirectory:
```
aleclearmind, line 20: Docs on the purpose of this class

share/mass-testing-report/CMakeLists.txt
aleclearmind, line 1: Can we move this to `python/revng/internal/cli/_commands/mass_testing/`?
`share` should only contain files being copied verbatim.

share/revng/mass-testing-revng-meta.yml
```diff
@@ -0,0 +1,13 @@
...
+
+extra_columns:
+  - name: text_size
+    label: .text size
+    type: int
+
+downloads:
+  - name: trace.json.gz
+    label: Trace
+    crash_related: true
```
aleclearmind, line 13: Why do we need `crash_related`?

aleclearmind (changes requested):

share/revng/mass-testing-revng-meta.yml
mrjackv: So that the download button only appears in the `crashes` and `timeouts` page and not in the raw data table. Unless you want to have all 3 have the same set of buttons

share/mass-testing-report/CMakeLists.txt
mrjackv: Eh ok, but it's kind of surprising to have some npm project stuffed inside the python subdirectory.
I've looked at the folder structure but there doesn't seem to be anything "that makes sense" where to put `mass-testing-report` that doesn't look out of place (other than `share`, that's why I put it there)

libexec/mass-testing/readobj-json
mrjackv: 1. Eh, no
2. Not quite, the `ELF` output is proper JSON, but the PE/Mach-O gets "broken" by `--elf-output-style=JSON` enough that it is valid YAML

libexec/mass-testing/readobj-json
mrjackv: 1. OK
2. I'd avoid polluting the `libexec/revng` directory with a whole bunch of executables which are mass-testing related

share/mass-testing-report/.gitignore
mrjackv: Sure, however without the `node_modules` directory I don't get any autocompletion while editing `main.ts`, which is very annoying.
I can add `node_modules` in my global gitignore, but the next person that has to edit `main.ts` will also be annoyed.

libexec/mass-testing/symbolizer_wrapper
mrjackv: Ok for both

libexec/mass-testing/test-harness
mrjackv: Eh, in general the sleep there is to account for `time` & the other wrappers to gracefully terminate. I can cut it down to say 5 seconds, but the reality is that it's a random number that "looks good".

libexec/mass-testing/test-harness
mrjackv: Ok, but then this makes creating a reproducible `commands.json` impossible. Copying `argv` 1:1 is not pratical.

libexec/mass-testing/test-harness
mrjackv: 1. I'd prefer that, it solves a whole bunch of problems. I'd use `%INPUT%` for added flexibility
2. Eh yes, I'd keep it in `test-harness`

libexec/mass-testing/text-size
mrjackv: The file produced by `readobj-json`

python/revng/internal/cli/_commands/mass_testing/configure.py
mrjackv: Yes

python/revng/internal/cli/_commands/mass_testing/generate_report/stacktrace_transform.py
mrjackv: > I'd consider having these arguments passed in from the user

Do you mean `generate_report/__main__.py` or the actual user? I'd find it kind of cumbersome for `mass-testing generate-report` to provide all the command-line options.

Also, in both situations it forces the user to know what `flamegraph.pl` options are
The only thing that changes is the title, I'd rather add it as a parameter

libexec/mass-testing/readobj-json
aleclearmind: 2. > PE/Mach-O gets "broken" by --elf-output-style=JSON enough that it is valid YAML

   What do you mean?

libexec/mass-testing/readobj-json
aleclearmind: 2. You can make a subdirectory and extend the lookup mechanism. But I don't want mass testing to introduce new namespaces (module names, "top level" directories).

share/mass-testing-report/.gitignore
aleclearmind: For now, let's go with your global gitignore (you can also have local ones, `.git/info/exclude`).

libexec/mass-testing/test-harness
aleclearmind: Can't at least check every second if the root process is terminated *up to* 10 seconds?

libexec/mass-testing/test-harness
aleclearmind: Elaborate, like "if we don't do this, this specifically will happen".

libexec/mass-testing/test-harness
aleclearmind: Let's go with #1.

libexec/mass-testing/text-size
aleclearmind: Mmh, maybe we should either add this as a feature of `readobj-json` or have this script invoke `readobj-json`.

python/revng/internal/cli/_commands/mass_testing/generate_report/stacktrace_transform.py
aleclearmind: OK

share/mass-testing-report/CMakeLists.txt
aleclearmind: `CMakeLists.txt` can stay in `python/revng/internal/cli/_commands/mass_testing` but the `npm` stuff (which is in the build directory) can be wherever you prefer. The build directory is already quite dirty, I'm not too sad.

share/revng/mass-testing-revng-meta.yml
aleclearmind: Didn't we say no button but just a link to the "Index of /lol" page?

libexec/mass-testing/test-harness
mrjackv: Basically, in order to have a proper `reproducer.tar`, it's needed to store somewhere something like `argv`. But if the argv was copied 1:1 you would have:
```
test-harness --output <some directory on the build machine> --input <some path on the build machine> ... -- revng-test <options> %INPUT%
```

In an ideal world, the reproducer script would call:
```
revng <options> %INPUT%
```

So that none of the `revng-test` logic get in the way of the reproducer.

libexec/mass-testing/test-harness
mrjackv: That sounds sensible, I'll do it that way.

share/revng/mass-testing-revng-meta.yml
mrjackv: We did, but considering that we later devised the entire `meta.yml` structure it seems a waste not to leverage it to also have direct links to files of note

aleclearmind: Having lots of buttons is not the nicest, I'm more than happy with index of, but as you prefer.

libexec/mass-testing/revng-test
mrjackv: Do you mean for the yaml keys? they tend to be annoying to handle in both python and typescript, I'd avoid them.

aleclearmind: Why? Don't we use `obj["text-size"]`?
No biggie anyway, in this case.

libexec/mass-testing/readobj-json
mrjackv: Done both

share/mass-testing-report/.gitignore
mrjackv: Done

libexec/mass-testing/text-size
mrjackv: Merged into `dump-sections`

python/revng/internal/cli/_commands/mass_testing/run.py
mrjackv: Done

share/mass-testing-report/CMakeLists.txt
mrjackv: Moved

libexec/mass-testing/test-harness
mrjackv: Changed to positional

libexec/mass-testing/test-harness
mrjackv: Fixed by:
* Using `%INPUT%` so that it can be easily replaced
* Having `hooks.pre` and `hooks.post` in `meta.yml` to mangle `$PATH` so that `revng` is a symlink to `revng-test`

libexec/mass-testing/test-harness
mrjackv: Implemented

libexec/mass-testing/test-harness
mrjackv: Added

libexec/mass-testing/test-harness
mrjackv: Added

python/revng/internal/cli/_commands/mass_testing/generate_report/stacktrace_transform.py
mrjackv: Done

python/revng/internal/cli/_commands/mass_testing/generate_report/test_directory.py
mrjackv: Added

libexec/mass-testing/test-harness
mrjackv: Added

aleclearmind (changes requested):

python/revng/internal/cli/support.py
```diff
@@ -88,7 +88,7 @@ def popen(command, options: Options, environment: OptionalEnv = None, **kwargs)
     return Popen(command, env=environment, **kwargs)
 
 
-def try_run(command, options: Options, environment: OptionalEnv = None) -> int:
+def try_run(command, options: Options, environment: OptionalEnv = None, **kwargs) -> int:
```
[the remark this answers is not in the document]
mrjackv: >  What's the specific need for this?

Specifically I needed to pass `stdout=DEVNULL, stderr=DEVNULL` to an inkscape invocation

> We've been actively trying to keep the surface to spawn new processes small and uniform otherwise things becomes a mass real quick.

`revng.internal.cli.support.popen` has kwargs, I don't see a reason why its wrappers (run, try_run) should not.

libexec/revng/mass-testing/revng-dump-sections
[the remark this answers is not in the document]
mrjackv: The `revng-` prefix is needed also in the subdirectories to be consistent with the rest of the executables in `libexec`. So you have `libexec/revng/revng-artifact` and `libexec/revng/mass-testing/revng-dump-sections`.

The prefix is stripped, so the invocation is `revng mass-testing dump-sections`

libexec/revng/mass-testing/revng-revng-test
```diff
@@ -0,0 +1,31 @@
...
+# with test-harness. In particular it does the following:
+# * Saves `sections.json` and adds `text_size` to meta.yml
+# * Saves the trace output as `trace.json.gz`
+# * Suppresses output from the `revng` command
+# * Detects if it's being used as a wrapper for `revng` and calls the real one
+#   properly
+
+SECTIONS_JSON="$TEST_OUTPUT_DIR/sections.json"
+
+dump-sections "$TEST_INPUT" > "$SECTIONS_JSON"
```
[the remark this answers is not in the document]
mrjackv: > How is dump-sections in PATH? o_O

`revng mass-testing run` adds the `mass-testing` executables to `PATH` without the prefix [Link](https://github.com/revng/revng/pull/396/commits/a533af74d1fc19b5026f88f1d3f57c3f73b84ef7#diff-9b1da3fad311f65152340658285d7a8adf1a9f03e79fa5949720db0e87c364f7R44-R55)

libexec/revng/mass-testing/revng-test-harness
```diff
@@ -0,0 +1,270 @@
...
+    try:
+        process.wait(args.timeout)
+        timed_out = False
+    except TimeoutExpired:
+        handle_timeout(process)
+        timed_out = True
+
+    # Wait for things to settle and then take care of children
+    with suppress(TimeoutExpired):
+        process.wait(10)
```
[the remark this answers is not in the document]
mrjackv: Yes

python/revng/internal/cli/_commands/mass_testing/configure.py
```diff
@@ -0,0 +1,70 @@
...
+
+        file_list = []
+        input_path = Path(args.input)
+        for dirpath, _, filenames in os.walk(args.input):
+            dirpath_path = Path(dirpath)
+            for filename in filenames:
+                file_list.append(str((dirpath_path / filename).relative_to(input_path)))
+
+        # Shuffle the input files, this avoids having a limited variety of
+        # test runs for a partial run
```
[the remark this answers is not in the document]
mrjackv: Basically, modulo filesystem ordering (which is usually sorted), the order of execution of the individual "tests" will be lexicographic, e.g.

```
bar/file_0001
bar/file_0002
...
bar/file_1000
foo/file_0001
foo/file_0002
...
foo/file_1000
```

If we don't `shuffle` that means that a partial run (e.g. where we ctrl-c mid-way) might have (in this example) an over-representation of `bar` and few or no members of `foo`.

python/revng/internal/support/collect.py
```diff
@@ -31,6 +33,35 @@ def collect_files_gen(
                 yield str(library.resolve())
 
 
+def collect_files_recursive(
```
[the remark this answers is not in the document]
mrjackv: Done

libexec/revng/mass-testing/revng-test-harness
```diff
@@ -0,0 +1,270 @@
...
+    # Setup as child subreaper, this avoids processes "escaping"
+    # the test harness
+    set_child_subreaper()
+
+    final_cmd = [*time_cmd, *[p if p != "%INPUT%" else args.input for p in cmd]]
+    global_meta_path = Path.cwd() / "meta.yml"
+    if global_meta_path.exists():
+        with open(global_meta_path) as f:
+            global_meta: dict = yaml.safe_load(f)
+        if "hooks" in global_meta:
```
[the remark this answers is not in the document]
mrjackv: Done

typescript/mass-testing-report/src/main.ts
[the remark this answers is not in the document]
mrjackv: Done

libexec/revng/mass-testing/revng-dump-sections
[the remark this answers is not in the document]
aleclearmind: This made sense as long as there were no subdirectories.
I'd suggest to drop the `revng-` prefix from everywhere and enumerate by the simple fact that they are `+x`.

libexec/revng/mass-testing/revng-revng-test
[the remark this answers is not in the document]
aleclearmind: Broken link

python/revng/internal/cli/support.py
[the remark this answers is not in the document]
mrjackv: I was under the impression that all subprocess invocations had to go through `support.run` & friends. If that's not the case then I'm 100% ok with using `subprocess.run`

libexec/revng/mass-testing/revng-revng-test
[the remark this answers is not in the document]
mrjackv: Here's the new one: https://github.com/revng/revng/pull/396/commits/b6c636b81789ae016835d3d9ea5784dffe996972#diff-9b1da3fad311f65152340658285d7a8adf1a9f03e79fa5949720db0e87c364f7R44-R52

python/revng/internal/cli/_commands/mass_testing/configure.py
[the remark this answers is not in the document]
aleclearmind: Mmmh, what's the use case to Ctrl+C midway?

This reduces determinism, if there's a bug, it will reproduce at different latencies, which can be annoying.
I'm OK with the idea of shuffling, but let's make it deterministic. Get some shuffling routine which enables you do set the seed, so it's "random" but predictably so.

libexec/revng/mass-testing/revng-dump-sections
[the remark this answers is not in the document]
mrjackv: >  I'd suggest to drop the revng- prefix from everywhere and enumerate by the simple fact that they are +x.

Do you mean also in `libexec/`? (In the sense, renaming `revng-artifact` into `artifact`?

python/revng/internal/cli/_commands/mass_testing/configure.py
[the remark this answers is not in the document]
mrjackv: > Mmmh, what's the use case to Ctrl+C midway?

A couple that come to mind:
* Mass testing takes too long, we get impatient and want to see results
* The test container/machine crashes, we want to salvage the work that has been done

> This reduces determinism, if there's a bug, it will reproduce at different latencies, which can be annoying.

This basically only effects the order in which `ninja` runs things, I doubt there's a bug that's inter-job.

libexec/revng/mass-testing/revng-revng-test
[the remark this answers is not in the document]
aleclearmind: I'm confused, what's the ratio for creating symlinks without `revng-`?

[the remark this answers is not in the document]
mrjackv: So that you can write:
```
test-harness ... ---
  revng ...
```
in `commands.yml` without `revng mass-testing`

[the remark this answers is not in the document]
mrjackv: With:
```yaml
commands:
- type: revng-regression.decompile
  from:
  - type: source
  suffix: /
  command: |-
    test-harness "$INPUT" "$OUTPUT" --timeout 300 --
    revng
      artifact
      decompile-to-single-file
      --analyze %INPUT%
```

Without:
```yaml
commands:
- type: revng-regression.decompile
  from:
  - type: source
  suffix: /
  command: |-
    revng mass-testing test-harness "$INPUT" "$OUTPUT" --timeout 300 --
    revng mass-testing revng-test
      artifact
      decompile-to-single-file
      --analyze %INPUT%
```

(I've also changed the `revng` invocation because I have no idea how `revng` would be symlinked to `revng-test`, since it would no longer be on `PATH`, modulo implementing some lookup in `libexec`)

python/revng/internal/cli/support.py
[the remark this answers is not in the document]
mrjackv: Dropped use of `support.run` from `mass-testing`.

libexec/revng/mass-testing/revng-dump-sections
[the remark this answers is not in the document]
mrjackv: Done

python/revng/internal/cli/_commands/mass_testing/configure.py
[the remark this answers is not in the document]
mrjackv: Implemented

PR #397, opened 2024-07-18
Rebase lifter on new libtcg [WIP]

AntonJohansson:


lib/Lift/InstructionTranslator.cpp
```diff
@@ -733,47 +626,72 @@ IT::TranslationResult IT::translateCall(PTCInstruction *Instr) {
...
-    auto *Load = Variables.load(Builder, TemporaryId);
+  for (unsigned I = 0; I < Instr->nb_iargs; ++I) {
+    auto *Load = Variables.load(Builder, &Instr->input_args[I]);
     if (Load == nullptr)
       return Abort;
     InArgs.push_back(Load);
   }
 
-  auto ConstArgs = TheInstruction.ConstArguments;
+  // TODO(anjo): Remove this, constant args to eg add are not passed as const but const temps
```
aleclearmind, line 658: Can you elaborate?

lib/Lift/InstructionTranslator.cpp
```diff
@@ -826,60 +744,80 @@ IT::translateOpcode(PTCOpcode Opcode,
...
+    revng_assert(ConstArguments.size() > 0 and
+                 ConstArguments[0].kind == LIBTCG_ARG_COND);
     Value *Compare = createICmp(Builder,
-                                ConstArguments[0],
+                                ConstArguments[0].cond,
                                 InArguments[0],
                                 InArguments[1]);
     // TODO: convert single-bit registers to i1
+    // NOTE(anjo) I don't think this is a good idea, semantics of QEMU require
+    // these to be full size, this is especially true for negsetcond
```
aleclearmind, line 767: Is this comment about the TODO above?

lib/Lift/InstructionTranslator.cpp
```diff
@@ -1143,56 +1088,43 @@ IT::translateOpcode(PTCOpcode Opcode,
...
-    Value *MaskedBase = Builder.CreateAnd(InArguments[0], BaseMask);
-    Value *Deposit = Builder.CreateAnd(InArguments[1], Bits);
-    Value *ShiftedDeposit = Builder.CreateShl(Deposit, Position);
-    Value *Result = Builder.CreateOr(MaskedBase, ShiftedDeposit);
-
+  case LIBTCG_op_deposit_i32:
+  case LIBTCG_op_deposit_i64: {
+    Value *Result = genDeposit(Builder, RegisterSize,
+                               InArguments[0], InArguments[1],
+                               InArguments[2], InArguments[3]);
```
aleclearmind, line 1095: Just outlining or functional changes?

lib/Lift/InstructionTranslator.cpp
```diff
@@ -1202,48 +1134,57 @@ IT::translateOpcode(PTCOpcode Opcode,
...
   }
-  case PTC_INSTRUCTION_op_not_i32:
-  case PTC_INSTRUCTION_op_not_i64:
+  case LIBTCG_op_extrl_i64_i32: {
+      return v{ Builder.CreateTrunc(InArguments[0], Builder.getInt32Ty()) };
+  }
+  case LIBTCG_op_extrh_i64_i32: {
+      Value *Shifted = Builder.CreateAShr(InArguments[0], ConstantInt::get(Builder.getInt64Ty(), 32));
+      return v{ Builder.CreateTrunc(Shifted, Builder.getInt32Ty()) };
+  }
```
aleclearmind, line 1161: These are new instructions right?
Can you make a list of new instructions so I can stare at them intensely for a while?

lib/Lift/VariableManager.cpp
```diff
@@ -410,6 +411,7 @@ void VariableManager::finalize() {
       Builder.CreateBr(ReturnBB);
 
       // Add the case to the switch
+      // TODO(anjo): TMP
```
aleclearmind, line 414: What's this?

lib/Lift/VariableManager.cpp
```diff
@@ -516,71 +517,79 @@ VariableManager::getByCPUStateOffsetInternal(intptr_t Offset,
...
-    } else {
-      GlobalsMap::iterator It = OtherGlobals.find(TemporaryId);
-      if (It != OtherGlobals.end()) {
+  // TODO(anjo): I guess we map TCG args to LLVM values here?
+  //             Where are constants handled?
+  switch (Arg->kind) {
+  case LIBTCG_ARG_TEMP:
+    switch (Arg->temp->kind) {
+    case LIBTCG_TEMP_EBB: {
+      // Temporary is dead at the end of the Extended Basic Block (EBB), the single entry,
```
aleclearmind, line 535: What's an EBB?

lib/Lift/CodeGenerator.cpp
```diff
@@ -706,6 +751,14 @@ void CodeGenerator::translate(optional<uint64_t> RawVirtualAddress) {
       if (not F->isDeclaration() and not F->isIntrinsic())
         F->setLinkage(GlobalValue::InternalLinkage);
 
+  //
+  // Create the libtcg context
+  //
+  // TODO(anjo): Move to Lift.cpp?
+  LibTcgDesc Desc = {};
```
aleclearmind, line 758: What's the use case for `LibTcgDesc`?
We're not using it.

lib/Lift/InstructionTranslator.cpp
```diff
@@ -38,438 +39,339 @@ static cl::opt<bool> RecordASM("record-asm",
...
-
-template<>
-inline uint64_t
-InstructionArgumentsIterator<In, true>::get(unsigned Index) const {
-  return ptc_call_instruction_in_arg(&ptc, TheInstruction, Index);
+static uint64_t pc(LibTcgInstruction *Instr) {
+  revng_assert(Instr->opcode == LIBTCG_op_insn_start);
+  uint64_t PC = Instr->constant_args[0].constant;
+  if (Instr->nb_cargs > 1)
+    PC |= Instr->constant_args[1].constant << 32;
```
aleclearmind, line 46: Can you elaborate on why we need this?

aleclearmind (changes requested):

lib/Lift/InstructionTranslator.cpp
```diff
@@ -681,7 +576,7 @@ IT::newInstruction(PTCInstruction *Instr,
     }
   }
 
-  Variables.newBasicBlock();
+  //Variables.newBasicBlock();
```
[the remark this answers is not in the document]
AntonJohansson: This is a consequence of the lifetime change in QEMU. Smallest scope for variable lifetime pre-TCG optimization is now an extended basic block, which is a basic block that allows for early exits using conditional branches.

BB < EBB <TB

The allocated variables are instead reset on encountering an unconditional jump or exittb.

I left the comment around because I was expecting a bug related to variable lifetimes and I wanted to find the old reset point easily

lib/Lift/InstructionTranslator.cpp
AntonJohansson: History, 

a previous version of libtcg (QEMU 6.0?) stored immediate values to instructions in the block of `LibTcgInstruction::constant_args[]`. What the block of code below this line _did_ was separate out immediate values from other constants such as labels/comparison ops/memory ops. Immediate values would go in `InArgs`, and other constants into `ConstArgs`.

But this changed somewhere between 6.0 and 8.2.1, and now immediate values are stored on the TCG Temp struct instead (LibTcgTemp::val), and so the path pushing to `InArgs` is _rarely_ taken. There are still a few cases, op_extract2 and op_st have some argument there at the very least, but the distinction is not necessary anymore, as it's not as common.

So instead we can simply pass in `Instr->constant_args` directly, maybe wrapped in some fancy C++ std::array, and the _few_ instructions which need constants can access them directly.

lib/Lift/InstructionTranslator.cpp
AntonJohansson: Yes exactly, QEMU has no concept of i1, so to match semantics we should just use i32 or i64

lib/Lift/InstructionTranslator.cpp
AntonJohansson: Only outlining, we need genDeposit to implement extract2 in a later commit

lib/Lift/InstructionTranslator.cpp
AntonJohansson: The major commit which updates InstructionTranslator is only to maintain parity in terms of instruction support. All new instructions are handled in separate following commits. Look at:

  "Lift/InstructionTranslator: handle ext[rl] + exitTB",
  "Lift/InstructionTranslator: handle clz_i[32|64],[s]extract[2]_i32",
  "Lift/InstructionTranslator: handle [s]extract_i64"
  "Lift/InstructionTranslator: handle ctz_i[32|64],negsetcond_[32|64]"

lib/Lift/VariableManager.cpp
AntonJohansson: No idea TBH xD

Probably I wanted to look at this again so I left a marker in the code, or there was some hacky solution there previously but the comments stuck around.

You can just remove it I think.

lib/Lift/VariableManager.cpp
AntonJohansson: The comment explains, Extended Basic Block (EBB) it's a single-entry, multiple-exit basic block where you also allow conditional branches.

It appeared in QEMU as a part of the TCG lifetime change.

lib/Lift/CodeGenerator.cpp
AntonJohansson: This is from a common C pattern to allow for default initialized and named arguments, it's very useful when it comes to configuring something such as the LibTcgContext in this case.

Currently LibTcgDesc only has two fields
```c
typedef struct LibTcgDesc {
    void *(*mem_alloc)(size_t);
    void (*mem_free)(void *);
} LibTcgDesc;
```

for overriding the memory allocator used for translation blocks/temporaries, initializing it to 0, means "use the default" which is malloc/free in this case.

A real-world use-case is in the libtcg example analyzes where we override the allocator to be a simple bump allocator
```c
    LibTcgInterface libtcg;
    LibTcgContext *context;
    libtcg_open(arch, &(LibTcgDesc){
        .mem_alloc = libtcg_alloc,
    }, &libtcg, &context);
```
Here we are overriding the allocator to a simple bump allocator which uses a single allocation and partitions out the memory, we then free all memory at once later on. Note this uses the libtcg-loader API (libtcg_open), which revng doesnt use ATM.

At the time of designing the libtcg API I was expecting needing to add way more configuration options. This pattern is not as useful in our case when we only have two parameters. I see this being more useful if we in the future decide to add more configuration options to initializing the LibTcgContext.

Here's a more complicated example which uses the same pattern along with designated array initializers in C to construct complicated data in readable ways (from https://github.com/floooh/sokol)
```c
    state.pip = sg_make_pipeline(&(sg_pipeline_desc){
        .shader = sg_make_shader(triangle_shader_desc(sg_query_backend())),
        .layout = {
            .attrs = {
                [ATTR_triangle_position].format = SG_VERTEXFORMAT_FLOAT3,
                [ATTR_triangle_color0].format = SG_VERTEXFORMAT_FLOAT4
            }
        },
    });
```

lib/Lift/InstructionTranslator.cpp
AntonJohansson: IIRC we are not forcing libtcg to an x64 host, and if you're lifting 64-bit code on a 32-bit host, the PC would be split into two registers for the `op_insn_start` instruction.

But we should force libtcg to lift as if it was running on a x64 host no matter what, for consistency of IR. I'll take a look at this.

lib/Lift/InstructionTranslator.cpp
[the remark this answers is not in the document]
aleclearmind: I wonder if it really makes sense to distinguish them in our context?
Under the assumption that the TCG we get is valid, I don't think we really care about the distinction.

Are they in different "namespaces"? As in, can you have `LIBTCG_TEMP_EBB` with ID 1 and a `LIBTCG_TEMP_TB` with the same ID?

lib/Lift/CodeGenerator.cpp
aleclearmind: Got it.

PR #401, opened 2024-09-02
Introduce `ptml::CTypeBuilder` (sister PR)

Cvelth:


include/revng/PTML/Doxygen.h
```diff
@@ -13,7 +13,7 @@ namespace ptml {
 
 /// Emits PTML containing a free-form comment passed in a string reformatted
 /// by appending a specified indicator as well maintaining a specific offset.
-std::string freeFormComment(const ::ptml::PTMLBuilder &B,
+std::string freeFormComment(const ::ptml::TagBuilder &B,
```
aleclearmind, line 16: Maybe it should be `MarkupBuilder` or just  `Builder`. AFAIU a "tag" is `<div>` (opening tag) or `</div>` (closing tag).

Cvelth: To be fair, tags are all it builds. All of our stuff is encoded like `<div data-some-name="some-value">`.

That being said, I don't have strong preferences (`Builder` is the best because it's the shortest, but that's the extent of it).

On that note, @pfez, since renaming was your idea, what do you think?

pfez: `Builder` seems too general. I wanted something that conveys the meaning of "this builder only builds very simple markup tag, not an entire PTML file". `MarkupBuilder` works fine.

Cvelth: Done.

PR #406, opened 2024-09-20
Remove leading `_` from generated names

Cvelth:
Sister PR: https://github.com/revng/revng-c/pull/36


include/revng/Model/Binary.h
```diff
@@ -282,15 +283,51 @@ class model::Binary : public model::generated::Binary {
...
+  }
+  const model::TypeOfGlobalNameCache *GlobalNameCache() const debug_function {
+    return GlobalNameCache(false);
+  }
+
+  void discardGlobalNameCache() const { TheGlobalNameCache = std::nullopt; }
+
+public:
+  inline std::string
+  getLLVMFunctionName(const model::Function &Function) const {
```
aleclearmind, line 322: 1. `inline` is not necessary
2. In LLVM we don't have all the issues we have in C, I think we can just use `CustomName` directly, if available.

include/revng/Model/Binary.h
```diff
@@ -282,15 +283,51 @@ class model::Binary : public model::generated::Binary {
...
 public:
   void dumpTypeGraph(const char *Path) const debug_function;
 
 public:
   MetaAddressRangeSet executableRanges() const;
+
+private:
+  // It's defined here instead of the TTG because we don't really want it to
+  // ever get serialized.
+  mutable std::optional<model::TypeOfGlobalNameCache> TheGlobalNameCache = {};
```
aleclearmind, line 295: Oh no. Not `mutable`.

I'm starting to think that all the naming business should be an external object that inspects the model.
There's no real reason to embed it in the model.

If you want the names, you create your object, you take care of its lifetime and make sure you don't change the model while such object is alive.

```c++
class Namer {
private:
  GlobalNamespace;
public:
  Namer(const model::Binary &Binary) { /* Initialize GlobalNamespace */ }

public:
  std::string getArgumentName(...);
};

File: lib/Model/TypeDefinition.cpp [hunk 3]
```diff
@@ -115,17 +115,42 @@ model::Identifier model::StructField::name() const {
...
+  model::Identifier Result;
+
+  if (Argument.CustomName().empty()) {
+    const std::string &Prefix = Configuration.unnamedFunctionArgumentPrefix();
+    (Prefix + llvm::Twine(Argument.Index())).toVector(Result);
+
+    for (const auto &AnotherArgument : Arguments())
+      if (not AnotherArgument.CustomName().empty())
+        if (AnotherArgument.CustomName() == Result)
+          return std::nullopt; // Argument name collides with another argument!
```
aleclearmind, line 133: This is not valid in the model.
My expectations is that we won't need any `tryGet...Name`, they are supposed to *always* succeed (on a valid model, which we can assume).

lib/Model/Binary.cpp
```diff
@@ -173,15 +173,27 @@ MetaAddressRangeSet Binary::executableRanges() const {
...
+    revng_assert(Names.contains(CustomName()), "Cache is outdated?");
     return CustomName();
+
   } else {
-    auto AutomaticName = (Twine("_function_") + toIdentifier(Entry())).str();
+    std::string AutomaticName = Configuration.unnamedFunctionPrefix()
+                                + toIdentifier(Entry());
+    if (Names.contains(AutomaticName)) {
+      // COLLISION!!! EVERYONE, PANIC!
+      // jokes aside, I'm not sure what to do here.
```
aleclearmind, line 188: You need to keep adding `_`.
No point in having `tryGetName`, let's just have `name` which adds as many `_` as necessary.

It seems to me that you're trying to avoid handling this case, but this is a perfectly normal situation that we need to handle silently.

include/revng/Model/Binary.h
Cvelth: 1. A left-over, my bad. This was a free function and I didn't pay attention.
2. Yes, but if we do that, we aren't going to get `_function_0x_$SOMEWHERE_Code_$SOMETHING` labeled correctly. And I'd rather keep that logic in one place.

But considering this is going to become a member of the new `Namer` thing, I think it will work out nicely.

aleclearmind:
Please undraft once ready for review.

include/revng/Model/Binary.h
Cvelth: 1. Done.

Cvelth: 1. Done.

lib/Model/Binary.cpp
Cvelth: Done.

include/revng/Model/Binary.h
Cvelth: Please take a look.

pfez:
@aleclearmind who is supposed to review this, you or me?

include/revng/Model/NamingHelper.h
```diff
@@ -0,0 +1,171 @@
...
+  }
+};
+
+} // namespace detail
+
+class VerifyHelper;
+
+class NamingHelper {
+private:
+  using GlobalNameCacheType = std::unordered_map<std::string,
```
aleclearmind, line 47: `GlobalNameCacheMap`

aleclearmind (changes requested):

aleclearmind:
> @aleclearmind who is supposed to review this, you or me?

I'll review it

include/revng/Model/NamingConfiguration.h
```diff
@@ -0,0 +1,181 @@
...
+  - name: UnnamedFunctionPrefix
+    doc: The default value is `function_`.
+    type: string
+    optional: true
+  - name: UnnamedDynamicFunctionPrefix
+    doc: The default value is `dynamic_`.
+    type: string
+    optional: true
+
+  - name: UnnamedTypeDefinitionPrefix
```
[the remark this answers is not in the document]
Cvelth: Prefix before the type name, for example
```
THIS_THING_struct_0
THIS_THING_typedef_1
```

lib/ABI/Analyses/ConvertFunctionsToCABI.cpp
```diff
@@ -249,6 +249,11 @@ class ConvertFunctionsToCABI {
     auto ToConvert = filterTypes<RawFD>(Model->TypeDefinitions(),
                                         TypesToIgnore);
 
+    using NH = model::NamingHelper;
+    auto NamingHelper = Log.isEnabled() ?
+                          std::make_optional<const NH>(Model->namingHelper()) :
+                          std::nullopt;
```
aleclearmind, line 255: Use an `if` statement

lib/FunctionIsolation/IsolateFunctions.cpp
```diff
@@ -183,14 +184,15 @@ class IsolateFunctionsImpl {
     Context(TheModule->getContext()),
     GCBI(GCBI),
     Binary(Binary),
+    NamingHelper(Binary),
```
aleclearmind, line 187: Does this mean we're constructing the global namespace just to then call `LLVMFunction` which, unless we have `DebugNames` turned on, doesn't use it?

include/revng/Model/NamingHelper.h
```diff
@@ -0,0 +1,171 @@
...
+  NamingHelper(const NamingConfiguration &Configuration) :
+    Configuration(Configuration) {}
+
+  std::optional<std::string_view> isNameForbidden(std::string_view Name);
+  bool populateGlobalNameCache(VerifyHelper &VH, const Binary &Binary);
+
+public:
+  static std::optional<NamingHelper> tryMake(VerifyHelper &VH,
+                                             const Binary &Binary);
+  NamingHelper(const Binary &Binary);
```
aleclearmind, line 67: Push the constructors at the top.

I think the pattern for using `NamingHelper` should be:

1. Lazily call `populateGlobalNameCache` so that we can call `LLVMFunction(...)` without invoking it (if `DebugNames` is turned off).
2. When calling it lazily from the various methods, assert it's successful.
3. Drop `tryMake`. Instead, construct the regular object and, if a certain call sites expects failures, let it call `populateGlobalNameCache` proactively and check the result.

include/revng/Model/VerifyHelper.h
```diff
@@ -144,9 +145,11 @@ class VerifyHelper {
...
 
 public:
-  [[nodiscard]] bool isGlobalSymbol(const model::Identifier &Name) const;
-  [[nodiscard]] bool registerGlobalSymbol(const model::Identifier &Name,
-                                          const std::string &Path);
+  [[nodiscard]] bool isGlobalSymbol(std::string_view Name) const;
+  [[nodiscard]] bool registerNamingHelper(const model::Binary &Binary);
+  [[nodiscard]] const model::NamingHelper *getNamingHelper() const {
+    return NamingHelper.get();
+  }
```
aleclearmind, line 152: Unused

include/revng/Model/VerifyHelper.h
```diff
@@ -144,9 +145,11 @@ class VerifyHelper {
   }
 
 public:
-  [[nodiscard]] bool isGlobalSymbol(const model::Identifier &Name) const;
-  [[nodiscard]] bool registerGlobalSymbol(const model::Identifier &Name,
-                                          const std::string &Path);
+  [[nodiscard]] bool isGlobalSymbol(std::string_view Name) const;
+  [[nodiscard]] bool registerNamingHelper(const model::Binary &Binary);
```
aleclearmind, line 149: As mentioned elsewhere, would it make sense to have `model::Binary` be an argument of `VerifyHelper` itself?

include/revng/Model/NamingConfiguration.h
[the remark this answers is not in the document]
aleclearmind: Ah, documentation.

lib/FunctionIsolation/EnforceABI.cpp
```diff
@@ -151,8 +151,6 @@ bool EnforceABI::prologue() {
 
 bool EnforceABI::runOnFunction(const model::Function &FunctionModel,
                                llvm::Function &OldFunction) {
-  revng_assert(not FunctionModel.name().empty());
-
```
aleclearmind, line 156: OKGO

include/revng/Model/NamingHelper.h
```diff
@@ -0,0 +1,171 @@
...
+  GlobalNameCacheType GlobalNameCache = {};
+
+public:
+  const NamingConfiguration &Configuration;
+
+private:
+  NamingHelper(const NamingConfiguration &Configuration) :
+    Configuration(Configuration) {}
+
+  std::optional<std::string_view> isNameForbidden(std::string_view Name);
```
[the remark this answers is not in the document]
Cvelth: It is, technically a boolean, because options is implicitly convertible. The string it returns when `true` is just an error message.

include/revng/Model/NamingHelper.h
```diff
@@ -0,0 +1,171 @@
...
+class VerifyHelper;
+
+class NamingHelper {
+private:
+  using GlobalNameCacheType = std::unordered_map<std::string,
+                                                 std::string,
+                                                 detail::TransparentStringHash,
+                                                 std::equal_to<>>;
+
+  GlobalNameCacheType GlobalNameCache = {};
```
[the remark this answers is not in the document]
Cvelth: Yeah, the name made more sense while it was a field of the Binary. Good catch!

include/revng/Model/NamingHelper.h
```diff
@@ -0,0 +1,171 @@
...
+                                             const Binary &Binary);
+  NamingHelper(const Binary &Binary);
+
+public:
+  [[nodiscard]] bool isGlobalSymbol(std::string_view Name) const {
+    return GlobalNameCache.contains(Name);
+  }
+
+public:
+  Identifier segment(const model::Segment &Segment) const;
```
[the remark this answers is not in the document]
Cvelth: I did consider that. But also it seemed like I was loosing some explicitness at the revng-c side so I rolled it back.

But honestly, in the hindsight, it wasn't as bad of an idea as it seemed.

Let's go with overloads!

include/revng/Model/NamingHelper.h
```diff
@@ -0,0 +1,171 @@
...
+    else
+      Result += Function.Entry().toString();
+
+    return Result;
+  }
+
+public:
+  Identifier type(const model::TypeDefinition &Definition) const;
+
+  Identifier enumEntry(const model::EnumEntry &Entry,
```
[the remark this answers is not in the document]
Cvelth: I kind of disagree 🤔.

I feel like the most important thing should go first, in this case - we're emitting an entry - so _which_ entry we're emitting is the most important bit. All the other arguments are just there for context.

But, if you feel strongly about "a big thing before a small one" rule, we should add it to the code conventions pad or something.

include/revng/Model/NamingHelper.h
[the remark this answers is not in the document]
aleclearmind: You need to use `llvm::Error` for this. Let's not make up new error reporting mechanisms.

include/revng/Model/NamingHelper.h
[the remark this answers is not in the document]
aleclearmind: I don't have super strong opinions, but usually I like arguments in decreasing order of how generic they are.
Usually "Context" is the first argument, then you have Module, then Function...

Considering the implicit `this` argument, also that is rather generic.
But no biggie.

include/revng/Model/VerifyHelper.h
```diff
@@ -144,9 +145,11 @@ class VerifyHelper {
   }
 
 public:
-  [[nodiscard]] bool isGlobalSymbol(const model::Identifier &Name) const;
-  [[nodiscard]] bool registerGlobalSymbol(const model::Identifier &Name,
-                                          const std::string &Path);
+  [[nodiscard]] bool isGlobalSymbol(std::string_view Name) const;
```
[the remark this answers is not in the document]
aleclearmind: I'd prefer we stick with `StringRef` all over the place, unless we globally decide to switch to `string_view` (and retain `StringRef` for LLVM API only).

include/revng/Yield/PTML.h
```diff
@@ -58,7 +59,8 @@ std::string functionNameDefinition(const ::ptml::MarkupBuilder &B,
 /// ```
 std::string functionLink(const ::ptml::MarkupBuilder &B,
                          std::string_view Location,
-                         const model::Binary &Binary);
+                         const model::Binary &Binary,
+                         const model::NamingHelper &NamingHelper);
```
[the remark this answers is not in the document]
Cvelth: Not yet, but we definitely planned to.

But honestly, I'd rather wait until we merge revng-c into revng so that we end up with a single set of `ptml::XBuilder`s encapsulating everything. (I feel like I provided this same argument last time we discussed this).

Not much point changing them now considering that not _too_ bad just yet _and_ the whole thing is going to change sooner rather than later.

lib/FunctionIsolation/IsolateFunctions.cpp
Cvelth: Yep, and it pains me to do that. I guess we should sneak one more `DebugNames` check which is also sad.

Or, maybe, provide another way to make `NamingHelper` (template argument) which ignores global namespace. We don't care about trailing `_`s as much in debug names.

include/revng/Model/VerifyHelper.h
Cvelth: It was at one point. But the thing is, it would still need a separate `NamingHelper`. So we kind of don't win anything - hence why I get rid of it.

include/revng/Model/NamingHelper.h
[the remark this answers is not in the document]
Cvelth: I guess it's the total opposite of how I see things :D

The way I see it, the most generic argument is going to appear a lot more often, as such, no need to draw attention to it as much - it can be towards the end. While the first argument is a good way to indicate "hey, this is a really important bit".

But, yeah, not a strong stance, hence why I suggesting deciding on a single approach in the conventions pad - for consistency sake.

include/revng/Model/VerifyHelper.h
aleclearmind: Can you elaborate?
If it's in the constructor, we can avoid having the `registerNamingHelper` method and avoid `VerifyHelper::NamingHelper` to be a `unique_ptr`.

include/revng/Yield/PTML.h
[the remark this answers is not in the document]
aleclearmind: Uff, refactoring never ends. OK.

include/revng/Model/VerifyHelper.h
Cvelth: fuck, I misread it! All good, building the VH from the binary works!

include/revng/Model/NamingHelper.h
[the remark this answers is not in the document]
aleclearmind: I `git grep`'d the LLVM codebase and it seems that it's a bit more common to have the more general thing first.
I will add this to the best pratice.

lib/ABI/Analyses/ConvertFunctionsToCABI.cpp
Cvelth: Ah, wait, this cannot be done as an `if` statement because the naming helper is not assignable (it has a reference member). Should I try to think of another trick? Or can a ternary operator stay?

include/revng/Model/Binary.h
```diff
@@ -282,15 +284,18 @@ class model::Binary : public model::generated::Binary {
...
+  bool verifyConfiguration() const debug_function;
 
 public:
   void dumpTypeGraph(const char *Path) const debug_function;
 
 public:
   MetaAddressRangeSet executableRanges() const;
+
+public:
+  NamingHelper namingHelper() const { return NamingHelper(*this); }
```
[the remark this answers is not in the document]
Cvelth: Gone!

include/revng/Model/NamingConfiguration.h
```diff
@@ -0,0 +1,181 @@
+#pragma once
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+/* TUPLE-TREE-YAML
+name: NamingConfiguration
```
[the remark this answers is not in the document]
Cvelth: Please take a look!

include/revng/Model/NamingConfiguration.h
```diff
@@ -0,0 +1,181 @@
...
+    doc: The default value is `dynamic_`.
+    type: string
+    optional: true
+
+  - name: UnnamedTypeDefinitionPrefix
+    doc: |
+      The default value is ``.
+      Note that the type kind (like `struct`, or `typedef`) is going to be
+      inserted automatically after this Prefix
+      See `model::TypeDefinitionKind::automaticNamePrefix` helper.
```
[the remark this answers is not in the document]
Cvelth: I just dropped the line :D

I don't feel like explaining all the supported type kinds here is a good idea.

include/revng/Model/NamingConfiguration.h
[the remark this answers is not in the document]
Cvelth: Please take a look!

lib/Model/Verification.cpp
```diff
@@ -782,6 +706,63 @@ bool Binary::verifyTypeDefinitions(VerifyHelper &VH) const {
   return true;
 }
 
+//
+// Configuration
+//
+
+bool Binary::verifyConfiguration(VerifyHelper &VH) const {
```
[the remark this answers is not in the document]
Cvelth: Done!

include/revng/Model/NamingHelper.h
Cvelth: I went with `GlobalNamespaceMap`.

include/revng/Model/NamingHelper.h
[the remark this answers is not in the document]
Cvelth: Done.

Note, while at it I also added a helper for making those and adopted it everywhere. Please take a look at the `Introduce an error creation helper` commit.

include/revng/Model/NamingHelper.h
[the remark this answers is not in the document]
Cvelth: Done!

include/revng/Model/NamingHelper.h
```diff
@@ -0,0 +1,171 @@
...
+  [[nodiscard]] bool isGlobalSymbol(std::string_view Name) const {
+    return GlobalNameCache.contains(Name);
+  }
+
+public:
+  Identifier segment(const model::Segment &Segment) const;
+  Identifier function(const model::Function &Function) const;
+  Identifier dynamicFunction(const model::DynamicFunction &Function) const;
+
+  std::string LLVMFunction(const model::Function &Function) const {
```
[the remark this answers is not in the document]
Cvelth: I went with `llvmName`.

include/revng/Model/NamingHelper.h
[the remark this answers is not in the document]
Cvelth: Done!

include/revng/Model/NamingHelper.h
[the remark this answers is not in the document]
Cvelth: Done!

include/revng/Model/NamingHelper.h
```diff
@@ -0,0 +1,171 @@
...
+
+  Identifier field(const model::UnionField &Field,
+                   const model::UnionDefinition &Definition) const;
+  Identifier field(const uint64_t &Index,
+                   const model::UnionDefinition &Definition) const {
+    revng_assert(Index <= Definition.Fields().size());
+    return field(Definition.Fields().at(Index), Definition);
+  }
+
+  Identifier padding(const uint64_t &Offset) const {
```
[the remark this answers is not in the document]
Cvelth: Gone, good catch!

include/revng/Model/NamingHelper.h
```diff
@@ -0,0 +1,171 @@
...
+
+  Identifier returnValue(const model::NamedTypedRegister &ReturnValue,
+                         const model::RawFunctionDefinition &Function) const;
+  Identifier returnValue(const model::Register::Values &Register,
+                         const model::RawFunctionDefinition &Function) const {
+    return argument(Function.ReturnValues().at(Register), Function);
+  }
+
+private:
+  using RFT = model::RawFunctionDefinition;
```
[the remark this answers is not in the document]
Cvelth: Done!

include/revng/Model/VerifyHelper.h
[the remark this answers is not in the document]
Cvelth: Done, see the new dedicated commit (`Replace std::string_view with llvm::StringRef`)

Note that I had to roll some back to `std::string_view` because `llvm::StringRef` lacks some constexpr capabilities (for example, you cannot even compare strings).

lib/FunctionIsolation/IsolateFunctions.cpp
Cvelth: Closing this as it kind of duplicates the next thread.

include/revng/Model/NamingHelper.h
Cvelth: Done!

include/revng/Model/VerifyHelper.h
Cvelth: Gone!

include/revng/Model/VerifyHelper.h
```diff
@@ -74,18 +76,17 @@ class VerifyHelper {
   std::map<const model::TypeDefinition *, uint64_t> SizeCache;
   std::set<const model::TypeDefinition *> InProgress;
   bool AssertOnFail = false;
-  std::map<model::Identifier, std::string> GlobalSymbols;
+  std::unique_ptr<const model::NamingHelper> NamingHelper = nullptr;
```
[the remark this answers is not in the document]
Cvelth: Closing this in favor of the other similar thread: https://github.com/revng/revng/pull/406#discussion_r1804670878

include/revng/Model/VerifyHelper.h
Cvelth: The constructor worked great!

The only issue is that even this didn't let us avoid a pointer in here. There was one more reason for it I overlooked initially: this header cannot include `NameBuilder.h` because `NameBuilder.h` has to include all the different headers (`Function.h`, `TypeDefinition.h` and so on) that include this header - leading to a cycle.

lib/FunctionIsolation/EnforceABI.cpp
Cvelth: Done!

include/revng/Model/NamingConfiguration.h
```diff
@@ -0,0 +1,223 @@
...
+    doc: |
+      The prefix for a segment without a name.
+
+      The default value is `segment_`.
+    type: string
+    optional: true
+
+  - name: UnnamedFunctionPrefix
+    doc: |
+      The prefix for a local function without a name.
```
aleclearmind, line 21: Indentation

include/revng/Support/Error.h
```diff
@@ -0,0 +1,21 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Support/Error.h"
+
+namespace revng {
+
+template<typename... Ts>
+inline llvm::Error createLLVMError(char const *Fmt, const Ts &...Vals) {
```
aleclearmind, line 13: I would call this `revng::createError` or `revng::createStringError`, it's not something we only use with LLVM APIs.

include/revng/Model/NamingHelper.h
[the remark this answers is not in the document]
aleclearmind: Ah, very much needed. Thanks!

include/revng/Model/VerifyHelper.h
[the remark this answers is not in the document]
aleclearmind: Thanks, I know you love `string_view` but I really want to strive for coherency in the codebase.

include/revng/Model/VerifyHelper.h
aleclearmind: What pointer survived? Please give me a specific reference.

include/revng/Model/NamingConfiguration.h
Cvelth: Wait, what do you mean? These are not checked by the check-conventions?!

include/revng/Support/Error.h
Cvelth: The reason I put `LLVM` in there is so that it is easy to replace :wink:

That being said, do you want to keep it in `revng` namespace? We don't use it that much after all. I only added it to differentiate from the LLVM alternative.

include/revng/Model/NamingHelper.h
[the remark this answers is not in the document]
Cvelth: Actually, I don't care that much. It's not like `llvm::Twine` which I'd rather never appeared. ` llvm::StringRef` is perfectly good and nice to work with helper (hopefully they attach some more `constexpr` to its methods though 🤞)

include/revng/Model/VerifyHelper.h
Cvelth: The fact that the `NameBuilder` member is a `unique_ptr` inside the `VerifyHelper`.

include/revng/Model/NamingConfiguration.h
aleclearmind: No, these are comments. I think we run prettify on `.yml` files, but no in the middle of comments.

Cvelth: These should really become their own files sooner rather that later :cursed_parrot:

Cvelth: That being said, I just copied what was there before: https://github.com/revng/revng/blob/develop/include/revng/Model/ABI.h#L15

How do you see us formatting these?

include/revng/Model/VerifyHelper.h
aleclearmind: > NameBuilder.h has to include all the different headers (Function.h, TypeDefinition.h and so on)

Does it? It seems to me that you could forward declare most of the things and outline implementations.

include/revng/Model/NamingConfiguration.h
aleclearmind: I'm fine with copying things over, but here the indentation is different from the next line.

include/revng/Support/Error.h
Cvelth: Ah, that's what you meant. The `llvm` was there more to reflect that it makes an `llvm::Error`, and not about where it's supposed to be used. That being said, in hindsight, I can clearly see your point!

Let's go with `revng::createError`.

include/revng/Model/VerifyHelper.h
Cvelth: I guess it can be done, but then we're losing some inlining.

But even putting that aside, I feel like it's weird to dictate the requirements on a widely-used interface based on a verification routine. `VerifyHelper` feels pretty "internal, for the lack of a better word. I'd rather it be ugly, than the interface we get all of our names from.

include/revng/Model/NamingConfiguration.h
Cvelth: Ah, fuck, a typo. Sorry for the noise and a good catch!

include/revng/Model/VerifyHelper.h
aleclearmind: OK, I'll leave the decision to you.

If you keep `unique_ptr`, add a thorough comment describing why we need that and what would need to be done in order to switch to a regular instance.

Cvelth: It was even deeper than expected: even `model::Identifier` depends on the `model::VerifyHelper`, so yeah, the other way around would be very awkward.

Comment I added:
```cpp
  // Name builder is stored as a pointer in order to avoid an inclusion loop:
  // NameBuilder methods return `model::Identifier`, which depends on this
  // helper.
  std::unique_ptr<model::NameBuilder> NameBuilder = nullptr;
```

include/revng/Model/NamingConfiguration.h
Cvelth: Fixed.

include/revng/Support/Error.h
Cvelth: Done.

scripts/tuple_tree_generator/tuple_tree_generator/templates/struct.h.tpl
```diff
@@ -123,6 +123,22 @@ public:
       /**- endif -**/
     }
 
+  /** if struct.abstract **/
```
Cvelth, line 126: @aleclearmind I added this to the PR, please take a look (the very last commit: `Model: forbid copies and moves of abstract structs`). Shouldn't be controversial, but just in case.

aleclearmind: If the code still compiles, seems reasonable!

PR #412, opened 2024-10-17
Attaching comments to statements

Cvelth:
~~Sister PR: https://github.com/revng/revng-c/pull/44~~


include/revng/Yield/ControlFlow/Comments.h
aleclearmind, line 1: Let's not add any new features to `Yield`. There's a tendency to put it here all the new work you do. But we should already split it off.

We can either to a new library or move this to PTML maybe, since this is probably going to be used by anyone emitting PTML.

Eventually, PTML should probably become something like `BackendSupport`.

Cvelth: Done.

aleclearmind (changes requested):

aleclearmind:
NOTETOSELF: resume from `PTML: add contextual comment support`.

include/revng/PTML/ContextualCommentMap.h
```diff
@@ -0,0 +1,223 @@
...
+              ResultMap[Statement].emplace_back(Index,
+                                                IsLocationExact,
+                                                &Comment.Location());
+              CommentWasPlaced = true;
+              break;
+            }
+          }
+
+          if (CommentWasPlaced)
+            break;
```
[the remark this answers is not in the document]
Cvelth: Wait, `B` _and_ `D`? That could be very awkward in some cases with ties.

Ah, but wait, if we lock it to specific lists, it might actually be what we want! I'll need to think about this a bit more, but now I get why you were so adamant about domination earlier.

include/revng/Model/VerifyHelper.h
```diff
@@ -77,10 +77,7 @@ class VerifyHelper {
   std::set<const model::TypeDefinition *> InProgress;
   bool AssertOnFail = false;
 
-  // Name builder is stored as a pointer in order to avoid an inclusion loop:
-  // NameBuilder methods return `model::Identifier`, which depends on this
-  // helper.
-  std::unique_ptr<model::NameBuilder> NameBuilder = nullptr;
+  std::optional<model::NameBuilder> NameBuilder = std::nullopt;
```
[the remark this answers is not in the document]
Cvelth: Because we use `VerifyHelper` all over the place. And in a lot of cases, we don't care about names, think `model::Function::verify()` (the overload without arguments).

include/revng/ADT/STLExtras.h
```diff
@@ -437,6 +437,16 @@ constexpr auto takeAsTuple(RangeType &&R) {
   });
 }
 
+// TODO: This will no longer be needed when optionals are accepted as ranges
+//       (C++26)
+template<typename T>
+llvm::SmallVector<T, 1> optionalToRange(std::optional<T> &&Value) {
```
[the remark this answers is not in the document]
Cvelth: Ah, right, I think I moved all the usages to small-vectors right from the bat. Good catch!

include/revng/Yield/ControlFlow/Graph.h
```diff
@@ -23,11 +24,11 @@ class Node {
 
 public:
   Node() = default;
-  Node(BasicBlockID Address,
+  Node(const yield::BasicBlock &Block,
```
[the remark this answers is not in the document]
Cvelth: Because otherwise making the new graph type that allows iterating on node instructions would be very awkward.

include/revng/Model/ContextualComment.h
```diff
@@ -0,0 +1,33 @@
...
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "revng/Support/MetaAddress.h"
+
+/* TUPLE-TREE-YAML
+
+name: ContextualComment
```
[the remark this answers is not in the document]
Cvelth: The big thing I want is for this to be distinct from the comments attached to types.

I don't mind calling this a `Comment`, but then we should call those other ones something else.

include/revng/PTML/ContextualCommentMap.h
```diff
@@ -0,0 +1,223 @@
...
+
+template<typename NodeType>
+struct StatementTraits {
+  // See `HasStatementTraits` for the list of what this trait should provide.
+};
+
+namespace detail {
+template<typename NodeType>
+using ST = StatementTraits<NodeType>;
+}
```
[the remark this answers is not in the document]
Cvelth: Ah, that's exactly what I tried to do, but `typename detail::ST::whatever<X>` is still too long.

Good catch!

include/revng/PTML/ContextualCommentMap.h
```diff
@@ -0,0 +1,223 @@
...
+  using CommentType = model::ContextualComment;
+  using CommentLocationType = decltype(std::declval<CommentType>().Location());
+
+  struct NumericScore {
+    uint64_t Numerator = 0;
+    uint64_t Denominator = 1;
+
+    auto operator<=>(const NumericScore &Another) const {
+      uint64_t AdjustedLHS = this->Numerator * Another.Denominator;
+      uint64_t AdjustedRHS = Another.Numerator * this->Denominator;
```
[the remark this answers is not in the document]
Cvelth: The most "stupid" way to compare two fractions:

Let's say we are trying to compare `a/b` and `c/d`. _A_ common denominator is `b*d`, so we can just compare `(a*d)/(b*d)` and `(c*b)/(d*b)`, or, rather, only the nominators (`a*d` and `c*b`), which is what happens here.

include/revng/Model/VerifyHelper.h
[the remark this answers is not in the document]
aleclearmind: OK, add a comment.

include/revng/PTML/ContextualCommentMap.h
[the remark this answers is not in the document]
Cvelth: Yeah, I'm aware of that, but as far as I remember, you mentioned that the plan is to duplicate the blocks _after_ the comments are already there.

include/revng/Model/ContextualComment.h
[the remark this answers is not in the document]
aleclearmind: I think these are statement comments, given the flexible definition of "statement" have. Therefore, what about `StatementComment`?

include/revng/PTML/ContextualCommentMap.h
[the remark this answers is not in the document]
aleclearmind: OK, I trust you, add a comment.

include/revng/PTML/ContextualCommentMap.h
```diff
@@ -0,0 +1,223 @@
...
+  std::unordered_map<typename Trait::StatementType, CommentList> ResultMap;
+  CommentList HomelessComments;
+
+public:
+  ContextualCommentMap() = default;
+  ContextualCommentMap(const model::Function &Function, NodeType EntryNode) {
+    using LocationType = std::decay_t<typename Trait::LocationType>;
+    std::vector<std::pair<NumericScore, LocationType>> Scores;
+    Scores.resize(Function.Comments().size());
+    if (Scores.empty()) {
```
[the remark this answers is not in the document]
Cvelth: It's not for optimization sake. I'm just using the vector as an array - it has exactly one entry per comment - this way I don't need to care about resizing it ever again. Don't forget that we are iterating nodes first.

That being said, moving it up would make me want to put `Scores` in as well - and we need that later on.

But honestly, this check should probably just live at the very top of the function, that way it's not going to confuse anyone.

include/revng/PTML/ContextualCommentMap.h
```diff
@@ -0,0 +1,223 @@
...
+
+          if (Score.has_value() && *Score > Scores[Index].first)
+            Scores[Index] = { *Score, Location };
+        }
+      }
+    }
+
+    // Then translate it into the mode-to-comment-list map taking into account
+    // the possibility of multiple nodes having the same location.
+    for (auto [Index, Comment] : llvm::enumerate(Function.Comments())) {
```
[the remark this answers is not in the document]
Cvelth: `[]` thing is special in a sense that it is a universal reference by definition (`auto [X, Y]` is indistinguishable from `auto &&[X, Y]`). As such, it makes it so `Index` is copied (since it's an integer) and `Comment` is a reference, since that's what iterator's `operator*` returns.

include/revng/PTML/ContextualCommentMap.h
```diff
@@ -0,0 +1,223 @@
...
+
+          if (CommentWasPlaced)
+            break;
+        }
+      }
+    }
+  }
+
+public:
+  CommentList getComments(Trait::StatementType Node) const {
```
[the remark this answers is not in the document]
Cvelth: Wait, right. The compiler probably complained and I just thoughtlessly removed the reference. My bad.

include/revng/Model/ContextualComment.h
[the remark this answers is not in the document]
Cvelth: If we're fine with pushing `Statement` terminology this far up (into the model) - sure, that's perfect.

But I have a bit of FUD with naming it so without a clear definition what a statement is way into not-`revng-c`-anymore-side of the pipeline.

include/revng/PTML/ContextualCommentMap.h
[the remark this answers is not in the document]
aleclearmind: I was just suggesting
```diff
-if (Scores.empty()) {
+if (Functions.Comments().empty())
```

include/revng/PTML/ContextualCommentMap.h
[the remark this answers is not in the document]
Cvelth: > we'll just run comment injection _after_ duplication

Apparently I cannot read :D My bad.

include/revng/PTML/ContextualCommentMap.h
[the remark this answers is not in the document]
aleclearmind: So `auto []` unpacking never creates a copy?

include/revng/PTML/ContextualCommentMap.h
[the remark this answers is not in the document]
Cvelth: Yeah, of course, moving the condition _before_ the vector definition entails that.

I guess I just used the shorter way to spell the same thing :D

include/revng/Model/ContextualComment.h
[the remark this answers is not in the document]
aleclearmind: I come to you and ask "what does the set of `MetaAddress`es define?".

The only possible answer is "a statement". So I think it makes sense to push the definition up here.

include/revng/PTML/ContextualCommentMap.h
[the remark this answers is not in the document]
Cvelth: Nope, the same way `auto &&` arguments never do. That's why it's the preferred way to accept "everything" in a heavily templated function.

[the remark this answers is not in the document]
Cvelth: Ok, on the second thought, there are some special cases, think copiable-but-not-movable type returned by value. But, as you can probably imagine, those are insanely rare!

include/revng/Model/ContextualComment.h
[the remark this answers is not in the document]
Cvelth: Yeah, from that angle it makes a lot of sense! It was a really good idea!

include/revng/PTML/ContextualCommentMap.h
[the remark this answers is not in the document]
Cvelth: Okay, it looks like I was wrong! Sorry about misinformation and good catch!

https://godbolt.org/z/771e1Ejhs

include/revng/Model/VerifyHelper.h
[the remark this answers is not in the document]
Cvelth: Done:
```cpp
  // This field a only gets populated if this object was created with a binary,
  // as such, no name-space related checks will take place, say, inside
  // a `model::Function::verify`.
```

include/revng/ADT/STLExtras.h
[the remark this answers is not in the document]
Cvelth: Gone!

include/revng/Model/ContextualComment.h
```diff
@@ -0,0 +1,33 @@
...
+type: struct
+
+fields:
+  - name: Location
+    doc: |
+      The point this comment is attached to encoded as a set of addresses
```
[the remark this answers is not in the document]
Cvelth: How about something like this?
```
      The point this comment is attached to encoded as a set of addresses.

      When emitted artifact contains a statement with a set of addresses exactly
      matching the set provided here, the comment is emitted before that
      statement.

      If there's no such statement, one is chosen based on how similar its
      address set is to the address set of the comment.

      Note that the same comment can be emitted multiple times if there are
      multiple statements (that do _not_ dominate each other) with the same
      address set.
```

include/revng/Model/ContextualComment.h
```diff
@@ -0,0 +1,33 @@
...
+    sequence:
+      type: SortedVector
+      elementType: MetaAddress
+  - name: Text
```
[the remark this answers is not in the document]
Cvelth: Done.

include/revng/PTML/ContextualCommentMap.h
```diff
@@ -0,0 +1,223 @@
...
+                                      StatementTraits<NodeType>::StatementType
+                                        Statement) {
+  // using StatementType = /* your type */
+  typename StatementTraits<NodeType>::StatementType;
+
+  // using AddressType = /* your type */
+  typename StatementTraits<NodeType>::AddressType;
+
+  // using LocationType = /* your _set_ of `AddressType`s */
+  typename StatementTraits<NodeType>::LocationType;
```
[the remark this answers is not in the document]
Cvelth: Gone.

include/revng/PTML/ContextualCommentMap.h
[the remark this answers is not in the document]
Cvelth: Gone.

include/revng/PTML/ContextualCommentMap.h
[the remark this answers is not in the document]
Cvelth: Done.

include/revng/PTML/ContextualCommentMap.h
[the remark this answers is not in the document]
Cvelth: Done.

include/revng/PTML/ContextualCommentMap.h
```diff
@@ -0,0 +1,223 @@
...
+    // Then translate it into the mode-to-comment-list map taking into account
+    // the possibility of multiple nodes having the same location.
+    for (auto [Index, Comment] : llvm::enumerate(Function.Comments())) {
+      if (Scores[Index].first.Numerator == 0) {
+        // No feasible location for this comment, mark it as homeless.
+        HomelessComments.emplace_back(Index, false, &Comment.Location());
+
+      } else {
+        bool CommentWasPlaced = false;
+        for (NodeType Node : llvm::depth_first(EntryNode)) {
```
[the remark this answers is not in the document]
Cvelth: Done.

include/revng/PTML/ContextualCommentMap.h
```diff
@@ -0,0 +1,223 @@
...
+          using StatementType = Trait::StatementType;
+          for (const StatementType &Statement : Trait::getStatements(Node)) {
+            if (Trait::getAddresses(Statement) == Scores[Index].second) {
+              bool IsLocationExact = std::ranges::equal(Comment.Location(),
+                                                        Scores[Index].second);
```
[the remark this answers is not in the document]
Cvelth: Attached.

include/revng/PTML/ContextualCommentMap.h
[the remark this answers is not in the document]
Cvelth: Done.

include/revng/PTML/ContextualCommentMap.h
```diff
@@ -0,0 +1,223 @@
...
+/// To build an object of this type, you need to pass a model function as well
+/// as the is by point of the graph.
+///
+/// Then you just call \ref getComments with a given node on its emission to
+/// get a (possibly empty) list of comments.
+///
+/// Don't forget to also call \ref getHomelessComments to get a (possibly empty)
+/// list containing all the comments that don't match a single node.
+template<HasStatementTraits NodeType>
+class ContextualCommentMap {
```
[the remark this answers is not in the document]
Cvelth: I went with `CommentPlacementHelper`.

include/revng/Model/ContextualComment.h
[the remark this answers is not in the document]
aleclearmind: Minor changes. OKGO.
```
      The point this comment is attached to, encoded as a set of addresses.

      When emitted artifact contains a statement with a set of addresses exactly
      matching the set provided here, the comment is emitted before that
      statement.

      If there's no such statement, one is chosen based on how similar its
      address set is to the address set of the comment.

      Note that the same comment can be emitted multiple times if there are
      multiple statements (that do *not* dominate each other) with the same
      address set.
```

include/revng/PTML/ContextualCommentMap.h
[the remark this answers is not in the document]
aleclearmind: My rule was "in `for` loops, use `const auto &[...]` if you don't want to change it, use `auto &[...]` if you want to change it, never use `auto [...]`".

Can I stick with it?

include/revng/PTML/ContextualCommentMap.h
```diff
@@ -0,0 +1,223 @@
...
+///         `StatementTraits`.
+///
+/// To build an object of this type, you need to pass a model function as well
+/// as the is by point of the graph.
+///
+/// Then you just call \ref getComments with a given node on its emission to
+/// get a (possibly empty) list of comments.
+///
+/// Don't forget to also call \ref getHomelessComments to get a (possibly empty)
+/// list containing all the comments that don't match a single node.
```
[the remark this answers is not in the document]
Cvelth: Done:
```cpp
/// This is a helper for deciding which statement (within a given graph) is
/// best suited for emission of a given comment.
///
/// \tparam NodeType the type of the graph node.
///         Note that it has to specialize `StatementTraits`.
///
/// A typical usage would be something like:
///
/// ```cpp
///   // 1. Make a helper object from your function and your graph:
///   CommentPlacementHelper<MyNodeType> CommentPlacement(MyModelFunction,
///                                                       MyGraph);
///   // note: `MyNodeType` must either have a `getParent()` method returning
///   //       `decltype(MyGraph) *` OR provide a specialization of 
///   //       `llvm::DomTreeNodeTraits`.
///
///   // 2. When emitting your artifact, add calls to `getComments`:
///   for (const auto &Statement : myWayOfIteratingTheGraph(MyGraph)) {
///     // . . .
///     for (const auto &Comment : CommentPlacement.getComments(Statement)) {
///       emitTheComment(Comment);
///       // note: see `PTML/Doxygen` for the standard way of emitting comments
///     }
///     // . . .
///   }
///
///   // 3. Don't forget to emit homeless comments somewhere.
///   for (const auto &Comment : CommentPlacement.getHomelessComments())
///     emitTheComment(Comment);
/// ```
///
/// With the interface out of the way, lets now discuss _how_ statements are
/// selected:
///
/// 1. For each comment-statement pair a similarity score is computed
/// (see \ref tverskyIndex) by traversing the graph.
/// 2. Then, for each comment a single _address set_ (note: not a statement yet)
/// is selected.
/// 3. Then graph is traversed again, this time looking for all the statements
/// whose location exactly matches the selected one, except for statements that
/// are dominated by an already selected statement.
/// 4. All these statements are recorded in a \ref ResultMap and will be emitted
/// once requested by a \ref getComments call.
```

What do you think?

include/revng/PTML/ContextualCommentMap.h
```diff
@@ -0,0 +1,223 @@
...
+// clang-format on
+
+/// This is a helper for deciding on the best location (within a given graph)
+/// to emit each comment from the given set at.
+///
+/// \tparam NodeType the type of the graph node. Note that it has to specialize
+///         `StatementTraits`.
+///
+/// To build an object of this type, you need to pass a model function as well
+/// as the is by point of the graph.
```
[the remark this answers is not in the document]
Cvelth: Fixed.

include/revng/PTML/ContextualCommentMap.h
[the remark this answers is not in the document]
Cvelth: I adopted blanket approach and fixed it everywhere formatting wasn't an issue. See the last commit I just pushed.

include/revng/Model/ContextualComment.h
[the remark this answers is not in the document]
Cvelth: Done.

include/revng/Yield/ControlFlow/Graph.h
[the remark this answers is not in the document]
aleclearmind: What do you mean?
I see you're taking a `BasicBlock &` argument but then using only `.ID()`.

Is there some implicit convention on how the constructor should look like?
If so, add a comment.

include/revng/PTML/ContextualCommentMap.h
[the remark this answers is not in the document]
aleclearmind: 1. Indent:
```
/// 1. For each comment-statement pair a similarity score is computed
///    (see \ref tverskyIndex) by traversing the graph.
/// 2. Then, for each comment a single _address set_ (note: not a statement yet)
///    is selected.
/// 3. Then graph is traversed again, this time looking for all the statements
///    whose location exactly matches the selected one, except for statements that
///    are dominated by an already selected statement.
/// 4. All these statements are recorded in a \ref ResultMap and will be emitted
///    once requested by a \ref getComments call.
```
2. I see no mention to dominator tree.

include/revng/Yield/ControlFlow/Graph.h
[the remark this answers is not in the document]
Cvelth: That's the current usages. The basic block itself (or rather, its `Instructions()` field) is used by the new graph introduced further in the PR.

include/revng/PTML/CommentPlacementHelper.h
```diff
@@ -0,0 +1,339 @@
...
+
+      void operator()(const DTNode &Node) {
+        llvm::SmallVector<bool> ShouldSkip;
+        ShouldSkip.resize(Scores.size());
+
+        return impl(Node, std::move(ShouldSkip));
+      }
+
+    private:
+      void impl(const DTNode &Node, llvm::SmallVector<bool> ShouldSkip) {
```
aleclearmind, line 219: I'd rename `ShouldSkip` to `AssignedInThisBranch`

include/revng/PTML/CommentPlacementHelper.h
```diff
@@ -0,0 +1,339 @@
...
+        for (const auto &Statement : Trait::getStatements(Node->getBlock())) {
+          StatementLocationType Location = Trait::getAddresses(Statement);
+          for (auto &&[Index, Comment] : llvm::enumerate(Function.Comments())) {
+            if (not Location.empty() and Location == Scores[Index].second) {
```
aleclearmind, line 223: Why do you check `Location.empty()`?

Cvelth: Because the graph we produce for the assembly has location-less entry and exit points. This is just a way to ignore those.

lib/Yield/PTML.cpp
```diff
@@ -264,22 +265,72 @@ class InstructionPrefixManager {
...
+
+  using StatementType = const MetaAddress;
+
+  // TODO: switch to `std::optional` after updating to c++26.
+  using LocationType = const llvm::SmallVector<MetaAddress, 1>;
+
+  static auto getStatements(const StatementGraph::Node *Node) {
+    constexpr static SortedVector<yield::Instruction> Empty{};
+    return (Node->Block->ID().isValid() ? Node->Block->Instructions() : Empty)
+           | std::views::transform([](auto &&R) { return R.Address(); });
```
aleclearmind, line 288: Why did you choose to have `StatementType` to be a `MetaAddress` as opposed to a `yield::Instruction`?
If we ever add additional methods to this trait it will probably make sense to have `yield::Instruction`.

lib/Yield/PTML.cpp
```diff
@@ -264,22 +265,72 @@ class InstructionPrefixManager {
...
+  for (const auto &Comment : CM.getComments(Instruction.Address())) {
+    auto &&ModelComment = ModelFunction.Comments().at(Comment.CommentIndex);
+    Result += "\n"
+              + ::ptml::statementComment(B,
+                                         ModelComment,
+                                         Binary,
+                                         Instruction.Address().toString(),
+                                         CommentIndicator,
+                                         Prefixes.totalPrefixSize(Binary),
+                                         80);
```
aleclearmind, line 326: No magic numbers

lib/Yield/PTML.cpp
```diff
@@ -416,8 +479,36 @@ std::string yield::ptml::functionAssembly(const ::ptml::MarkupBuilder &B,
...
-    Result += labeledBlock<true>(B, BasicBlock, Function, Binary, std::move(P));
+
+  const model::Function &MFunction = Binary.Functions().at(Function.Entry());
+
+  auto [G, _] = efa::buildControlFlowGraph<StatementGraph>(Function.Blocks(),
+                                                           Function.Entry(),
+                                                           Binary);
+  ::CommentPlacementHelper CM(MFunction, G);
+
+  for (const auto &BBlock : Function.Blocks())
```
aleclearmind, line 491: Add `{ ... }`

lib/Yield/PTML.cpp
```diff
@@ -431,7 +522,16 @@ std::string yield::ptml::controlFlowNode(const ::ptml::MarkupBuilder &B,
   auto Iterator = Function.Blocks().find(BasicBlock);
   revng_assert(Iterator != Function.Blocks().end());
 
-  auto Result = labeledBlock<false>(B, *Iterator, Function, Binary);
+  const model::Function &MFunction = Binary.Functions().at(Function.Entry());
+
+  // TODO: enable comments in the CFG too after the size computation rework.
```
aleclearmind, line 528: What do you mean here?

Did you consider pushing the injection of comments at `process-assembly` level?

include/revng/Support/DecompilationHelpers.h
```diff
@@ -117,3 +117,41 @@ inline const llvm::CallInst *isCallToNonIsolated(const llvm::Value *I) {
 inline bool isHelperAggregateLocalVarDecl(const llvm::Value *I) {
   return isCallToNonIsolated(I) and I->getType()->isAggregateType();
 }
+
+inline bool isStatement(const llvm::Instruction &I) {
```
aleclearmind, line 125: @pfez can you confirm that this is the One True Way™ to say if an `Instruction` is a statement or not, in the context of the old backend after `switch-to-statements`?

lib/Canonicalize/EmbedStatementComments.cpp
```diff
@@ -0,0 +1,170 @@
...
+  if (not Addresses.empty()) {
+    for (const MetaAddress &Address : Addresses)
+      Result += Address.toString() + " + ";
+    Result.resize(Result.size() - 3);
+  }
+
+  return Result;
+}
+
+struct EmbedStatementComments : public llvm::FunctionPass {
```
aleclearmind, line 76: Mmmh, you need to do things as described [in this pad](https://pad.rev.ng/s/EVGvDPdZA#).

include/revng/Model/MiscellaneousConfiguration.h
```diff
@@ -0,0 +1,32 @@
...
+
+/* TUPLE-TREE-YAML
+name: MiscellaneousConfiguration
+doc: |
+  Configuration options that don't belong anywhere else find themselves here.
+  As this list grows, they should be split into their own sub-sections.
+
+type: struct
+fields:
+  - name: EmitImpreciseCommentPositionWarning
```
aleclearmind, line 15: We need to review configuration options in there.

We need to add a configuration option only if we reasonably expect the user to actually customize it.
Configuration should never be a way to push onto the user a problem in our implementation.

include/revng/ADT/GenericGraph.h
```diff
@@ -695,13 +695,13 @@ class MutableEdgeNode
 
 public:
   EdgeView addSuccessor(MutableEdgeNode *NewSuccessor, EdgeLabel EL = {}) {
-    auto [Owner, View] = constructEdge(this, NewSuccessor, std::move(EL));
+    auto &&[Owner, View] = constructEdge(this, NewSuccessor, std::move(EL));
```
aleclearmind, line 698: I'm not so sure about all these changes. We might be moving things that previously were copied.

I'd be much more in favor of adding `const auto &` wherever possible, but this seems risky.

include/revng/PTML/CommentPlacementHelper.h
aleclearmind: I understand, but I think that case will fail the second test anyway. I don't think `Scores[...].second` is ever `empty()`.

Cvelth: > I don't think `Scores[...].second` is ever `empty()`

`Scores` is, de-facto, an array. So, before a single score is assigned for a given index, it is, in fact, empty.

lib/Yield/PTML.cpp
Cvelth: My thinking was simple: being extra explicit makes it easier to understand what's going on - this way there's no doubt that it never uses anything about the instruction but its address - because it _doesn't_ have anything in the first place.

If we ever need more information about the instruction, it's pretty easy to change the `StatementType`.

lib/Canonicalize/EmbedStatementComments.cpp
Cvelth: Ah, right, the pipe rework, there was a rev.ng hour about it at some point.

My bad!

include/revng/Model/MiscellaneousConfiguration.h
Cvelth: I agree. But it's always easier to remove a configuration option (i.e. hard code it as a `inline constexpr` variable in a header somewhere) that to add one.

My go-to is to add anything that I would want to customize and then, when we actually gather enough options for them to warrant some attention, just do a pass "demoting" some of them as a part of the inevitable hierarchy restructuring and whatnot.

include/revng/ADT/GenericGraph.h
Cvelth: I'll take another look!

include/revng/PTML/CommentPlacementHelper.h
aleclearmind: So this is to prevent assigning homeless comments to statements without addresses associated?
If so, add a comment.

lib/Yield/PTML.cpp
aleclearmind: It's not too bad, but I think it kinda spoils the point of the trait, i.e., being able to operate on an opaque data structure, which in this case is a `yield::Instruction`. At this point one could wonder why not just have a single function that returns a vector of set of addresses.

No biggie though, up to you.

lib/Yield/PTML.cpp
aleclearmind: I don't remember, can you provide some context with specific examples?

include/revng/PTML/ContextualCommentMap.h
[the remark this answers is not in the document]
Cvelth: 2.
```diff
/// 3. Then graph is traversed again, this time looking for all the statements
///    whose location exactly matches the selected one, except for statements that
-///    are dominated by an already selected statement.
+///    are reachable in a dominator tree from an already selected statement.
```
?

include/revng/ADT/GenericGraph.h
Cvelth: Wait, I think there's a misunderstanding. `auto &&` is not a _rvalue_ reference, it's a _universal_ reference.

Here's a bit of an illustration: https://godbolt.org/z/Ghd7qh6ca

TL;DR is that because of decay, moving will never occur without an explicit `std::move`, but it still lets the compiler avoid unnecessary moves (compare `by-value` to `by-rvalue-reference`).

aleclearmind: > auto && is not a rvalue reference, it's a universal reference

Ah OK.
But I get the exact same results without `&&`... Can you make an example where using `auto &&` is beneficial over just `auto`?

Cvelth: Yep, that's how how this entire conversation started: I thought `auto []` was always sufficient, but that was not the case:

https://github.com/revng/revng/pull/412#discussion_r1858212292

include/revng/PTML/CommentPlacementHelper.h
```diff
@@ -0,0 +1,339 @@
...
+
+public:
+  CommentPlacementHelper() = default;
+  CommentPlacementHelper(const model::Function &Function, auto &&Graph) {
+    if (Function.Comments().empty()) {
+      // No comments in this function, nothing to map
+      return;
+    }
+
+    // Compute a domination tree for the given graph to reduce duplicate
```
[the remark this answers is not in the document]
Cvelth: Done.

include/revng/PTML/CommentPlacementHelper.h
Cvelth: Done.

include/revng/PTML/CommentPlacementHelper.h
```diff
@@ -0,0 +1,339 @@
...
+      }
+    }
+
+    // Process homeless comments so they don't interfere with the others
+    for (auto &&[I, Comment] : llvm::enumerate(Function.Comments()))
+      if (Scores[I].first.Numerator == 0)
+        HomelessComments.emplace_back(I, false, Score{}, &Comment.Location());
+
+    // Use the tree above to build a node-to-comment-list map.
+    struct TreeVisitor {
```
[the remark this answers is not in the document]
Cvelth: Done.

include/revng/PTML/CommentPlacementHelper.h
Cvelth: Done.

include/revng/PTML/CommentPlacementHelper.h
```diff
@@ -0,0 +1,339 @@
...
+        return impl(Node, std::move(ShouldSkip));
+      }
+
+    private:
+      void impl(const DTNode &Node, llvm::SmallVector<bool> ShouldSkip) {
+        for (const auto &Statement : Trait::getStatements(Node->getBlock())) {
+          StatementLocationType Location = Trait::getAddresses(Statement);
+          for (auto &&[Index, Comment] : llvm::enumerate(Function.Comments())) {
+            if (not Location.empty() and Location == Scores[Index].second) {
+              if (not ShouldSkip[Index]) {
```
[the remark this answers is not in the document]
Cvelth: Done.

lib/Yield/PTML.cpp
Cvelth: Done.

lib/Yield/PTML.cpp
Cvelth: Done.

lib/Canonicalize/EmbedStatementComments.cpp
Cvelth: Done (I went with a pure pipe, so no passes).

lib/Yield/PTML.cpp
```diff
@@ -239,43 +241,97 @@ class InstructionPrefixManager {
...
+  using StatementType = const yield::Instruction *;
+
+  // TODO: switch to `std::optional` after updating to c++26.
+  using LocationType = const llvm::SmallVector<MetaAddress, 1>;
+
+  static auto getStatements(const StatementGraph::Node *Node) {
+    constexpr static SortedVector<yield::Instruction> Empty{};
+    return (Node->Block->ID().isValid() ? Node->Block->Instructions() : Empty)
+           | std::views::transform([](auto &&R) { return &R; });
+  }
```
aleclearmind, line 289: 1. I've no idea what this method returns at this point.
   Can you use the explicit type instead of `auto`?
2. I suspect `->Instructions()` will be empty if `not ->ID().isValid()`.

include/revng/Yield/ControlFlow/Graph.h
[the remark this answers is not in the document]
aleclearmind: > The basic block itself (or rather, its Instructions() field) is used by

Here we take a `const yield::BasicBlock &Block` but then just use `Block.ID()`.
No reference to `Block` is saved. How can some other part of the code use it?

lib/Yield/PTML.cpp
aleclearmind: Ping

lib/Yield/PTML.cpp
aleclearmind: Ping

include/revng/Model/MiscellaneousConfiguration.h
aleclearmind: I understand but I still think this is pushing the problem onto the user.
I think we should introduce such option only after the fact, once we see how things turn out.

Please remove it.

include/revng/ADT/GenericGraph.h
aleclearmind: Either we have an obvious rule to tell everyone on when to use  `auto [...]` vs `auto &&[...]`, or I'd stick with `auto [...]`.

The changeset that passes to `auto &&` seem more like a liability than beneficial. I'd take it out. Iterators are cheap to copy and after inlining the copy will probably go away.

lib/Yield/PTML.cpp
Cvelth: 1. That's defined in the trait:
```cpp
  // static /* range_with_value_type<StatementType> auto */
  // getStatements(NodeType);
  {
    StatementTraits<NodeType>::getStatements(Node)
  } -> range_with_value_type<typename StatementTraits<NodeType>::StatementType>;
```

2. Yeah, but better be safe than sorry. I'd rather we didn't assume anything about a basic block without a valid address.

include/revng/Yield/ControlFlow/Graph.h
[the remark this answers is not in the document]
Cvelth: This change is about `buildControlFlowGraph` (in `revng/EarlyFunctionAnalysis/CFGHelpers.h`), not about this callee. This callee just needs to change to account for that.

include/revng/ADT/GenericGraph.h
Cvelth: > Iterators are cheap to copy

That's the thing, this is not copying iterators! It dereferences them and then copies all the fields.

include/revng/PTML/CommentPlacementHelper.h
```diff
@@ -0,0 +1,354 @@
...
+///    once requested by a \ref getComments call.
+template<HasStatementTraits NodeType>
+class CommentPlacementHelper {
+private:
+  using Trait = StatementTraits<NodeType>;
+  using InternalNodeType = std::remove_pointer_t<NodeType>;
+
+  using StatementComment = model::StatementComment;
+  using CLT = decltype(std::declval<StatementComment>().Location());
+  using CommentLocationType = std::decay_t<CLT>;
```
aleclearmind, line 116: Let's just have a `using CommentLocation = ...` somewhere, no need for this much abstraction. Makes things less obvious to the newcomer without much benefit.

include/revng/PTML/CommentPlacementHelper.h
```diff
@@ -0,0 +1,354 @@
...
+      return operator<=>(Another) == std::strong_ordering::equal;
+    }
+  };
+
+public:
+  struct CommentView {
+    uint64_t CommentIndex;
+    bool LocationMatchesExactly;
+    Score Score;
+    const CommentLocationType *ExpectedLocation;
```
aleclearmind, line 149: Can this be `nullptr`?
In any case, always initialize primitive types in `struct`s.

include/revng/PTML/CommentPlacementHelper.h
```diff
@@ -0,0 +1,354 @@
...
+      uint64_t AdjustedRHS = Another.Numerator * this->Denominator;
+      return AdjustedLHS <=> AdjustedRHS;
+    }
+    bool operator==(const Score &Another) const {
+      return operator<=>(Another) == std::strong_ordering::equal;
+    }
+  };
+
+public:
+  struct CommentView {
```
aleclearmind, line 145: In what sense is this a *view*? Maybe `CommentPlacement` or `CommentAssignment`?

include/revng/PTML/CommentPlacementHelper.h
```diff
@@ -0,0 +1,354 @@
...
+    uint64_t CommentIndex;
```
aleclearmind, line 146: Why not a reference to the comment?

include/revng/PTML/CommentPlacementHelper.h
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    // Process homeless comments so they don't interfere with the others
+    for (auto &&[I, Comment] : llvm::enumerate(Function.Comments()))
+      if (Scores[I].first.Numerator == 0)
+        HomelessComments.emplace_back(I, false, Score{}, &Comment.Location());
+
+    TreeVisitor Visitor{ ResultMap, Function, Scores };
+    for (const auto &N : llvm::children<DTNode>(DominatorTree.getRootNode()))
+      Visitor(N);
```
aleclearmind, line 208: Can't you `Visitor(DominatorTree.getRootNode())`? 🤔 
`TreeVisitor` is a bit cumbersome, it be nice to inline it/make a lambda.

include/revng/PTML/CommentPlacementHelper.h
```diff
@@ -0,0 +1,354 @@
...
+private:
+  // Use a dominator tree to build a node-to-comment-list map.
+  using DTNode = const llvm::DomTreeNodeBase<InternalNodeType> *;
+  struct TreeVisitor {
+    std::unordered_map<StatementType, CommentList> &ResultMap;
+    const model::Function &Function;
+    const std::vector<std::pair<Score, StatementLocationType>> &Scores;
+
+    void operator()(const DTNode &Node) {
+      llvm::SmallVector<bool> AssignedInThisBranch;
```
aleclearmind, line 220: Have you considered any of the [bit storage containers](https://llvm.org/docs/ProgrammersManual.html#bit-storage-containers)?

lib/Canonicalize/EmbedStatementComments.cpp
```diff
@@ -0,0 +1,163 @@
...
+  return Result;
+}
+
+struct EmbedStatementComments {
+public:
+  static constexpr auto Name = "embed-statement-comments";
+
+private:
+  std::optional<llvm::FunctionCallee> LLVMIRComment = std::nullopt;
+  const llvm::FunctionCallee &llvmIRComment(llvm::Module &M) {
```
aleclearmind, line 80: Seems like you're trying not to define a function if it's not necessary.

For simplicity, this is not a practice we follow, let's create it in any case. Global DCE will purge it if unnecessary.

lib/Canonicalize/EmbedStatementComments.cpp
```diff
@@ -0,0 +1,163 @@
...
+      llvm::BasicBlock &EntryBlock = LLVMFunction->getEntryBlock();
+      for (llvm::BasicBlock *Node : llvm::depth_first(&EntryBlock)) {
+        using Trait = yield::StatementTraits<llvm::BasicBlock *>;
+        for (llvm::Instruction *I : Trait::getStatements(Node))
+          for (const MapT::CommentView &Comment : CM.getComments(I))
+            EmitAComment(I, Comment, addressesToString(Trait::getAddresses(I)));
+      }
+
+      for (const MapT::CommentView &Comment :
+           std::views::reverse(CM.getHomelessComments())) {
```
aleclearmind, line 152: Why reverse?

aleclearmind (changes requested):

include/revng/Yield/ControlFlow/Graph.h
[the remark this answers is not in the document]
aleclearmind: Can you be more specific? It seems that this `Node` here doesn't need `yield::BasicBlock`, but we're passing it since there's a "contract" the constructor of this `Node` needs to respect.

Which other class similar to `Node` needs the basic block?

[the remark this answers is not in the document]
Cvelth: The new one: [`9e6c549` (#412)](https://github.com/revng/revng/pull/412/commits/9e6c5494a19c7038c65bac65f7bcf0e4f9a2071d#diff-3c5c99ec28e2502fce6d3cde9223dd03fe83f6471bad02ed96f7df1f410733eeR484-R486)

include/revng/ADT/GenericGraph.h
aleclearmind: I don't understand. Let's start from scratch. Do not assume anything about the previous discussion.

1. What specific problem are we trying to solve? What is the source of the problem? Using `auto [a, b] = ...`?
2. Can you provide an instance of the problem in the code?
3. Does using `auto &&[...]` solve the problem? If so, how?
4. What is the general rule you propose to use `auto [...]` vs `auto &&[...]`.

lib/Yield/PTML.cpp
aleclearmind: 1. That's far from being clear. What is the actual return type of `getStatements` in that specific trait specialization? I think it's a range of pointer to something. Let's try to be explicit, if the types are not too ugly.
2. If you want to be safe `assert`, do not try to handle "weird" cases.

Cvelth: 1.1. The thing is that we don't really care what the type is, as long as it can be iterated upon.
1.2. It is kind of ugly, `std::ranges::transform_view</* something something*/>`. I'm not sure how nice it is to rely on the `range` internals though.
2. 👌, assert works, I think.

include/revng/PTML/CommentPlacementHelper.h
```diff
@@ -0,0 +1,354 @@
...
+///
+/// A typical usage would be something like:
+///
+/// ```cpp
+///   // 1. Make a helper object from your function and your graph:
+///   CommentPlacementHelper<MyNodeType> CommentPlacement(MyModelFunction,
+///                                                       MyGraph);
+///   // note: `MyNodeType` must either have a `getParent()` method returning
+///   //       `decltype(MyGraph) *` OR provide a specialization of
+///   //       `llvm::DomTreeNodeTraits`.
```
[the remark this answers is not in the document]
Cvelth: > Why `DomTreeNodeTraits`? I don't see it used.

That's some LLVM-specific weirdness. TL;DR this is a hint for the case if someone ever needs to use this with a graph that doesn't provide `getParent` (so not LLVM-related thing that is also not a `GenericGraph`).

In short, these are dominator tree requirements. It wasn't me who set them, I just mentioned them here so they are present somewhere.

include/revng/PTML/CommentPlacementHelper.h
Cvelth: It the same sense `string_view` is a view. As in a smart "reference" to something owned elsewhere. That being said, I have no qualms about renaming it.

include/revng/PTML/CommentPlacementHelper.h
Cvelth: I didn't go that route because it felt like a premature optimization. But yeah, let's use that.

lib/Yield/PTML.cpp
aleclearmind: 1. I agree that we don't care about what's the container we iterate on, but I don't know if what I'm iterating on is a pointer, a reference or what. That's confusing.
   Can you report the exact type? You can put `int` there to force the compiler giving you the exact type.

lib/Canonicalize/EmbedStatementComments.cpp
Cvelth: I don't remember from the top of my head, I'll look into it.

include/revng/PTML/CommentPlacementHelper.h
aleclearmind: Yeah, but if it's simple to adopt it, why not?

Up to you anyway.

lib/Yield/PTML.cpp
aleclearmind: Maybe I'm overdoing this, but I think it's worth getting a proper solution given the increased usage of ranges in our codebase.

Can we make it return `RangeOf<Type *> auto` ([example](https://godbolt.org/z/6d4jzrdT5))?
We erase the details of the actual type, but we still convey to the user what the function returns without looking at the function body.

Cvelth: Ah, wait, you can do that? For some reason I was assuming it causes problems on return values. Yeah, my bad! Sorry you had to spell it out for me. This is definitely a huge improvement!

aleclearmind: It wasn't obvious for me either this was doable, but I tried and it just worked `:D`

include/revng/Support/DecompilationHelpers.h
pfez: Yeah it is, but I have a couple of comments:
1. I'm not sure `isComment` should qualify as a statement
2. The comments at line 136,137 on stack frame declarations and stack arguments declarations are leftovers and should be removed

include/revng/Yield/ControlFlow/Graph.h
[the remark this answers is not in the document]
aleclearmind: The answer to my question was `StatementGraphNode`.
Maybe we should have a `concept` for `buildControlFlowGraph`-compatible nodes.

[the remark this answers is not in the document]
Cvelth: > The answer to my question was `StatementGraphNode`.

Sorry.

> Maybe we should have a concept for `buildControlFlowGraph`-compatible nodes.

We definitely should!

lib/Yield/PTML.cpp
Cvelth: Done, see the new `ModelConfiguration: introduce CommentLineWidth` commit

include/revng/Support/DecompilationHelpers.h
Cvelth: 1. Closing this as per the mattermost discussion.
2. Done!

include/revng/Model/MiscellaneousConfiguration.h
Cvelth: Done.

lib/Yield/PTML.cpp
Cvelth: Done!

include/revng/PTML/CommentPlacementHelper.h
```diff
@@ -0,0 +1,354 @@
...
+  : public DomTreeGraphTraitsBase<
+      llvm::DomTreeNodeBase<NodeType>,
+      typename llvm::DomTreeNodeBase<NodeType>::const_iterator> {};
+
+template<typename NodeType>
+  requires yield::HasStatementTraits<NodeType *>
+struct GraphTraits<const llvm::DomTreeNodeBase<NodeType> *>
+  : public DomTreeGraphTraitsBase<
+      const llvm::DomTreeNodeBase<NodeType>,
+      typename llvm::DomTreeNodeBase<NodeType>::const_iterator> {};
```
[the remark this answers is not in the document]
Cvelth: Done:
```
// These are necessary for llvm graph helpers (such as `llvm::depth_first`)
// to be usable on domination graphs.
//
// In this case, we don't need to implement them, since `llvm::DomTreeNodeBase`
// already does so (under a different name), we just need to point them in
// the right direction.
```

include/revng/PTML/CommentPlacementHelper.h
```diff
@@ -0,0 +1,354 @@
...
+///       emitTheComment(Comment);
+///       // note: see `PTML/Doxygen` for the standard way of emitting comments
+///     }
+///     // . . .
+///   }
+///
+///   // 3. Don't forget to emit homeless comments somewhere.
+///   for (const auto &Comment : CommentPlacement.getHomelessComments())
+///     emitTheComment(Comment);
+/// ```
```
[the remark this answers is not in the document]
Cvelth: Done!

include/revng/PTML/CommentPlacementHelper.h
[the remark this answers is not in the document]
Cvelth: Ping.

include/revng/PTML/CommentPlacementHelper.h
```diff
@@ -0,0 +1,354 @@
...
+/// ```cpp
+///   // 1. Make a helper object from your function and your graph:
+///   CommentPlacementHelper<MyNodeType> CommentPlacement(MyModelFunction,
+///                                                       MyGraph);
+///   // note: `MyNodeType` must either have a `getParent()` method returning
+///   //       `decltype(MyGraph) *` OR provide a specialization of
+///   //       `llvm::DomTreeNodeTraits`.
+///
+///   // 2. When emitting your artifact, add calls to `getComments`:
+///   for (const auto &Statement : myWayOfIteratingTheGraph(MyGraph)) {
```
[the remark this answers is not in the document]
Cvelth: Added.

include/revng/PTML/CommentPlacementHelper.h
```diff
@@ -0,0 +1,354 @@
...
+///     }
+///     // . . .
+///   }
+///
+///   // 3. Don't forget to emit homeless comments somewhere.
+///   for (const auto &Comment : CommentPlacement.getHomelessComments())
+///     emitTheComment(Comment);
+/// ```
+///
+/// With the interface out of the way, lets now discuss _how_ statements are
```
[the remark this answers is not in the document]
Cvelth: Replaced here and everywhere else.

include/revng/PTML/CommentPlacementHelper.h
```diff
@@ -0,0 +1,354 @@
...
+///     // . . .
+///     for (const auto &Comment : CommentPlacement.getComments(Statement)) {
+///       emitTheComment(Comment);
+///       // note: see `PTML/Doxygen` for the standard way of emitting comments
+///     }
+///     // . . .
+///   }
+///
+///   // 3. Don't forget to emit homeless comments somewhere.
+///   for (const auto &Comment : CommentPlacement.getHomelessComments())
```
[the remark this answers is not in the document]
Cvelth: Done:
```cpp
/// // 3. Don't forget to emit homeless comments somewhere.
/// //    These are the comments that do not have even a single match among
/// //    all the statements in the provided graph.
```

include/revng/PTML/CommentPlacementHelper.h
```diff
@@ -0,0 +1,354 @@
...
+/// selected:
+///
+/// 1. For each comment-statement pair a similarity score is computed
+///    (see \ref tverskyIndex) by traversing the graph.
+/// 2. Then, for each comment a single _address set_ (note: not a statement yet)
+///    is selected.
+/// 3. Then graph is traversed again, this time looking for all the statements
+///    whose location exactly matches the selected one, except for statements
+///    that are reachable in a dominator tree from an already selected
+///    statement.
```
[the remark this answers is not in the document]
Cvelth: Done:
```/// With the interface out of the way, lets now discuss *how* statements are
/// selected:
///
/// 1. For each comment-statement pair a similarity score is computed
///    (see \ref tverskyIndex) based on the similarity of the set of addresses
///    recorded in the comment to the set of addresses associated to
///    the statement.
/// 2. Then, for each comment a single *address set* (note: not a statement yet)
///    is selected.
/// 3. Then graph is traversed again, this time looking for all the statements
///    whose location exactly matches the selected one.
///    When there's an exact match, the comment is assigned to that statement.
///    This happens *unless* the current statement is dominated by another
///    statement that is already associated to the comment currently being
///    considered. This prevents emitting the same comment multiple times
///    in cases where it's obviously not beneficial.
/// 4. All these statements are recorded in a \ref ResultMap and will be emitted
///    once requested by a \ref getComments call.
///
/// For some specific examples on how this helper works, see the corresponding
/// unit tests, but let me also provide a trivial example. Let's say we have
/// a *really* sequential graph with three nodes: `(A) -> (B) -> (C)`.
/// And let's say each node has a single statement with addresses attached:
///   - A: { 0x001:Generic64 }
///   - B: { 0x001:Generic64, 0x002:Generic64 }
///   - C: { 0x001:Generic64, 0x002:Generic64 }
///
/// And let's say we're trying to decide where to emit a comment with
/// `{ 0x001:Generic64, 0x002:Generic64, 0x003:Generic64 }`
///
/// After running step 1 (above) on it, the resulting scores for locations(!)
/// are as follows (note that there are only two locations):
/// - `{ 0x001:Generic64 }`: 1/3
/// - `{ 0x001:Generic64, 0x002:Generic64 }`: 2/3
///
/// So the best one is selected. And there are two statements that correspond
/// to it: `B` and `C`.
///
/// If these two statement were not related, the comment would end up duplicated
/// BUT based on the graph, the only way to reach `C` is through `B`! As such,
/// there's not much downside to supressing the `C` comment and only emitting
/// one before `B`.
```

include/revng/PTML/CommentPlacementHelper.h
```diff
@@ -0,0 +1,354 @@
...
+      // `b*d`, which means we can just compare `(a*d)/(b*d)` and `(b*c)/(b*d)`.
+      // And since "adjusted" denominators are the same, we can also omit them
+      // and just compare the numerators.
+      uint64_t AdjustedLHS = this->Numerator * Another.Denominator;
+      uint64_t AdjustedRHS = Another.Numerator * this->Denominator;
+      return AdjustedLHS <=> AdjustedRHS;
+    }
+    bool operator==(const Score &Another) const {
+      return operator<=>(Another) == std::strong_ordering::equal;
+    }
```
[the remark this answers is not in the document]
Cvelth: Done, good catch!

include/revng/PTML/CommentPlacementHelper.h
Cvelth: Actually we already do: `model::StatementComment::TypeOfLocation`.

My bad on forgetting it exists.

include/revng/PTML/CommentPlacementHelper.h
Cvelth: Done.

include/revng/PTML/CommentPlacementHelper.h
Cvelth: Done (`CommentAssignment`).

include/revng/PTML/CommentPlacementHelper.h
Cvelth: Because it's trivial to go `index -> reference` (`MyFunction.Comments().at(MyIndex)`), and the other way is impossible, since `Comments` is a function and we place no limitation on duplication and whatnot.

As for why we need indexes, I feel like emitting `@comment(i64 INDEX, ...)` is better design than embedding model data in, after all model is constant throughout the entire pipeline and is always available.

Bonus point, we don't touch comment body until the very end - so editing the comment doesn't really invalidate anything (unless I messed up somewhere, but that would be a separate problem).

include/revng/PTML/CommentPlacementHelper.h
Cvelth: Actually this was a bug! This extra loop was never necessary.

include/revng/PTML/CommentPlacementHelper.h
Cvelth: Done!

lib/Canonicalize/EmbedStatementComments.cpp
Cvelth: Ah, this was an artifact of a _very_ old design - because of how homeless comments were gathered initially they needed reversed. Not needed anymore, Good catch!

Cvelth:
@aleclearmind, I mentioned some commits by name in several threads, but you should probably just give a fresh round of reviews to everything starting from `PTML: introduce a way to disable comment warnings` (including it), delta the last `auto &&[]` commit, there's a separate thread for it.

include/revng/ADT/GenericGraph.h
Cvelth: First of all, this commit should not block this PR - it's not related, so no rush solving this to get comments in. (and the commit is low effort anyway, I have both the regex I ran to produce it and a link to the single point that needs manual intervention noted down somewhere, so reproducing this commit at any stage is trivial).

That being said:

1. The simplest example would be something like `std::map<CheapKey, ExpensiveObject>`. If it's iterated as
```cpp
for (auto [Key, Object] : MyMap)
```
A copy of the `ExpensiveObject` is created at every iteration.

2. Can you provide an instance of the problem in the code?

Here's the original code that started this discussion:
```cpp
    for (auto [Index, Comment] : llvm::enumerate(Function.Comments())) {
```
(from [here](https://github.com/revng/revng/pull/412#discussion_r1854123014))

It's not critical, but not nice. There are most likely others, I haven't checked manually - since automagic replacement did the trick.

3. With `auto &&` there are never _unintended_ copies. So, for example, if the original object is a reference, the variable is also a reference to it, but if it was returned by value - it's copied.

A good example is (let's stick with `std::map` for now) `std::map<A, B>::iterator::operator*` that returns `std::pair<A &, B &>` - so a copy doesn't happen, but if we look at `std::map<A, B>::try_emplace` that returns `std::pair<std::map<A, B>::iterator, bool>` (both by value) - we get two copies of these trivial objects.

Of course it works in a mixed scenario: `std::pair<Expensive &, Cheap> my_function()` would do the right thing (tm) if called as
```
auto &&[E, C] = my_function();
```
but will do an expensive unintended copy with
```
auto [E, C] = my_function();
```

4. Before answering, let me summarize what I mentioned above:

`auto &&[a, b]` - use when whatever the function (including stuff like `operator*`, you get the point) is returning works for you (sensible default).
`const auto &[a, b]` - when, for some reason you have to _enforce_ constness (as in, function returns non-const stuff but you need to explicitly mark it as const (flashbacks to the model that does different things based on whether it's const or not)).
`auto &[a, b]` - not much reason to use this one if not to signify that you're planning on mutating the original values - and want an error the signature of the function ever changes (say, it starts returning something by copy).
`auto [a, b]` - when you want to _force_ a copy.

As you have probably already noticed, it's 100% consistent with usage of `auto` in templates (both arguments and variables). So, do what you would have done in a template still applies.

(template is significant because of difference between a universal reference and a r-value reference in non-templated context, I _really_ wish they adopted different syntax for the two, but hey, this is the C++ we ended up with).

P. S. Github is getting really laggy again, We should really get this PR in already!

include/revng/Yield/ControlFlow/Graph.h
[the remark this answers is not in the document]
aleclearmind: OK. Maybe you can check the concept as a static assertion in the body of the function then?

lib/Yield/PTML.cpp
aleclearmind: Great

include/revng/PTML/CommentPlacementHelper.h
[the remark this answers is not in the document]
aleclearmind: OK, but it's "dominator tree", not "domination tree"

include/revng/PTML/CommentPlacementHelper.h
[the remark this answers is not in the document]
aleclearmind: I thought that in order to compute the `DomaintorTree` you *have to* to have `getParent()`.
But I now see that this is not mandatory and you can implement `DomTreeNodeTraits` and have something else other than `getParent()`.

Reformulate the comment as: "We need to compute the dominator tree. In order to do this you either need to implement the `DomTreeNodeTraits` trait or have your node have a `getParent()` method to use the default implementation".

include/revng/PTML/CommentPlacementHelper.h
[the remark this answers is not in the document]
aleclearmind: Works.

include/revng/PTML/CommentPlacementHelper.h
aleclearmind: Add a comment. This was not obvious to me.

lib/Canonicalize/EmbedStatementComments.cpp
aleclearmind: > The reason this is in a separate helper [...]

When I mentioned "defining a function" I didn't mean the`llvmIRComment` function, but the `llvm::Function` you're creating (`comment`).

> As for the std::optional<llvm::FunctionCallee> thing

I'm saying: let's create it unconditionally before the first call to `llvmIRComment`.
There's no real benefit in not creating it. If we create it and then not use it, global dce will garbage collect it.

lib/PTML/Doxygen.cpp
```diff
@@ -585,8 +585,11 @@ std::string ptml::functionComment(const ::ptml::MarkupBuilder &PTML,
   return Builder.emit(std::move(Result));
 }
 
+static Logger ImprecisePositionWarning("imprecise-comment-position-warnings");
```
aleclearmind, line 588: Too specific, I'd just go for `comment-emission`

include/revng/Model/Configuration.h
```diff
@@ -17,6 +17,9 @@ type: struct
   - name: Naming
     type: NamingConfiguration
     optional: true
+  - name: Miscellaneous
+    type: MiscellaneousConfiguration
+    optional: true
```
aleclearmind, line 22: I'd flatten `Miscellaneous` in here, we can pack things into a sub-struct in the future if there's a couple of options that can be grouped together.

tests/unit/CMakeLists.txt
```diff
@@ -608,3 +608,25 @@ add_test(NAME test_pointer_array_emission COMMAND test_pointer_array_emission)
 add_library(test_scopegraph SHARED "${SRC}/ScopeGraphLoggerPass.cpp")
 target_link_libraries(test_scopegraph revngScopeGraphUtils revngSupport
                       ${LLVM_LIBRARIES})
+
+#
+# test_statement_comment_placement
```
aleclearmind, line 613: Nice

include/revng/Yield/ControlFlow/Graph.h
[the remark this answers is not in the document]
Cvelth: Done!

include/revng/ADT/GenericGraph.h
Cvelth: Done!

include/revng/PTML/CommentPlacementHelper.h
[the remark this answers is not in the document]
Cvelth: I ended up going with
```cpp
/// // note: We need to be able to compute the dominator tree. In order to allow
/// //       that, you need to provide a `llvm::DomTreeNodeTraits<MyNodeType>`
/// //       trait OR `MyNodeType::getParent()` method.
```

lib/PTML/Doxygen.cpp
Cvelth: I went with `statement-comment-emission`.

include/revng/Model/Configuration.h
Cvelth: Done!

PR #437, opened 2025-01-10
Introduce `DotGraphTraits` for `ScopeGraph`

andrealinux1:
With this PR we introduce the `DOTGraphTraits` for the `ScopeGraph`, so that we are now able to obtain the `dot` of it for easier debugging and visualization.

We also integrate a fix in order to make the constructor of the `ScopeGraph` more robust with respect to temporary references going out of scope.


include/revng/RestructureCFG/ScopeGraphGraphTraits.h
```diff
@@ -306,6 +307,90 @@ struct llvm::GraphTraits<llvm::Inverse<Scope<const llvm::Function *>>>
...
+getEdgeAttributes(const llvm::BasicBlock *N,
+                  llvm::GraphTraits<
+                    Scope<llvm::BasicBlock *>>::ChildIteratorType EI,
+                  const Scope<llvm::Function *> G) {
+
+  // If we are printing a block which has a `scope_closer` edge, which by
+  // construction must be unique, if present, and the last in the
+  // `MaterializedRange` representing the successors, we print such edge as
+  // dashed
+  if (getScopeCloserTarget(N) and EI.atLastElement()) {
```
pfez, line 341: I'm not excited by `atLastElement`. Can't you check if the target block of EI is the scope closer target of `N` instead?

If so, I'd also drop the `atLastElement` method.

andrealinux1: I fixed this with the solution we discussed IRL, not the one proposed in this comment:
https://github.com/revng/revng/blob/10dcf9302e9e2d88a344b05b9764d19909fd31a9/include/revng/RestructureCFG/ScopeGraphGraphTraits.h#L342

Dropped the `atLastElement` method from the `EagerMaterializationRangeIterator`

pfez:
Merged in 255e7ac30b336a1be5743fe426345f9ad2a3de06

Thanks!

PR #439, opened 2025-01-17
`master` promotion 

mrjackv:


python/revng/internal/cli/_commands/mass_testing/configure.py
```diff
@@ -24,6 +24,7 @@ def register_arguments(self, parser: argparse.ArgumentParser):
         parser.description = "Collect binary files and run test-configure"
         parser.add_argument("--meta", help="Metadata definition")
         parser.add_argument("--seed", help="Seed to use in the RNG for shuffling")
+        parser.add_argument("--sample", help="Sample the inputs to the specified amount")
```
aleclearmind, line 27: Is this for debugging purposes?

python/revng/internal/cli/_commands/mass_testing/configure.py
```diff
@@ -40,9 +40,8 @@ def run(self, options: Options):
 
         # Shuffle the input files, this avoids having a limited variety of
         # test runs for a partial run
-        if args.seed is not None:
-            seed = args.seed
-        else:
+        seed = os.environ.get("MASS_TESTING_CONFIGURE_SEED", args.seed)
```
aleclearmind, line 42: What's the use for these?

python/revng/internal/cli/_commands/mass_testing/run.py
```diff
@@ -27,16 +27,24 @@ def register_arguments(self, parser: argparse.ArgumentParser):
...
             cpu_count = args.jobs
         else:
             cpu_count = len(os.sched_getaffinity(0))
+            memory_limits = [
+                x["memory_limit"] for x in meta_data["configurations"] if "memory_limit" in x
+            ]
+            if len(memory_limits) > 0:
+                max_ram = max(memory_limits)
+                mem_bytes = os.sysconf("SC_PAGE_SIZE") * os.sysconf("SC_PHYS_PAGES")
+                cpu_count = min(int((mem_bytes * 0.9) // max_ram), cpu_count)
```
aleclearmind, line 46: Comment what you're doing and why

python/revng/internal/cli/_commands/mass_testing/configure.py
mrjackv: Yes, I needed to check if `mass-testing` was working but didn't want to wait for 2 days. I've left it in the PR since it might be useful in the future.

python/revng/internal/cli/_commands/mass_testing/configure.py
mrjackv: That's the seed passed to the PRNG, if you use the same seed and the same inputs you'll get the same files (if you use `--sample`) or the same order (due to `shuffle`)

python/revng/internal/cli/_commands/mass_testing/run.py
mrjackv: Done

python/revng/internal/cli/_commands/mass_testing/configure.py
aleclearmind: Usually arguments take precedence over environment variables, here is the other way around.

Is there a reason to have a non-hardcoded seed?
I understand we want determinism. Do we ever want a *different* set of samples?

Even if we do, setting the seed to `0` might saving us from having to specifiy a seed via environment variables.

mrjackv: > Usually arguments take precedence over environment variables, here is the other way around.

Yes, but since these "variables" need to be overridden from outside orchestra it's needed for the environment variable counterpart to have precedence over the command line one

> Is there a reason to have a non-hardcoded seed?

Since the ordering can "hide" bugs (e.g. a crash should happen, but it times out before that) it's beneficial for runs to use a random seed

> Even if we do, setting the seed to 0 might saving us from having to specifiy a seed via environment variables.

That's only needed in "controlled" situations, e.g. when running regression. In general I'm not sure we want that, having variety means that some different bugs will show up (which IMHO is good)

aleclearmind: I think no one is going to use this other than the mass testing itself.

Let's hardcode a seed and call it a day, if you really want you can use `--seed=$RANDOM` to test more stuff.

mrjackv: > if you really want you can use --seed=$RANDOM to test more stuff.

Having only the command-line options is basically useless, if anything then only the env should exist.

> Let's hardcode a seed and call it a day

I don't see any benefit in that, if there's an ordering bug it could get hidden. Or, put another way, there isn't any benefit (from the POV of mass-testing) to **not** shuffle things differently each run.
Note: the seed used for the run is recorded in `meta.yml`, so in any case the run **can** be "reproduced" if needed.

aleclearmind: OK.

> Usually arguments take precedence over environment variables, here is the other way around.

Please do this though.

mrjackv: Dropped the command-line options.

aleclearmind:
Merged in 0be8ef9. Thanks.

PR #440, opened 2025-01-20
Refactor creation of local variables for new decompilation pipeline

pfez:


share/revng/pipelines/revng-pipelines.yml
```diff
@@ -660,6 +660,61 @@ Branches:
...
+            UsedContainers: [module.bc.zstd]
+            Passes:
+              - remove-llvmassume-calls
+              - dce
+              - exit-ssa
+              - switch-to-statements
+        Artifacts:
+          Container: module.bc.zstd
+          Kind: stack-accesses-segregated
+          SingleTargetFilename: stack_access_segregate.ll
```
aleclearmind, line 713: Do not reuse `SingleTargetFilename`.

share/revng/pipelines/revng-pipelines.yml
```diff
@@ -660,6 +660,61 @@ Branches:
...
+          Docs: |-
+            This artifact contains an LLVM function for each function in the
+            input program.
+            TODO: fill this up
```
aleclearmind, line 717: Do not leave TODOs for documentation, this is user facing!

share/revng/test/configuration/revng/decompilation.yml
```diff
@@ -11,6 +11,14 @@ commands:
     suffix: /
     command: |-
       revng artifact --resume "$OUTPUT" --analyze decompile-to-single-file "$INPUT" -o /dev/null
+  - # Run new decompilation pipeline
+    type: revng.new-decompilation-pipeline
```
aleclearmind, line 15: I think we should settle with either the `legacy-` prefix for the old stuff or `new-` for the new stuff.
If we don't do this now, we'll have some and some and it will be a mess.

aleclearmind (changes requested):

share/revng/pipelines/revng-pipelines.yml
```diff
@@ -660,6 +660,61 @@ Branches:
...
+    # New clift-based decompilation pipeline
+  - From: detect-stack-size
+    Steps:
+      - Name: segregate-stack-accesses
+        Pipes:
+          - Type: llvm-pipe
+            UsedContainers: [module.bc.zstd]
+            Passes:
+              - hoist-struct-phis
+              - segregate-stack-accesses
```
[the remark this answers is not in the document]
pfez: I did this to mimic what `legacy-segregate-stack-accesses` does. So I think I should fix the old one too.

In both cases, I'd go for squashing `{legacy-,}segregate-stack-accesses` with the subsequent `{legacy-,}late-optimize`, keeping the name of the former and basically dropping the latter.
This saves a full roundtrip from disk, compared to the current pipeline in `develop`.
Ok?

As an aside. The fact that the container is saved to disk at the end of a step even if it doesn't have an associated artifact seems to leave a gap open for suboptimal specifications of pipelines (like I did). Why even leave the option of having steps without associated artifacts? If that's forbidden people wouldn't be able to do the same mistake as I did.

share/revng/pipelines/revng-pipelines.yml
pfez: It's not reused. It's the only one.

share/revng/test/configuration/revng/decompilation.yml
pfez: Fair enough. This isn't even the full new decompilation pipeline.

I'm switching the old tests to `legacy.

share/revng/pipelines/revng-pipelines.yml
```diff
@@ -622,7 +622,7 @@ Branches:
         Artifacts:
           Container: module.bc.zstd
           Kind: stack-accesses-segregated
-          SingleTargetFilename: stack_access_segregate.ll
+          SingleTargetFilename: legacy_stack_access_segregate.ll
```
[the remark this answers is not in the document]
pfez: Done!

share/revng/test/configuration/revng/decompilation.yml
pfez: Done. You can check out the last three commits of the patchset.

This also made me realize that there were a few additional end-to-end tests that could be added. I'm working on those now.

share/revng/pipelines/revng-pipelines.yml
pfez: Done!

share/revng/pipelines/revng-pipelines.yml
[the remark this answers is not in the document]
aleclearmind: > I did this to mimic what `legacy-segregate-stack-accesses` does. So I think I should fix the old one too.

Yup.

> In both cases, I'd go for squashing `{legacy-,}segregate-stack-accesses` with the subsequent `{legacy-,}late-optimize`, keeping the name of the former and basically dropping the latter. This saves a full roundtrip from disk, compared to the current pipeline in `develop`. Ok?

Yup.

> As an aside. The fact that the container is saved to disk at the end of a step even if it doesn't have an associated artifact seems to leave a gap open for suboptimal specifications of pipelines (like I did). Why even leave the option of having steps without associated artifacts? If that's forbidden people wouldn't be able to do the same mistake as I did.

This will be a non-problem in the new pipeline. Let's ignore the problem.

share/revng/pipelines/revng-pipelines.yml
aleclearmind: OK, then give it the name of the step.

share/revng/test/configuration/revng/end-to-end.yml
```diff
@@ -120,6 +120,20 @@ commands:
...
+  - type: revng.decompiled-c
+    from:
+      - type: revng-qa.compiled
+        filter: one-per-architecture
+      - type: revng.analyzed-model
+    suffix: /
+    command: |-
+      WORKDIR="$OUTPUT/check-decompiled-c";
+      mkdir -p "$$WORKDIR";
+      revng artifact --model "$INPUT2" --resume "$OUTPUT" switch-to-statements "$INPUT1"
```
aleclearmind, line 135: `WORKDIR` is unused. 🤔

share/revng/pipelines/revng-pipelines.yml
```diff
@@ -660,6 +660,65 @@ Branches:
           SingleTargetFilename: mlir-llvm-dialect-with-clift-types.mlir
           Docs: |-
             A test artifact to import types into a Clift module.
+    # New clift-based decompilation pipeline
```
aleclearmind, line 663: I'd settle for "clift-based" vs "legacy" decompilation pipeline.
Let's avoid new, otherwise it will stick around forever.

share/revng/test/configuration/revng/decompilation.yml
```diff
@@ -3,26 +3,34 @@
...
-    type: revng.decompile-to-single-file
+  - # Run legacy decompilation pipeline up to decompile-to-single-file
+    type: revng.legacy.decompile-to-single-file
     from:
       - type: revng-qa.compiled-with-debug-info
         filter: for-decompilation
     suffix: /
     command: |-
       revng artifact --resume "$OUTPUT" --analyze decompile-to-single-file "$INPUT" -o /dev/null
+  - # Run initial portion of the new decompilation pipeline up to switch-to-statements
```
aleclearmind, line 14: `s/new/clift-based/`

share/revng/test/configuration/revng/end-to-end.yml
pfez: Leftover from copy-paste

share/revng/pipelines/revng-pipelines.yml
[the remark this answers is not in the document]
pfez: Done. I'll close the conversation when I push the branch.

share/revng/test/configuration/revng/end-to-end.yml
pfez: Ah no, I actually changed the test to put the result in `WORKDIR`, which is helpful for debug. I pushed for you to see. Tell me if you want to drop it, or leave it.

share/revng/test/configuration/revng/decompilation.yml
pfez: The tests now pass. There's not much to say left here. Closing the conversation.

share/revng/test/configuration/revng/end-to-end.yml
aleclearmind: What stops you from using `$OUTPUT/prepare-for-clift`? No need to save it again.

pfez: Fair enough. Dropped this.

PR #441, opened 2025-01-28
Fixed for UI QA iteration of 2025-01-16

aleclearmind:


lib/InitModelTypes/InitModelTypes.cpp
```diff
@@ -95,13 +97,23 @@ static RecursiveCoroutine<bool> addOperandType(const llvm::Value *Operand,
         if (OperandType->isPointer()) {
           // If the operand is already a pointer, just forward it
           TypeMap.insert({ Operand, OperandType.copy() });
-
         } else if (not PointersOnly) {
           auto PS = model::Architecture::getPointerSize(Model.Architecture());
           TypeMap.insert({ Operand, model::PrimitiveType::makeGeneric(PS) });
+        } else {
+          // WIP: ??
```
aleclearmind, line 104: @pfez I will drop this. OK?

lib/InitModelTypes/InitModelTypes.cpp
```diff
@@ -95,13 +97,23 @@ static RecursiveCoroutine<bool> addOperandType(const llvm::Value *Operand,
...
         }
         rc_return true;
       }
+    } else if (Expr->getOpcode() == Instruction::PtrToInt) {
+      auto It = TypeMap.find(Expr->getOperand(0));
+      if (It != TypeMap.end()) {
+        const model::UpcastableType &OperandType = It->second;
+        // WIP: PointersOnly?
```
aleclearmind, line 112: @pfez This is not a pointer, shall I insert it only `if (not PointersOnly)`?

lib/InitModelTypes/InitModelTypes.cpp
```diff
@@ -140,6 +152,27 @@ static RecursiveCoroutine<bool> addOperandType(const llvm::Value *Operand,
     auto PtrSize = model::Architecture::getPointerSize(Model.Architecture());
     TypeMap.insert({ Operand, model::PrimitiveType::makeGeneric(PtrSize) });
     rc_return true;
+  } else if (auto *ReferencedFunction = dyn_cast<llvm::Function>(Operand)) {
+    // WIP: PointersOnly
+    if (PointersOnly)
+      rc_return true;
```
aleclearmind, line 158: @pfez I will drop this. OK?

lib/InitModelTypes/InitModelTypes.cpp
pfez: Yes you can drop this.

lib/InitModelTypes/InitModelTypes.cpp
pfez: Forget about all this handling. The handling of `PtrToInt` should be exactly the same as `IntToPtr` AFAICT.

Basically if the operand type is a pointer, you add it to the map regardless.

Otherwise, you `makeGeneric` and add it to the map, but only when `not PointersOnly`, because if `PointersOnly` is true you would be adding a non-pointer to the map which is wrong.

lib/InitModelTypes/InitModelTypes.cpp
pfez: Yes.

PR #448, opened 2025-02-24
 UI QA 2025-02-11

mrjackv:


python/revng/internal/cli/_commands/fetch_debuginfo/common.py
```diff
@@ -26,10 +27,18 @@ def log_warning(self, message):
 log_warning = logger.log_warning
 
 
+session = requests.Session()
+adapter = requests.adapters.HTTPAdapter(
+    max_retries=Retry(total=5, backoff_factor=1.0, status_forcelist=[408, 429, 500, 502, 503, 504])
```
aleclearmind, line 32: How did you put together the HTTP status list? What are they?

At least add a comment.

lib/Pipeline/Step.cpp
```diff
@@ -494,6 +494,8 @@ Step::storeInvalidationMetadata(const revng::DirectoryPath &Path) const {
 
     ZstdCompressedOstream OS(File->get()->os(), 5);
     ::serialize(OS, ToStore);
+    OS.flush();
```
aleclearmind, line 497: > Without flushing the output, the invalidation metadata could end up being half-written

How? In case of a crash?

The next few lines don't seem to be crash-prone.

python/revng/internal/daemon/event_manager.py
```diff
@@ -54,7 +54,7 @@ def run(self):
             # related to speeding up the pipeline manager from saving (e.g. detecting dirtiness in
             # containers) need to be implemented before this can be dropped
             if self.last_save + self._interval < time.time() and (
-                self.last_event is not None and self.last_event < self.last_save
+                self.last_event is not None and self.last_event > self.last_save
```
aleclearmind, line 57: 🤦🏿

aleclearmind (changes requested):

lib/Pipeline/Step.cpp
mrjackv: > How? In case of a crash?

No, it's a bit subtle, basically the order of events is:
1. Get `WritableFile`, which has a `llvm::raw_fd_ostream`
2. Wrap ostream in `ZstdCompressedOstream`
3. Serialize stuff on the zstdOstream
4. Run `WritableFile.commit`. It calls the `raw_fd_ostream.flush` but does not flush the wrapping zstdOstream, so the last block of zstd data is never written to the file.
5. The storage stuff takes the partially written file and does its business, unaware that the file is not complete

python/revng/internal/cli/_commands/fetch_debuginfo/common.py
mrjackv: Added comments to the codes
Basically I compiled the list by looking at the list of error codes and adding the ones that "made sense", e.g. related to timeouts, throttling or the server being in some way unavailable.

lib/Pipeline/Step.cpp
aleclearmind: OK, usually in these cases I wrap `OS` into a scope just to force triggering the destructor.

It's not necessarily better, just a pattern we already adopt around the code base.

mrjackv: Switched to the scope.

PR #459, opened 2025-03-11
UI QA 2025-03-04

mrjackv:


include/revng/Support/Error.h
```diff
@@ -25,6 +25,7 @@ inline void cantFail(std::error_code EC) {
 }
 
 template<std::ranges::range T>
+  requires std::is_same_v<typename T::value_type, llvm::Error>
```
aleclearmind, line 28: Use `RangeOf`.

lib/Support/ProgramRunner.cpp
```diff
@@ -75,6 +75,7 @@ int ProgramRunner::run(llvm::StringRef ProgramName,
   Log << DoLog;
 
   int ExitCode = llvm::sys::ExecuteAndWait(StringRefs[0], StringRefs);
+  Log << "Program exited with code " << ExitCode << "\n";
```
aleclearmind, line 78: This is illegal!

Either use `revng_log(Log, ...)` or...

```c
if (Log.isEnabled()) {
  Log << ... << DoLog;
}
```
If you don't use `revng_log` you have to finalize with `DoLog`. If you don't, the stuff you're emitting will be flushed in the next message.

Also, the part of the code above, would benefit to be split and have the logging part be wrapped in a `if (Log.isEnabled())`. Not urgent, but let's not add more un-guarded usages of `Log`.

lib/PipelineC/Tracing/Runner.cpp
```diff
@@ -489,6 +489,13 @@ llvm::Error Trace::run(const revng::tracing::RunTraceOptions Options) const {
...
 
+    TraceRunnerLogger << "Running #" << Command.ID << ": " << Command.Name;
+    for (const revng::tracing::Argument &Argument : Arguments) {
+      TraceRunnerLogger << ", " << Argument.toString();
+    }
+    TraceRunnerLogger << "\n";
+    TraceRunnerLogger.flush();
+
     CommandHandler[Command.Name](Context, Arguments, Command.Result);
   }
```
aleclearmind, line 509: The `;` below should be removed

lib/PipelineC/Tracing/Runner.cpp
```diff
@@ -489,6 +489,13 @@ llvm::Error Trace::run(const revng::tracing::RunTraceOptions Options) const {
       continue;
     }
 
+    TraceRunnerLogger << "Running #" << Command.ID << ": " << Command.Name;
+    for (const revng::tracing::Argument &Argument : Arguments) {
+      TraceRunnerLogger << ", " << Argument.toString();
+    }
+    TraceRunnerLogger << "\n";
+    TraceRunnerLogger.flush();
```
aleclearmind, line 497: 1. Do not put in a logger `\n`, unless you want a multiline message.
2. Always terminate the message with `<< DoLog`.
3. Always wrap writing to a logger in a `if (Log.isEnabled())` (unless you use `revng_log`).

include/revng/PipelineC/Tracing/Trace.h
```diff
@@ -138,6 +138,8 @@ struct RunTraceOptions {
   // Instead of using a temporary directory, the first invocation will use
   // these directory instead and subsequent ones will abort
   std::string ResumeDirectory;
+  // Is enabled, the manager will save after every analysis is run
```
aleclearmind, line 141: 1. `s/Is/If/`
2. When you document a field in a `struct` use Doxygen (`///`)

lib/PipelineC/Tracing/Runner.cpp
```diff
@@ -497,6 +497,14 @@ llvm::Error Trace::run(const revng::tracing::RunTraceOptions Options) const {
     TraceRunnerLogger.flush();
 
     CommandHandler[Command.Name](Context, Arguments, Command.Result);
+
+    // If an analysis was just run and the SaveAfterAnalyses is enabled, save
+    if ((Command.Name == "rp_manager_run_analysis"
+         || Command.Name == "rp_manager_run_analyses_list")
+        && Options.SaveAfterAnalyses) {
```
aleclearmind, line 504: `s/||/or/`, `s/&&/and/`

lib/PipelineC/Tracing/Runner.cpp
```diff
@@ -497,6 +497,14 @@ llvm::Error Trace::run(const revng::tracing::RunTraceOptions Options) const {
...
+      std::vector<revng::tracing::Argument> SaveArguments{ Arguments[0] };
+      CommandHandler["rp_manager_save"](Context, SaveArguments, "true");
```
aleclearmind, line 506: 1. The following should work:

   ```c++
   CommandHandler["rp_manager_save"](Context, { Arguments[0] }, "true");
   ```
2. Does `rp_manager_save` saves everything or only the model?
   Ideally we'd just like to save the model.

include/revng/Support/CommandLine.h
```diff
@@ -9,6 +9,10 @@
 #include "llvm/Support/CommandLine.h"
 
 extern llvm::cl::OptionCategory MainCategory;
+
+// This is the path of the input binary. CLI tools should populate this on
+// every invocations. Non-CLI users of this variable need to account for it
+// being possibly empty.
```
aleclearmind, line 15: 1. Use Doxygen
2. > Non-CLI users of this variable need to account for it being possibly empty

   Users should not be aware whether they are being used by the CLI or not, they should just assume this might be empty.

lib/Model/Importer/Binary/ELFImporter.cpp
```diff
@@ -445,7 +445,9 @@ void ELFImporter<T, HasAddend>::findMissingTypes(object::ELFFile<T> &TheELF,
   unsigned MaximumRecursionDepth = 1;
 
   LDDTree Dependencies;
-  lddtree(Dependencies, InputPath, MaximumRecursionDepth);
+  std::string BinaryPath = InputPath.empty() ? TheBinary.getFileName().str() :
+                                               InputPath;
```
aleclearmind, line 449: 1. Use `const std::string &` if you can, save one copy.
2. We also need to do this for `PDBImporter`.

aleclearmind (changes requested):

lib/PipelineC/Tracing/Runner.cpp
mrjackv: 1. From line 492 to line 495 it's a single long line, there's no newline. It is added on line 496
2. Ok
3. Isn't the point of the logger to do that on its own? If it's a performance concern I doubt the trace runner is really somewhere where the logging overhead matters.

aleclearmind: 1. Do not put the newline. `<< DoLog` terminates the message for you and it's conceptually a newline.
2. .
3. It's most likely not a concern, but this is a policy and we do that everywhere.
   Let's not deviate from the policy unless there's a good reason.

lib/Model/Importer/Binary/ELFImporter.cpp
mrjackv: 1. `getFileName()` returns a `StringRef`, so `.str()` is a temporary
2. `PDBImporter` does not have an equivalent instance, there's only [this use](https://github.com/revng/revng/blob/develop/lib/Model/Importer/DebugInfo/PDBImporter.cpp#L490) of `InputPath` and it's already guarded

lib/PipelineC/Tracing/Runner.cpp
mrjackv: 1. Ok
2. It saves everything. I've double check but we already *do* save the model after analyses, so the flag would be useful for e.g. looking at a `module.ll` pre-crash

lib/Model/Importer/Binary/ELFImporter.cpp
aleclearmind: 1. `const std::string &` works well with temporaries. It extends their lifetime.
2. Mmmh, yeah, checking again that's fine. We'll eventually need to get rid of this.

lib/PipelineC/Tracing/Runner.cpp
aleclearmind: 2. > I've double check but we already do save the model after analyses

   Are you 100% sure? I think when I was running the trace the resume directory remained empty.
   I'm not that interested in saving `module.ll` as of now. If we already save the model, this feature is not necessary. Saving containers after analysis is arbitrary and not in line with the original intent of this change.

include/revng/Support/Error.h
mrjackv: Done

lib/PipelineC/Tracing/Runner.cpp
mrjackv: Removed

lib/PipelineC/Tracing/Runner.cpp
mrjackv: Fixed, added guard for enabled, and used `DoLog`

lib/PipelineC/Tracing/Runner.cpp
mrjackv: Fixed

include/revng/Support/CommandLine.h
mrjackv: 1. Done
2. Reworded

lib/Model/Importer/Binary/ELFImporter.cpp
mrjackv: 1. Fixed

include/revng/PipelineC/Tracing/Trace.h
mrjackv: 1. Ok
2. Ok

lib/PipelineC/Tracing/Runner.cpp
mrjackv: Sorry, brain fart, `PipelineManager` saves the model **after** an analysis is run.
We can keep things as-is or I can add another function in `PipelineC.cpp` that only saves the model.

aleclearmind:
Remember to remove the two `// TODO: we should set InputPath` comments, they should be solved.

lib/PipelineC/Tracing/Runner.cpp
mrjackv: Added the command-line option `--save-after-every-analysis` to `PipelineManager`

mrjackv:
> Remember to remove the two `// TODO: we should set InputPath` comments, they should be solved.

Done

aleclearmind:
Merged in 6465a75. Thanks.

PR #466, opened 2025-04-03
Clifter

vasama-work:


tests/unit/mlir_lit_tests/expressions/cast-extend.mlir
```diff
@@ -15,3 +15,9 @@
 
 clift.cast<extend> %i16 : !int16_t -> !int32_t
 clift.cast<extend> %u16 : !uint16_t -> !uint32_t
+
+!ptr32_int32_t = !clift.ptr<4 to !int32_t>
+!ptr64_int32_t = !clift.ptr<8 to !int32_t>
+
+%p32_i32 = clift.undef : !ptr32_int32_t
+clift.cast<extend> %p32_i32 : !ptr32_int32_t -> !ptr64_int32_t
```
pfez, line 23: I'd add add a test for expected failures when `extend` doesn't extend. Same for `truncate`.

include/revng/mlir/Dialect/Clift/IR/CliftOps.td
```diff
@@ -132,6 +132,10 @@ def Clift_FunctionOp : Clift_Op<"func",
       return mlir::cast<clift::FunctionType>(getFunctionType());
     }
 
+    clift::ValueType getCliftReturnType() {
```
pfez, line 135: Why not just `getReturnType`?

Actually the same applies for the above `getCliftFunctionType`, which could be `getFunctionType`.

include/revng/mlir/Dialect/Clift/IR/CliftOps.td
```diff
@@ -534,7 +534,7 @@ class Clift_ComparisonOp<string mnemonic>
 
   let assemblyFormat = [{
     $lhs `,` $rhs attr-dict `:`
-    custom<CliftOpTypes>(type($result), type($lhs), type($rhs))
```
pfez, line 515: What was the motivation for this?

share/revng/pipelines/revng-pipelines.yml
```diff
@@ -689,6 +689,7 @@ Branches:
               - inline-divergent-scopes
               - enforce-single-exit
               - materialize-trivial-goto
+              - remove-constant-array-returns
```
pfez: Actually I think it's better to place this before `dagify`. It doesn't really change anything at this stage. But it may change things later on, when the various passes on the `ScopeGraph` start doing smart heuristics that may use the weight of a block.

share/revng/include/primitive-types.h
```diff
@@ -233,4 +233,18 @@ extern float80_t undef_float80_t();
 extern float96_t undef_float96_t();
 extern float128_t undef_float128_t();
 
+//
+// Pointers
+//
+
+#ifdef LEGACY_BACKEND
+#define pointer16_t(T) __typeof__((T *){ 0 })
```
pfez, line 241: wait what? 16-bits pointers? really?

not that there is any harm in having them around. but do you have any specific practical reason to pulling them in today?

lib/HeadersGeneration/ModelToHeaderPipe.cpp
```diff
@@ -48,13 +48,24 @@ class ModelToHeader {
...
+    const auto &Model = *revng::getModelFromContext(EC);
+
+    auto Arch = Model.Architecture();
+    uint64_t ExplicitTargetPointerSize = Arch == model::Architecture::Invalid ?
+                                           0 :
+                                           getPointerSize(Arch);
+
+    if (ExplicitTargetPointerSize == 8) // Hardcoded 64, same as in Clift.
+      ExplicitTargetPointerSize = 0;
+
```
pfez: This is basically the same code as above in `HelpersToHeader.cpp`. Please turn it into an helper function that takes the model (or even the context) and returns the `ExplicitTargetPointerSize`.

pfez: Also IIUC the logic here is: if the pointer size dictated by the `Architecture` in the model (i.e. the input binary) is different from the pointer size in the target C implementation (which as of now is always the `Default`, so hardcoded to 8 bytes) then we instruct the builder to make pointer sizes explicit.

I think it's better to make the comment more pedantic about it, to avoid confusion in the future.

lib/mlir/Dialect/Clift/IR/CliftAttributes.cpp
```diff
@@ -98,19 +114,24 @@ llvm::StringRef ClassAttrImpl<AttrT>::getHandle() const {
 
 template<typename AttrT>
 bool ClassAttrImpl<AttrT>::hasDefinition() const {
-  return static_cast<bool>(Base::getImpl()->getDefinition());
+  return static_cast<bool>(Base::getImpl()->getDefinitionOrNull());
```
pfez: I'd make this `return Base::getImpl()->getDefinitionOrNull() != nullptr`

include/revng/mlir/Dialect/Clift/IR/CliftOps.td
vasama-work: Ah. That's because those names are already taken up by the `FunctionOpInterface` names returning `mlir::Type`. I think we might want to get rid of that interface altogether, but I didn't want to do that right now, so this is the simplest solution.

include/revng/mlir/Dialect/Clift/IR/CliftAttributes.td
```diff
@@ -90,14 +90,4 @@ def Clift_TypedefAttr
   let genVerifyDecl = 1;
 }
 
-def Clift_ClassDefinitionAttr
```
pfez, line 93: Make sense, but this needs an update to the pad documenting the clift types.

include/revng/mlir/Dialect/Clift/IR/CliftOps.td
vasama-work: Maybe you made the comment before I updated the commit message:

>    Make comparison op assembly syntax more explicit

    // Old syntax:
    // Argument types are !int32_t.
    // The boolean result type is implicitly !int32_t.
    clift.eq %0, %1 : !int32_t
    
    // New syntax:
    // The boolean result type must be specified explicitly.
    clift.eq %0, %1 : !int32_t -> !int32_t
    
 To elaborate even further, the boolean result type of a comparison really has nothing to do with the operands, so it doesn't make sense to deduce it.

share/revng/include/primitive-types.h
vasama-work: Nope. I can remove it if you like.

Currently we don't actually need `pointer64_t` either because we only support 64-bit target. Let me know if you would like to remove just `pointer16_t` or both.

lib/mlir/Dialect/Clift/Utils/CBackend.cpp
```diff
@@ -106,23 +107,28 @@ class CEmitter {
     Builder.setOutputStream(this->Out);
   }
 
-  const model::Segment &getModelSegment(GlobalVariableOp Op) {
+  static llvm::StringRef getDefaultName(llvm::StringRef Name,
```
pfez: `getNameOrDefault` maybe?

tests/unit/mlir_lit_tests/expressions/cast-extend.mlir
vasama-work: Done, in a separate commit, pending squash.

share/revng/pipelines/revng-pipelines.yml
vasama-work: Done, in a separate commit, pending squash. Ran the end-to-end test and it didn't seem to break anything.

lib/HeadersGeneration/ModelToHeaderPipe.cpp
vasama-work: Done, in a separate commit, pending squash.

tests/unit/mlir_lit_tests/types/union-with-large-tree.mlir
```diff
@@ -0,0 +1,110 @@
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
```
pfez, line 2: Commenting here for lack of a better place.

The message of the commit introducing this change reads

> Fixes bytecode emission of class types with very large trees of members.
> See the new unit test: union-with-large-tree.mlir .

I'm looking at `union-with-large-tree.mlir` and I see the test, and I know it works now, but I have no way of knowing what kind of problems this test was causing before this commit.

Can you elaborate in the commit message what problem is this commit fixing?

include/revng/mlir/Dialect/Clift/IR/CliftTypes.td
```diff
@@ -452,6 +452,10 @@ def Clift_StructType : Clift_ClassTypeBase<"Struct", "struct"> {
       return $_get($_ctxt, StructAttr::get($_ctxt, Handle), /*IsConst=*/false);
     }]>,
 
+    TypeBuilder<(ins "mlir::clift::StructAttr":$Definition), [{
+      return $_get($_ctxt, Definition, /*IsConst=*/false);
+    }]>,
+
```
pfez, line 458: I think this builder belongs to the commit that introduces `ClassDefinition` in place of `ClassDefinitionAttr`.

Same for the builder for unions with the same API.

lib/mlir/Dialect/Clift/IR/CliftAttributes.cpp
```diff
@@ -493,3 +574,89 @@ void CliftDialect::printAttribute(mlir::Attribute Attr,
                                   mlir::DialectAsmPrinter &Printer) const {
   revng_abort("cannot print attribute");
 }
+
+namespace {
+
+enum class CliftAttrKind : uint8_t {
```
pfez, line 580: Isn't this more like `CliftDefinedTypeKind`?

lib/mlir/Dialect/Clift/IR/CliftAttributes.cpp
```diff
@@ -493,3 +574,89 @@ void CliftDialect::printAttribute(mlir::Attribute Attr,
...
+      return mlir::success();
+    }
+    if (auto T = mlir::dyn_cast<clift::UnionType>(A.getType())) {
+      WriteKind(CliftAttrKind::Union);
+      writeUnionDefinition(T, Writer);
+      return mlir::success();
+    }
+  }
+
+  return mlir::failure();
```
pfez, line 661: Commenting here for lack of a better place. How does the error reporting work for this kind of stuff, from a user perspective?

lib/mlir/Dialect/Clift/Utils/ModelVerify.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+    return mlir::success();
+  }
+
+  const model::Binary &Model;
+};
+
+} // namespace
+
+mlir::LogicalResult clift::verifyAgainstModel(mlir::ModuleOp Module,
```
pfez, line 233: Would be nice to have some unit tests for this. With some small hand-written model and some small hand-written clift.

Please add them, testing both for expected success and expected failure.

include/revng/mlir/Dialect/Clift/IR/CliftOps.td
pfez: I thought so. Thank you.

lib/mlir/Dialect/Clift/IR/CliftAttributes.cpp
vasama-work: Done, in a separate commit, pending squash.

lib/mlir/Dialect/Clift/Utils/CBackend.cpp
vasama-work: Looks like this isn't actually used anymore. I've removed it instead.

tests/unit/mlir_lit_tests/types/union-with-large-tree.mlir
vasama-work: Okay. I won't change the commit message for now since that requires force pushing. Instead I will write the new commit message here:

```
When emitting bytecode, and when a custom bytecode format is not defined
for dialect types, the emitter falls back to embedding the textual
representation for the type. Usually each type would be emitted in a
separate embedding, but in the case of recursive types, the entire graph
of referenced types must be emitted in one go. For types with a very
large graph (see the new unit test: union-with-large-tree.mlir), this is
not feasiable as the number of types printed may grow to be in the
millions. This was observed during testing of some ARM binaries. This
commit works around the issue by implementing a custom bytecode format
for Clift types.
```

Is that better?

include/revng/mlir/Dialect/Clift/IR/CliftTypes.td
vasama-work: No, it's in the correct commit. This builder is required by the bytecode reading code.

lib/mlir/Dialect/Clift/IR/CliftAttributes.cpp
vasama-work: Only because the attributes currently supporting custom bytecode representation are related to defined types. But note that `EnumFieldAttr` is not itself a defined type or a description of one. This enum just disambiguates which Clift *attribute* kind is represented by the subsequent bytecode. In the future we may add support for other attributes having nothing to do with defined types.

lib/mlir/Dialect/Clift/IR/CliftAttributes.cpp
vasama-work: Since you commented on this particular `return mlir::failure()`, while it sounds scary, in this case it just means that we don't support bytecode emission for the given attribute, in which case the MLIR bytecode emission machinery simply falls back to using the textual representation.

If there's a discrepancy between the writing and reading code, or malformed bytecode is otherwise produced, the error message you get is of course pretty cryptic. Something like "unexpected data at the end of block" for example.

include/revng/mlir/Dialect/Clift/IR/CliftAttributes.td
vasama-work: Done.

lib/mlir/Pipes/LLVMToCliftPipe.cpp
```diff
@@ -0,0 +1,43 @@
+//
```
aleclearmind, line 1: Copy the header from other `.cpp` files, it's slightly different from the `.h` ones.
Review all newly introduced `.cpp` files.

lib/mlir/Pipes/LLVMToCliftPipe.cpp
```diff
@@ -0,0 +1,43 @@
...
+public:
+  static constexpr auto Name = "llvm-to-clift";
+
+  std::array<pipeline::ContractGroup, 1> getContract() const {
+    using namespace pipeline;
+    using namespace revng::kinds;
+
+    return { ContractGroup({ Contract(StackAccessesSegregated,
+                                      0,
+                                      MLIRFunctionKind,
```
aleclearmind: About the `...Kind` suffix, review what's the convention. `StackAccessesSegregated` does not have it.

lib/mlir/Pipes/LLVMToCliftPipe.cpp
```diff
@@ -0,0 +1,43 @@
...
+  void run(pipeline::ExecutionContext &EC,
+           const pipeline::LLVMContainer &LLVMContainer,
+           revng::pipes::CliftContainer &CliftContainer) {
+    CliftContainer.getContext()->loadDialect<clift::CliftDialect>();
+
+    clift::importLLVM(CliftContainer.getModule(),
+                      *revng::getModelFromContext(EC),
+                      &LLVMContainer.getModule());
+
+    EC.commitAllFor(CliftContainer);
```
aleclearmind: Why are you commit everything altogether?

This is a function-wise pipe and should work function-wise and commit things function-wise.

You are in case [1.2](https://pad.rev.ng/s/EVGvDPdZA#).

share/revng/pipelines/revng-pipelines.yml
```diff
@@ -701,6 +702,45 @@ Branches:
...
+            UsedContainers: [module.bc.zstd, module.mlir]
+        Artifacts:
+          Container: module.mlir
+          Kind: mlir-module
+          SingleTargetFilename: llvm_to_clift.mlir
+          Docs: |-
+            This artifact contains a Clift function for each function in the
+            input program.
+  - From: llvm-to-clift
+    Steps:
```
aleclearmind: 1. Drop these two lines.

   If you have a branch of the pipeline the nests itself directly in the one you previously declared, you can avoid creating a new branch.
2. Considering merging in a single step these pipes.
   Each step requires to save things on disk.
   I understand that you might want that for debugging purposes, but this is not the criterion we use to split list of pipes into passes.
   We split a list of pipes into passes if we think it's beneficial to resume things from a certain point, for invalidation purposes.
   I would make a single step instead of 3.
   If you need artifacts for debugging purposes, add them temporarily. I know this is annoying, but that's how it is right now.

include/revng/mlir/Dialect/Clift/IR/CliftOps.td
pfez: Maybe add a `TODO` comment to those 2 APIs explaining this for future reference.

lib/mlir/Pipes/Legalization.cpp
```diff
@@ -0,0 +1,61 @@
...
+      }
+    });
+
+    for (const model::Function &Function :
+         getFunctionsAndCommit(EC, CliftContainer.name())) {
+      auto It = Functions.find(Function.Entry());
+      revng_check(It != Functions.end()
+                  and "Requested Clift function not found");
+
+      revng_check(legalizeForC(It->second, Target).succeeded());
```
aleclearmind: Split the check for the result on a separate statement.
Assume `revng_check` might be turned into a `revng_assert` or go away.

Also, it makes the code more readable. I usually skip over `revng_{check,assert}` while reading the code.

share/revng/test/configuration/revng/end-to-end.yml
```diff
@@ -158,3 +158,44 @@ commands:
     suffix: /
     command: |-
       revng artifact --model "$INPUT2" --resume "$OUTPUT" prepare-for-clift "$INPUT1" -o /dev/null
+
+  #
+  # Produce the emit-c artifact and check the resulting C code
+  #
+  - type: revng.emit-c
```
aleclearmind: This is too much code and it 's in large part duplicated from the previous test.

@fez: why are we not using `emit-recompilable-archive`?

In general, when the test body becomes this big, it's good to make a script, like `check-invalidations.

lib/mlir/Pipes/Legalization.cpp
```diff
@@ -0,0 +1,61 @@
...
+                                      InputPreservation::Preserve) }) };
+  }
+
+  void run(pipeline::ExecutionContext &EC,
+           pipes::CliftContainer &CliftContainer) {
+    const auto &Target = clift::TargetCImplementation::Default;
+
+    mlir::ModuleOp Module = CliftContainer.getModule();
+
+    std::unordered_map<MetaAddress, clift::FunctionOp> Functions;
```
aleclearmind, line 39: We usually follow the LLVM's programmer manual which suggests against `unordered_map`.

I think we could use a `DenseMap`, but maybe we'd need to implement a certain trait on `MetaAddress`.

No biggie though.

include/revng/mlir/Dialect/Clift/IR/CliftOps.td
pfez: > the boolean result type of a comparison really has nothing to do with the operands, so it doesn't make sense to deduce it.

Fair. Put this sentence in the commit message too, please.

share/revng/include/primitive-types.h
pfez: Just leave them both, they're not doing any harm. I just wanted to have more context.

lib/mlir/Dialect/Clift/Utils/CBackend.cpp
pfez: Even better then.

tests/unit/mlir_lit_tests/types/union-with-large-tree.mlir
pfez: Much better, yes!

lib/mlir/Dialect/Clift/IR/CliftAttributes.cpp
pfez: Fair enough.

lib/mlir/Dialect/Clift/IR/CliftAttributes.cpp
```diff
@@ -466,3 +574,89 @@ void CliftDialect::printAttribute(mlir::Attribute Attr,
                                   mlir::DialectAsmPrinter &Printer) const {
   revng_abort("cannot print attribute");
```
aleclearmind: Use capital for the first letter

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
```diff
@@ -0,0 +1,1815 @@
...
+  abi::FunctionType::Layout CurrentLayout;
+
+  mlir::OpBuilder::InsertPoint LocalDeclarationInsertPoint;
+
+  ScopeGraphPostDomTree PostDomTree;
+
+  clift::ValueType VoidTypeCache;
+
+  uint64_t PointerSize;
+  clift::ValueType PointerTypeCache;
```
pfez: `VoidPointerTypeCache`

include/revng/mlir/Dialect/Clift/IR/CliftOps.td
vasama-work: I'll do that when rebasing, once you're done with your review.

lib/mlir/Dialect/Clift/Utils/ModelVerify.cpp
vasama-work: Done, in a separate commit, pending squash.

lib/mlir/Pipes/LLVMToCliftPipe.cpp
vasama-work: I don't follow. Can you elaborate?

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
```diff
@@ -0,0 +1,1815 @@
...
+
+    // True if the AssignLabelOp has already been created.
+    bool HasAssignLabel = false;
+  };
+
+  struct ValueMappingInfo {
+    mlir::Value Value;
+
+    // True if the address of the value should be taken before use in Clift.
+    bool Addressof;
```
pfez: `TakeAddressOnUse`.

I think making an example in the comment is most useful to explain when we expect to use this.

This is the case for `AllocaInst` for example, that return a pointer on LLVM IR, but that we translate to local variables in clift. Users of the `AllocaInst` on LLVM still see a pointer though, hence when we translate the uses of the `AllocaInst` to clift we have to inject an `AddressOf` to make type checking happy.

lib/mlir/Pipes/Legalization.cpp
vasama-work: Honestly I don't like this whole dance of first mapping meta-addresses to Clift functions in each pipe. How would you feel about standardising the naming of isolated Clift functions? E.g. using just the meta-address:
```
clift.func @"0x400177:Code_x86_64"<!cabifunction_123_>()
```

Then we could look them up using the symbol name as needed.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
```diff
@@ -0,0 +1,1815 @@
...
+  uint64_t PointerSize;
+  clift::ValueType PointerTypeCache;
+  clift::ValueType IntptrTypeCache;
+
+  struct ArgumentMappingInfo {
+    mlir::Value Argument;
+
+    // True if the argument is a by-address argument in the LLVM IR and should
+    // be dereferenced before use in Clift.
+    bool IsByAddress;
```
pfez: `DereferenceOnUse`

This also would benefit from some examples in the comments, to describe *when* we need to dereference on the use: how do the input LLVM IR and the model look like.

IIRC this is used when the LLVM IR passes an argument by address *but the argument is passed by copy on the model*. This is an important distinction, because the situation where the argument is pointer-typed both on llvm and on the model is not a problem. Actually I'm not even sure it can happen, but if it did, it would be fine.

There are some comments around the uses of `isByAddressParameter` to explain this, but I think it would be beneficial to comment directly the data structure here.

lib/mlir/Pipes/LLVMToCliftPipe.cpp
aleclearmind: You need to add something like

```
/// \file CodeGenerator.cpp
/// This file handles the whole translation process from the input assembly to
/// LLVM IR.
```

lib/mlir/Pipes/Legalization.cpp
aleclearmind: To do that we'd need to move the references to the model (which AFAIU are in attribute) into the name.

AFAIU there you have a location. having a function named "/functions/0x400177:Code_x86_64" is not the nicest. Also, we'll use name for proper names from the model, won't we?

If your problem is having the same code all over the place, you can factor it out.

I don't think we have a performance problem, if we had, we could have an analysis caching the mapping used by various pipes, but I don't think it's called for.

lib/mlir/Pipes/LLVMToCliftPipe.cpp
vasama-work: I've changed the importer API to work per function. See this commit:
`4e2adbbce AMEND: Change llvm-to-clift to one function at a time`

share/revng/pipelines/revng-pipelines.yml
vasama-work: Do the changes in this commit seem okay to you?
`AMEND: Multiple steps for emit-c artifact`

lib/mlir/Pipes/Legalization.cpp
vasama-work: Done, in a separate commit, pending squash.
(`78aa43c74 AMEND: CliftFunction rename in legalization`)

lib/mlir/Dialect/Clift/IR/CliftAttributes.cpp
vasama-work: Done, in a separate commit, pending squash.
(`c6110711e Fix CliftAttributes abort message`)

lib/mlir/Pipes/LLVMToCliftPipe.cpp
```diff
@@ -0,0 +1,50 @@
...
+           const pipeline::LLVMContainer &LLVMContainer,
+           revng::pipes::CliftContainer &CliftContainer) {
+    CliftContainer.getContext()->loadDialect<clift::CliftDialect>();
+    auto const &Model = *revng::getModelFromContext(EC);
+
+    model::CNameBuilder NameBuilder(Model);
+    const llvm::Module &M = LLVMContainer.getModule();
+
+    auto Importer =
+      clift::LLVMToCliftImporter::make(CliftContainer.getModule(), Model);
```
aleclearmind: This is a bit dangerous.

Can we guarantee that this part of the code does not read any (or reads very little) parts of the model?

Anything being read here, will be a dependency of *all* emitted targets.
Also, if they have lazily populated caches, invalidation will break.

Double check if this happens. If not, this is OK and add a comment about the fact that.

Ideally, `Importer` and `NameBuilder` should be defined in the loop below.

lib/mlir/Pipes/Legalization.cpp
vasama-work: I'm not proposing to change or get rid of the handle. Just to use a specific symbol naming for isolated functions so we can easily look them up. This old backend pipe does something similar:
https://github.com/revng/revng/blob/64446d0453fbae77bbad34950cdc0f2362f3cbe7/lib/Backend/DecompilePipe.cpp#L41

lib/mlir/Pipes/LLVMToCliftPipe.cpp
vasama-work: I changed this to `CliftFunction`. Is that okay with you?
(`9b5eebbec Rename MLIRFunctionKind to CliftFunction`)

share/revng/pipelines/revng-pipelines.yml
aleclearmind: Not really. Please try to address my comment in its entirety and if something is not clear ask for clarification. If the author of the PR "probes" the reviewer with "is this OK?", they're pushing on the reviewer the burden of verifying if they actually fully addressed the comment, which is primarily the duty of the author of the PR.

This said:

1. There's still two superfluous branches. Merge the `From: prepare-for-clift` and the `From: llvm-to-clift` branches into the one containing `prepare-for-clift`. There is no reason to split them.
2. You still have two new steps: either justify why you think it's important to cache results there, or merge them into the `prepare-for-clift` step.
   If you need a step for testing an artifact, let's discuss it.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
```diff
@@ -0,0 +1,1815 @@
...
+
+  auto getOrEmitSymbol(const llvm::GlobalObject *G, auto Emit) {
+    using ReturnType = std::invoke_result_t<decltype(Emit)>;
+    auto [Iterator, Inserted] = SymbolMapping.try_emplace(G);
+    if (Inserted)
+      Iterator->second = Emit();
+    return mlir::cast<ReturnType>(Iterator->second);
+  }
+
+  clift::ValueType emitGlobalObject(const llvm::GlobalObject *G) {
```
pfez, line 400: I'd go for `emitFunction`, since IIUC this only ever emits functions, isn't it?

In fact, I think the argument type could be `const llvm::Function *`.

lib/mlir/Pipes/Legalization.cpp
aleclearmind: Names are introduced later in the pipeline.
The names will be different depending on the stage of the MLIR pipeline.

vasama-work: Currently we never modify the *symbol names*, and I don't think we should ever modify them or use them for the name emitted by the C backend. Instead once we implement the model-agnostic backend, we should insert new discardable attributes on functions:
```
clift.func @"0x400177:Code_x86_64"<!cabifunction_123_>() attributes {
  handle = "/function/0x400177:Code_x86_64",
  clift.name = "some_function"
}
```

The symbol names are really just used to reference module-level operations in Clift code.

share/revng/pipelines/revng-pipelines.yml
vasama-work: Ah, I get it now. Get rid of the `prepare-for-clift` artifact which is no longer needed and also the `llvm-to-clift` artifact. I find this very useful for testing, but I guess we can get rid of it for now and think about whether it's needed later.

lib/mlir/Pipes/Legalization.cpp
aleclearmind: Mmmh, I'm not super fan of this: we'd be duplicating information and we're not using names for... names.

Imagine that you want to use clift for a C frontend and you're linking two modules, you want to link by the name of the operation itself. Not by the "clift.name".

I think the name of the operation should reflect the one we intend to emit in C. Concerns about the code having to build a map are secondary.

vasama-work: > Imagine that you want to use clift for a C frontend and you're linking two modules, you want to link by the name of the operation itself. Not by the "clift.name".

You can still do that if you want. Clift itself doesn't care. For a bit of context:
This hypothetical `clift.name` attribute is what we call *discardable*. What that means is that it is attached to an operation from the outside, rather than being an intrinsic part of the operation the way the handle and type of the function are. This would be more like an override we could use for the backend, because we don't necessarily want to use the *symbol name*. It could be inserted in the pre-backend comment and name insertion pipe we've envisioned, alongside comments and whatever other information is needed by the then model-agnostic backend to do its job. I don't think repurposing symbol names for this is the right thing for *us* to do. But again, nothing about what we do in the rev.ng pipeline prevents some other hypothetical user of Clift from doing that.

And really, for sure, the most important purpose of the *symbol name* is referring to module-level operations within the module, e.g. in `UseOp`:
```
// Turn the symbol name into an SSA value:
%0 = clift.use @"0x400177:Code_x86_64" : !f
// Call the function via that SSA value:
%1 = clift.call %0() : !f
```

> I think the name of the operation should reflect the one we intend to emit in C.

To be very explicit, I disagree. The purpose of the symbol name is lookup. Right now the symbol name comes from LLVM and the emitted name comes directly from the model. In the future when we have the model-agnostic backend, we will have to attach more information than just names as attributes. There is no reason to conflate this with symbol names.

lib/mlir/Pipes/LLVMToCliftPipe.cpp
vasama-work: > Can we guarantee that this part of the code does not read any (or reads very little) parts of the model?

The only thing read by the importer construction is: `Model.Architecture()`

> Also, if they have lazily populated caches, invalidation will break.

I'm not entirely sure what you mean. Lazily populated caches in the importer or on the model?

lib/mlir/Pipes/Legalization.cpp
aleclearmind: My perspective is:

1. We want to link by the name of the operation, not by `clift.name`.
2. If we have a lifted module and an hypothetical module coming from C (maybe written by the user), we want to link the `lol` function declared by the user with the `lol` function in the module, not with `0x400177:Code_x86_64`.

> I don't think repurposing symbol names for this is the right thing for us to do. 

Why? I don't see it as "repurposing", I see at its natural role.
Just like in LLVM IR the function name becomes the symbol name in the object file.

> The purpose of the symbol name is lookup

The concept of "lookup" is an implementation detail.
We're designing a language, and we should focus on semantics (e.g., linking semantics), not on implementation details.
I don't see "otherwise I've to build a `map<MetaAddress, FunctionOp>`" as a very strong argument in this context.

I see no reason not to use the name of something as the name we emit in C.
I also see no reason to duplicate the `MetaAddress` (`handle` and `FunctionOp`'s name), it's unnecessarily verbose and prone to go out of sync.

vasama-work: A bit of further clarification: as it is right now, because we get the function symbol names directly from the imported LLVM functions, we could simply do the same thing we do to look up LLVM function names without any other changes:
```cpp
clift::FunctionOp F = SymbolTable::lookupSymbolIn(CliftModule,
                                                  B.NameBuilder.llvmName(Function));
```
So really it is just a matter of deciding to do this (or something like this, e.g. just using the stringified meta-address directly).

lib/mlir/Pipes/LLVMToCliftPipe.cpp
aleclearmind: > The only thing read by the importer construction is: Model.Architecture()

Nice. Add a comment about this so I won't get scared when I pass by.

> > Also, if they have lazily populated caches, invalidation will break.
>
> I'm not entirely sure what you mean. Lazily populated caches in the importer or on the model?

By the importer or `CNameBuilder`.

If you have something like:

```c++
struct MySupportClass {
  std::map<FunctionOp, std::string> LazyNameCache;
}

void mypipe() {
  MySupportClass Cache.
  for (FunctionOp F : getAndCommit(...)) {
    for (CallOp C : F) {
      if (not Cache.LazyNameCache.contains(C.callee())) {
        Cache.LazyNameCache[C.callee()] = Model.Functions[C.address()].Name();
      }

      use(Cache.LazyNameCache.at(C.callee()));
    }
  }
}
```

We have a problem due to the fact that the first caller of function F1 will be recorded as depending on the name of F1, but the second one won't because it did not call `model::Function::Name()`, due to the cache.

This breaks invalidation tracking.

Does this make sense?

lib/mlir/Pipes/Legalization.cpp
vasama-work: > If we have a lifted module and an hypothetical module coming from C (maybe written by the user), we want to link the `lol` function declared by the user with the `lol` function in the module, not with `0x400177:Code_x86_64`.

It's not so simple. We'll need to involve the model in such an operation anyway to map `lol` to `/function/0x400177:Code_x86_64` for the handle, so there's no reason we can't also map the C name used by the user to `local_0x400177:Code_x86_64` or whatever.

> Just like in LLVM IR the function name becomes the symbol name in the object file.

I think that's a very different scenario. In our case the C name comes from the model.

> The concept of "lookup" is an implementation detail.

It is not. It is very much part of the Clift language. See the `UseOp` example I made.

aleclearmind: Even in LLVM, we do (or should) not rely on names for mapping `model::Function`s to `llvm::Function`s, we use the metadata.

There should be only one source of truth as what `model::Function` a `FunctionOp` is associated to.

On top of this, if we go the way I suggest above, the name is going to change along the pipeline and looking up things that way is not reliable.

aleclearmind: > It's not so simple. We'll need to involve the model in such an operation anyway to map lol to /function/0x400177:Code_x86_64 for the handle, so there's no reason we can't also map the C name used by the user to local_0x400177:Code_x86_64 or whatever.

The clift at the end of the pipeline will have as name `lol`. `clift-link` could very well work on it, being entirely model agnostic.

```
clang -emit-clift user.c -o user.clift
revng artifact clift program -o revng.clift # This uses lol as name
clift-link user.clift revng.clift -o linked.clift
```

lib/mlir/Pipes/LLVMToCliftPipe.cpp
vasama-work: I see. We have a few caches in the importer that are persisted across functions. Of those, the one that might be problematic is the symbol mapping, which maps `llvm::GlobalObject` to `clift::GlobalOpInterface` (a module-level entity in the Clift module, it's generic but in practice it's just LLVM functions to Clift functions). When importing function F, for each function G called by F, we import a declaration for G. For isolated/imported functions this requires accessing the prototype of G. In the case that we later import another function H, which also calls G, we might not access the prototype of G again while importing H.

I guess that's problematic for invalidation?

The other persistent mappings are for caching string literal detection (does not touch the model) and for caching helper function name to a Clift function (also does not touch the model).

aleclearmind: > When importing function F, for each function G called by F, we import a declaration for G. For isolated/imported functions this requires accessing the prototype of G. In the case that we later import another function H, which also calls G, we might not access the prototype of G again while importing H.
>
> I guess that's problematic for invalidation?

Yes, that's exactly the typical case. The right thing to do is to re-compute the prototype every time.

As for other caches, if you don't touch the model, it's not a problem.

lib/mlir/Pipes/Legalization.cpp
vasama-work: Alright, I'm not sure how useful this discussion is right now. Going back to the original issue, since `llvm::DenseMap` requires implementing some traits, would you prefer I do that or just keep using `unordered_map`?

aleclearmind: It's not critical to use `DenseMap`, but keep in mind that we prefer it to other STL data structures.

If you're curious to understand more of it, implement the trait.
Only issue is that I think you need *two* reserved encodings.

Maybe it's better skip over this for now.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
```diff
@@ -0,0 +1,1815 @@
...
+    // Clear the function-specific mappings once this function is emitted.
+    auto MappingGuard = llvm::make_scope_exit([&]() {
+      ArgumentMapping.clear();
+      BlockMapping.clear();
+      AllocaMapping.clear();
+      ValueMapping.clear();
+    });
+
+    llvm::ArrayRef LayoutArguments = getLayoutArguments(CurrentLayout);
+    revng_assert(F->arg_size() == Op.getArgumentTypes().size());
```
pfez, line 1663: This assertion is redundant from above. I'd keep this one and drop the previous, so that all assertions on numbers of arguments are close together.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
```diff
@@ -0,0 +1,1815 @@
...
+  clift::ValueType IntptrTypeCache;
+
+  struct ArgumentMappingInfo {
+    mlir::Value Argument;
+
+    // True if the argument is a by-address argument in the LLVM IR and should
+    // be dereferenced before use in Clift.
+    bool IsByAddress;
+
+    mlir::Type CastType;
```
pfez, line 1740: Please also elaborate a bit about why we need `CastType` in a comment.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
```diff
@@ -0,0 +1,1815 @@
...
+
+    // Compute the scope graph. While the computation should not modify the
+    // function IR, it is not const-correct and so a const_cast must be used.
+    PostDomTree.recalculate(const_cast<llvm::Function &>(*F));
+
+    mlir::OpBuilder::InsertionGuard BuilderGuard(Builder);
+    Builder.setInsertionPointToEnd(&BodyBlock);
+
+    // Finally emit the function body starting at the entry block.
+    emitScope(&F->getEntryBlock(), /*OuterPostDom=*/nullptr);
```
pfez: Here it is better to pass a pointer to the root of the `PostDomTree` instead of `nullptr` as `OuterPostDom`.
Thanks to the properties of the `ScopeGraph`, that node should always be guaranteed to be a real node, and never an artificial sink node introduced by the LLVM PostDom machinery.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
```diff
@@ -0,0 +1,1815 @@
...
+
+namespace {
+
+static Logger<> BasicBlockLog{ "llvm-to-clift-basic-blocks" };
+static Logger<> ExpressionLog{ "llvm-to-clift-expressions" };
+
+[[nodiscard]] static mlir::OpBuilder::InsertPoint
+makeInsertionPointAfterStart(mlir::Block *Block) {
+  return mlir::OpBuilder::InsertPoint(Block, Block->end());
+}
```
pfez, line 53: :thinking: this doesn't do what's advertised by the name. It always makes an insert point at the end. It just happens to only be called in a place where "after start" and `end` coincide, because the block is always empty.

Change the name to `makeInsertionPointAtEnd`, or just inline it in the only call site.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
```diff
@@ -0,0 +1,1815 @@
...
+
+  mlir::Block *Block = InsertPoint.getBlock();
+  mlir::Block::iterator Point = InsertPoint.getPoint();
+
+  // Convert an end iterator back into begin and advance non-end iterators. This
+  // is because the builder inserts operations *before* the specified iterator.
+  Point = Point == Block->end() ? Block->begin() : std::next(Point);
+
+  Builder.setInsertionPoint(Block, Point);
+}
```
pfez, line 80: The whole thing of "wrapping back" from `end` to `begin` is not clear. What's not clear about it is if you're doing this on purpose, as a feature some of the callers rely upon in some situation (and if so when), or for what other reason. If I had to guess without looking at callers and reasoning about them you may be silently relying on the fact that callers call these APIs in the `begin`/`end` corner cases only when the insertion block is empty. That's probably the only case I can think of where your trick makes sense. If that's what you're relying upon advertise it comments and even assert it. Otherwise, please explain.

pfez: But then looking at users it seems that you're explicitly relying on `end` to actually mean at the beginning of the block. Weird.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
```diff
@@ -0,0 +1,1815 @@
...
+  /* LLVM control flow import */
+
+  template<typename OpT, typename... ArgsT>
+  OpT emitLocalDeclaration(mlir::Location Loc, ArgsT &&...Args) {
+    mlir::OpBuilder::InsertionGuard Guard(Builder);
+    restoreInsertionPointAfter(Builder, LocalDeclarationInsertPoint);
+    OpT Op = Builder.create<OpT>(Loc, std::forward<ArgsT>(Args)...);
+    LocalDeclarationInsertPoint = saveInsertionPointAfter(Builder);
+    return Op;
+  }
```
pfez, line 1251: AFAIU `LocalDeclarationInsertPoint` is always ever set at the beginning and in here.
At the beginning you set it at the `end()` of the block, which means that the first call to this `emitLocalDeclaration` makes it wrap to the `begin()` of the block (even if `begin() == end()`), due to how `restoreInsertionPointAfter` works.

Also then you call `saveInsertionPointAfter`, which is guaranteed to never hit the quirky corner case of `begin()` here, because you've just called `Builder.create<OpT>` that implicitly moved the insertion point from `begin()` to `std::next(begin())` (assuming the creation methods of the `Builder` work as in plain LLVM). So in principle here you could just do `LocalDeclarationInsertPoint = std::prev(Builder.getInsertPoint())`.

Now that I think about it, `LocalDeclarationInsertPoint` could just be left unset at the beginning, and `emitLocalDeclaration` could just use it plainly if set, or set it the first time it's called. On the first call, the insertion block is always the entry block of the current function, and the insert point is either `end()` (if the block is empty, or `begin()` if the block isn't empty. This would enable to drop the call to `restoreInsertionPoint` and enable you to just turn the line before `return Op` into simply `LocalDeclarationInsertPoint = Builder.saveInsertionPoint()` (no `std::prev` required).

The reason why I want to do this is that I've looked ahead at other users of `restoreInsertPointAfter` and `saveInsertPointAfter` and I think I've finally figured out why you need the quirky corner cases (that you really need to document), and I think that it would be overall beneficial to call those only where really needed to rely on the quirky corner case itself. Here we can do without it and the overall logic becomes less entangled with the corner case, which makes it easier to reason about and understand.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
```diff
@@ -0,0 +1,1815 @@
...
+    LoggerIndent Indent(BasicBlockLog);
+
+    // Map BB to the MLIR block, emit label if necessary:
+    {
+      auto [Iterator, Inserted] = BlockMapping.try_emplace(BB);
+
+      // Save the current insertion point in case a label assignment must later
+      // be emitted at the start of this block.
+      revng_assert(not Iterator->second.InsertPoint.isSet());
+      Iterator->second.InsertPoint = saveInsertionPointAfter(Builder);
```
pfez, line 1330: Oh wait, now I see why you need the quirky corner cases. It's all due to the fact that we accept an `AssignLabel` in the middle of a single `mlir::Block` that represents multiple consecutive `llvm::BasicBlock`s, isn't it?

Please, remind me why we wanted that?

In any case make sure you document this thoroughly on `restoreInsertionPointAfter` and `saveInsertionPointAfter`. Given that after my suggested changes these would be used only for insertion points of `AssignLabel`, perhaps we can even find a better names for them, that makes it clearer what they are used for.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
```diff
@@ -0,0 +1,1815 @@
...
+    if (llvm::isa<llvm::UnreachableInst>(Terminal)) {
+      // Do nothing.
+    } else if (auto *Return = llvm::dyn_cast<llvm::ReturnInst>(Terminal)) {
+      auto Op = Builder.create<ReturnOp>(TerminalLoc);
+
+      if (const llvm::Value *Value = Return->getReturnValue()) {
+        auto FunctionType = CurrentFunction.getCliftFunctionType();
+
+        auto FuncReturnType = FunctionType.getReturnType();
+        auto LLVMReturnType = FuncReturnType;
```
pfez: Make the type of these variables explicit please.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
```diff
@@ -0,0 +1,1815 @@
...
+}
+
+template<typename OpT, typename... ArgsT>
+static mlir::OwningOpRef<OpT> createOperation(mlir::MLIRContext *Context,
+                                              mlir::Location Location,
+                                              ArgsT &&...Args) {
+  OpT Op = mlir::OpBuilder(Context).create<OpT>(Location,
+                                                std::forward<ArgsT>(Args)...);
+  return mlir::OwningOpRef<OpT>(Op);
+}
```
pfez: This is used only at the very end of the file. Move it closer to the user.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
```diff
@@ -0,0 +1,1815 @@
...
+      revng_log(ExpressionLog, "LHS subexpression");
+      mlir::Value Lhs = (LoggerIndent(ExpressionLog),
+                         rc_recur emitExpression(I->getOperand(0), Loc));
+
+      revng_log(ExpressionLog, "RHS subexpression");
+      mlir::Value Rhs = (LoggerIndent(ExpressionLog),
+                         rc_recur emitExpression(I->getOperand(1), Loc));
+
+      PrimitiveKind Kind = PrimitiveKind::GenericKind;
+      switch (I->getOpcode()) {
```
pfez, line 878: 1. `Mul` should set `PrimitiveKind::Number`
2. All bitwise operators should set `PrimitiveKind::PointerOrNumber`

I may be missing some others. Let's see what I can find out below.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
```diff
@@ -0,0 +1,1815 @@
...
+    }
+
+    if (auto I = llvm::dyn_cast<llvm::ICmpInst>(V)) {
+      revng_log(ExpressionLog, "llvm::ICmpInst");
+      LoggerIndent Indent(ExpressionLog);
+
+      using enum llvm::ICmpInst::Predicate;
+
+      PrimitiveKind Kind = PrimitiveKind::GenericKind;
+      switch (I->getPredicate()) {
```
pfez, line 946: `ICMP_EQ` and `ICMP_NE` should also set `Kind` to `PointerOrNumber`, because they definitely can't be floating points.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
```diff
@@ -0,0 +1,1815 @@
...
+        return emitIntegerCast(Loc,
+                               Operand,
+                               getIntegerSize(V->getType()),
+                               Kind);
+      };
+
+      switch (I->getOpcode()) {
+        using Operators = llvm::CastInst::CastOps;
+      case Operators::Trunc:
+        rc_return EmitIntegerCast(PrimitiveKind::GenericKind);
```
pfez, line 1033: Again  `PointerOrNumber`, you can't truncate something and expect a floating point out.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
```diff
@@ -0,0 +1,1815 @@
...
+        rc_return Operand;
+      case Operators::IntToPtr:
+        if (getIntegerSize(I->getSrcTy()) > PointerSize) {
+          Operand = emitCast(Loc,
+                             Operand,
+                             getPrimitiveType(PointerSize),
+                             CastKind::Truncate);
+        }
+
+        rc_return emitCast(Loc, Operand, getVoidPointerType());
```
pfez, line 1055: Too bad we emit so many `void *`, but any sane user will first bitcast these to some sensible `Type *`, and then we can write a beautifier that turns `(Type *)((void *) something)` into `(Type *) something`.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
```diff
@@ -0,0 +1,1815 @@
...
+      Value = Builder.create<CastOp>(Loc, TargetType, Value, Kind);
+
+    return Value;
+  }
+
+  // Apply an implicit conversion to the requested type. The conversion is only
+  // applied in the case that neither the source nor target values have array
+  // type, and as long as both have the same size. If the cast is possible, it
+  // is naturally performed using the bitcast mode. If the cast is not possible,
+  // not cast is emitted and the original value is returned instead. In the case
```
pfez: Typo. `s/not/no`.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
```diff
@@ -0,0 +1,1815 @@
...
+  // of failure to convert LLVM IR to Clift, this is a likely point of failure.
+  // When that happens, invalid Clift is emitted, causing subsequent failure in
+  // verification.
+  //
+  // During development of the LLVM IR to Clift converter, this behaviour was
+  // found to be more useful for debugging than simply aborting. It is generally
+  // easier to find the problem when full context of the surrounding Clift code
+  // is available.
```
pfez, line 434: Sensible approach.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
vasama-work: I added this general comment before the three functions:
```
// These helper functions are used to save insertion points *after* an
// operation, while the usual operating mode of `OpBuilder::InsertPoint` is
// *before* and operation. On save, the saved iterator is moved backwards by one
// position, conversely moved forward on restore. For the begin/end positions,
// it can be imagined that they loop around to the other. For this reason the
// end position actually signifies insertion at the start of the block.
//
// The scenario where saving *after* vs *before* an operation (or end) makes a
// difference is as follows: We emit some operation X in block B and want to
// save the position *after* X. If X is at the end B, saving the position
// normally would result in saving an end iterator for B. If some new operation
// Y is subsequently inserted at the end of B, the meaning of the previously
// saved position is changed. The iterator still points to the end of B, but now
// the preceding operation (and so the operation *after* which we insert) is no
// longer X but rather Y. For the same reason the usual mechanism does not allow
// saving the position at the beginning of a block (or *after* an imagined
// operation before the first operation in the block, if any).
```

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
vasama-work: LLVM IR basic blocks do not map one-to-one to Clift blocks. That's because in Clift we no longer have a CFG. For Clift the existence of blocks is just an artifact of MLIR's CFG nature. We never have more than one block per region. Each statement region represents a C scope and a single C scope (Clift region and block) may be composed of many LLVM IR basic blocks. When we have a goto targeting a particular LLVM IR basic block, the matching label emitted in Clift may end up being in the middle of a scope. In that case, as explained in the new comment I added before the `InsertionPointAfter` helpers, an `InsertionPoint` saved normally at the *then* end of the scope would be invalidated upon any further insertion of new operations at the end of that scope.

> It's all due to the fact that we accept an AssignLabel in the middle of a single `mlir::Block` that represents multiple consecutive `llvm::BasicBlocks`, isn't it?

It's not clear to me what you mean by this. Hopefully the previous explanation clarified it for you, but if not, let me know. We certainly don't map gotos targeting multiple different LLVM IR basic blocks to a single shared label (even in the case that there happened to be no operations in between the two labels).

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
vasama-work: It seems to have gone unused with the change ale requested to make the importer incremental, so I've removed it instead.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
vasama-work: Okay, there's clearly a misunderstanding on the purpose of these kind conversions on one of our parts. The reason I've inserted any conversion of kind at all is only to preserve semantics. The operations handled in this switch have different behaviour depending on the signedness of the operand. When the kind is anything other than `Generic`, `emitIntegerOp` will insert conversions to the requested kind before applying the operation and back to `Generic` afterwards. Even if we converted to `Number` or `PointerOrNumber`, it would have no effect and be converted right back to `Generic` again. And furthermore, the future Clift beautifications would probably remove those pointless casts again and simply convert the arithmetic back to `Generic` because it is followed by a cast to `Generic`. If we want some kind of deduction of type based on usage, I'm all for it, but I strongly disagree that this is the place to do it.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
vasama-work: Yes, the intent of the Clifter is not to emit beautiful code, only to preserve semantics. Already with the beautifications I've made in a branch you see a huge reduction in useless casts. They're very easy to remove.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
vasama-work: Done.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
vasama-work: I renamed it and added more comments in this area.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
vasama-work: The previous comment was incorrect. I renamed it to `TakeAddressOnUse` and added more comments in this area.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
vasama-work: I got rid of this function entirely and inlined the logic into the only caller.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
vasama-work: I removed it.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
vasama-work: I added more comments in this area.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
vasama-work: I applied this change and everything still seems to work. 👍

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
vasama-work: Done.

lib/mlir/Pipes/Legalization.cpp
vasama-work: Yeah, I'd rather keep using the `unordered_map` for now.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
pfez: > the matching label emitted in Clift may end up being in the middle of a scope

This clarifies it. Thanks.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
pfez: This clarifies the whole quite a lot. Thank you.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
pfez: You still decided to leave this be in the old form as of now, but as I explained this doesn't really need to use the quirky wraparound helper functions, so I think it shouldn't use it.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
pfez: Nice!

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
vasama-work: Ah. As I see it, there's two options for how to emit local variables:
1. Always at the beginning of the function block.
2. In order at the beginning of the function block (how it works now).

The reason I like 2 over 1, even if it's a slight bit more complicated due to having to save the position, is that it doesn't reverse the order of emitted locals. Instead the locals are emitted in the order in which they are imported, which I prefer for the clarity when debugging issues in the Clifter output. And it's not a large amount of complexity since these helpers already exist.

share/revng/test/configuration/revng/end-to-end.yml
pfez: > why are we not using `emit-recompilable-archive`?

We aren't using that because `emit-recompilable-archive` uses the legacy backend code as of now. But I agree that we should turn `emit-recompilable-archive` into `emit-legacy-recompilable-archive`, and make a new `emit-recompilable-archive` that works with the new clift-based C backend. This will initially still use the legacy headers generation, but then as we move towards clift-based header generation, it will adopt the new clift ways.

lib/mlir/Dialect/Clift/IR/CliftAttributes.cpp
pfez: > simply falls back to using the textual representation

Ok good. Thanks

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
pfez: Neat

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
pfez: I see. Let's keep this.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
pfez: > When the kind is anything other than `Generic`, `emitIntegerOp` will insert conversions to the requested kind before applying the operation and back to `Generic` afterwards.

Then my question is: why always back to `Generic` afterwards? It seems perfectly fine to me to just cast e.g. the result of a `Mul` back to just `Number`. And then if some of its user will need something more `Generic` the casts will be introduced there.

share/revng/test/configuration/revng/decompilation.yml
```diff
@@ -18,7 +18,7 @@ commands:
         filter: for-decompilation
     suffix: /
     command: |-
-      revng artifact --resume "$OUTPUT" --analyze prepare-for-clift "$INPUT" -o /dev/null
+      revng artifact --resume "$OUTPUT" --analyze llvm-to-clift "$INPUT" -o /dev/null
```
pfez: Now the `llvm-to-clift` artifact is gone. This fails.

pfez: You should put `emit-c` here instead.
And change the comment at line 14, because now this isn't the initial portion of the new pipeline. It's the whole thing.

share/doc/revng/references/mime-types.md
```diff
@@ -27,6 +27,8 @@ USAGE: revng-artifact [options] <artifact> <binary>
   emit-type-definitions       - text/x.c+tar+gz
   cleanup-ir                  - application/x.llvm.bc+zstd
   prepare-for-clift           - application/x.llvm.bc+zstd
+  llvm-to-clift               - application/x.mlir.bc
```
pfez: This is gone AFAIK.

lib/mlir/Pipes/LLVMToCliftPipe.cpp
vasama-work: I kept the symbol cache but made sure to access the pertinent model properties each time. Also added comments clarifying this in each relevant location.

share/revng/test/configuration/revng/end-to-end.yml
vasama-work: Okay, I added a new `emit-c-as-single-file` artifact using the same pipe as the legacy backend, and use that in the end-to-end test. Also added the script.

share/revng/pipelines/revng-pipelines.yml
vasama-work: I removed the branches as requested and the only artifacts left now are `emit-c` and `emit-c-as-single-file`.

share/revng/test/configuration/revng/decompilation.yml
vasama-work: Done.

share/doc/revng/references/mime-types.md
vasama-work: I removed it.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
vasama-work: It was decided to leave this as is and revisit it when we do expression rewrites.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
vasama-work: It was decided to leave this as is and revisit it when we do expression rewrites.

lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp
vasama-work: It was decided to leave this as is and revisit it when we do expression rewrites.

lib/mlir/Pipes/Legalization.cpp
pfez: The discussion about symbol names was resolved offline.

We can keep an `unordered_map` here for now.

share/revng/test/configuration/revng/end-to-end.yml
aleclearmind: 1. We can postpone `emit-recompilable-archive` to when we emit headers from clift.
   Add a TODO about the fact that we should base this test on `emit-recompilable-archive` instead of mimic it by hand.
2. The point of the script was more to unify `revng.legacy.decompiled-c` and `revng.emit-and-check-c`, which are almost identical. As it is, it's not very useful.

   It's a bit an hack, but I think that if you define the `./emit-and-check-c` script in `revng.emit-and-check-c`, you can use it from other tests as well (i.e., `revng.legacy.decompiled-c`).

lib/mlir/Pipes/LLVMToCliftPipe.cpp
```diff
@@ -0,0 +1,55 @@
...
+    // properties (e.g. architecture) to avoid creating dependencies for all
+    // imported functions. The importer does some caching, but care is taken to
+    // make sure that the pertinent model properties are queried within each
+    // function import process regardless of caching.
+    auto Importer = clift::LLVMToCliftImporter::make(CliftContainer.getModule(),
+                                                     Model);
+
+    for (const model::Function &Function :
+         revng::getFunctionsAndCommit(EC, CliftContainer.name())) {
+      Importer->import(M.getFunction(NameBuilder.llvmName(Function)));
```
aleclearmind: Do not look up by name. You need to go through `revng.entry`.

lib/mlir/Pipes/LLVMToCliftPipe.cpp
aleclearmind: > but care is taken to make sure that the pertinent model properties are queried within each function import process regardless of caching.

1. Can you point me at the point where you make sure you query the model even if cached?
2. Are you 100% sure it's sensible to keep the cache at this point?
   The typical solution here is not to have the cache. It's just a footgun.

vasama-work: 1. Here: https://github.com/revng/revng/blob/48c488bcf2bcbb6324398f56cc0bc298f07dd035/lib/mlir/Dialect/Clift/Utils/ImportLLVM.cpp#L348
2. I wouldn't necessarily call it a cache. It maps `llvm::Function` to `clift::FunctionOp`. That's needed because we don't want to generate the same function declaration multiple times. This applies to helpers too.

share/revng/test/configuration/revng/end-to-end.yml
vasama-work: Done.

lib/mlir/Pipes/LLVMToCliftPipe.cpp
vasama-work: Done.

share/revng/test/configuration/revng/end-to-end.yml
aleclearmind: Nice!

lib/mlir/Pipes/LLVMToCliftPipe.cpp
aleclearmind: 2. OK then. Just please specify a little bit better that's not a *real* cache in the comment so I don't freak out when I pass by there.
   Unfortunately this kind of stuff cannot be automatically enforced so we have to be very careful with proper documentation.

vasama-work: Done.

PR #467, opened 2025-04-03
 S3 storage in python - revision 2

mrjackv:


python/revng/internal/api/rich_manager.py
```diff
@@ -0,0 +1,226 @@
...
+        super().__init__()
+        self.manager = manager
+        self.synchronizer = synchronizer
+        self.save_hooks = save_hooks
+        self.stopping = False
+        self.shutdown_event = Event()
+
+        self.lock = Lock()
+        self.submit_event = Event()
+        self.dir_list: list[str] = []
```
aleclearmind, line 34: No abbreviations

python/revng/internal/api/rich_manager.py
```diff
@@ -0,0 +1,226 @@
...
+            if self.next_trigger_time is None:
+                self.next_trigger_time = time.time() + self.interval
+
+
+class EventType(enum.StrEnum):
+    VOLATILE_CHANGE = enum.auto()
+    SAVE = enum.auto()
+
+
+class RichManager(Manager):
```
aleclearmind, line 118: `Rich` doesn't mean much. Find a better name.

python/revng/internal/api/rich_manager.py
```diff
@@ -0,0 +1,226 @@
...
+            self._storage_saver.start()
+
+            self._temporary_workdir = TemporaryDirectory(prefix="revng-manager-workdir-")
+            workdir = self._temporary_workdir.name
+
+        if synchronizer is not None:
+            synchronizer.load(Path(workdir))
+
+        self.save_lock = Lock()
+        super().__init__(workdir, flags)
```
aleclearmind, line 145: I'd outline everything above this line to a function `get_workdir` and ensure `super().__init__` is called in the first line of `__init__`.

python/revng/internal/api/rich_manager.py
aleclearmind, line 1: 1. Add comments
2. Describe thoroughly the role of each lock. What do they protect?

python/revng/internal/api/rich_manager.py
```diff
@@ -0,0 +1,226 @@
...
+        self.workdir_path = Path(self.workdir)
+
+        if self._storage_saver is not None:
+            self._wrap_storage_function("save")
+            self._wrap_storage_function("run_analysis")
+            self._wrap_storage_function("run_analyses_list")
```
aleclearmind, line 151: Why are wrapping these functions twice (here and below)?

I also think using inheritance + override is much cleaner than this `setattr` business. It's quite hacky.

python/revng/internal/api/rich_manager.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def start_countdown(self):
+        with self.lock:
+            if self.next_trigger_time is None:
+                self.next_trigger_time = time.time() + self.interval
+
+
+class EventType(enum.StrEnum):
+    VOLATILE_CHANGE = enum.auto()
+    SAVE = enum.auto()
```
aleclearmind, line 115: I'm not sure we need this. Too much decoupling.

Can't you simply wrap certain functions with a function and certain other with another?

aleclearmind, line 115: I'm not sure we need this. Too much decoupling.

Can't you simply wrap certain functions with a function and certain other with another?

The event system seems overkill.

include/revng/Pipeline/Container.h
```diff
@@ -54,6 +54,7 @@ class ContainerBase {
 
   const char *ID;
   std::string Name;
+  bool Dirty = false;
```
aleclearmind, line 57: Docs

include/revng/Pipeline/Container.h
```diff
@@ -102,26 +108,49 @@ class ContainerBase {
   /// not after(this)->enumerate().contains(Targets);
   ///
   /// returns false if nothing was removed
-  virtual bool remove(const TargetsList &Targets) = 0;
+  virtual bool removeImpl(const TargetsList &Targets) = 0;
+
+  bool remove(const TargetsList &Targets) {
+    bool Result = removeImpl(Targets);
+    Dirty = true;
```
aleclearmind, line 115: I'd set Dirty at the entry and call it a day. No intermediate vars.

No biggie though.

lib/Pipeline/ContainerSet.cpp
```diff
@@ -83,8 +83,13 @@ llvm::Error ContainerSet::store(const revng::DirectoryPath &Directory) const {
     if (Container == nullptr)
       continue;
 
+    if (not Container->isDirty())
+      continue;
```
aleclearmind, line 87: Please verify that this work and that we're actually writing less stuff.

python/revng/internal/api/rich_manager.py
mrjackv: Calling `__init__` on the first line is not a strict requirement, we're outlining a piece of code that's only called in `__init__` which adds class attributes (which makes linters unhappy) just so `__init__` can be on the first line. I'd argue to keep the code there.

python/revng/internal/api/rich_manager.py
mrjackv: Closing since it's a duplicate of the next one

mrjackv: > Can't you simply wrap certain functions with a function and certain other with another?

I'm not sure I follow, the objectives of the `RichManager` class are:
* To allow using a synchronizer if needed
* To allow triggering the event handler with some events, to trigger the autosave thread

lib/Pipeline/ContainerSet.cpp
mrjackv: I've already checked, less things are written.

python/revng/internal/api/rich_manager.py
aleclearmind: > A lot of repetition

I'm a bit skeptical that using inheritance is worse than what we have now.
`setattr`, `getattr`, two functions for wrapping declaring a function on the fly...

```python
class RichManager(Manager):

  @override
  def save(...)
    notify_save()
    super().save(...)

  def save(...)
    notify_save()
    super().save(...)

  ...

  def notify_save(self):
    if self._storage_saver is not None:
      ...
```

Pretty simple and intuitive.

> *args, **kwargs

Well... let's not do that. It's literally 5 functions.

> Having extra code because the synchronizer might not be there

Wrap unconditionally and set invoke the synchronizer only if it's there (or make a dummy synchronizer).

python/revng/internal/api/rich_manager.py
mrjackv: > I understand it's not a strict requirement, but, unless you're doing something to prepare an argument for the base class constructor, it's better to put it afterwards.

I'm not sure I understand the meaning of this sentence, we're preparing arguments for the parent's constructor, it needs to show up above `super().__init__`

python/revng/internal/api/rich_manager.py
mrjackv: `SaveAwareManager`? I'm a bit stuck on the naming

python/revng/internal/api/rich_manager.py
aleclearmind: Yes, in fact that part is fine, but I'd outline it in a function.

This way, next time someone makes a change is not tempted to add stuff before `super().__init__()`, given that the first line will be `super().__init__(get_workdir(), flags)`.

python/revng/internal/api/rich_manager.py
aleclearmind: `SyncingManager`?

python/revng/internal/api/rich_manager.py
mrjackv: 1. Done
2. Done

python/revng/internal/api/rich_manager.py
mrjackv: Sure, renamed it to `SyncingManager`

python/revng/internal/api/rich_manager.py
mrjackv: Managed to use a decorator

include/revng/Pipeline/Container.h
mrjackv: Reworked

include/revng/Pipeline/Container.h
mrjackv: Added

aleclearmind:
Merged in d1a7eff. Thanks.

PR #475, opened 2025-04-16
Remove S3StorageClient

mrjackv:
Sister PRs:
* [orchestra](https://github.com/revng/orchestra/pull/173)


include/revng/Support/OnQuit.h
```diff
@@ -12,17 +12,31 @@
...
   /// termination
   void add(std::function<void()> &&Handler) {
-    Registry.push_back(std::move(Handler));
+    Registry.push_back({ std::move(Handler), false });
+  }
+
+  /// Registers a function that can be called repeatedly
+  void addRepeatable(std::function<void()> &&Handler) {
+    Registry.push_back({ std::move(Handler), true });
   }
```
aleclearmind, line 31: What problem are you trying to solve specifically?

lib/Storage/S3StorageClient.cpp
aleclearmind, line 1: I guess we could drop much more than just the S3 storage backend.
At least add a TODO somewhere about the fact that we no longer need all this flexibility and that it could be dropped.

mrjackv: > I guess we could drop much more than just the S3 storage backend.

That would require basically reverting back to `std::string` or at the very least re-engineer a bit of stuff, but yeah.

> At least add a TODO somewhere

I will

mrjackv: Update: I managed to drop another big chunk of stuff with a little bit of effort, now we don't have any `virtual` calls anymore for storage.
@aleclearmind do check the last commit as it's new and removed all the indirection.

include/revng/Storage/LocalFile.h
aleclearmind, line 1: This file is called `LocalFile` but contains nothing called like that.

Move it to its only includer.

mrjackv: Merged into `StorageClient.h`

include/revng/Support/OnQuit.h
mrjackv: As discussed IRL, I've reworked the API so that users of `OnQuit` can additionally ask to be called on `USR1` and `USR2`

aleclearmind:
Merged in 5ab9606. Thanks.

PR #480, opened 2025-04-23
IDS on multiple Divergent Scopes

andrealinux1:
Implement IDS performing the Divergent Scopes Inlining for multiple scopes at the same time. This improves both the performances of the pass and the layout of the resulting `ScopeGraph`.

Prepended to the PR, some commits that are targeted to improve the performances by reducing the instances of `ScopeGraphBuilder`s across various cfg restructuring passes.

### DIFF
--- a/lib/RestructureCFG/DAGifyPass.cpp
+++ b/lib/RestructureCFG/DAGifyPass.cpp
@@ -20,45 +20,45 @@ using namespace llvm;
 // Debug logger
 Logger<> DAGifyPassLogger("dagify");
 
-/// Helper function which transform a retreating edge into a `goto` edge
-static void processRetreating(const revng::detail::EdgeDescriptor<BasicBlock *>
-                                &RetreatingEdge) {
-  BasicBlock *Source = RetreatingEdge.first;
-  BasicBlock *Target = RetreatingEdge.second;
-  revng_log(DAGifyPassLogger,
-            "Found retreating " << Source->getName().str() << " -> "
-                                << Target->getName().str() << "\n");
-
-  Function *F = Source->getParent();
-
-  // We should not insert a `goto` in a block already containing a
-  // `goto_block` marker
-  revng_assert(not isGotoBlock(Source));
-
-  // It must be that `Target` is a successor in the `Terminator` of `Source,`
-  // and should not be the additional, optional and unique successor
-  // representing the `scope_closer` edge, since we are not able to handle it
-  // with the following code. In principle we may support this situation, but we
-  // have now an invariant of the `ScopeGraph` which states that
-  // `scope_closer` edges should not happen to be retreating edges. This
-  // is due to the guarantee we have in how we insert such edges in all
-  // the passes of our pipeline.
-  revng_assert(llvm::any_of(graph_successors(Source),
-                            [&Target](const auto &Elem) {
-                              return Elem == Target;
-                            }));
-
-  ScopeGraphBuilder SGBuilder(F);
-  SGBuilder.makeGotoEdge(Source, Target);
-}
-
 class DAGifyPassImpl {
   Function &F;
+  const ScopeGraphBuilder SGBuilder;
 
 public:
-  DAGifyPassImpl(Function &F) : F(F) {}
+  DAGifyPassImpl(Function &F) : F(F), SGBuilder(&F) {}
 
 public:
+  /// Helper function which transform a retreating edge into a `goto` edge
+  void processRetreating(const revng::detail::EdgeDescriptor<BasicBlock *>
+                           &RetreatingEdge) {
+    BasicBlock *Source = RetreatingEdge.first;
+    BasicBlock *Target = RetreatingEdge.second;
+    revng_log(DAGifyPassLogger,
+              "Found retreating " << Source->getName().str() << " -> "
+                                  << Target->getName().str() << "\n");
+
+    Function *F = Source->getParent();
+
+    // We should not insert a `goto` in a block already containing a
+    // `goto_block` marker
+    revng_assert(not isGotoBlock(Source));
+
+    // It must be that `Target` is a successor in the `Terminator` of `Source,`
+    // and should not be the additional, optional and unique successor
+    // representing the `scope_closer` edge, since we are not able to handle it
+    // with the following code. In principle we may support this situation, but
+    // we have now an invariant of the `ScopeGraph` which states that
+    // `scope_closer` edges should not happen to be retreating edges. This
+    // is due to the guarantee we have in how we insert such edges in all
+    // the passes of our pipeline.
+    revng_assert(llvm::any_of(graph_successors(Source),
+                              [&Target](const auto &Elem) {
+                                return Elem == Target;
+                              }));
+
+    SGBuilder.makeGotoEdge(Source, Target);
+  }
+
   bool run() {
 
     // Build the `ScopeGraph` on which the `GenericRegionInfo` analysis should
--- a/lib/RestructureCFG/EnforceSingleExitPass.cpp
+++ b/lib/RestructureCFG/EnforceSingleExitPass.cpp
@@ -55,9 +55,10 @@ static bool isInfiniteLoop(const scc_iterator<Scope<BasicBlock *>> &SCCIt) {
 /// blocks, and infinite loops too.
 class EnforceSingleExitPassImpl {
   Function &F;
+  const ScopeGraphBuilder SGBuilder;
 
 public:
-  EnforceSingleExitPassImpl(Function &F) : F(F) {}
+  EnforceSingleExitPassImpl(Function &F) : F(F), SGBuilder(&F) {}
 
 public:
   bool run() {
@@ -157,7 +158,6 @@ public:
     if (TrivialExits.size() != 0) {
       for (BasicBlock *TrivialExit : skip_front(TrivialExits)) {
         revng_assert(TrivialExit != OneTrueExit);
-        ScopeGraphBuilder SGBuilder(TrivialExit->getParent());
         SGBuilder.addScopeCloser(TrivialExit, OneTrueExit);
       }
     }
@@ -167,7 +167,6 @@ public:
 
       // Build a `scope_closer` edge from each identified exit block to the
       // `sink_block`
-      ScopeGraphBuilder SGBuilder(NonTrivialExit->getParent());
       SGBuilder.addScopeCloser(NonTrivialExit, OneTrueExit);
     }
 
--- a/lib/RestructureCFG/InlineDivergentScopesPass.cpp
+++ b/lib/RestructureCFG/InlineDivergentScopesPass.cpp
@@ -11,6 +11,7 @@
 #include "llvm/IR/LLVMContext.h"
 #include "llvm/Support/GenericDomTree.h"
 
+#include "revng/ADT/Concepts.h"
 #include "revng/RestructureCFG/InlineDivergentScopesPass.h"
 #include "revng/RestructureCFG/ScopeGraphGraphTraits.h"
 #include "revng/RestructureCFG/ScopeGraphUtils.h"
@@ -49,7 +50,7 @@ static bool isConditional(BasicBlock *Node) {
   }
 }
 
-/// Helper function which detects
+/// Helper function which detects if `Node` is an exit for the `ScopeGraph`
 static bool isExit(BasicBlock *Node) {
   auto Successors = children<Scope<BasicBlock *>>(Node);
   return std::ranges::empty(Successors);
@@ -120,6 +121,26 @@ getImmediateDominator(BasicBlock *N,
     return Node->getBlock();
   } else {
     return nullptr;
+  }
+}
+
+/// Define a concept to restrict the usage of `replaceSuccessors` over a
+/// `SuccessorsToRemove` parameter of type the `SmallSet`s and `SmallVector`
+template<typename T, typename ValueType>
+concept IterableOfBasicBlockPtrs = requires(T Container) {
+  { std::begin(Container) } -> std::input_iterator;
+  { std::end(Container) };
+  { *std::begin(Container) } -> std::convertible_to<ValueType>;
+};
+
+/// Helper function which substitutes some successors in the `Terminator` with
+/// `NewTarget`
+template<IterableOfBasicBlockPtrs<BasicBlock *> Container>
+static void replaceSuccessors(Instruction *Terminator,
+                              Container &SuccessorsToRemove,
+                              BasicBlock *NewTarget) {
+  for (BasicBlock *Successor : SuccessorsToRemove) {
+    Terminator->replaceSuccessorWith(Successor, NewTarget);
   }
 }
 
@@ -280,15 +301,15 @@ electDivergence(BasicBlock *Candidate,
 }
 
 /// Helper function that performs the IDS transformation
-static void performIDS(DivergenceDescriptor &DivergenceDescriptor,
-                       BasicBlock *PlaceHolderTarget) {
+static void
+performMultipleIDS(const ScopeGraphBuilder &SGBuilder,
+                   SmallVector<DivergenceDescriptor> &MultipleDivergences,
+                   BasicBlock *PlaceHolderTarget) {
 
   // Create the new `BasicBlock` representing the `C'` conditional
   // inserted by the IDS transformation. We will refer to this block as the
   // `Tail` (of the IDS group of nodes).
-  BasicBlock *Conditional = DivergenceDescriptor.Conditional;
-  auto DivergentSuccessors = DivergenceDescriptor.DivergentSuccessors;
-  BasicBlock *Exit = DivergenceDescriptor.Exit;
+  BasicBlock *Conditional = MultipleDivergences[0].Conditional;
   LLVMContext &Context = getContext(Conditional);
   Function *F = Conditional->getParent();
   BasicBlock *Tail = BasicBlock::Create(Context,
@@ -305,6 +326,14 @@ static void performIDS(DivergenceDescriptor &DivergenceDescriptor,
   IRBuilder<> TailBuilder(Tail);
   TailBuilder.Insert(TailTerminator);
 
+  // Populate the `DivergentSuccessors` summing all ones present in each
+  // `DivergenceDescriptor`
+  SmallSet<BasicBlock *, 4> DivergentSuccessors;
+  for (auto Divergence : MultipleDivergences) {
+    DivergentSuccessors.insert(Divergence.DivergentSuccessors.begin(),
+                               Divergence.DivergentSuccessors.end());
+  }
+
   // We connect the `Conditional` to `Tail`, by making sure that all
   // the previous slots and cases going to the nondivergent exits, are now
   // connected to the `Tail` block, in order to preserve the original semantics.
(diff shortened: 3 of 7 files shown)


lib/RestructureCFG/DAGifyPass.cpp
```diff
@@ -21,7 +21,8 @@ using namespace llvm;
 Logger<> DAGifyPassLogger("dagify");
 
 /// Helper function which transform a retreating edge into a `goto` edge
-static void processRetreating(const revng::detail::EdgeDescriptor<BasicBlock *>
+static void processRetreating(ScopeGraphBuilder &SGBuilder,
```
pfez, line 24: Instead of adding an argument, just make this a method of `DAGifyPassImpl`.

lib/RestructureCFG/DAGifyPass.cpp
```diff
@@ -48,15 +49,15 @@ static void processRetreating(const revng::detail::EdgeDescriptor<BasicBlock *>
                               return Elem == Target;
                             }));
 
-  ScopeGraphBuilder SGBuilder(F);
   SGBuilder.makeGotoEdge(Source, Target);
 }
 
 class DAGifyPassImpl {
   Function &F;
+  ScopeGraphBuilder SGBuilder;
```
pfez, line 57: I haven't checked, but it should be possible to make this a `const ScopeGraphBuilder`.

You probably have to change `ScopeGraphBuilder` itself to make all methods `const` but from a cursory look that's feasible too.

Let's strive for making `ScopeGraphBuilder` and its users more `const`-correct.

lib/RestructureCFG/EnforceSingleExitPass.cpp
```diff
@@ -55,9 +55,10 @@ static bool isInfiniteLoop(const scc_iterator<Scope<BasicBlock *>> &SCCIt) {
 /// blocks, and infinite loops too.
 class EnforceSingleExitPassImpl {
   Function &F;
+  ScopeGraphBuilder SGBuilder;
```
pfez, line 58: Same here. It should be possible to make this `const`.

lib/RestructureCFG/SelectScopePass.cpp
```diff
@@ -77,12 +77,13 @@ static size_t electMaxScopeID(SmallSetVector<BasicBlock *, 2> &Predecessors,
 class SelectScopePassImpl {
   Function &F;
   PostDomTreeOnView<BasicBlock, Scope> PDT;
+  ScopeGraphBuilder SGBuilder;
```
pfez, line 80: This can be `const` too.

lib/RestructureCFG/InlineDivergentScopesPass.cpp
```diff
@@ -366,31 +507,28 @@ static bool tryIDS(Function &F, BasicBlock *PlaceHolderTarget) {
...
 
-  // If no change has been performed, we return `false` to signal this fact
-  return false;
+  return tryMultipleIDS(SGBuilder, DivergenceDescriptors, PlaceHolderTarget);
 }
 
 /// Implementation class used to run the `IDS` transformation
 class InlineDivergentScopesImpl {
   Function &F;
+  ScopeGraphBuilder SGBuilder;
```
pfez, line 583: I think this can be `const` too.

tests/unit/llvm_lit_tests/InlineDivergentScopes.ll
```diff
@@ -40,6 +40,8 @@ block_e:
 ; CHECK: block_c_ids:
 ; CHECK-NEXT:   br label %block_b
 
+; IDS on switch test
+
```
pfez, line 44: Please squash this commit with the one implementing the new IDS on multiple divergent scopes.

lib/RestructureCFG/InlineDivergentScopesPass.cpp
```diff
@@ -358,6 +401,18 @@ performMultipleIDS(ScopeGraphBuilder &SGBuilder,
     MultipleDivergences.erase(MultipleDivergences.begin() + LastElementIndex);
   }
 
+  // Decide if we need a new `Head` that chokes the entry of the divergent
```
pfez, line 404: Perhaps "collects" is better than "choke". You've used this both in comments and in the commit message. Please fix.

lib/RestructureCFG/InlineDivergentScopesPass.cpp
```diff
@@ -279,6 +279,49 @@ electDivergence(BasicBlock *Candidate,
...
+                                        Conditional->getName() + "_head_ids",
+                                        F);
+  Instruction *HeadTerminator = ConditionalTerminator->clone();
+  IRBuilder<> HeadBuilder(Head);
+  HeadBuilder.Insert(HeadTerminator);
+
+  // Collect the `Successor`s composing the local `Divergence`
+  SmallSet<BasicBlock *, 4> LocalDivergentSuccessors;
+  LocalDivergentSuccessors.insert(Divergence.DivergentSuccessors.begin(),
+                                  Divergence.DivergentSuccessors.end());
```
pfez, line 300: Doesn't copy-assignment work?
```c++
SmallSet<BasicBlock *, 4> LocalDivergentSuccessors = Divergence.DivergentSuccessors;
```

lib/RestructureCFG/InlineDivergentScopesPass.cpp
```diff
@@ -279,6 +279,49 @@ electDivergence(BasicBlock *Candidate,
...
+      LocalNonDivergentSuccessors.push_back(Successor);
+    }
+  }
+
+  // The `Head` only connects the `LocalDivergentSuccessors`, we simplify
+  // away the other successors
+  for (BasicBlock *Successor : LocalNonDivergentSuccessors) {
+    HeadTerminator->replaceSuccessorWith(Successor, PlaceHolderTarget);
+  }
+  simplifyTerminator(Head, PlaceHolderTarget);
```
pfez, line 316: The only other use of `simplifyTerminator` follows this same pattern: first you replace successors with placeholder target, then you call `simplifyTerminator`. Please refactor.

lib/RestructureCFG/InlineDivergentScopesPass.cpp
```diff
@@ -334,24 +376,123 @@ static void performIDS(DivergenceDescriptor &DivergenceDescriptor,
...
+  // We start by collecting all the `Conditional`s involved in at least one
+  // divergence
+  SmallSet<BasicBlock *, 4> Conditionals;
+  for (auto DivergenceDescriptor : DivergenceDescriptors) {
+    Conditionals.insert(DivergenceDescriptor.Conditional);
+  }
+
+  // For each conditional, we try to collect all the multiple divergences that
+  // involves such `Conditional`
+  for (BasicBlock *Conditional : Conditionals) {
```
pfez, line 488: The key of `Conditionals` is a pointer, so this iteration is non-deterministic. Make `Conditionals` a `SetVector` please.

lib/RestructureCFG/InlineDivergentScopesPass.cpp
```diff
@@ -334,24 +376,123 @@ static void performIDS(DivergenceDescriptor &DivergenceDescriptor,
...
+static bool
+canAddDescriptor(const SmallVector<DivergenceDescriptor> &Collection,
+                 const DivergenceDescriptor &NewDescriptor) {
+  // Skip checking if collection is empty
+  if (Collection.empty())
+    return true;
+
+  // Check that Conditional matches
+  if (Collection[0].Conditional != NewDescriptor.Conditional)
+    return false;
```
pfez, line 453: Can this even happen?

lib/RestructureCFG/InlineDivergentScopesPass.cpp
```diff
@@ -334,24 +376,123 @@ static void performIDS(DivergenceDescriptor &DivergenceDescriptor,
...
+  }
+}
+
+/// Helper function which returns if a `DivergenceDescriptor` is compatible with
+/// all the ones already present in a `Collection`. Being compatible means that
+/// they insist on the same conditional, have non-overlapping `Exit` blocks, and
+/// do not have any overlapping `DivergentSuccessors`
+static bool
+canAddDescriptor(const SmallVector<DivergenceDescriptor> &Collection,
+                 const DivergenceDescriptor &NewDescriptor) {
```
pfez, line 446: In general, it's clear what this function does, but it's not obvious the larger perspective of why you need to do this. What is the bad case you're guarding against?

lib/RestructureCFG/InlineDivergentScopesPass.cpp
```diff
@@ -305,6 +305,14 @@ static void performIDS(DivergenceDescriptor &DivergenceDescriptor,
   IRBuilder<> TailBuilder(Tail);
   TailBuilder.Insert(TailTerminator);
 
+  // Populate the `DivergentSuccessors` summing all ones present in each
+  // `DivergenceDescriptor`
+  SmallSet<BasicBlock *, 4> DivergentSuccessors;
+  for (auto Divergence : MultipleDivergences) {
+    DivergentSuccessors.insert(Divergence.DivergentSuccessors.begin(),
+                               Divergence.DivergentSuccessors.end());
+  }
```
pfez, line 357: Doesn't copy-assigning work?

pfez (changes requested):

lib/RestructureCFG/DAGifyPass.cpp
andrealinux1: I made this a method of `DAGifyPassImpl` here https://github.com/revng/revng/blob/3a9379ef740d1d2b9f7bcad02cc962719251e02a/lib/RestructureCFG/DAGifyPass.cpp#L32

lib/RestructureCFG/DAGifyPass.cpp
andrealinux1: OK, I made all the methods of `ScopeGraphBuilder` in this commit https://github.com/revng/revng/pull/480/commits/98f9098ddd7d7a2a9c4e2c6654e9e7696141d815.

Then, I made the field `const` in `DAGifyPass.cpp` here https://github.com/revng/revng/blob/3a9379ef740d1d2b9f7bcad02cc962719251e02a/lib/RestructureCFG/DAGifyPass.cpp#L25

lib/RestructureCFG/EnforceSingleExitPass.cpp
andrealinux1: I made this field `const` here https://github.com/revng/revng/blob/3a9379ef740d1d2b9f7bcad02cc962719251e02a/lib/RestructureCFG/EnforceSingleExitPass.cpp#L58

lib/RestructureCFG/SelectScopePass.cpp
andrealinux1: I made this member `const` here https://github.com/revng/revng/blob/3a9379ef740d1d2b9f7bcad02cc962719251e02a/lib/RestructureCFG/SelectScopePass.cpp#L80

lib/RestructureCFG/InlineDivergentScopesPass.cpp
andrealinux1: This member is now `const` here https://github.com/revng/revng/blob/3a9379ef740d1d2b9f7bcad02cc962719251e02a/lib/RestructureCFG/InlineDivergentScopesPass.cpp#L597

lib/RestructureCFG/InlineDivergentScopesPass.cpp
andrealinux1: Yeah, you are right, changed this here https://github.com/revng/revng/blob/3a9379ef740d1d2b9f7bcad02cc962719251e02a/lib/RestructureCFG/InlineDivergentScopesPass.cpp#L319

lib/RestructureCFG/InlineDivergentScopesPass.cpp
andrealinux1: Since the `for` and `replaceSuccessorWith` pattern was used multiple times in the file, I crafted a new `replaceSuccessors` helper here https://github.com/revng/revng/blob/3a9379ef740d1d2b9f7bcad02cc962719251e02a/lib/RestructureCFG/InlineDivergentScopesPass.cpp#L333C3-L333C20 and substituted all the uses

lib/RestructureCFG/InlineDivergentScopesPass.cpp
andrealinux1: Oh good catch, changed this here https://github.com/revng/revng/blob/3a9379ef740d1d2b9f7bcad02cc962719251e02a/lib/RestructureCFG/InlineDivergentScopesPass.cpp#L333C3-L333C20

lib/RestructureCFG/InlineDivergentScopesPass.cpp
andrealinux1: Mmmm I'm not sure how you want to use the copy-assignment here. In `DivergentSuccessors` I need to put all the elements inside the `DivergentSuccessors` field of each `Divergence` I iterate on.

Using the copy-assignment would be possible only once, what am I missing?

lib/RestructureCFG/InlineDivergentScopesPass.cpp
andrealinux1: Yes, we have no guarantee that the divergence we are trying to add insists on the same conditional. More on this in the following comment explaining the rationale behind `canAddDescriptor`.

lib/RestructureCFG/InlineDivergentScopesPass.cpp
pfez: Nevermind.

lib/RestructureCFG/InlineDivergentScopesPass.cpp
```diff
@@ -316,42 +385,183 @@ static void performIDS(DivergenceDescriptor &DivergenceDescriptor,
...
+    }
+  }
+
+  return true;
+}
+
+/// Helper function which attempts to perform multiple IDS transformations
+static bool
+tryMultipleIDS(const ScopeGraphBuilder &SGBuilder,
+               SmallVector<DivergenceDescriptor> &DivergenceDescriptors,
```
pfez: Can this be `const`?

lib/RestructureCFG/InlineDivergentScopesPass.cpp
pfez: It's clearer now. But it's still not clear what we do when we cannot apply the transformations simultaneously. My current understanding is that we only do the ones that are compatible and leave the others to the next iteration of `tryIDS`. But do we have a guarantee that at the end all divergent exits will be handled? If so, how?

pfez: Also, `canAddDescriptor` should probably be called something like `isCompatible`, or something like that.

lib/RestructureCFG/InlineDivergentScopesPass.cpp
```diff
@@ -316,42 +385,182 @@ static void performIDS(DivergenceDescriptor &DivergenceDescriptor,
...
   DomTreeOnView<BasicBlock, Scope> DomTree;
   DomTree.recalculate(F);
 
+  SmallVector<DivergenceDescriptor> DivergenceDescriptors;
+
   // Collect the exit nodes
   SmallVector<BasicBlock *> Exits = getExits(F, PlaceHolderTarget);
 
-  // Attempt IDS on each exit node
+  // Collect all the IDS opportunities
```
pfez: Explain why we do this in order of exit nodes and not iterating on all conditionals.

lib/RestructureCFG/InlineDivergentScopesPass.cpp
```diff
@@ -316,42 +385,182 @@ static void performIDS(DivergenceDescriptor &DivergenceDescriptor,
...
+    for (BasicBlock *ExistingSucc : Existing.DivergentSuccessors) {
+      for (BasicBlock *NewSucc : NewDescriptor.DivergentSuccessors) {
+        if (ExistingSucc == NewSucc)
+          return false;
+      }
+    }
+  }
+
+  return true;
+}
```
pfez: This should always be true by construction, as per our in-person discussion.

If anything, turn this into some logic for assertions.

lib/RestructureCFG/InlineDivergentScopesPass.cpp
```diff
@@ -316,42 +385,182 @@ static void performIDS(DivergenceDescriptor &DivergenceDescriptor,
...
+        // If we find a `DivergenceDescriptor` compatible with the current
+        // `Conditional`, we add it to the pool of candidates
+        if (isCompatible(CompatibleDivergenceDescriptors,
+                         DivergenceDescriptor)) {
+          CompatibleDivergenceDescriptors.push_back(DivergenceDescriptor);
+        }
+      }
+    }
+
+    if (CompatibleDivergenceDescriptors.size() >= 1) {
```
pfez: This should always be true.

lib/RestructureCFG/InlineDivergentScopesPass.cpp
pfez: Also you always do this once, for the first `Conditional`. The loop is pointless, just take `Conditionals.begin()`.

pfez: Unless with the latest modifications we can keep going and do all of them. If possible, that would be great.

lib/RestructureCFG/InlineDivergentScopesPass.cpp
andrealinux1: I changed the `canAddDescriptor` to `isCompatible`, and added a comment specifying that now this function is simply used as a failsafe check in an assertion here https://github.com/revng/revng/blob/1b4cabba417d846b3f4e47ca33b3f6b7b0a15234/lib/RestructureCFG/InlineDivergentScopesPass.cpp#L454

lib/RestructureCFG/InlineDivergentScopesPass.cpp
andrealinux1: The argument is now `const` here https://github.com/revng/revng/blob/1b4cabba417d846b3f4e47ca33b3f6b7b0a15234/lib/RestructureCFG/InlineDivergentScopesPass.cpp#L495

lib/RestructureCFG/InlineDivergentScopesPass.cpp
andrealinux1: I added a comment explaining the logic of iterating over the `Exit`s instead of iterating over the `Conditional`s https://github.com/revng/revng/blob/1b4cabba417d846b3f4e47ca33b3f6b7b0a15234/lib/RestructureCFG/InlineDivergentScopesPass.cpp#L557

lib/RestructureCFG/InlineDivergentScopesPass.cpp
andrealinux1: I improved the comments, and added a comment to the function specifying that this is used only inside an `assert` function https://github.com/revng/revng/blob/1b4cabba417d846b3f4e47ca33b3f6b7b0a15234/lib/RestructureCFG/InlineDivergentScopesPass.cpp#L454

lib/RestructureCFG/InlineDivergentScopesPass.cpp
andrealinux1: I completely removed this `if`, and left this check as an `assert` as a failsafe here https://github.com/revng/revng/blob/1b4cabba417d846b3f4e47ca33b3f6b7b0a15234/lib/RestructureCFG/InlineDivergentScopesPass.cpp#L525

lib/RestructureCFG/InlineDivergentScopesPass.cpp
andrealinux1: The latest iteration of the design led to the new criterion implemented in this separate commit https://github.com/revng/revng/pull/480/commits/77669464a984f9314e6b830ed5fa3d468236f9b0

lib/RestructureCFG/InlineDivergentScopesPass.cpp
```diff
@@ -516,26 +522,46 @@ tryMultipleIDS(const ScopeGraphBuilder &SGBuilder,
...
+    // corresponding exit node, which is a `goto` exit, may become a new
+    // divergent exit for another `Conditional` upper in the `ScopeGraph`. In
+    // such case, if we do not restart the collection of the divergent exits, we
+    // may process divergences already collected, before the newly introduced
+    // one. And this violates the invariant that we always process divergent
+    // `goto` exits before the non `goto` exits. And this in turn may cause
+    // suboptimalities when `IDS` is combined with `MaterializeTrivialGoto` in
+    // order to reduce the number of emitted `goto`s.
+    if (AllExitGotos
+        and CoveredSuccessors == getScopeGraphSuccessors(Conditional).size())
```
pfez: Please compute `getScopeGraphSuccessors` on `Conditional` before `performMultipleIDS` otherwise they could go out of sync.

Also please add unit tests with the examples we've seen these days.

lib/RestructureCFG/InlineDivergentScopesPass.cpp
pfez: This comment is obsolete.

lib/RestructureCFG/InlineDivergentScopesPass.cpp
pfez: :+1:

lib/RestructureCFG/InlineDivergentScopesPass.cpp
pfez: Great thanks. I'll close this conversation and make a final review.

lib/RestructureCFG/InlineDivergentScopesPass.cpp
```diff
@@ -316,42 +386,209 @@ static void performIDS(DivergenceDescriptor &DivergenceDescriptor,
...
 
-  // Attempt IDS on each exit node
+  // Collect all the IDS opportunities.
+  // We iterate over all the `Exit`s, and then search for a `Candidate`
+  // conditional node for which such `Exit` is divergent wrt. We do this by
+  // walking upward the dominator tree until we find (if it is present) the
+  // divergent `Conditional`. An alternative, which would also make the code
+  // more straightforward, would be to iterate over the `Conditional`s and start
+  // the search from there, but this would mean to perform multiple visits to
+  // the `Exit`s. With out techniques instead, even if less intuitive, we
```
pfez: Typo `s/out techniques/this technique`

andrealinux1: Fixed this here https://github.com/revng/revng/blob/c0bb76d89a9c6061d6dfd843454b522729575818/lib/RestructureCFG/InlineDivergentScopesPass.cpp#L592

lib/RestructureCFG/InlineDivergentScopesPass.cpp
andrealinux1: Updated the computation here https://github.com/revng/revng/blob/c0bb76d89a9c6061d6dfd843454b522729575818/lib/RestructureCFG/InlineDivergentScopesPass.cpp#L516.

Regarding the unit test, the correct ordering which enables us to prune away the additional `goto` is already tested in this unit test https://github.com/revng/revng/blob/develop/tests/unit/llvm_lit_tests/InlineDivergentScopesAndMaterializeTrivialGoto.ll, which is the reason why I even noticed this behavior in the first place! Without it, I would have implemented the change we discussed without caring about restarting, and introduced a suboptimal situation.

pfez: Ah very nice!

PR #487, opened 2025-06-17
Introduce backend failure handling

andrealinux1:
Handle a backend decompilation failure by printing an error message in place of the function body.

Also, fix a minor logic bug on exit dispatcher deduplication.

### DIFF
--- a/include/revng/RestructureCFG/ASTTree.h
+++ b/include/revng/RestructureCFG/ASTTree.h
@@ -69,7 +69,13 @@ private:
   links_container_expr CondExprList = {};
 
 public:
-  ASTTree() = default;
+  ASTTree() :
+    ASTNodeList(),
+    BBASTMap(),
+    ASTBBMap(),
+    RootNode(nullptr),
+    IDCounter(0),
+    CondExprList() {}
 
   // Default movable
   ASTTree(ASTTree &&) = default;
--- a/include/revng/RestructureCFG/RegionCFGTreeImpl.h
+++ b/include/revng/RestructureCFG/RegionCFGTreeImpl.h
@@ -30,6 +30,7 @@
 #include "revng/RestructureCFG/MetaRegionBB.h"
 #include "revng/RestructureCFG/RegionCFGTree.h"
 #include "revng/RestructureCFG/Utils.h"
+#include "revng/Support/Assert.h"
 #include "revng/Support/GraphAlgorithms.h"
 #include "revng/Support/IRHelpers.h"
 
@@ -287,7 +288,8 @@ inline void RegionCFG<NodeT>::dumpCFGOnFile(const std::string &FuncName,
 }
 
 template<class NodeT>
-inline bool RegionCFG<NodeT>::purgeIfTrivialDummy(BBNodeT *Dummy) {
+inline RegionCFG<NodeT>::PurgeDummyReturnCode
+RegionCFG<NodeT>::purgeIfTrivialDummy(BBNodeT *Dummy) {
   RegionCFG<NodeT> &Graph = *this;
 
   revng_assert(not Dummy->isEmpty() or Dummy->predecessor_size() != 0);
@@ -301,17 +303,24 @@ inline bool RegionCFG<NodeT>::purgeIfTrivialDummy(BBNodeT *Dummy) {
     BasicBlockNode<NodeT> *Successor = *Dummy->successors().begin();
 
     // Connect directly predecessor and successor, and remove the dummy node
-    // under analysis
+    // under analysis.
+    // The following should be an assert, but since the backend is in
+    // maintenance mode, we have an early return to propagate an early failure.
+    if (Predecessor->hasSuccessor(Successor)) {
+      return PurgeDummyReturnCode::Failure;
+    }
+
     moveEdgeTarget({ Predecessor, Dummy }, Successor);
     Graph.removeNode(Dummy);
-    return true;
+    return PurgeDummyReturnCode::Simplification;
   }
 
-  return false;
+  return PurgeDummyReturnCode::NoSimplification;
 }
 
 template<class NodeT>
-inline bool RegionCFG<NodeT>::purgeTrivialDummies() {
+inline RegionCFG<NodeT>::PurgeDummyReturnCode
+RegionCFG<NodeT>::purgeTrivialDummies() {
   RegionCFG<NodeT> &Graph = *this;
   bool RemovedNow = true;
   bool Removed = false;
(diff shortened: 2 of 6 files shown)


lib/Backend/DecompileFunction.cpp
```diff
@@ -2017,6 +2017,44 @@ static ASTVarDeclMap computeVariableDeclarationScope(const llvm::Function &F,
...
+  for (BasicBlock *BB : BlocksToDelete) {
+    BB->replaceAllUsesWith(NewBB);
+  }
+
+  // We remove all the `BasicBlock`s, in reverse order so that the old entry
+  // block which contained the local variables is deleted last (so no live uses
+  // are left around)
+  for (BasicBlock *BB : llvm::reverse(BlocksToDelete)) {
+    BB->eraseFromParent();
+  }
```
pfez: All this part can just be replaced by `F.deleteBody()`. And then you create the new block and the call to `abort`.

Just make sure that you preserve the linkage because IIRC `deleteBody` sets the linkage to external.

include/revng/RestructureCFG/ASTTree.h
```diff
@@ -69,7 +69,14 @@ class ASTTree {
   links_container_expr CondExprList = {};
 
 public:
-  ASTTree() = default;
+  ASTTree() {
```
pfez: I guess the problem here is that the `= default` constructor did not set `RootNode` to `nullptr` nor `IDCounter` to `0`.
If that's the case, fine. But the more idiomatic syntax for this is:

```c++
ASTTree() :
    ASTNodeList(),
    BBASTMap(),
    ASTBBMap(),
    RootNode(nullptr),
    IDCounter(0),
    CondExprList() {
}
```

lib/Backend/DecompileFunction.cpp
```diff
@@ -2017,6 +2017,44 @@ static ASTVarDeclMap computeVariableDeclarationScope(const llvm::Function &F,
   return computeVarDeclMap(GHAST, PendingVariables);
 }
 
+/// Helper function which blanks a `llvm::Function` whose backend decompilation
+/// failed, replaces its body with a single `BasicBlock` containing an error
+/// message
+static void blankFunction(llvm::Function &F) {
```
pfez: I'd call this `turnBodyIntoError` or something like that.

lib/RestructureCFG/RestructureCFG.cpp
```diff
@@ -1476,5 +1476,6 @@ bool restructureCFG(Function &F, ASTTree &AST) {
                  << UntanglePerformedCounter << "," << InitialWeight << "\n";
   }
 
+  // We return false to notify that not restructuring error arose
```
pfez: I think we should return `true` on success, so the call sites look like:
```c++
if (not restructureCFG()) {
  // handle error
}
```

include/revng/RestructureCFG/RegionCFGTreeImpl.h
```diff
@@ -1089,9 +1105,11 @@ inline void RegionCFG<NodeT>::inflate() {
           revng_assert(DuplicatedSuccSize == 0 or DInl == false);
 
           // Notice: after this call Duplicated is invalid if the call returns
-          // true, meaning that dereferencing it is bad. You can still use it as
-          // a key or value into maps though.
-          if (not purgeIfTrivialDummy(Duplicated)) {
+          // `Simplification`, meaning that dereferencing it is bad. You can
+          // still use it as a key or value into maps though.
+          auto ReturnCode = purgeIfTrivialDummy(Duplicated);
+          revng_assert(ReturnCode != PurgeDummyReturnCode::Failure);
```
pfez, line 1111: How come you're sure this cannot fail?

include/revng/RestructureCFG/RegionCFGTreeImpl.h
```diff
@@ -1113,7 +1131,9 @@ inline void RegionCFG<NodeT>::inflate() {
           // Notice: after this call Candidate is invalid if the call returns
           // true, meaning that dereferencing it is bad. You can still use it as
           // a key or value into maps though.
-          if (purgeIfTrivialDummy(Candidate)) {
+          ReturnCode = purgeIfTrivialDummy(Candidate);
+          revng_assert(ReturnCode != PurgeDummyReturnCode::Failure);
```
pfez, line 1135: Same here. How can you be sure this can never fail?

lib/Backend/DecompileFunction.cpp
andrealinux1: OK, I now use `deleteBody`.

In addition to restoring the `LinkageType`, you also need to restore the `Metadata`, we should remember this, it may be handful in the future.
https://github.com/revng/revng/blob/12bf4e7db02bd415387e51de8e94964c1e562576/lib/Backend/DecompileFunction.cpp#L2032

include/revng/RestructureCFG/ASTTree.h
andrealinux1: Yeah, the `default`ed constructor did left the object in a dirty state, I updated the idiomatic syntax as you suggested:
https://github.com/revng/revng/blob/12bf4e7db02bd415387e51de8e94964c1e562576/include/revng/RestructureCFG/ASTTree.h#L72

lib/Backend/DecompileFunction.cpp
andrealinux1: I renamed this as you suggested:
https://github.com/revng/revng/blob/12bf4e7db02bd415387e51de8e94964c1e562576/lib/Backend/DecompileFunction.cpp#L2024

lib/RestructureCFG/RestructureCFG.cpp
andrealinux1: Mmmm ok, I went with this convention to follow the standard UNIX way, `return 0` everything is good.

Anyway I changed this:
https://github.com/revng/revng/blob/12bf4e7db02bd415387e51de8e94964c1e562576/lib/Backend/DecompileFunction.cpp#L2068

Also, I changed the convention for `generateAST` too:
https://github.com/revng/revng/blob/12bf4e7db02bd415387e51de8e94964c1e562576/lib/RestructureCFG/RestructureCFG.cpp#L1424

And also for `inflate` too:
https://github.com/revng/revng/blob/12bf4e7db02bd415387e51de8e94964c1e562576/include/revng/RestructureCFG/GenerateAst.h#L464

So that all the error handling functions follow the same convention.

include/revng/RestructureCFG/RegionCFGTreeImpl.h
andrealinux1: Ok, the reasoning is the following:
- I needed to add the `ReturnCode` for the `purgeIfTrivialDummy` function because I found a callsite where it may fail, which is this one:

https://github.com/revng/revng/blob/12bf4e7db02bd415387e51de8e94964c1e562576/include/revng/RestructureCFG/RegionCFGTreeImpl.h#L334

If there is an error there, the error is propagated upwards.

In the other two callsites (the ones you highlighted in these comments), even though they still use `purgeIfTrivialDummy`, I did not find any case where it fails.
Therefore I left these assertions that check that no failure is raised. If they will in the future, I'll modify them when necessary.

It is a similar reasoning which led to not introduce error handling for `beautifyAST`, when I added the error handling preventively even though no failure was happening there, and you suggested to remove it, and wait for an actual failure to be triggered in order to insert the failure handling.

include/revng/RestructureCFG/RegionCFGTreeImpl.h
andrealinux1: See the reasoning in https://github.com/revng/revng/pull/487#discussion_r2156621985

include/revng/RestructureCFG/RegionCFGTreeImpl.h
pfez: Ok makes sense

PR #498, opened 2025-07-11
pypi

mrjackv:


python/pyproject_revng.toml
```diff
@@ -9,6 +9,8 @@ dependencies = [
     "PyYAML>=6.0.1,<7.0",
     "yachalk>=0.1.7,<1.0",
     "llvmcpy>=0.2.1,<1.0",
+    "gql[aiohttp]>=3.5.3,<4.0",
+    "aiohttp>=3.12.11,<4.0",
```
aleclearmind, line 13: Is `requests` pulled by any of these? I had an error about it missing.

python/revng/model/mixins.py
```diff
@@ -32,6 +37,43 @@ def get_artifact(self, artifact_name: str):
             return result
 
 
-BinaryMixin = _ArtifactMixin
+class BinaryMixin(_ArtifactMixin):
+    def get_artifact(self, artifact_name: str, targets=None):
```
aleclearmind: At this point I'd also move `analyze` and `analyses_list`.

python/revng/support/artifacts.py
```diff
@@ -189,11 +189,29 @@ class LLVMArtifact(Artifact):
...
-        return context.parse_ir(buffer)
+        try:
+            return context.parse_ir(buffer)
+        except llvmcpy.LLVMException as e:
+            if self._data[:4] != b"\x28\xb5\x2f\xfd":
+                raise e
+            # If outside of revng parse_ir cannot parse bitcode, decompress it
+            # via the zstd tool
+            with TemporaryFile("w+b") as output:
+                proc = Popen(["zstd", "-dc", "-"], stdin=PIPE, stdout=output)
```
aleclearmind: Why this over `import zstd`?

mrjackv: 1. One less package to pull in
2. `zstd` in particular uses a native module, so it either compiles its own `libzstd.so` uselessly or 99% requires making a new orchestra module just to convince it to compile with the existing library
3. We don't need any "advanced" feature (e.g. compressing individual chunks), running `zstd` does everything we need and uses the `zstd` executable, which we already use

python/pyproject_revng.toml
mrjackv: It was, but if you look at the commit I've removed the only use of `requests` in favor of `urllib.request.urlopen`

python/revng/support/artifacts.py
aleclearmind: > One less package to pull in

Assuming it's available, if not the user needs to manually install it.
This is supposed to run *out* of orchestra.

mrjackv: We have `zstd` as part of the revng root, [reference](https://github.com/revng/orchestra/blob/develop/.orchestra/config/components/revng.yml#L36)

aleclearmind: > This is supposed to run out of orchestra.

This also means we don't expect the user to do `source environment`.

python/revng/model/mixins.py
mrjackv: Done in [this PR](https://github.com/revng/revng/pull/500)

python/revng/support/artifacts.py
mrjackv: Done in [this PR](https://github.com/revng/revng/pull/500)

PR #499, opened 2025-07-14
The second iteration of debug information preservation effort

Cvelth:


lib/Canonicalize/ExitSSAPass.cpp
```diff
@@ -293,8 +293,11 @@ buildStore(BasicBlock *StoreBlock, Value *Incoming, AllocaInst *Alloca) {
          llvm::make_range(std::next(S->getIterator()), StoreBlock->end())) {
       for (Use &Operand : NextInBlock.operands()) {
         if (Operand.get() == IncomingInst) {
-          if (not LoadFromStore)
+          if (not LoadFromStore) {
             LoadFromStore = Builder.CreateLoad(IncomingInst->getType(), Alloca);
+            if (auto *IncomingInst = dyn_cast<Instruction>(Incoming))
+              LoadFromStore->setDebugLoc(IncomingInst->getDebugLoc());
```
pfez, line 299: Seems to me that the load should get the debug info from the PHINode.

lib/FunctionIsolation/RemoveExceptionalCalls.cpp
```diff
@@ -40,11 +41,16 @@ bool RemoveExceptionalCalls::runOnModule(llvm::Module &M) {
...
+    //       was dropped.
+    DILocation *DI = BB->getTerminator()->getDebugLoc();
+
     // Drop terminator of the old basic block
     eraseFromParent(BB->getTerminator());
 
     // Terminate with an unreachable
-    new UnreachableInst(C, BB);
+    auto Unreachable = new UnreachableInst(C, BB);
+    Unreachable->setDebugLoc(DI);
```
pfez: It shouldn't be unsafe to use `DI`, but if you want to be sure just do:
```cpp
   DILocation *DI = BB->getTerminator()->getDebugLoc();
   auto *OldTerminator = BB->getTerminator();
   auto Unreachable = new UnreachableInst(C, BB);
   Unreachable->setDebugLoc(DI);
   eraseFromParent(OldTerminator);
   ```

File: include/revng/LocalVariables/LocalVariableBuilder.h [hunk 3]
```diff
@@ -276,7 +276,7 @@ class LocalVariableBuilder {
   ///
   /// TODO: This method can be made const whenever we drop legacy mode.
   AssignType *createAssignmentBefore(llvm::Value *LocationToAssign,
-                                     llvm::Value *ValueToAssign,
```
pfez: Don't do this. The fact that we're not using this with anything but `Instruction`s today doesn't mean that we are not going to need a more general `Value` in the future.

lib/LocalVariables/LocalVariableBuilder.cpp
```diff
@@ -243,7 +243,7 @@ LegacyVB::createCopyFromAssignedOnUse(AssignType *Assign, Use &U) {
 template<>
 LegacyVB::AssignType *
 LegacyVB::createAssignmentBefore(Value *LocationToAssign,
-                                 Value *ValueToAssign,
+                                 Instruction *ValueToAssign,
```
pfez: Leave `Value` also here please.

lib/LocalVariables/LocalVariableBuilder.cpp
```diff
@@ -331,10 +331,11 @@ VB::CopyType *VB::createCopyFromAssignedOnUse(AssignType *Assign, Use &U) {
 
 template<>
 VB::AssignType *VB::createAssignmentBefore(Value *LocationToAssign,
-                                           Value *ValueToAssign,
+                                           Instruction *ValueToAssign,
```
pfez: Leave `Value` also here please.

lib/RestructureCFG/ScopeGraphUtils.cpp
```diff
@@ -166,6 +166,7 @@ BasicBlock *ScopeGraphBuilder::makeGotoEdge(BasicBlock *Source,
                                              F);
   IRBuilder<> Builder(Context);
   Builder.SetInsertPoint(GotoBlock);
+  Builder.SetCurrentDebugLocation(Source->getTerminator()->getDebugLoc());
```
pfez: @andrealinux1 is taking care of this in a PR of its own. Anyway this is fine. Andrea is just going to add a few others in his PR that are currently not in `develop` and that you couldn't have catched.

lib/Canonicalize/ExitSSAPass.cpp
Cvelth: 1. What do you mean? I don't see any `PHINode`s mentioned in this function? Should I dig into its callees or something?

2. I just copied what is done for the store just above it. Should the store behavior also change in the same fashion?

lib/FunctionIsolation/RemoveExceptionalCalls.cpp
Cvelth: Ah, this was a todo for myself 😅.

This should have never been pushed, my bad.

include/revng/LocalVariables/LocalVariableBuilder.h
Cvelth: Then what would you think about an additional `DebugLoc` argument?

lib/RestructureCFG/ScopeGraphUtils.cpp
Cvelth: So should I drop this commit?

lib/Canonicalize/ExitSSAPass.cpp
Cvelth: @pfez ping

lib/FunctionIsolation/RemoveExceptionalCalls.cpp
Cvelth: This is gone.

include/revng/LocalVariables/LocalVariableBuilder.h
Cvelth: @pfez ping

lib/RestructureCFG/ScopeGraphUtils.cpp
Cvelth: @pfez ping

lib/Canonicalize/ExitSSAPass.cpp
pfez: Caution: verbose explanation follows.

TLDR: you can leave the code like it is. My comment was overzealous.



1. You don't see `PHINode`s mentioned in this function, but one of the arguments is `Value *Incoming`, which is a value that is always guaranteed to be an incoming values of a `PHINode`. That's the `PHINode` I'm referring to.

2. You should leave the store as it is.

Let me give you some context. Here a `PHINode` is being decomposed into a) an `AllocaInst` (done before this function is called, passed in as the third argument), b) a `StoreInst` (one for each incoming value of the PHI, first argument of the function), c) a bunch of loads (the logic of which I'm not going into yet) that mostly represent the used of the `PHINode`.

The fundamental ideas are the following.
* The `StoreInst` represents the incoming value of the `PHINode` being written into the `AllocaInst`. So the debug info of the `StoreInst`, in principle, should be the same as the instruction computing the value being stored, because the store doesn't exist in the binay: in the binary that value is just computed.
* The `LoadInst`s represent uses of the `PHINode`, so I think they should have the same debug info as the PHI itself, pretty much like what happens [here](https://github.com/revng/revng/blob/82d52cb362251b602f601c291651b36952f325bf/lib/Canonicalize/ExitSSAPass.cpp#L458).

This is the reasoning that motivated my comment initially.

However, on a closer look, the specific loads that we're generating there do not count as uses of the `PHINode`. Instead, they are an optimization. Basically they are always ever injected in the block where the store is, so they would load the stored value directly, and in this case it is actually correct to use the same debug info as the store also on the loads.

The only place where it is correct to use the PHI's debug locations is still [this one](https://github.com/revng/revng/blob/82d52cb362251b602f601c291651b36952f325bf/lib/Canonicalize/ExitSSAPass.cpp#L458).

include/revng/LocalVariables/LocalVariableBuilder.h
pfez: That's probably the most general thing ever that you could do, but in practice I really think we only ever want to use the debug location from `ValueToAssign` if it's an instruction. So leave the prototype as it is, and just fix the body to use the debug info conditionally, when `ValueToAssign` is an `Instruction`.

lib/RestructureCFG/ScopeGraphUtils.cpp
pfez: Yes, I already have another PR from @andrealinux1 that does all this and some more.

lib/Canonicalize/HoistStructPhis.cpp
```diff
@@ -70,10 +70,12 @@ class HoistStructPhis : public llvm::FunctionPass {
...
           for (Type *ArgumentType : Call->getFunctionType()->params()) {
-            Phis.push_back(cast<PHINode>(PHINode::Create(ArgumentType,
-                                                         PhiSize,
-                                                         "",
-                                                         Phi)));
+            llvm::Instruction *NewPhi = PHINode::Create(ArgumentType,
+                                                        PhiSize,
+                                                        "",
+                                                        Phi);
+            NewPhi->setDebugLoc(Phi->getDebugLoc());
```
aleclearmind, line 77: I'm almost thinking to: introduce a `create<T>(...)` template function that can instantiate any `llvm::Instruction` (forwarding the arguments to `T::Create` or `new T`) and adds a mandatory `DebugLoc` argument.

We could enforce its usage all over the revng codebase.

What do you think?

lib/Support/IRHelpers.cpp
```diff
@@ -26,8 +26,6 @@
 #include "revng/Support/StringOperations.h"
 #include "revng/Support/Tag.h"
 
-// TODO: including GeneratedCodeBasicInfo.h is not very nice
```
aleclearmind, line 29: :+1:

lib/EarlyFunctionAnalysis/AttachDebugInfo.cpp
```diff
@@ -74,78 +78,102 @@ static bool isTrue(const llvm::Value *V) {
...
 }
 
-static void handleFunction(DIBuilder &DIB,
-                           llvm::Function &F,
-                           DISubprogram *TheSubprogram,
-                           efa::ControlFlowGraph &FM,
-                           GeneratedCodeBasicInfo &GCBI) {
-  namespace ranks = revng::ranks;
+class DebugInformationBuilder {
+  DIBuilder &DIB;
```
aleclearmind, line 84: 1. Add `private:`
2. Zero/`nullptr` initialize all the pointers, enums, scalars.

lib/EarlyFunctionAnalysis/AttachDebugInfo.cpp
```diff
@@ -149,8 +149,14 @@ class DebugInformationBuilder {
...
     for (auto *BB : ReversePostOrderTraversal(&F)) {
-    if (not GCBI.isTranslated(BB))
+      if (not GCBI.isTranslated(BB)) {
+        // This is a dispatcher-related basic block, fill its debug information
+        // with function entry point to avoid leaving it empty.
+        for (auto &I : *BB)
+          I.setDebugLoc(DefaultDI);
+
         continue;
+      }
```
aleclearmind: This is doing more than you think it does.

`not GCBI.isTranslated(BB)` implies that the block type (see `BlockType`) is not `JumpTargetBlock` or `TranslatedBlock`.

I'd say most of the other options are not attachable to any meaningful debug location, therefore using a default (as you're doing is fine).

However, `IndirectBranchDispatcherHelperBlock` is a dispatcher introduced in the code to handle an indirect jump (`jmp rax`). So all of the body of the `SwitchInst` and the related blocks should probably be attached to the debug location of the instruction of the jump instruction.

aleclearmind:
What's the status of this PR? I see several comments opened by @pfez :+1:'d but not resolved.

lib/Canonicalize/HoistStructPhis.cpp
Cvelth: Sure, but didn't we also decide that we want a wrapper over a builder? Why not hit this second bird with the same stone and add this in there right away?

lib/EarlyFunctionAnalysis/AttachDebugInfo.cpp
Cvelth: Sounds good!

Cvelth:
> I see several comments opened by @pfez 👍'd but not resolved.

They are all about the same issue which is trivial (change the type of the argument in a couple of places), I was planning to take care of it after the next review iteration.

lib/Canonicalize/HoistStructPhis.cpp
aleclearmind: I feel like people will find it boring to create a builder for having a single instruction.

You probably go from one statement to 2, maybe 3? Also, I think sometimes we might want to create an instruction without attaching to any `BasicBlock`.

Maybe it's better to leave things as they are. Don't know.

Cvelth: > Maybe it's better to leave things as they are. Don't know.

I think we should either go all the way and provide one single interface for creating _all the_ instructions so that we can in fact unify all the uses and forbid "raw llvm" altogether. Be it builder based, free function, some combination of both, whatever.

Or don't touch it too much (a very basic builder wrapper is still fine, since it's more or less trivial), but then these corner cases (not putting into a block, etc) stay flakey.

------------------

> You probably go from one statement to 2, maybe 3?

I don't see a reason not to have a _wrapper_ over the builder similar to what you're suggesting above. What I'm saying is that if we're unifying instruction creation, we should probably unify it properly and _enforce_ that **all** the instructions get created that way.

aleclearmind: I'm not sure what to do here, I'd backlog it. Let's contain refactoring and focus on features.

include/revng/LocalVariables/LocalVariableBuilder.h
Cvelth: Done.

lib/LocalVariables/LocalVariableBuilder.cpp
Cvelth: Done.

lib/LocalVariables/LocalVariableBuilder.cpp
Cvelth: Done.

lib/RestructureCFG/ScopeGraphUtils.cpp
Cvelth: Dropped.

lib/Canonicalize/HoistStructPhis.cpp
Cvelth: Backlogged.

lib/EarlyFunctionAnalysis/AttachDebugInfo.cpp
Cvelth: Done.

lib/Support/IRHelpers.cpp
Cvelth: Squashed into `Minor Improvements`.

lib/EarlyFunctionAnalysis/AttachDebugInfo.cpp
Cvelth: Done, please take a fresh look.

Cvelth:
@aleclearmind please review new (or heavily changed) commits:
- `Rename remove-broken-debug-information`
- `Introduce checkDebugLocationValidity helper`
- `AttachDebugInfo: use checkDebugLocationValidity`
- `Reduce using namespace abuse`
- `Do not open unused file streams`
- `Adopt checkDebugLocationValidity`

lib/EarlyFunctionAnalysis/AttachDebugInfo.cpp
```diff
@@ -74,78 +79,132 @@ static bool isTrue(const llvm::Value *V) {
...
+
+          revng_assert(Address.inliningIndex() == CurrentBB.inliningIndex());
+          CurrentDI = buildDI(FM.Entry(), CurrentBB, Address.start());
+
+          if (llvm::Error Error = checkDebugLocationValidity(CurrentDI)) {
+            std::string ErrorMessage = revng::unwrapError(std::move(Error));
+            revng_log(Log,
+                      "Discarding debug information from:\n"
+                        << dumpToString(I) << "Because "
+                        << std::move(ErrorMessage) << '\n');
```
aleclearmind: 1. `ErrorMessage` is not changed, no need to `std::move`.
2. Do not introduce new usages of `dumpToString`, they are super slow. Remember [this](https://pad.rev.ng/p/BC9HPOvVm#/11).

lib/EarlyFunctionAnalysis/AttachDebugInfo.cpp
```diff
@@ -74,78 +79,132 @@ static bool isTrue(const llvm::Value *V) {
...
+public:
+  void handleFunction(llvm::Function &F,
+                      efa::ControlFlowGraph &FM,
+                      GeneratedCodeBasicInfo &GCBI) {
+    BasicBlockID CurrentBB = BasicBlockID(FM.Entry());
+    DILocation *DefaultDI = buildDI(FM.Entry(), CurrentBB, FM.Entry());
+    DILocation *CurrentDI = DefaultDI;
+
+    for (auto *BB : ReversePostOrderTraversal(&F)) {
+      bool CanBeHandledNormally = GCBI.isTranslated(BB);
```
aleclearmind: 1. I'd negate this and call it `UseGenericDebugLocation`.
2. Add a comment about what we do otherwise, i.e., color instructions depending on the last call to `newpc`. Stress that visiting in RPOT is important.

lib/EarlyFunctionAnalysis/IRHelpers.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "revng/Pipeline/Location.h"
+#include "revng/Pipes/Ranks.h"
+
+// Unless your helpers need to depend on `revngPipeline` (for example, if they
+// create/parse `Location`s), you should place them in
+// `include/revng/Support/IRHelpers.h` instead.
+
+llvm::Error checkDebugLocationValidity(const llvm::DebugLoc &DebugLocation) {
+  if (not DebugLocation)
+    return revng::createError("the debug location is empty");
```
aleclearmind: I think we start with a capital letter usually.

lib/EarlyFunctionAnalysis/IRHelpers.cpp
aleclearmind, line 1: Let's not increment the number of files called `IRHelpers.h`, they are not particularly good.
Can't you move it to `Support/IRHelpers.cpp`?

lib/EarlyFunctionAnalysis/CFGAnalyzer.cpp
```diff
@@ -106,12 +106,6 @@ blockFromIndirectBranchInfo(CallBase *CI, SortedVector<efa::BasicBlock> &CFG) {
 static std::unique_ptr<llvm::raw_ostream>
 streamFromOption(const opt<std::string> &Option) {
   if (Option.getNumOccurrences() == 1) {
-    std::ifstream File(Option.c_str());
-    if (File.is_open()) {
-      int Status = std::remove(Option.c_str());
-      revng_assert(Status == 0);
-    }
-
```
aleclearmind: Can you elaborate why we were doing this in the commit message? I'm a bit confused.

lib/Canonicalize/DiscardBrokenDebugInformation.cpp
```diff
@@ -21,16 +22,15 @@ struct DiscardBrokenDebugInformation : public llvm::FunctionPass {
...
-          } else if (VerifyLog.isEnabled()) {
-            const auto &Serialized = I.getDebugLoc()->getScope()->getName();
-            revng_assert(pipeline::locationFromString(revng::ranks::Instruction,
-                                                      Serialized.str()));
-          }
+        if (llvm::Error Error = checkDebugLocationValidity(I)) {
+          std::string ErrorMessage = revng::unwrapError(std::move(Error));
+          revng_log(Log,
+                    "Discarding debug information from:\n"
+                      << dumpToString(I) << "Because "
```
aleclearmind: Never use `dumpToString` in loggers.

lib/EarlyFunctionAnalysis/IRHelpers.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "llvm/IR/DebugLoc.h"
+#include "llvm/IR/Instruction.h"
+
+#include "revng/EarlyFunctionAnalysis/IRHelpers.h"
+#include "revng/Pipeline/Location.h"
+#include "revng/Pipes/Ranks.h"
+
+// Unless your helpers need to depend on `revngPipeline` (for example, if they
+// create/parse `Location`s), you should place them in
+// `include/revng/Support/IRHelpers.h` instead.
```
aleclearmind: Uh?

lib/EarlyFunctionAnalysis/IRHelpers.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+llvm::Error checkDebugLocationValidity(const llvm::DebugLoc &DebugLocation) {
```
aleclearmind: 1. I'm not sure when these situations happen.
   Are these bugs or what?
   Would it make sense to assert on any of these?
2. Rename to `isDebugLocationInvalid`.

lib/EarlyFunctionAnalysis/AttachDebugInfo.cpp
Cvelth: 1. 🙆.
2. I'd argue that this is a kind of logger you never want to run on big binaries anyway - and making it a lot easier to use (it provides all the information you want when debugging this) on small one is worth it in my book.

lib/EarlyFunctionAnalysis/IRHelpers.cpp
Cvelth: This error is never emitted as is, it's "unwrapped" and concatenated. If anything, I'd prefer to use something like `std::optional<std::string>` here instead, but we have already gone through that discussion in a different PR.

lib/EarlyFunctionAnalysis/IRHelpers.cpp
Cvelth: > Can't you move it to `Support/IRHelpers.cpp`?

This depends on `revngPipeline` and `revngPipes`. I tried to bring a more systematic discussion about this on mattermost. Placing this in EFA for now is the most reasonable workaround I could come up with.

> Let's not increment the number of files called `IRHelpers.h`

I'm not attached to the file name, we can call this `DebugLocationChecks.h` or something, doesn't matter.

lib/EarlyFunctionAnalysis/CFGAnalyzer.cpp
Cvelth: `File` is never used - which seemed like a huge red flag. I guess this asserts that it exists. But we already assert the exact same thing right after this, when opening it as `raw_fd_ostream`.

lib/EarlyFunctionAnalysis/IRHelpers.cpp
Cvelth: What I'm trying to say in this comment is "Don't use `EFA/IRHelpers` unless you have to. Go use `Support/IRHelpers` instead".

lib/EarlyFunctionAnalysis/IRHelpers.cpp
Cvelth: 1. Similar checks were already done by `discard-broken-debug-information` pass. I just pulled them up, cleaned them up a bit, and, for consistency sake, checked all the debug locations at attachment point too. Because the problem is that all the broken locations get _silently_ discarded anyway (so discarding them earlier is better - that way they don't get propagated / never replace anything else).

> Are these bugs or what?

If anything were to be broken here, it would certainly be a bug, maybe we should in fact assert on this here.

But as soon as any llvm passes are run, we cannot afford to assert anymore - as we cannot assume debug information being correct anymore.

2. 🙆.

lib/EarlyFunctionAnalysis/AttachDebugInfo.cpp
aleclearmind: 2. > this is a kind of logger you never want to run on big binaries anyway

   In loggers we don't go by that assumption. Loggers need to be usable on big binaries as well.
   
   > making it a lot easier to use
   
   I'm not arguing on how useful it is, but if you want to use it, you have to use `ModuleSlotTracker`. Please let's always follow this rule, no exceptions.

lib/EarlyFunctionAnalysis/IRHelpers.cpp
aleclearmind: > This error is never emitted as is, it's "unwrapped" and concatenated

Let's not make non-local assumptions.
Wherever you emit it do something like:

```
We failed due to the following error:
$OTHER_ERROR
```

Cvelth: Sad, but ok

lib/EarlyFunctionAnalysis/IRHelpers.cpp
aleclearmind: Can't you move it to `revngPipes`?

All of this will require some love eventually. Let's not add dependencies between unrelated libraries (`revngCanonicalize` and `revngEarlyFunctionAnalysis`).

lib/EarlyFunctionAnalysis/CFGAnalyzer.cpp
aleclearmind: OK, add this to the commit message.

lib/EarlyFunctionAnalysis/IRHelpers.cpp
aleclearmind: If you want to have something like this add a directive first like "Don't add functions here..."

lib/EarlyFunctionAnalysis/IRHelpers.cpp
Cvelth: Sure, `revngPipes` also works!

I am just trying to avoid touching "soon-to-be-discarded" parts unnecessarily. Because, from the past experience, when those are discarded all the "still used" things just get placed somewhere random.

lib/EarlyFunctionAnalysis/IRHelpers.cpp
Cvelth: Noted!

That being said, unless we end up keeping the `IRHelpers.h` name (doubt it), this comment will go away anyway.

lib/EarlyFunctionAnalysis/IRHelpers.cpp
aleclearmind: I think I'm missing the definition of "valid" debug info.
I don't think LLVM passes create "invalid" debug info in any general sense, I guess there's some implicit definition of what's valid in this context.
It also needs to be added to the code.

Cvelth: What about:
```cpp
// In revng modules, valid debug information location is one that is:
// - non-empty (`!!DebugLoc`),
// - has a scope (`DebugLoc->getScope()`) with a non-empty name,
// - where the name is a valid `/instruction/...` location.
// (the last one is subject to change when we start attaching more than one
//  address per llvm instruction).
```
?

aleclearmind: Works

lib/EarlyFunctionAnalysis/AttachDebugInfo.cpp
Cvelth: Done.

2, in couple of other places too, see `Introduce ModuleSlotTracker-aware dumpToString` and `Adopt ModuleSlotTracker-aware dumpToString` commits.

lib/EarlyFunctionAnalysis/AttachDebugInfo.cpp
Cvelth: 1. I went with `UseFallbackDebugLocation` to stress even further that this is not good.
2. Done:
```cpp
      // There are two options of how we can handle basic block addresses:
      //
      // - in the general case, addresses provided by the `newpc` helper calls
      //   are used to fill in the metadata of all the instructions in-between
      //   (note that the reverse post order traversal is important).
      //
      // - when that is not possible (which most likely means that the block
      //   in question is artificial), the fallback is to use the most basic
      //   possible location:
      //   ```
      //   /instruction/<function-entry>/<function-entry>/<function-entry>
      //   ```
      //
      // The following flag decides which approach is used for this basic block:
      bool UseFallbackDebugLocation = !GCBI.isTranslated(BB);

      if (getType(BB) == BlockType::IndirectBranchDispatcherHelperBlock) {
        // These helper blocks are introduced to handle indirect jumps (for
        // example, `jmp rax`). But, since CFG around them is reasonable AND
        // because we're traversing them in the reverse post order, we can let
        // normal `newpc`-based address setter do its job for them too.
        UseFallbackDebugLocation = false;
      }

      // TODO: keep a close eye on this, especially if we ever add more basic
      //       block types, as using the default location is pretty much
      //       the worst option available.
      if (UseFallbackDebugLocation) {
        for (auto &I : *BB)
          I.setDebugLoc(DefaultDI);

        continue;
      }
```

lib/EarlyFunctionAnalysis/IRHelpers.cpp
Cvelth: Done.

lib/EarlyFunctionAnalysis/IRHelpers.cpp
Cvelth: Done.

P. S. Turns out it wasn't even a new dependency, `revngCanonicalize` already depends on `revngEarlyFunctionAnalysis` through `revngInitModelTypes`.

lib/EarlyFunctionAnalysis/CFGAnalyzer.cpp
Cvelth: Actually, my bad, this was _removing_ the existing file to then open it in "append mode" for whatever reason (instead of just opening it normally).

But seeing how this does weird stuff, I ended up leaving it alone :D

lib/Canonicalize/DiscardBrokenDebugInformation.cpp
Cvelth: Done, see `Adopt ModuleSlotTracker-aware dumpToString`.

lib/EarlyFunctionAnalysis/IRHelpers.cpp
Cvelth: Gone.

lib/EarlyFunctionAnalysis/IRHelpers.cpp
Cvelth: Done.

aleclearmind:
Merged in ccb0180. Thanks.

PR #500, opened 2025-07-15
scripting improvements

mrjackv:
Sister PRs:
* [orchestra](https://github.com/revng/orchestra/pull/187)


python/revng/tupletree/__init__.py
```diff
@@ -259,23 +259,23 @@ def __str__(self):
...
+        self._accessor: Optional[Callable[[str], StructBase]] = None
 
-    @classmethod
-    def create(cls, root_type, obj):
-        ref_str = root_type.get_reference_str(obj)
-        return cls(ref_str, referenced_obj=obj)
+    def get(self) -> StructBase:
+        assert self.is_valid(), "Reference is not valid"
+        assert self._accessor is not None
+        return self._accessor(self._ref_str)
```
aleclearmind, line 272: I've this bug:

```
address = MetaAddress(0x400930, Type=MetaAddressType.Code_x86_64)
function = project.model.Functions[address]
function.Prototype.Definition.get().Arguments[0].Type.PointeeType.Definition.get()
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[45], line 1
----> 1 function.Prototype.Definition.get().Arguments[0].Type.PointeeType.Definition.get()

File ~/venv/lib/python3.12/site-packages/revng/tupletree/__init__.py:271, in Reference.get(self)
    269 def get(self) -> StructBase:
    270     assert self.is_valid(), "Reference is not valid"
--> 271     assert self._accessor is not None
    272     return self._accessor(self._ref_str)

AssertionError: 

```

I can provide the model/binary, if needed.

python/revng/support/artifacts.py
```diff
@@ -189,11 +190,21 @@ class LLVMArtifact(Artifact):
...
         buffer = llvmcpy.create_memory_buffer_with_memory_range_copy(
             self._data, len(self._data), name
         )
-        context = llvmcpy.get_global_context()
-        return context.parse_ir(buffer)
+        try:
+            return context.parse_ir(buffer)
+        except llvmcpy.LLVMException as e:
+            if self._data[:4] != zstd.FRAME_HEADER:
+                raise e
```
aleclearmind: Can't we preemptively check this instead of doing the trial and error thing?

python/revng/tupletree/__init__.py
```diff
@@ -310,10 +308,39 @@ def _field_is_default(field, value):
     return value == factory()
 
 
+class TypedListDescriptor:
```
aleclearmind, line 311: Docs

python/revng/tupletree/__init__.py
```diff
@@ -726,8 +726,7 @@ def _make_diff_subtree(
     differ we add to the `result`.
     """
     result = []
-    if type(obj_old) is not type(obj_new):
-        return []
+    assert type(obj_old) is type(obj_new)
```
aleclearmind: Maybe raise `ValueError`?

python/revng/support/artifacts.py
```diff
@@ -217,6 +219,25 @@ def show(self):
...
+class RecompilableArchiveArtifact(Artifact):
+    MIMES = ("application/x.recompilable-archive",)
+
+    def parse(self) -> pycparser.c_ast.FileAST:
+        with TemporaryDirectory() as temp_dir:
+            with tarfile.open(fileobj=BytesIO(self._data), mode="r:*") as tar:
+                tar.extractall(path=temp_dir)
+
+            c_file = Path(temp_dir) / "decompiled/functions.c"
+            args = ["-E", "-DDISABLE_ATTRIBUTES", "-DDISABLE_FLOAT16"]
```
aleclearmind: I get 

```
clang-16: warning: argument unused during compilation: '-E' [-Wunused-command-line-argument]
```

If I drop `-E`, I get even more messages :thinking:

share/doc/revng/references/mime-types.md
```diff
@@ -21,7 +21,7 @@ USAGE: revng-artifact [options] <artifact> <binary>
   make-segment-ref            - application/x.llvm.bc+zstd
   decompile                   - text/x.c+ptml+tar+gz
   decompile-to-single-file    - text/x.c+ptml
-  emit-recompilable-archive   - application/x-tar
+  emit-recompilable-archive   - application/x.recompilable-archive
```
aleclearmind: Mmmh, I'm not super happy to drop the MIME of tar.

Is this compressed? Below we do `+tar+gz`.

share/revng/include/primitive-types.h
```diff
@@ -166,7 +166,8 @@ extern uint128_t undef_uint128_t();
 // Float
 //
 
-#if __ARM_FP16_ARGS == 1 || defined(__FLT16_MIN__)
+#if (__ARM_FP16_ARGS == 1 || defined(__FLT16_MIN__)) \
+  && !defined(DISABLE_FLOAT16)
```
aleclearmind: I'd prefer to hoist the `DISABLE_XXX` macros at the top, as a form of documentation, as in my original proposal.

python/revng/support/artifacts.py
mrjackv: The check is backwards: it's unknown whether the llvm version is ours which is patched to read `zstd`s or not, the only way is to try to read it and then do something if an exception is thrown

python/revng/tupletree/__init__.py
mrjackv: Please do, together with the script.

share/doc/revng/references/mime-types.md
mrjackv: Would `application/x.recompilable-archive+tar+gz` work?

share/revng/include/primitive-types.h
mrjackv: Sure

python/revng/support/artifacts.py
mrjackv: That's kind of weird, clang explicitly says that `-E` is supported. If you drop the `-E` it tries to compile, so it's normal that it produces more errors.
The only hunch I have is that it's in reality `clang-cpp` and it's complaining that it got `-E` twice or something like that.

python/revng/support/artifacts.py
aleclearmind: My idea was no to rely on LLVM supporting this at all.

This library is primarily designed to be used *out of* orchestra.

python/revng/support/artifacts.py
aleclearmind: `strace -f -e execve`?

share/doc/revng/references/mime-types.md
aleclearmind: Yup

python/revng/support/artifacts.py
mrjackv: Sure, but if we *happen* to be within orchestra we save having to call zstd.
If you prefer we do this regardless, then that's fine

aleclearmind: I don't think we'll ever use this within orchestra.

But in any case, we're just decompressing in Python (but still using C code, I assume) as opposed to do the same within our LLVM fork. I don't think there's a performance penalty.

If we can avoid dealing with exceptions, let's do it.

There are many reasons why there could be a `llvmcpy.LLVMException`. Here we are suppressing the exception and trying to decompress it.
If the reason for the failure was *not* due to the compression (let's assume it was not compressed), we're hiding the error and reporting the zstd failure to decompress (it wasn't compressed). Quite sad.

share/doc/revng/references/mime-types.md
mrjackv: Done

python/revng/tupletree/__init__.py
mrjackv: Done

python/revng/tupletree/__init__.py
mrjackv: Done

python/revng/support/artifacts.py
mrjackv: Fixed, now it check and uncompressed regardless if needed

python/revng/support/artifacts.py
mrjackv: I've dug a little bit but it seems that both `clang` and `clang-cpp` report the warning if invoked with it. Strace showed that both are run as:
```
[pid 3669642] execve("$ORCHESTRA_ROOT/root/bin/clang-cpp", ["clang-cpp", "-E", "-DDISABLE_ATTRIBUTES", "-DDISABLE_FLOAT16", "/tmp/tmpp619leay/decompiled/functions.c"], 0x7ffc71613170 /* 74 vars */) = 0
```

So there's nothing "strange" going on. I've dug a bit in the llvm commit history and found [this commit](https://github.com/llvm/llvm-project/commit/9478f661c26fbc22491218477917df5d8d73c51c#diff-b8f2c49144b39e839dc1da0805dd7561e1a03dbcc9d5ead4cb373e2978c641b1R784) (see `Options.td` if it doesn't open immediately) where the visibility of the `-E` option changed, but I'm not sure if that's the reason it reports the error.

In any case I can remove both or them or add the `-Wno` for both

python/revng/tupletree/__init__.py
mrjackv: Fixed

python/revng/support/artifacts.py
aleclearmind: I did a bit of testing, I think you either use `["clang", "-E"]` or just `["clang-cpp"]`. `["clang-cpp", "-E"]` does not make sense. `-E` instructs the compiler driver to invoke the preprocessor. `cpp` *is* the preprocessor, no `-E` is needed.

mrjackv: Ok, I must have forgot to `orc install` or something else because I would have sworn that also `clang -E` reported the error.
I've pushed the fixed version that only adds `-E` to `gcc` and `clang`.

PR #501, opened 2025-07-18
ScopeGraph: remove the `ScopeGraphLogger` library

andrealinux1:


include/revng/RestructureCFG/ScopeGraphGraphTraits.h
```diff
@@ -391,35 +391,45 @@ struct llvm::DOTGraphTraits<Scope<const llvm::Function *>>
...
-  llvm::dbgs() << "Serializing the dot file representing the ScopeGraph\n";
+  Output = Output + "Serializing the dot file representing the ScopeGraph\n";
   llvm::WriteGraph<Scope<llvm::Function *>>(&F, "ScopeGraph-" + F.getName());
+
+  return Output;
+}
+
+// Debug function used to dump a serialized representation of the `ScopeGraph`
+// on stderr
+debug_function inline void dumpScopeGraph(llvm::Function &F) {
```
aleclearmind, line 433: 1. Move `debug_function` at the end of the line.
2. Use doxygen for documentation
3. Never use `llvm::dbgs()`, use `dbg <<` ([C.7](https://pad.rev.ng/s/SJNUet_ku#))

lib/RestructureCFG/ScopeGraphLoggerPass.cpp
```diff
@@ -17,7 +27,10 @@ class ScopeGraphLoggerPassImpl {
 
 public:
   bool run() {
-    dumpScopeGraph(F);
+    std::ofstream Output;
+    std::ostream &OutputStream = pathToStream(OutputPath, Output);
```
aleclearmind: Drop all the usages of `pathToStream` and adopt `llvm::ToolOutputFile` instead.

include/revng/RestructureCFG/ScopeGraphGraphTraits.h
andrealinux1: 1. Are you sure about this? The only way I understood to have this is to put the `-Wno-gcc-compat` for `cmake`, otherwise a portability error with `gcc` is raised...

lib/RestructureCFG/ScopeGraphLoggerPass.cpp
andrealinux1: - I dropped the current PR usage of `pathToStream`, and adopted `llvm::ToolOutputFile` (super nice btw, didn't know about it).
- Do you also want me to remove all the other usages of `pathToStream` in the codebase?
- Do you also want me to drop the `pathToStream` definition in `include/revng/Support/CommandLine.h`?

include/revng/RestructureCFG/ScopeGraphGraphTraits.h
Cvelth: 1. Ale means like [this](https://github.com/revng/revng/blob/develop/include/revng/ABI/FunctionType/Layout.h#L199).

(sorry for butting in)

andrealinux1: Eh but that's a class method, in the case of a free function like this I get the 
```home/andrea/big-data/orchestra-clift/sources/revng/include/revng/RestructureCFG/ScopeGraphGraphTraits.h:433:47: error: GCC does not allow 'used' attribute in this position on a function definition [-Werror,-Wgcc-compat]
inline void dumpScopeGraph(llvm::Function &F) debug_function {
                                              ^
/home/andrea/big-data/orchestra-clift/sources/revng/include/revng/Support/Debug.h:20:39: note: expanded from macro 'debug_function'
#define debug_function __attribute__((used, noinline))
```
error, unless we specify the `-Wno-gcc-compat` directive for `cmake`

Cvelth: Wow, I just grepped the codebase and have indeed only ever seen on methods 😲. Today I learned!

andrealinux1: Eh the error puzzled me too at first glance, didn't understand why the compiler was complaining only on my function:)

lib/RestructureCFG/ScopeGraphLoggerPass.cpp
aleclearmind: 1. :+1:
2. Yes
3. Yes

lib/RestructureCFG/RestructureCFG.cpp
```diff
@@ -1492,16 +1493,23 @@ bool restructureCFG(Function &F, ASTTree &AST) {
...
-    OutputStream << "function,"
-                    "duplications,percentage,tuntangle,puntangle,iweight\n";
-    OutputStream << F.getName().data() << "," << DuplicationCounter << ","
-                 << Increase << "," << UntangleTentativeCounter << ","
-                 << UntanglePerformedCounter << "," << InitialWeight << "\n";
+    llvm::ToolOutputFile OutputFile(MetricsOutputPath + "/" + FunctionName,
+                                    EC,
+                                    llvm::sys::fs::OF_Text);
+    if (EC)
+      ExitOnError(llvm::createStringError(EC, EC.message()));
```
aleclearmind: `ExitOnError` is OK only on `Main.cpp`. Here, using the message.
Same thing below.

andrealinux1: It's not clear what you mean with `Here, using the message`. Do you want me to `revng_abort` with `EC.message`?

aleclearmind: Yes

andrealinux1: I suspected so it's already done

lib/RestructureCFG/ScopeGraphLoggerPass.cpp
```diff
@@ -17,7 +28,15 @@ class ScopeGraphLoggerPassImpl {
 
 public:
   bool run() {
-    dumpScopeGraph(F);
+    ExitOnError ExitOnError;
+    std::error_code EC;
+
+    llvm::ToolOutputFile OutputFile(OutputPath, EC, llvm::sys::fs::OF_Text);
+    if (EC)
+      revng_abort(EC.message().c_str());
```
aleclearmind:
```
revng_assert(!EC, EC.message().c_str());
```

andrealinux1: Changed in both the spots

PR #505, opened 2025-07-28
boto3: disable multipart uploads

mrjackv:


aleclearmind:
Merged in 615b2a6. Thanks.

PR #506, opened 2025-07-28
Fixes 2025-07-28

mrjackv:


python/revng/internal/daemon/__init__.py
```diff
@@ -189,6 +202,7 @@ def shutdown():
         routes=routes,
         on_startup=[startup],
         on_shutdown=[shutdown],
+        exception_handlers={ClientDisconnect: client_disconnect_handler},  # type: ignore
```
aleclearmind, line 224: Have you been able to test this?

python/revng/internal/daemon/__init__.py
```diff
@@ -124,6 +125,27 @@ async def client_disconnect_handler(request: Request, exc: ClientDisconnect) ->
     return NoopResponse()
 
 
+def all_500_starlette(hooks: PluginHooks) -> Starlette:
```
aleclearmind, line 120: Dumb question again, have you been to test this?

python/revng/internal/daemon/__init__.py
mrjackv: I've tried but I haven't been able to yank the connection hard enough to make it trigger "naturally"
Manually triggering the exception does however trigger it, so I'm confident it works

python/revng/internal/daemon/__init__.py
mrjackv: Yes, it works

PR #508, opened 2025-08-05
Introduce the `MaterializeLoopScopes` pass

andrealinux1:
The `MaterializeLoopScopes` pass, does the following for each `GenericRegion`:
1) Normalize the abnormal entries.
2) Tries to find a suitable unique exit point for every `GenericRegion`, and inserts a `scope_closer_ edge from the elected entry to it.

In this PR, we also introduce:
- The debug metadata preservation for the `dagify`, `ids` and `select-scope` passes, with the goal of improving the preservation of the metadata.
- Restructure the `DAGify` pass in order to include the abnormal entry phase.
- Various minor fixes.

### DIFF
--- a/share/revng/pipelines/revng-pipelines.yml
+++ b/share/revng/pipelines/revng-pipelines.yml
@@ -686,6 +686,7 @@ Branches:
               - inline-divergent-scopes
               - enforce-single-exit
               - materialize-trivial-goto
+              - materialize-loop-scopes
               - select-scope
               - inline-divergent-scopes
               - enforce-single-exit


lib/RestructureCFG/DAGifyPass.cpp
```diff
@@ -61,14 +61,11 @@ class DAGifyPassImpl {
...
-    Scope<Function *> ScopeGraph(&F);
-
-    // Build and run the `GenericRegionInfo` analysis on the `ScopeGraph`
-    GenericRegionInfo<Scope<Function *>> RegionInfo;
+    // We instantiate and run the `GenericRegionInfo` analysis on the raw CFG,
+    // and not on the `ScopeGraph`
+    GenericRegionInfo<Function *> RegionInfo;
     RegionInfo.clear();
-    RegionInfo.compute(ScopeGraph);
+    RegionInfo.compute(&F);
```
pfez, line 123: The commit message for this doesn't describe what the commit does.

The `GenericRegionInfo` was *already* computed inside the `DAGify` pass. What's changing is that now we do it on the raw CFG instead of on the `ScopeGraph`.
Please fix the commit message.
And add comments to the code to explain why we decided to do it on the CFG and what are the invariants that need to be verified for it to work.

tests/unit/llvm_lit_tests/DAGify.ll
```diff
@@ -112,7 +112,7 @@ block_e:
 
 ; CHECK-LABEL: define void @h(i1 noundef %a, i1 noundef %b)
 ; CHECK: block_a:
-; CHECK-NEXT:   br i1 %a, label %block_b, label %block_c
+; CHECK-NEXT:   br i1 %a, label %goto_block_b, label %block_c
```
pfez, line 115: Squash this fix to the unit tests in the commit that introduces late entry normalization.

lib/RestructureCFG/ScopeGraphUtils.cpp
```diff
@@ -292,3 +292,99 @@ void verifyScopeGraphAnnotations(const BasicBlock *BB) {
...
 }
+
+template<IterableOfBasicBlockPtrs<BasicBlock *> Container>
+void replaceSuccessors(Instruction *Terminator,
+                       Container &SuccessorsToRemove,
+                       BasicBlock *NewTarget) {
+  for (BasicBlock *Successor : SuccessorsToRemove) {
+    Terminator->replaceSuccessorWith(Successor, NewTarget);
+  }
+}
```
pfez: Just leave this in the header so you don't need the explicit instantiations here in the `.cpp`, and if at some point we need this on another kind of container we don't need to add yet another instantiation.

include/revng/RestructureCFG/ScopeGraphUtils.h
```diff
@@ -76,3 +78,24 @@ void verifyScopeGraphAnnotationsImpl(FunctionTags::Tag &Tag,
                                      const llvm::BasicBlock *BB);
 
 void verifyScopeGraphAnnotations(const llvm::BasicBlock *BB);
+
+/// Define a concept to restrict the usage of `replaceSuccessors` over a
+/// `SuccessorsToRemove` parameter of type the `SmallSet`s and `SmallVector`
+template<typename T, typename ValueType>
```
pfez: `ValueType` is not necessary here. Just use `llvm::BasicBlock *`, which BTW is also already what the name of the concept implies.

pfez: Actually, in `include/revng/ADT/Concepts.h` we already have the `RangeOf` concept, which is just what you want here. No need to make a new separate dedicated concept for this.

lib/RestructureCFG/DAGifyPass.cpp
```diff
@@ -21,6 +21,19 @@ using namespace llvm;
...
+/// later checked for consistency
+static void insertHeadMD(BasicBlock *Head) {
+
+  // We attach a metadata to the terminator instruction of the `Head`
+  // block. Later on, we will check during `MaterializeLoopScopes` that
+  // the `Head` block of each `GenericRegion` will be the same
+  Instruction *HeadTerminator = Head->getTerminator();
+  QuickMetadata QMD(getContext(HeadTerminator));
+  auto *HeadMD = QMD.tuple("genericregion-head");
+  HeadTerminator->setMetadata("genericregion-head", HeadMD);
```
pfez: This ends up as:
```llvm
!2 = !{!"genericregion-head"}
; [...]
!genericregion-head !2
```

It's kind of ugly because the `genericregion-head` thing is duplicated as both the name of the `NamedMetadata` and as the string it contains.

Either we have some better information to embed in the tuple than just the `"genericregion-head"` string, or just emit an empty tuple for the named metadata.

lib/RestructureCFG/MaterializeLoopScopes.cpp
```diff
@@ -0,0 +1,397 @@
...
+  // We must find the metadata by design
+  revng_assert(MD);
+  auto *Tuple = dyn_cast_or_null<MDTuple>(MD);
+  revng_assert(Tuple);
+
+  // We check that the metadata contains the expected tag
+  Function *F = Head->getParent();
+  QuickMetadata QMD(getContext(F));
+  auto HeadMD = QMD.extract<StringRef>(Tuple, 0);
+  revng_assert(HeadMD == "genericregion-head");
```
pfez: This additional check may turn out to be redundant if we get rid of the string, which doesn't actually convey any meaning.

lib/RestructureCFG/MaterializeLoopScopes.cpp
```diff
@@ -0,0 +1,397 @@
...
+char MaterializeLoopScopes::ID = 0;
+static constexpr const char *Flag = "materialize-loop-scopes";
+using Reg = llvm::RegisterPass<MaterializeLoopScopes>;
+static Reg X(Flag, "Perform the materialization of loop scopes transformation");
+
+bool MaterializeLoopScopes::runOnFunction(llvm::Function &F) {
+
+  // Instantiate and call the `Impl` class
+  MaterializeLoopScopesImpl IDSImpl(F);
+  bool FunctionModified = IDSImpl.run();
```
pfez: `IDSImpl`?

lib/RestructureCFG/MaterializeLoopScopes.cpp
```diff
@@ -0,0 +1,397 @@
...
+public:
+  MaterializeLoopScopesImpl(Function &F) : F(F), SGBuilder(&F) {}
+
+public:
+  bool run() {
+
+    // We instantiate and run the `GenericRegionInfo` analysis on the raw CFG,
+    // and not on the `ScopeGraph`
+    GenericRegionInfo<Function *> RegionInfo;
+    RegionInfo.clear();
```
pfez: Calling `clear` shouldn't be needed. `RegionInfo` is just a local variable. If calling `clear` is needed then there's something fishy with the implementation of `GenericRegionInfo<Function *>`.

lib/RestructureCFG/MaterializeLoopScopes.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+        // If we reach this point, it means that all the preconditions are
+        // satisfied, and that we can elect the `UniqueSuccessor`
+
+        // We mark the current `Function` as modified
+        FunctionModified = true;
+
+        // We insert the `scope_closer` to `UniqueSuccessor`
+        enforceUniqueSuccessor(SGBuilder, Head, *UniqueSuccessor);
+      }
```
pfez: The logic of point 2.(A,B,C) is quite messy. It has 3 sub-cases, and for each of them it does an early `continue` in the loop.

Please restructure it so that it has just three stages:
* first phase collect exiting blocks and successors
* second phase tries to find a unique successor; it has to be possible for this phase to fail; this is where steps A, B, C will live
* third phase, when the unique successor is found, independently of the logic, just sets `FunctionModified` to `true` and enforces the single successor.

lib/RestructureCFG/MaterializeLoopScopes.cpp
```diff
@@ -0,0 +1,397 @@
...
+  }
+}
+
+/// Helper function used to elect the `GenericRegion` successor when a `Footer`
+/// is needed
+static BasicBlock *
+enforceFooterSuccessor(GenericRegion<BasicBlock *> *Region,
+                       SmallSet<BasicBlock *, 4> &ExitingBlocks,
+                       SmallSet<BasicBlock *, 4> &UniqueSuccessorCandidates) {
+  BasicBlock *UniqueExitingBlock = *ExitingBlocks.begin();
```
pfez, line 108: Assert that you only have a node in `ExitingBlocks`.

lib/RestructureCFG/MaterializeLoopScopes.cpp
```diff
@@ -0,0 +1,397 @@
...
+        // We collect all the blocks from which there is a exiting edge
+        // from the `GenericRegion`
+        ExitingBlocks.insert(RegionNode);
+      }
+    }
+  }
+}
+
+/// Helper function used to elect the `GenericRegion` successor when a `Footer`
+/// is needed
```
pfez: Mmh this doesn't just elect the successor. It actually creates it. Fix this comment.

pfez (changes requested):

lib/RestructureCFG/DAGifyPass.cpp
andrealinux1: You are right, I modified the commit message updating it https://github.com/revng/revng/pull/508/commits/a6fb1c3179eb83f5aaf62a2d6392c849ac5cd9b1.

However, in `DAGify` it doesn't matter if we do it on the raw CFG or on the `ScopeGraph`. We do it on the raw CFG just to be coherent with what we later do in `MaterializeLoopScopes`.

Therefore, I added a comment in `MaterializeLoopScopes` explain that we do it on the raw CFG.

tests/unit/llvm_lit_tests/DAGify.ll
andrealinux1: Squashed this in https://github.com/revng/revng/pull/508/commits/8afb755eb7e5762cecd4c09f0850a0153d9d71ba

lib/RestructureCFG/ScopeGraphUtils.cpp
andrealinux1: I moved this in the header to avoid the explicit instantiations https://github.com/revng/revng/blob/64bc362e7664a5a872f22768e0d73ad57162e8ac/include/revng/RestructureCFG/ScopeGraphUtils.h#L94

lib/RestructureCFG/DAGifyPass.cpp
andrealinux1: Ok, now the metadata is just an empty tuple https://github.com/revng/revng/blob/64bc362e7664a5a872f22768e0d73ad57162e8ac/lib/RestructureCFG/DAGifyPass.cpp#L33

lib/RestructureCFG/MaterializeLoopScopes.cpp
andrealinux1: The check has now been simplified https://github.com/revng/revng/blob/64bc362e7664a5a872f22768e0d73ad57162e8ac/lib/RestructureCFG/MaterializeLoopScopes.cpp#L62

lib/RestructureCFG/MaterializeLoopScopes.cpp
andrealinux1: Probably I used the IDS code as the skeleton for this pass for avoiding rewriting the `Impl` ballet thing, and forgot to update this variable name https://github.com/revng/revng/blob/64bc362e7664a5a872f22768e0d73ad57162e8ac/lib/RestructureCFG/MaterializeLoopScopes.cpp#L388

lib/RestructureCFG/MaterializeLoopScopes.cpp
andrealinux1: I don't know why this clear was here, I removed it https://github.com/revng/revng/blob/64bc362e7664a5a872f22768e0d73ad57162e8ac/lib/RestructureCFG/MaterializeLoopScopes.cpp#L334

lib/RestructureCFG/MaterializeLoopScopes.cpp
andrealinux1: Asserted this here https://github.com/revng/revng/blob/64bc362e7664a5a872f22768e0d73ad57162e8ac/lib/RestructureCFG/MaterializeLoopScopes.cpp#L107

lib/RestructureCFG/MaterializeLoopScopes.cpp
andrealinux1: Right, updated the comment here https://github.com/revng/revng/blob/64bc362e7664a5a872f22768e0d73ad57162e8ac/lib/RestructureCFG/MaterializeLoopScopes.cpp#L99

lib/RestructureCFG/MaterializeLoopScopes.cpp
andrealinux1: Oh right, this part has grown in various iterations and a bit inconsistently, and I forgot to take another look at this top down in order to check that it was sensible.

I've reshaped this integrating your suggestions, let me know if you need further changes.

https://github.com/revng/revng/blob/64bc362e7664a5a872f22768e0d73ad57162e8ac/lib/RestructureCFG/MaterializeLoopScopes.cpp#L352

include/revng/RestructureCFG/ScopeGraphUtils.h
andrealinux1: Actually, `RangeOf` does not bind well with `SmallSet`s. We had a discussion on this and I'll link it to you.

lib/RestructureCFG/MaterializeLoopScopes.cpp
pfez: Better, but `electUniqueSuccessor` should be called `tryEnforceUniqueSuccessor`, and point 1. should be part of it.
Then, `enforceUniqueSuccessor` should be renamed to `addScopeCloser`.

In this way the body of the loop is basically doing just:
* `tryEnforceUniqueSuccessor`
* if the previous point succeeds, do `addScopeCloser`.

include/revng/RestructureCFG/ScopeGraphUtils.h
pfez: None of the discussions on mattermost seems really conclusive. The TLDR IIUC is that the problem is just that `SmallSet` does not play well with `std::ranges::begin` and `std::ranges::end`.

Have you looked in upstream LLVM if there is some future commit making it compliant?

pfez: In any case, is it very important that we use `SmallSet` in this specific case?

lib/RestructureCFG/MaterializeLoopScopes.cpp
andrealinux1: I integrated your suggestions in the latest revision https://github.com/revng/revng/blob/cf5a88519be0919093d0d92027ffe9823bb8d1cb/lib/RestructureCFG/MaterializeLoopScopes.cpp#L356

include/revng/RestructureCFG/ScopeGraphUtils.h
andrealinux1: I don't see anything mentioning this in the commit history upstream (https://github.com/llvm/llvm-project/commits/main/llvm/include/llvm/ADT/SmallSet.h, the latest commit we have in our clone, is `f7dffc2` from 10/12/2022).

Regarding needing `SmallSet`, I need further clarifications of what you mean/imply:
- The `SmallSet` is useful here and in other places where we use the `replaceSuccessors` (also used in `IDS`), to perform automatic deduplication on the insertion phase, and quick look up.
- Therefore, I don't understand if you are:
- 1) Implying to move using a `SmallVector` everywhere, and perform the deduplication manually before the `push_back`, and paying the cost on lookup (basically faking that sets never existed).
- 2) Suggesting to use `std::set` in place of `SmallSet`, that I would understand.

pfez: I'm suggesting to use `std::set`, unless we have evidence (or strong reason to think) that not using `SmallSet` will cause observable slowdown.

andrealinux1: That we can do, I'm doing it

andrealinux1: Ok we now use `RangeOf` https://github.com/revng/revng/blob/852c1bb28ddbfc32d19c82aece759ae27963c0e2/include/revng/RestructureCFG/ScopeGraphUtils.h#L84.

This means using `std::set` in place of `SmallSet`, also in IDS (changes are in this commit fb8d4f9ebbc15fe7c8dd7b16bf6ec6f500353d47).

PR #514, opened 2025-08-29
Add options to override `_STARTS_AT` and `_SIZE` behavior

Cvelth:


aleclearmind:
Can you provide an example of how we're currently emitting `_STARTS_AT`? I'd like to move it.

lib/TypeNames/TypePrinters.cpp
```diff
@@ -74,6 +74,11 @@ void ptml::ModelCBuilder::printPadding(uint64_t FieldOffset,
          << tokenTag(NameBuilder.paddingFieldName(FieldOffset),
                      ptml::c::tokens::Field)
          << "[" << getNumber(NextOffset - FieldOffset) << "];\n";
+
+    if (Binary.Configuration().AlwaysPrintStructFieldOffsetAnnotation())
+      *Out << ptml::AttributeRegistry::getAnnotation<"_STARTS_AT">(NextOffset)
+           << "\n";
```
Cvelth: (Thread is easier to follow, so I'll make one, cc: @aleclearmind)

> Can you provide an example of how we're currently emitting _STARTS_AT? I'd like to move it.

It's emitted the same way (in the same place) `uint8_t padding_at_NNN[whatever];` because of implementation simplicity.

That being said, it's easy to move. What do you prefer? Placing it after the field like in the feedback pad?

aleclearmind: Can you provide an example of how we're currently emitting _STARTS_AT?

When I'm asking a question, I'm not implying anything beyond what I'm asking. Please provide an example of the output.

Cvelth:
```cpp
struct _PACKED _SIZE(248) struct_87 {
  generic64_t offset_0;
  generic64_t offset_8;
  uint8_t padding_at_16[204];
  _STARTS_AT(220)
  generic32_t offset_220;
  generic32_t offset_224;
  generic32_t offset_228;
  generic32_t offset_232;
  generic32_t offset_236;
  uint8_t padding_at_240[8];
  _STARTS_AT(248)
};
```

Here you go.

P. S. Good thing I actually looked at this, that final `_STARTS_AT` is definitely a bug. And it's been present for a while, from the looks of it. I guess it isn't harmful, but it's not nice either.

Cvelth: (this is a part of `emit-model-header` artifact with both new configuration options set)

aleclearmind: Let's go with
```c
  generic32_t offset_220 _STARTS_AT(220);
```

Cvelth: Wait, while at it, do we maybe want to _always_ print this when the option is set? Think
```cpp
struct _PACKED _SIZE(16) struct_87 {
  generic64_t offset_0 _STARTS_AT(0);
  generic64_t offset_8 _STARTS_AT(8);
};
```

Or only when necessary like it's done now?

Cvelth: Also, do we want the default state to be? `ON`? Or `OFF`?

aleclearmind: > Or only when necessary like it's done now?

Define "when necessary".

> Also, do we want the default state to be? ON? Or OFF?

I think we should not have an option. Options should be introduced upon user demand.
All these options are dubious and confusing, no one is ever going to touch them.
We need sane defaults and listen to user feedback, not tons of options "just in case".
I'd remove the options.

This should be on.

Cvelth: > when necessary

When offset of the field doesn't match the end of the previous field.

> This should be on.

👌

> We need sane defaults and listen to user feedback, not tons of options "just in case".

I don't know, I see the case for being able to output canonical C (or something really close to it), even if the difference is just an extra macro defined in one of our headers.

aleclearmind: > When offset of the field doesn't match the end of the previous field.

I thought we might omit it if the name of the field already shows it.
For simplicity, let's always emit it. If we get annoyed by it we can change.

> I don't know, I see the case for being able to output canonical C (or something really close to it), even if the difference is just an extra macro defined in one of our headers.

The idea of "canonical C" is loosely defined. Pursuing poorly defined goals is not a great idea.

You keep introducing new configuration options that no one will ever touch, this is not something I want and one day someone will have go through them and drop most of them.

Let's not add more, we're just booking extra work for the future.

Our policy is: have sane defaults and introduce options if users ask for it.
Options will need to be documented very well with examples and everything. Let's not add options "just in case".

Cvelth: Done, please re-review pretty much the entire PR (it's tiny). Everything but the first commit (`START_AT` -> `STARTS_AT`) was rewritten anew.

PR #515, opened 2025-09-02
Follow up fixes for `feature/rename-local-variables`

Cvelth:


include/revng/Model/NameBuilder.h
aleclearmind, line 1: Fix commit message

lib/Backend/DecompileFunction.cpp
aleclearmind, line 1: Explain in the commit message what was wrong

include/revng/Model/NameBuilder.h
Cvelth: What would you prefer? Something like `NameBuilder: fix use-after-free`?

aleclearmind: Yes, "fix-UAF" looks like a placeholder.

lib/Backend/DecompileFunction.cpp
Cvelth: Done:
```
Old backend: always emit debug information

Because of a sub-par heuristic designed to avoid emitting
*unnecessary* debug information tags, some of the *necessary*
tags were lost as well, leading to debug information mismatch
between the backend and the UI.

Note that this is not a fix, this commit simply removes
the heuristic in question, leading to ~2x increase in
output c file size.
```

include/revng/Model/NameBuilder.h
Cvelth: Done.

lib/Backend/DecompileFunction.cpp
aleclearmind: Works

aleclearmind:
Merged in 20b6f86. Thanks!

PR #517, opened 2025-09-09
Introduce a revng-specific wrapper over `llvm::IRBuilder`

Cvelth:


include/revng/Support/IRBuilder.h
aleclearmind, line 1: I'm sorry but we're not doing this.

I think you can implement this (or at least part of this) by simply specifying a custom `InserterTy` to `IRBuilder`: https://github.com/revng/llvm-project/blob/develop/llvm/include/llvm/IR/IRBuilder.h#L2549

Cvelth: Sadly, looking into this, I don't think `InserterTy` is going to work. It only have access to the information about the newly added instruction. But what we want to check is the _builder_ state.

Cvelth: We could extend `InserterTy` I guess. But that would mean llvm-side changes again. And I wonder whether it's worth it to modify llvm for something like this 🤔

aleclearmind: I'm fine with changing LLVM.
You could also try to hack things around with a `IRBuilder` with a stable pointer (e.g., `std::unique_ptr`) and in some hackish way let the inserter have a pointer to it *after* it has been constructed.

We're not going the generated code route.

Cvelth: Honestly, it might not even be _that_ hackish! `IRBuilder` already owns an `Inserter` object. So there's no need for any life-time hacks. We can just put a normal pointer in! It helps that `IRBuilder` does in fact manually create the `Inserter` in the constructor 🤔.

Actually, thinking about this, we might get away with only adding a simple `setInserter` helper. Yeah, ok, that could work, I see it now. Let's give this idea a spin!

Cvelth: Ok, after some fiddling with this, I managed to find a way to make it work! Didn't even have to patch llvm!

The only sad part is that all the `SetInsertPoint`s and constructors had to be re-implemented manually. But, I'd say, we can live with that.

Please take a fresh look!

P. S. This is based on `feature/split-llvm-container` which is marked `ready`. This PR starts from `Introduce IRBuilder wrapper`.

include/revng/Support/IRBuilder.h
```diff
@@ -0,0 +1,126 @@
...
+#include "revng/Support/Debug.h"
+
+namespace revng {
+
+class IRBuilder;
+
+namespace detail {
+
+// NOLINTNEXTLINE
+class RevngIRInsertionChecker : public llvm::IRBuilderDefaultInserter {
```
aleclearmind: 1. We never use `Revng` prefixes. We already have a namespace.
2. Ban `IRBuilder\b` instead of just `IRBuilder` and drop `NOLINTNEXTLINE`
3. `DebugInfoCheckingInserter`?

include/revng/Support/IRBuilder.h
```diff
@@ -0,0 +1,126 @@
...
+private:
+  const revng::IRBuilder *Builder;
```
aleclearmind: `= nullptr`

include/revng/Support/IRBuilder.h
```diff
@@ -0,0 +1,126 @@
...
+
+  // NOLINTNEXTLINE
+  IRBuilder(llvm::BasicBlock *BB, llvm::BasicBlock::iterator I) :
+    // NOLINTNEXTLINE
+    IRBuilder(BB->getContext()) {
+    SetInsertPoint(BB, I);
+  }
+};
+
+inline void detail::RevngIRInsertionChecker::checkImpl() const {
```
aleclearmind: It is a short method, why not inline it?

share/revng/rcc-config.yml
```diff
@@ -116,6 +116,12 @@ read_passes:
               - "//+.* clang-format o(?:ff|n)"
               - "/\\*+.* clang-format o(?:ff|n)"
               - "/\\*+.* \\brief"
+              - "(?<!/)(?<!class )(?<!revng::)IRBuilder"
```
aleclearmind: Ooofff. That's a lot.
This is going to be problematic if we write "pure LLVM" passes.
On the other hand, it seems quite useful.

Maybe we could mark it as `[[deprecated]]` in LLVM and suppress the warning with `pragma`s where appropriate?

>  `/IRBuilder` for includes,

Maybe rely on `.h` instead?

include/revng/Support/IRBuilder.h
```diff
@@ -51,6 +51,16 @@ using LLVMBuilderBase = llvm::IRBuilder<llvm::ConstantFolder,
...
 class IRBuilder : public detail::LLVMBuilderBase {
+public:
+  /// This is a way to disable checks for a specific instance of
+  /// `revng::IRBuilder`. Please use with care!
+  ///
+  /// Practically, this should only appear in one of two cases:
+  /// - in `revngLift` before debug information has been attached.
+  /// - in the old pipeline as a temporary workaround to avoid having to
+  ///   fix selected issues.
+  bool ChecksEnabled = true;
```
aleclearmind: What is the advantage of this over using the classic `IRBuilder`? :thinking:

mkdocs.yml
```diff
@@ -5,7 +5,7 @@
 site_name: rev.ng docs
 docs_dir: share/doc/revng
 theme:
-  name: material
+  name: mkdocs
```
aleclearmind: > (do not push) 

You pushed this

include/revng/Support/IRBuilder.h
Cvelth: 1. 👌 
2. That's intentional to disallow `IRBuilderBase`. And I guess `IRBuilderDefaultInserter` too! Since we don't want people introducing other inserters to override this one. And the cost of that is just this file being ugly.
3. I intentionally used a vague name in case there are other non-debug-information related checks we want to inject there. It's not like it's limited to _only_ debug information. That being said, as of now the name you're suggesting works!

include/revng/Support/IRBuilder.h
Cvelth: Mostly because it might grow in the future. And this makes it easy to do: "hey, that part is llvm-controlled, inside of this function - is our land, do what you want!".

That being said, as you noticed, currently it's pretty slim, so I don't mind inlining it for now.

share/revng/rcc-config.yml
Cvelth: 1. I'll give `[[deprecated]]` a spin, but IMO, I prefer conventions to fail in cases like this! As opposed to compiler warnings. That being said, yeah, I agree this is a pretty gnarly regex!

2. No idea why I didn't do that originally! Good catch!

include/revng/Support/IRBuilder.h
Cvelth: Allows us to ban classic builder altogether. IMO, doing "allowed here but not there" is a pretty slippery slope to walk.

mkdocs.yml
Cvelth: 😅.

include/revng/Support/IRBuilder.h
aleclearmind: I'd inline it or move it to a `.cpp` file.

share/revng/rcc-config.yml
aleclearmind: 1. > I prefer conventions to fail in cases like this

   Why? The compiler can be smarter than grepping and might save us some annoyances.

include/revng/Support/IRBuilder.h
aleclearmind: Maybe we can allow an alias of `IRBuilder` called `NonDebugInfoPreservingIRBuilder`?

I understand that you made all this work and you don't want to bitrot, but you're being quite strict here `:D`

include/revng/Support/IRBuilder.h
aleclearmind: 2. I don't honestly think it's necessary. Here we're trying to prevent sloppy people to adopt the vanilla `IRBuilder`, not to prevent advanced use cases (using `IRBuilderBase` or `IRBuilderDefaultInserter`).

include/revng/Support/IRBuilder.h
Cvelth: Ah, wait, I misunderstood you! I thought you mean to remove this `checkImpl` by inlining it into the virtual `InsertHelper`.

If you mean to put the body inside the class, that's impossible because it accesses a `revng::IRBuilder` field, so it has to come after the full definition.

That being said, yes, if we get rid of `ChecksEnabled`, this can go back up.

share/revng/rcc-config.yml
Cvelth: More or less a preference, I'd rather deal with things like this when running `check-conventions` (really early) than when building code (still kind early, but not nearly as early as convention check).

But yeah, that's a personal preference, all said and done, `[[deprecated]]` should work just fine.

include/revng/Support/IRBuilder.h
Cvelth: Sure, I guess. Do you see it as a separate entity? Or pretty much the same thing just `ChecksEnabled` being `protected` and set by an inheritor?

include/revng/Support/IRBuilder.h
Cvelth: Sure, I guess that works.

Cvelth: Done.

include/revng/Support/IRBuilder.h
Cvelth: (ping)

include/revng/Support/IRBuilder.h
Cvelth: (ping)

mkdocs.yml
Cvelth: Gone 😅

share/revng/rcc-config.yml
Cvelth: Ok, I looked into this.

And since `llvm::IRBuilder` is used all over the llvm, `[[deprecated]]` means a _lot_ of extra noise during an llvm rebuild, which is really sad.

Are you sure that's the route we want to go?

Is
```yaml
              - "(?<!class )(?<!revng::)IRBuilder\b(?!\\.h)"
                # this allows:
                # - `class IRBuilder` for forward declarations,
                # - `revng::IRBuilder` to emphasize the wrapper's existence,
                # - `IRBuilder.h` for includes,
                # and *no* other permutation of `IRBuilder` anywhere else.
```
really that bad?

include/revng/Support/IRBuilder.h
Cvelth: Done.

share/revng/rcc-config.yml
aleclearmind: It is rather strict, but let's give it a shot. We can change it if it bothers us.

include/revng/Support/IRBuilder.h
aleclearmind: I'm not a fan of `ChecksEnabled` that basically demotes you to its base `class`.
Let's just use the base class with a very explicit name.

share/revng/rcc-config.yml
Cvelth: IMO, it's rather to go as strict as possible early on, before people get used to this :P

We can relax it afterwards.

include/revng/Support/IRBuilder.h
Cvelth: The awkward part about using two different types is that we can no longer just accept `revng::IRBuilder &` from functions (for example, [this one (`develop`)](https://github.com/revng/revng/blob/c34aed5565e7ff330ec5acf33c0f6fa6f08a3640/include/revng/Support/FunctionTags.h#L283))

So we either have to

1. introduce a concept like
```cpp
std::same_as<T, revng::IRBuilder> || std::same_as<T, revng::NonDebugInfoPreservingIRBuilder>
```
(which means we have to include the header defining IRBuilder pretty much everywhere (and my recent adventures with model stuff made me see _really_ clearly why that's a bad idea)) or introduce a new header just for this concept.

2. relax the requirements and start accepting `llvm::IRBuilderBase` everywhere (including places that used to accept `llvm::IRBuilder`), making it more likely that someone finds a way to work around the enforcement and obtain an `llvm::IRBuilder` object somehow without spelling its name.

3. provide multiple overloads for functions that are actually intended to be used _before_ debug information is injected

4. something else?

-----------

And, IMO, all of these are worse than something like `ChecksEnabled`.

aleclearmind: It seems to me that accepting `llvm::IRBuilderBase` is the right thing to do.

> making it more likely that someone finds a way to work around the enforcement and obtain an llvm::IRBuilder object somehow without spelling its name

Check conventions goal is to stop accidental mistakes, not adversarial attempts at escaping it `:P`.

If you really really want to keep things as your proposed, at least make a factory method for the non-debugging version, I really don't want to write this:

```c++
      revng::IRBuilder Builder(replaceFunction(OldFunction), DLocation);
      Builder.ChecksEnabled = false; // Debug information hasn't been attached yet.
```

`ChecksEnabled` should not be public nor mutable.

Cvelth: The original idea was to keep it mutable specifically so you can do
```cpp
Builder.ChecksEnabled = false;
doSomethingNasty(Builder);
Builder.ChecksEnabled = true;
```

But since you are against this, I hid the flag inside the _inserter_ instead. Please take a look.

include/revng/Support/MetaAddress.h
```diff
@@ -26,11 +26,15 @@ class Module;
 class StructType;
 class GlobalVariable;
 class Instruction;
-class IRBuilderBase;
 template<typename T, typename Enable>
 struct DenseMapInfo;
 } // namespace llvm
 
+namespace revng {
+// NOLINTNEXTLINE
```
aleclearmind: Is this necessary?

lib/Canonicalize/MakeModelCastPass.cpp
```diff
@@ -217,7 +217,10 @@ void MakeModelCastPass::makeModelCast(const CastToEmit &ToEmit,
   const model::Type &OperandModelType = *TypeMap.at(Operand);
 
   auto *I = cast<Instruction>(OperandUse.getUser());
-  revng::IRBuilder Builder(I);
+
+  // This should definitely do the checks, but since it's going to go away soon,
+  // let it stay as is.
```
aleclearmind: `s/This should definitely do the checks/Here we should definitely use the builder preserving debug info/`

lib/RemoveExtractValues/RemoveExtractValuesPass.cpp
```diff
@@ -42,7 +42,10 @@ bool RemoveExtractValues::runOnFunction(llvm::Function &F) {
   auto OpaqueEVPool = FunctionTags::OpaqueExtractValue.getPool(*F.getParent());
 
   llvm::LLVMContext &LLVMCtx = F.getContext();
-  revng::IRBuilder Builder(LLVMCtx);
+
+  // TODO: checks are only omitted here because of unit tests.
+  revng::NonDebugInfoPreservingIRBuilder Builder(LLVMCtx);
```
aleclearmind: Not sure what to say here. I'm fine with using as many `NonDebugInfoPreservingIRBuilder` as we need to for now.

Also, we have plans to eventually dismiss this pass as well.

include/revng/Support/IRBuilder.h
```diff
@@ -0,0 +1,229 @@
...
+namespace detail {
+
+class DebugInfoCheckingInserter : public llvm::IRBuilderDefaultInserter {
+private:
+  const revng::IRBuilder *const Builder = nullptr;
+  const bool EnableChecks = true;
+
+public:
+  DebugInfoCheckingInserter(const revng::IRBuilder &Builder,
+                            bool EnableChecks = true) :
```
aleclearmind: Let's avoid default arguments, in particular in `detail` classes.

include/revng/Support/IRBuilder.h
aleclearmind: I'd leave `ChecksEnabled` out and just do:

```c++
using NonDebugInfoPreservingIRBuilder = llvm::IRBuilder;
```

It seems extra logic just to get the exact same result.

Cvelth: > the exact same result.

The difference is being able to pass `NonDebugInfoPreservingIRBuilder` as an `IRBuilder` into functions.

include/revng/Support/MetaAddress.h
Cvelth: Good catch!

lib/RemoveExtractValues/RemoveExtractValuesPass.cpp
Cvelth: Mostly me being extra "are we sure just debug-info-less unit tests are important enough to make _all_ the code worse" pedantic.

aleclearmind: OK, squash.

include/revng/Support/IRBuilder.h
aleclearmind: Right, add a comment.

lib/Canonicalize/MakeModelCastPass.cpp
Cvelth: > Here we should definitely use the builder preserving debug info

Actually, wait, this made me notice! The `revng::IRBuilder` barely does anything for debug information _preservation_. Should we go with `NonDebugInfoCheckingIRBuilder`? And similarly, in this comment?

include/revng/Support/IRBuilder.h
Cvelth: Done:
```cpp
/// The reason `NonDebugInfoCheckingIRBuilder` inherits from
/// `revng::IRBuilder` instead of just aliasing `llvm::IRBuilder<>` is to allow
/// passing it into functions accepting `revng::IRBuilder`.
// NOLINTNEXTLINE
class NonDebugInfoCheckingIRBuilder : public IRBuilder {
```

include/revng/Support/MetaAddress.h
Cvelth: Done.

lib/Canonicalize/MakeModelCastPass.cpp
Cvelth: Done (`checking` edition). Feel free to close if works.

lib/RemoveExtractValues/RemoveExtractValuesPass.cpp
Cvelth: Done.

include/revng/Support/IRBuilder.h
Cvelth: Done.

PR #519, opened 2025-09-11
Split `module.bc.zstd` into `root.bc.zstd` and `functions.bc.zstd`.

Cvelth:


lib/PipelineC/PipelineC.cpp
```diff
@@ -69,6 +69,8 @@ class ExistingOrNew {
 
 static void llvmErrorToRpError(llvm::Error Error, rp_error *Out) {
   if (Out == nullptr) {
+    // TODO: this is suspicions, isn't it an indication that callee is
+    // doing something nasty? Is it ok to just ignore it?
```
mrjackv: That's part of the `PipelineC` calling conventions, if you see the callers of this functions (e.g. `_rp_manager_run_analysis`), they pass through the `rp_error *` variable. The callee of these can opt-in to receive errors by passing the `rp_error *`, whereas they get ignored if `NULL` is provided as an argument. (Essentially the `rp_error *` constitutes an optional return value, but due to C structure that was one of the saner ways to implement it).
There's a section `Error Reporting` in `PipelineC/Prototypes.h` explaining this in depth.

lib/PipelineC/PipelineC.cpp
```diff
@@ -93,6 +95,8 @@ static void llvmErrorToRpError(llvm::Error Error, rp_error *Out) {
     *Out = rp_simple_error(Reason, "");
   };
 
+  // TODO: do we really need `handleAllErrors` here? It makes it seem as if
+  // we're overriding `*Out` a bunch of times, even though it only happens once.
```
mrjackv: The `All` here means that if an error is unhandled it aborts. We're guaranteed that that doesn't happen because the second handler accepts the base class `llvm::ErrorInfoBase`.

lib/Pipeline/Context.cpp
```diff
@@ -44,6 +44,7 @@ llvm::Error Context::load(const revng::DirectoryPath &Path) {
   if (not MaybeExists)
     return MaybeExists.takeError();
 
+  // TODO: this is suspicious, we should probably add a comment why.
```
mrjackv: The check below is basically `stat`, so if the file doesn't exist we stop and succeed. Either way a comment doesn't hurt.

share/revng/test/tests/daemon/test.py
```diff
@@ -129,9 +129,9 @@ def error_handler(e: BaseException):
 
     binary = pytestconfig.getoption("binary")
     tracing = TraceConfig()
-    tracing.on_request_headers_sent.append(header_trace)
-    tracing.on_request_chunk_sent.append(payload_trace)
-    tracing.on_request_end.append(response_trace)
+    tracing.on_request_headers_sent.append(header_trace)  # type: ignore
+    tracing.on_request_chunk_sent.append(payload_trace)  # type: ignore
+    tracing.on_request_end.append(response_trace)  # type: ignore
```
mrjackv: This commit is no longer required, this was fixed by the `feature/pypeline` branch, which has been merged

lib/Pipeline/Context.cpp
Cvelth: Yeah, that's what I mean, it's suspicious for a _load_ function to succeed when the file it's loading _from_ doesn't exist.

share/revng/test/tests/daemon/test.py
Cvelth: Ah, my bad, I didn't mean to push this! Thanks for catching this!

lib/PipelineC/PipelineC.cpp
Cvelth: Thanks for the explanation! I'll add a comment pointing to `PipelineC/Prototypes.h`

lib/Pipeline/Context.cpp
mrjackv: It makes sense in the context of `lib/Pipeline`, both context and containers are default constructible, so the logic is:
```c
Context C;
if (exists(context_file))
  C.deserialize(context_file.read());
 return C;
 ```

Comment (Cvelth) [reply lib/Pipeline/Context.cpp]: I see. I'll add a comment!

File: lib/PipelineC/PipelineC.cpp (hunk 2 above)
Comment (Cvelth) [reply lib/PipelineC/PipelineC.cpp]: Ah, wait, it looks like a misunderstanding of mine!

For some reason I assumed there's also a _single_ error handler that guarantees no additional errors are spawned (and as the result, it doesn't have to be recursive). But it looks like I made it up, my bad, sorry for the noise!

(keeping the thread open so that I don't forget to remove the comment)

File: share/revng/test/tests/daemon/test.py (hunk 4 above)
Comment (Cvelth) [reply share/revng/test/tests/daemon/test.py]: Dropped.

File: lib/PipelineC/PipelineC.cpp (hunk 1 above)
Comment (Cvelth) [reply lib/PipelineC/PipelineC.cpp]: Done:
```
    // By passing `Out == nullptr` the callee indicated that they do not care
    // about the error, so we can just discard it.
    //
    // For more context, see `include/revng/PipelineC/Prototypes.h`.
```

File: lib/PipelineC/PipelineC.cpp (hunk 2 above)
Comment (Cvelth) [reply lib/PipelineC/PipelineC.cpp]: Gone.

File: lib/Pipeline/Context.cpp (hunk 3 above)
Comment (Cvelth) [reply lib/Pipeline/Context.cpp]: Done:
```
  // In the context of `lib/Pipeline`, both context and containers are default
  // constructible. As such, we just skip loading files that do not exist,
  // ending up with the default state.
```

File: lib/Model/Importer/Binary/ELFImporter.cpp [hunk 5]
```diff
@@ -487,8 +487,8 @@ void ELFImporter<T, HasAddend>::findMissingTypes(object::ELFFile<T> &TheELF,
       if (auto E = importELF(DepModel, *TheBinary, AdjustedOptions)) {
         revng_log(ELFImporterLog,
                   "Can't import model for " << DependencyLibrary << " due to "
-                                            << E);
-        llvm::consumeError(std::move(E));
+                                            << Error);
+        llvm::consumeError(std::move(Error));
         ModelsOfLibraries.erase(DependencyLibrary);
         continue;
       }
```
aleclearmind: Missing rename of `E` in declaration?

include/revng/ADT/UpcastablePointer.h
```diff
@@ -64,6 +64,11 @@ ReturnT upcast(P &&Upcastable, const L &Callable, ReturnT &&IfNull) {
   if constexpr (I < std::tuple_size_v<concrete_types>) {
     using type = std::tuple_element_t<I, concrete_types>;
     if (auto *Upcasted = llvm::dyn_cast<type>(Pointer)) {
+      if constexpr (std::same_as<std::decay_t<ReturnT>, llvm::Error>)
+        llvm::consumeError(std::move(IfNull));
+      if constexpr (SpecializationOf<std::decay_t<ReturnT>, llvm::Expected>)
+        llvm::consumeError(std::move(IfNull).takeError());
```
aleclearmind: I don't think returning `Expected` ever makes sense.

include/revng/ADT/UpcastablePointer.h
```diff
@@ -75,27 +80,6 @@ ReturnT upcast(P &&Upcastable, const L &Callable, ReturnT &&IfNull) {
   }
 }
 
-template<typename L, UpcastablePointerLike P, size_t I = 0>
-llvm::Error upcast(P &&Upcastable, const L &Callable, llvm::Error IfNull) {
```
aleclearmind, line 79: Add a Doxygen comment about this use case on the remaining definition of this.

lib/Model/Importer/Binary/MachOImporter.cpp
```diff
@@ -282,6 +282,7 @@ Error MachOImporter::import() {
   }
 
   Error TheError = Error::success();
+  revng_check(TheError);
```
aleclearmind: `o_O`

lib/Support/ResourceFinder.cpp
```diff
@@ -28,8 +28,8 @@ std::string getComponentsHash() {
 
   std::string Result;
   for (std::string &File : Files) {
-    auto MaybeBuffer = llvm::MemoryBuffer::getFile(File);
-    revng_assert(MaybeBuffer);
+    auto MaybeBuffer = errorOrToExpected(llvm::MemoryBuffer::getFile(File));
+    revng_check(MaybeBuffer);
```
aleclearmind: `cantFail`?

tests/unit/Pipeline.cpp
```diff
@@ -657,7 +663,8 @@ BOOST_AUTO_TEST_CASE(SingleElementPipelineFailure) {
 
   auto Error = Pipeline.run("end", Targets);
   BOOST_TEST(!!Error);
-  consumeError(std::move(Error));
+  if (!!Error)
```
aleclearmind: The `!!` pattern seems suspicious to me. Can we avoid it or is there a reason?

tests/unit/Pipeline.cpp
```diff
@@ -534,6 +534,9 @@ BOOST_AUTO_TEST_CASE(SingleElementPipelineCanBeRunned) {
   Targets[CName].emplace_back(Target(RootKind2));
   auto Error = Pip.run("end", Targets);
   BOOST_TEST(!Error);
+  if (Error)
+    llvm::consumeError(std::move(Error));
```
aleclearmind: What's the point?

tests/unit/Pipeline.cpp
```diff
@@ -822,9 +833,7 @@ BOOST_AUTO_TEST_CASE(SingleElementPipelineForwardFinedGrained) {
   llvm::StringMap<ContainerToTargetsMap> Invalidations;
   Invalidations[Name].add(CName, {}, RootKind);
 
-  auto Error = Pipeline.getInvalidations(Invalidations);
-  BOOST_TEST(!Error);
-
+  Pipeline.getInvalidations(Invalidations);
```
aleclearmind, line 825: I think this commit belongs to a previous patch.
I'd squash them to be honest.

python/revng/internal/cli/_commands/override_by_name.py
```diff
@@ -68,7 +68,7 @@ def run(self, options: Options):
 
             # Extract model, if necessary
             result = run_revng_command(
-                ["model", "opt", "-Y", saved_file.name, "-o", model_file.name], options
+                ["model", "opt", saved_file.name, "-o", model_file.name], options
```
aleclearmind, line 71: Can we drop it from `model opt` then?

lib/ImportFromC/HeaderToModel.cpp
```diff
@@ -1265,6 +1265,8 @@ void HeaderToModelDiagnosticConsumer::HandleDiagnostic(Level DiagLevel,
   llvm::raw_string_ostream OS(Text);
   auto *DiagOpts = &Info.getDiags()->getDiagnosticOptions();
 
+  // Suspicious!
+  // TODO: investigate.
   uint64_t StartOfLocationInfo = OS.tell();
```
aleclearmind: Use `// WIP` if you don't want to merge.
Anyway, I can't help with this.

lib/Model/Importer/Binary/ELFImporter.cpp
```diff
@@ -224,6 +224,7 @@ Error ELFImporter<T, HasAddend>::import(const ImporterOptions &Options) {
 
       auto MaybeSectionName = TheELF.getSectionName(SectionHeader);
       if (auto Error = MaybeSectionName.takeError()) {
+        // TODO: is there anything useful we can extract from this error to log?
```
aleclearmind: Yes, we should *never* silently suppress errors! Log!

lib/Model/Importer/Binary/ELFImporter.cpp
```diff
@@ -461,6 +462,9 @@ void ELFImporter<T, HasAddend>::findMissingTypes(object::ELFFile<T> &TheELF,
       revng_log(ELFImporterLog, " Importing Model for: " << DependencyLibrary);
       auto BinaryOrErr = llvm::object::createBinary(DependencyLibrary);
       if (auto Error = BinaryOrErr.takeError()) {
+        // TODO: wouldn't user want to know we couldn't find one of
+        //       the dependencies?
+        //       Sounds like we want a better way to report this.
```
aleclearmind: I'd say this is a candidate for diagnostics that can get to the user.

As of now, we don't have it. I suggest to introduce a new convention for logs that should be turned into diagnostic:

```
// TODO: turn into diagnostic
```

lib/Pipeline/LLVMContainer.cpp
```diff
@@ -65,14 +65,14 @@ std::unique_ptr<ContainerBase>
 LLVMContainer::cloneFiltered(const TargetsList &Targets) const {
   using InspectorT = LLVMKind;
   auto ToClone = InspectorT::functions(Targets, *this->self());
-  auto ToClonedNotOwned = InspectorT::untrackedFunctions(*this->self());
+  auto ToCloneUntracted = InspectorT::untrackedFunctions(*this->self());
```
aleclearmind: Untracked?

include/revng/Pipeline/LLVMContainer.h
```diff
@@ -46,6 +46,9 @@ class LLVMContainer : public EnumerableContainer<LLVMContainer> {
                 std::unique_ptr<llvm::Module> M) :
     EnumerableContainer<ThisType>(*Context, Name), Module(std::move(M)) {}
 
+  LLVMContainer &cloneFrom(const LLVMContainer &Another);
+  LLVMContainer &swapWith(LLVMContainer &Another);
```
aleclearmind, line 50: I'd squash this commit in the first commit using it.
Same for `linkModules`.

share/revng/pipelines/revng-pipelines.yml
```diff
@@ -148,14 +152,10 @@ Branches:
             Therefore, they lack arguments and return values.
       - Name: enforce-abi
         Pipes:
-          - Type: llvm-pipe
-            UsedContainers: [module.bc.zstd]
-            Passes:
-              - drop-root
```
aleclearmind, line 154: I was like "nononono", but it's "yesyesyes".

We should have done this much sooner.

share/revng/pipelines/revng-pipelines.yml
```diff
@@ -506,15 +506,17 @@ Branches:
...
           - Type: embed-statement-comments
-            UsedContainers: [module.bc.zstd]
+            UsedContainers: [functions.bc.zstd]
       - Name: decompile
         Pipes:
           - Type: helpers-to-header
-            UsedContainers: [module.bc.zstd, helpers.h]
+            UsedContainers: [functions.bc.zstd, helpers.h]
           - Type: model-to-header
+            # **ivan** shouldn't we move this further up the pipeline?
```
aleclearmind: Use `WIP` for comments that should not be merged.
Don't let your name get merged.

share/revng/pipelines/revng-pipelines.yml
```diff
@@ -9,7 +9,11 @@ Containers:
   - Name: cross-relations.yml
     Type: binary-cross-relations
     Role: cross-relations
-  - Name: module.bc.zstd
+  - Name: functions.bc.zstd
+    Type: llvm-container
+  - Name: root.bc.zstd
+    Type: llvm-container
+  - Name: recompilable-root.bc.zstd
```
aleclearmind: The name seems to imply that `root.bc.zstd` is not recompilable, while it is, on another branch of the pipeline.

Let's go with `root-and-functions.bc.zstd` or `root-with-functions.bc.zstd`.

lib/FunctionIsolation/InvokeIsolatedFunctions.cpp
```diff
@@ -190,15 +163,21 @@ class InvokeIsolatedFunctions {
...
             auto Name = model::Register::getCSVName(Register);
-            GlobalVariable *CSV = M->getGlobalVariable(Name, true);
+            GlobalVariable *CSV = RootModule.getGlobalVariable(Name, true);
             revng_assert(CSV != nullptr);
             Arguments.push_back(createLoad(Builder, CSV));
           }
         }
       }
 
+      llvm::Function *
```
aleclearmind, line 172: You can use `auto *` here

lib/FunctionIsolation/InvokeIsolatedFunctions.cpp
```diff
@@ -23,78 +26,48 @@
...
 
-    for (BasicBlock &BB : *RootFunction) {
+    for (BasicBlock &BB : RootFunction) {
       revng_assert(not BB.empty());
 
       MetaAddress JumpTarget = getBasicBlockJumpTarget(&BB);
       auto It = Map.find(JumpTarget);
       if (It != Map.end()) {
+        revng_assert(get<1>(It->second) == nullptr,
+                     "are we ok with overriding basic blocks here?");
```
aleclearmind: Please do not leave questions in asserts, I find it quite confusing.

State your expectations, not questions.

lib/FunctionIsolation/InvokeIsolatedFunctions.cpp
```diff
@@ -190,15 +163,21 @@ class InvokeIsolatedFunctions {
...
-            GlobalVariable *CSV = M->getGlobalVariable(Name, true);
+            GlobalVariable *CSV = RootModule.getGlobalVariable(Name, true);
             revng_assert(CSV != nullptr);
             Arguments.push_back(createLoad(Builder, CSV));
           }
         }
       }
 
+      llvm::Function *
+        NewDeclaration = llvm::Function::Create(F->getFunctionType(),
```
aleclearmind, line 173: Using function types of another module scares me, but it should be fine :thinking:

include/revng/ADT/UpcastablePointer.h
Cvelth: Yes, but better safe than sorry, this does not have any impact and might be helpful if someone does put an expected in there.

That being said, if you prefer, I wouldn't mind
```cpp
static_assert(not SpecializationOf<std::decay_t<ReturnT>, llvm::Expected>);
```
either.

lib/Model/Importer/Binary/MachOImporter.cpp
Cvelth: Better safe than sorry :D

Weird copy-paste could lead to an assertion without this.

tests/unit/Pipeline.cpp
Cvelth: I just kept it because it was there. That's literally just `bool(Error)` but not as ugly, which could be relevant in some cases. Doesn't matter here though.

tests/unit/Pipeline.cpp
Cvelth: I'm not too familiar with Boost.Test in particular but I'd assume there's a way (like a CLI argument or something) to turn these into "print"-like statements instead of aborts or something akin to that.

But yeah, just safety for a corner case, I don't mind dropping it.

python/revng/internal/cli/_commands/override_by_name.py
Cvelth: I did in one of the previous commits.

The thing is it was done through a helper in a header, so it was a single action to drop it everywhere.

lib/Pipeline/LLVMContainer.cpp
Cvelth: For terminology consistency, the function name uses `untracked` prefix.

lib/Model/Importer/Binary/ELFImporter.cpp
Cvelth: Good catch! This was commit reordering fiasco, the declaration rename got stuck in a commit that came later.

Fixed.

include/revng/ADT/UpcastablePointer.h
Cvelth: Done:
```
/// This is an upcasting helper `UpcastablePointer` below takes advantage of
/// heavily. When possible, prefer the member function instead.
///
/// That being said, this free function is not tied to the object and can be
/// used on normal pointers, as long as LLVM-style RTTI conventions
/// are adhered to.
///
/// \note this also correctly handles `llvm::Error` and `llvm::Expected`
/// when they are passed into \param IfNull.
```

lib/Support/ResourceFinder.cpp
Cvelth: Done.

tests/unit/Pipeline.cpp
Cvelth: Ok, I was wrong, it is essential in `BOOST_TEST`s, as those do not convert the input type to bool AND they pass it as is into `operator<<` of the std-stream, which llvm types don't play nicely with.

I'd suggest a more clean approach to the error being checked and, more importantly, printed properly, but let's be honest, this is old pipeline tests - those go away soon anyway. I say let them be.

tests/unit/Pipeline.cpp
Cvelth: Squashed

lib/ImportFromC/HeaderToModel.cpp
Cvelth: I looked into this, and yes, this is clang infrastructure being awkward, not our fault.

Closing this.

lib/Model/Importer/Binary/ELFImporter.cpp
Cvelth: I went with
```
        // TODO: emit a diagnostic message for the user.
```

include/revng/Pipeline/LLVMContainer.h
Cvelth: Squashed.

share/revng/pipelines/revng-pipelines.yml
Cvelth: Whoops, this one also looks like rebasing fiasco, my bad and good catch.

Cvelth: (as for the comment itself, I got rid of it - it's for the old pipeline anyway, who cares, and I'll be the one adding the new equivalent of these pipes, so...)

share/revng/pipelines/revng-pipelines.yml
Cvelth: Done (`with` one).

lib/FunctionIsolation/InvokeIsolatedFunctions.cpp
Cvelth: It doesn't change anything.

lib/FunctionIsolation/InvokeIsolatedFunctions.cpp
Cvelth: Ah, my bad, this was a question to myself - "will I ever see this assert trigger or not?". But then I never saw it - and forgot about it :D

My bad, this should have been accompanied with a `WIP`.

Cvelth: Either way, I removed the message, just kept the assert, as I feel like it makes sense in the context even without extra words (check nothing is assigned to an object we're about to assign to).

lib/FunctionIsolation/InvokeIsolatedFunctions.cpp
Cvelth: I looked around and it looks like this is safe to do as long as both modules share the same context. Which we do here, I assume. Not to mention the two modules will be merged together right after this!

lib/Model/Importer/Binary/ELFImporter.cpp
Cvelth: Done.

include/revng/ADT/UpcastablePointer.h
aleclearmind: > That being said, if you prefer, I wouldn't mind

:+1:

include/revng/ADT/UpcastablePointer.h
```diff
@@ -52,6 +52,15 @@ concept UpcastablePointerLike = Dereferenceable<T> and Upcastable<pointee<T>>;
...
 
+/// This is an upcasting helper `UpcastablePointer` below takes advantage of
+/// heavily. When possible, prefer the member function instead.
+///
+/// That being said, this free function is not tied to the object and can be
+/// used on normal pointers, as long as LLVM-style RTTI conventions
+/// are adhered to.
+///
+/// \note this also correctly handles `llvm::Error` and `llvm::Expected`
+/// when they are passed into \param IfNull.
```
aleclearmind: What does it mean to "handle correctly"? Be explicit.

> \note `IfNull` can be a `llvm::Error`. In such case, in case of successful upcast, `IfNull` will be consumed.

include/revng/ADT/UpcastablePointer.h
```diff
@@ -52,6 +52,15 @@ concept UpcastablePointerLike = Dereferenceable<T> and Upcastable<pointee<T>>;
 template<typename T>
 concept NotUpcastablePointerLike = not UpcastablePointerLike<T>;
 
+/// This is an upcasting helper `UpcastablePointer` below takes advantage of
+/// heavily. When possible, prefer the member function instead.
```
aleclearmind: I think a comma is missing or something.

lib/Model/Importer/Binary/MachOImporter.cpp
aleclearmind: Mmmh, I don't understand. Can you make an example?

tests/unit/Pipeline.cpp
aleclearmind: Let's drop it. We'll leave boost test one day.

lib/Model/Importer/Binary/ELFImporter.cpp
aleclearmind: Please mark as solved, if you don't want me to take a look.

lib/Pipeline/LLVMContainer.cpp
aleclearmind: `s/Untracted/Untracked/`

libexec/revng/check-conventions
```diff
@@ -633,6 +633,12 @@ class CheckCommitPass(Pass):
         if len(subject) > 50:
             problems.append("Commit subject line exceeds 50 characters")
 
+        if "wip" in subject.lower():
```
aleclearmind: Maybe I'd go with WIP or use a regexp and do `\bwip\b`. Otherwise we ban "wipe" `:D`

lib/Model/Importer/DebugInfo/DwarfImporter.cpp
```diff
@@ -1254,6 +1254,7 @@ void DwarfImporter::import(StringRef FileName, const ImporterOptions &Options) {
 
       auto DebugFilePath = findDebugInfoFileByName(FileName, DebugFile, ELF);
       if (!DebugFilePath) {
+        // TODO: Should this not be an assert?
```
aleclearmind: Agreed

lib/Model/Importer/Binary/MachOImporter.cpp
Cvelth: I tried to make a godbolt example, but got it looks like it only works with ABI-breaking-checks disabled. And the assert we're interested in is hidden behind those.

So, the best I can do is paste some code from LLVM:
```cpp
// default constructor:
  /// Create a success value. Prefer using 'Error::success()' for readability
  Error() {
    setPtr(nullptr);
    setChecked(false);
  }

// is called from:
class ErrorSuccess final : public Error {};

inline ErrorSuccess Error::success() { return ErrorSuccess(); }
```

Note the `setChecked(false)` part. Which is then asserted on in the destructor:
```cpp
  ~Error() {
    assertIsChecked();
    delete getPtr();
  }

  void assertIsChecked() {
#if LLVM_ENABLE_ABI_BREAKING_CHECKS
    if (LLVM_UNLIKELY(!getChecked() || getPtr()))
      fatalUncheckedError();
#endif
  }
```

Which means:
```cpp
{
  Error TheError = Error::success();
}
```
would assert because `TheError` is unchecked.

tests/unit/Pipeline.cpp
Cvelth: Leave it for what? I'd assume most decent testing framework have something like that.

lib/Model/Importer/Binary/ELFImporter.cpp
Cvelth: Ah, my bad. I blame github because I'm sure I clicked the button (unless I missed or something) :P

lib/Pipeline/LLVMContainer.cpp
Cvelth: Ah! My bad.

include/revng/ADT/UpcastablePointer.h
Cvelth: Right, good catch:
```cpp
/// This is an upcasting helper, `UpcastablePointer` below takes advantage of
/// heavily. When possible, prefer the member function instead.
```

libexec/revng/check-conventions
Cvelth: Who needs wiping anyway? /j

Jokes aside, yeah, my bad. It looks like it's literally just `wipe` and `swipe` (also `swip[p]le`, but I have never heard that one :P), but let's go with the regex.

Cvelth: Speaking about which!!! Is there a reason why we don't check this in the CI?

I can see why we don't check the compilability, as that takes a while and, sometimes, for nastier PRs we actually want an ability to break it, but conventions?

I feel like we should really consider running the [top one](https://github.com/revng/orchestra/blob/develop/.orchestra/config/components/revng.yml#L57) before merging branches!

aleclearmind: We have:

```yaml
post_install_script: |
  if [ "$RUN_TESTS" -eq 1 ]; then
    cd "$SOURCE_DIR"
    ./libexec/revng/check-conventions
  fi
```

The CI runs `orc install revng --test`.

tests/unit/Pipeline.cpp
aleclearmind: I was thinking that it restart after aborts, but here it won't simply abort.

To be honest I don't think people usually write tests thinking that after a failure things continue to run and it's kinda hard to enforce this. I'm not user I want to pollute the tests with these `if`s. :thinking:

Have you ever found it useful? It's nice in theory, but did it ever helped you?
I usually fix one failure a time.

lib/Model/Importer/Binary/MachOImporter.cpp
aleclearmind: Ah I see. But the doc of `bindTable` does not suggest to do that.
Other users in LLVM do not do that.
What matters is that eventually we do check the error.

libexec/revng/check-conventions
Cvelth: Ah, so it's only `--check-commit` that's missing! Should we add it?

Cvelth: Ah, wait, no, that needs to be run on every commit! So yeah, I still feel like we should run something `check-branch`-like in CI.

tests/unit/Pipeline.cpp
Cvelth: Honestly, I don't think I even interacted with unit-tests to begin with :P

The failures we get from those are usually well telegraphed, so I don't think I ever actually needed to "debug" those in a traditional sense. Delta attaching the debugger and seeing which exact assert triggers.

So I don't have a strong stance on this in the first place. It's the matter of being technically-correct. But again, those are just tests, so who cares.

libexec/revng/check-conventions
aleclearmind: Eeeh eventually. Not today.

lib/Model/Importer/Binary/MachOImporter.cpp
Cvelth: I guess this is indeed hair-splitting. Because if there are no conditions - you're going to see the fact that this empty error is unchecked the very first time you run this. So maybe these aren't important after all.

include/revng/ADT/UpcastablePointer.h
Cvelth: (please don't close threads like these, or might miss them when doing the final pass and the real comment will be left without its comma :P)

lib/Model/Importer/Binary/MachOImporter.cpp
aleclearmind: Yeah, I'm just trying to figure out if this is a good practice we should extend elsewhere as well. 

If not, let's leave it out.

tests/unit/Pipeline.cpp
aleclearmind: Your call.

include/revng/ADT/UpcastablePointer.h
aleclearmind: Ah OK, I thought you reporting it meant it was already in the codebase.

Cvelth: I usually explicitly use a word "done" or "fixed" in that case!

include/revng/ADT/UpcastablePointer.h
Cvelth: Done.

lib/Model/Importer/Binary/MachOImporter.cpp
Cvelth: Gone.

tests/unit/Pipeline.cpp
Cvelth: I dropped this. Let's just drop the pipeline and be done with these tests!

lib/Pipeline/LLVMContainer.cpp
Cvelth: Fixed.

include/revng/ADT/UpcastablePointer.h
Cvelth: Done.

include/revng/ADT/UpcastablePointer.h
Cvelth: Done.

libexec/revng/check-conventions
Cvelth: Done (`\bwip\b`).

lib/Model/Importer/DebugInfo/DwarfImporter.cpp
Cvelth: Done.

PR #521, opened 2025-09-15
Pypeline: add pypeline-daemon

zommiommy:


python/revng/internal/cli/revng2.py
```diff
@@ -12,42 +12,58 @@
...
 
-from revng.pypeline.cli.utils import LazyGroup
-from revng.pypeline.main import import_pipebox
+from revng.pypeline.cli.project import project
+from revng.pypeline.main import pype
 
-logger = logging.getLogger("revng2")
+logger = logging.getLogger(__name__)
+handler = logging.StreamHandler(sys.stderr)
+formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
```
aleclearmind: Drop `%(asctime)s - %(name)s - `.

We'll probably drop `%(levelname)s` as well, let's see how the Mattermost discussion evolves.

python/revng/internal/cli/revng2.py
```diff
@@ -12,42 +12,58 @@
...
+    # Replace the name (needed for autocompletion and usage)
+    pype.name = "revng2"
+    pype.lazy_subcommands["quick"] = "revng.internal.cli.revng2:quick"
+    # Replace the default for pipebox
+    for param in pype.params:
+        if param.name == "pipebox":
+            param.default = os.environ.get("PIPEBOX", Path(__file__).parent.parent / "pipebox.py")
+
+    # Add `init` to project subcommand
+    project.lazy_subcommands["init"] = "revng.internal.cli.revng2:init"
```
aleclearmind: Is using `"revng.internal.cli.revng2:init"` as a string the only way? Rather sad.

python/revng/pypeline/main.py
```diff
@@ -10,13 +10,18 @@
...
 import click
+from click.shell_completion import get_completion_class
 
 from . import initialize_pypeline
 from .cli.utils import EagerParsedPath, LazyGroup
 
-logger = logging.getLogger("pypeline")
+logger = logging.getLogger(__name__)
+handler = logging.StreamHandler(sys.stderr)
+formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
```
aleclearmind: Centralize this logic, we'll change it soon and I don't want to get crazy.

python/revng/pypeline/main.py
```diff
@@ -63,34 +68,88 @@ def import_pipebox(module_path: str, is_complete: bool) -> object:
...
+    root = ctx.find_root()
+    command = root.command
+    assert command.name is not None, "Command name should not be None"
+    return command.name
+
+
+@click.command()
+@click.option(
+    "--shell",
+    type=click.Choice(["bash", "zsh", "fish"]),
```
aleclearmind, line 172: Are these those supported by `click`? What if in the future they add more?

python/revng/pypeline/main.py
```diff
@@ -63,34 +68,88 @@ def import_pipebox(module_path: str, is_complete: bool) -> object:
...
+    default=Path(os.environ.get("SHELL", "bash")).name,
```
aleclearmind: Looks weird to me that we have to take care of this. Doesn't `click` do this on its own? :thinking:

python/revng/pypeline/cli/project/analyze.py
```diff
@@ -106,17 +101,20 @@ def build_analysis_command(
...
-        logger.debug("model: `%s`", model)
         logger.debug("and kwargs: `%s`", kwargs)
 
         # Load the model
-        storage_provider = storage_provider_factory(model_path=model)
+        storage_provider_factory = storage_provider_factory_factory("local://")
+        storage_provider = storage_provider_factory.get(
+            project_id="WIP",
+            token="WIP",
+            cache_dir="WIP",
```
aleclearmind: Why?

python/revng/pypeline/storage/local_provider.py
```diff
@@ -112,6 +117,55 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        token: str | None,
+        cache_dir: str | None,
+    ) -> StorageProvider:
+        assert cache_dir is not None, "Cache directory must be provided"
+
+        if project_id in self.providers:
+            return self.providers[project_id]
+
+        # Figure out how the model should be name
+        model_ty = get_singleton(Model)  # type: ignore [type-abstract]
```
aleclearmind: Do not abbreviate, `model_type`

python/revng/pypeline/storage/local_provider.py
```diff
@@ -112,6 +117,55 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        model_name = model_ty.model_name()
+
+        # Find the model in the current directory or any of its parents
+        folder = Path.cwd().resolve()
+        while True:
+            logger.debug("Searching for model at '%s'", folder / model_name)
+            if (folder / model_name).exists():
+                break
+            if folder == folder.parent:
+                raise FileNotFoundError(f"Model '{model_name}' not found")
```
aleclearmind: Do not use `'` for user-visible messages. Use `"`.

python/revng/pypeline/storage/local_provider.py
```diff
@@ -112,6 +117,55 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+
+        if project_id in self.providers:
+            return self.providers[project_id]
+
+        # Figure out how the model should be name
+        model_ty = get_singleton(Model)  # type: ignore [type-abstract]
+        model_name = model_ty.model_name()
+
+        # Find the model in the current directory or any of its parents
+        folder = Path.cwd().resolve()
```
aleclearmind: Do not use the term "folder", they are called "directory" in the unix world `:P`

python/revng/pypeline/storage/local_provider.py
```diff
@@ -112,6 +117,55 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+            logger.debug("Searching for model at '%s'", folder / model_name)
+            if (folder / model_name).exists():
+                break
+            if folder == folder.parent:
+                raise FileNotFoundError(f"Model '{model_name}' not found")
+            folder = folder.parent
+
+        model_path = folder / model_name
+        logger.info("Model '%s' found at '%s'", model_name, model_path)
+        # Compute the hash of the model path as a tentative unique identifier
```
aleclearmind: In Python we allow for 100 cols.

python/revng/pypeline/storage/local_provider.py
```diff
@@ -112,6 +117,55 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        # for the project
+        hasher = hashlib.sha256()
+        hasher.update(str(model_path).encode())
+        db_name = hasher.hexdigest() + ".sqlite"
```
aleclearmind: 1. Outline hashing into a helper function.
2. I'm a bit concerned about using the absolute path as the database key.
   Obviously, if one moves the model file, all the caches will be lost.
   I'd be tempted to use as key the hash of the model file, but that would require continually renaming the DB, which is not super nice.
   We could create a symlink named after the last known hash of the model, to use as a fallback.
   We need to think a bit more here. cc/@mrjackv

python/revng/pypeline/storage/storage_provider.py
```diff
@@ -70,6 +78,64 @@ def __len__(self) -> int:
...
+        of them. The methods are not supposed to be thread-safe, and it's
+        responsibility of the user to ensure that this method is not called
+        concurrently.
+        """
+
+
+# WIP: find a more suitable name
+def storage_provider_factory_factory(url: str) -> StorageProviderFactory:
+    """Create a storage provider factory from an url."""
+    protocol, route = url.split("://", 1)
```
aleclearmind: Why not a proper URL parser?

python/revng/pypeline/storage/storage_provider.py
```diff
@@ -70,6 +78,64 @@ def __len__(self) -> int:
...
+    factories = get_registry(StorageProviderFactory)  # type: ignore [type-abstract]
+    for factory_ty in factories.values():
+        if factory_ty.protocol() == protocol:
+            return factory_ty(route)
+    available_protocols = ", ".join(factories.keys())
+    raise ValueError(
+        f"Unknown storage provider protocol {protocol}."
+        f" The available protocols are: {available_protocols}."
+    )
```
aleclearmind, line 156: Are these exception being caught and reported gracefully to the user?

python/revng/pypeline/cli/utils.py
```diff
@@ -141,6 +151,41 @@ def _lazy_load(self, cmd_name):
...
+                f"Expected a string, but got value of type {type(value).__name__}.", param, ctx
+            )
+
+        try:
+            parsed_url = urlparse(value)
+        except ValueError:
+            # urlparse can raise ValueError on rare malformed inputs
+            self.fail(f"'{value}' could not be parsed as a URL.", param, ctx)
+
+        # get all the registered providers
```
aleclearmind: Start comments with capital letter

python/CMakeLists.txt
```diff
@@ -24,64 +24,64 @@ execute_process(
...
 function(python_module)
-  set(options)
-  set(oneValueArgs TARGET_NAME WHEEL MODULE_INIT)
-  set(multiValueArgs MODULE_FILES MODULE_GENERATED_FILES)
-  cmake_parse_arguments(PYTHON_MODULE "${options}" "${oneValueArgs}"
+    set(options)
+    set(oneValueArgs TARGET_NAME WHEEL MODULE_INIT)
+    set(multiValueArgs MODULE_FILES MODULE_GENERATED_FILES)
+    cmake_parse_arguments(PYTHON_MODULE "${options}" "${oneValueArgs}"
                         "${multiValueArgs}" ${ARGN})
```
aleclearmind, line 31: Why these whitespace changes? Drop.

python/revng/pypeline/utils/__init__.py
```diff
@@ -2,6 +2,11 @@
 # This file is distributed under the MIT License. See LICENSE.md for details.
 #
 
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from . import cabc, default_dict_from_key, registry  # noqa: F401
```
aleclearmind: I don't understand why we need these :thinking:

python/CMakeLists.txt
```diff
@@ -24,64 +24,64 @@ execute_process(
...
 function(python_module)
-    set(options)
-    set(oneValueArgs TARGET_NAME WHEEL MODULE_INIT)
-    set(multiValueArgs MODULE_FILES MODULE_GENERATED_FILES)
-    cmake_parse_arguments(PYTHON_MODULE "${options}" "${oneValueArgs}"
+  set(options)
+  set(oneValueArgs TARGET_NAME WHEEL MODULE_INIT)
+  set(multiValueArgs MODULE_FILES MODULE_GENERATED_FILES)
+  cmake_parse_arguments(PYTHON_MODULE "${options}" "${oneValueArgs}"
                         "${multiValueArgs}" ${ARGN})
```
aleclearmind, line 31: Do not change whitespaces `:P`

python/revng/pypeline/daemon/notification_broker/local_broker.py
aleclearmind, line 1: > Pypeline: Rename PostOffice to NotificationBroker

We still have `postoffice.py` :thinking:
Am I missing something?

Also, this is a fixup commit. Squash all fixup commits.

aleclearmind:
NOTETOSELF: review from `Pypeline: add pypeline.daemon`

python/revng/pypeline/main.py
zommiommy: Yeah, those are [the shells supported by click](https://click.palletsprojects.com/en/stable/shell-completion/).

They don't directly expose the supported shells (there's a "private" variable [`_available_shells`](https://github.com/pallets/click/blob/309ce9178707e1efaf994f191d062edbdffd5ce6/src/click/shell_completion.py#L429)).

All the supported shell completions extend [`ShellComplete`](https://github.com/pallets/click/blob/309ce9178707e1efaf994f191d062edbdffd5ce6/src/click/shell_completion.py#L304C20-L304C33) and have a `name` attribute. Thus, using our own [`register_all_subclasses`](https://github.com/zommiommy/revng/blob/1398a4e404826bf9a891fc8f02edd02a7096c80c/python/revng/pypeline/utils/registry.py#L35) we can discover them with:

```python
from click.shell_complete import ShellComplete
from revng.pypeline.utils.register import register_all_subclasses, get_registry

register_all_subclasses(ShellComplete)
shells = [
    shell_complete.name
    for shell_complete in get_registry(ShellComplete).values()
]
```

If you want I'll add this, I just didn't think the extra complexity was worth.

python/revng/pypeline/main.py
zommiommy: Nope, the [docs says](https://click.palletsprojects.com/en/stable/shell-completion/) that to install the completion, you are supposed to run one of the following:
```
# bash
eval "$(_FOO_BAR_COMPLETE=bash_source foo-bar)"
# zsh
eval "$(_FOO_BAR_COMPLETE=zsh_source foo-bar)"
# fish
_FOO_BAR_COMPLETE=fish_source foo-bar | source
```
Or to add them to the respective profile files.

But during testing, I never remember the correct env var, so I do it this way so I can just run `eval "$(revng2 autocomplete)"` and _it just works_ ™ (for both `revng2` and `pype` with the same code).

Perhaps in the future, it won't be as important if `orc install revng` also adds the lines above to the orc shell profile files.

python/CMakeLists.txt
zommiommy: I think I ran the formatter on it by error, sorry, I'll revert it.

python/revng/pypeline/utils/__init__.py
zommiommy: That's me trying to get mypy (and other linters) to run on the full codebase. The idea was to force imports of all the modules at type checking time. I'm not 100% sure it's needed tbh but I was able to get more lints this way

python/revng/pypeline/daemon/notification_broker/local_broker.py
zommiommy: Forgot a `git rm` I guess, sorry

python/revng/internal/cli/revng2.py
zommiommy: No, I can modify the [`LazyGroup`](https://github.com/revng/revng/blob/6d1fbdca9723049f2b5ad6173cf7b621adc4cc12/python/revng/pypeline/cli/utils.py#L111) class to also accept "eager" groups, so I can pass directly the `init` object

zommiommy: But maybe we don't actually need `LazyGroup` anymore, it was needed so the `main` could import the pypebox before the subcommands were initialized. But now the pypebox is loaded with [`EagerParsedPath`](https://github.com/revng/revng/blob/6d1fbdca9723049f2b5ad6173cf7b621adc4cc12/python/revng/pypeline/cli/utils.py#L67) so I think we should be able to just use a normal `click.Group`. 

This would remove code, but would make imports eager so we would get slightly worse latency on the CLI.

python/revng/pypeline/storage/local_provider.py
zommiommy: 1. Done

python/revng/pypeline/storage/storage_provider.py
zommiommy: Because we don't have to parse the URL.
I need the protocol to select the correct storage provider factory and forward the protocol-less URL to the factory constructor.

I tried using `urlparse` as Jack already suggested, and I can certainly extract the protocol, but then there is no reasonable way to build the URL without a protocol.

To clarify, the cases I have in mind are:
- `local://cache/revng/db.sqlite`
- `hss://hss.revng.com/?`

Where the "route" can be used to configure the factory, such as specifying which SQLite database to use in the `local` case, or the HSS host, along with parameters that can be used to configure the HSS use.

python/revng/pypeline/utils/logger.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def enable_debug_logging(module_name: str) -> None:
+    """Enable debug logging for a given module name."""
+    ENABLED_MODULES.add(module_name)
+
+
+def disable_debug_logging(module_name: str) -> None:
+    """Disable debug logging for a given module name."""
+    ENABLED_MODULES.discard(module_name)
```
mrjackv: I would have gone for something more local, e.g.

```
class Logger:
   ...
   def enable_debug(self): ...
   def disable_debug(self): ...

pypeline_logger = Logger("pypeline")
```

Also, this avoids:
* Creating multiple `Logger` objects, one for each module where we call `get_logger`
* Possible typos, e.g. if one does `get_logger("pypelin")` it creates a new logger without errors (as opposed to an `ImportError`)

python/revng/pypeline/storage/local_provider.py
mrjackv: 2. The only sensible argument I can bring is that:
    * Relying on the model path is easy
    * it's a cache, if a user moves the model file (which is not something that common) then I think it's acceptable that the cache needs to be recomputed
    
    So doing it *currently* this way is short and sweets and unlocks 99% of what we need.
    That said, I agree that maybe in the long run we need to devise something more refined, but having to keep track of model contents changes seems overkill (w.r.t. the already big scope of this PR)

python/revng/pypeline/storage/storage_provider.py
zommiommy: No.
Exceptions are not meant to be caused by user actions, but by developers' errors.
This one indeed should not be an exception, but rather a helpful error message for the user.

python/revng/pypeline/main.py
aleclearmind: Not worth it.

python/revng/pypeline/storage/local_provider.py
aleclearmind: 2. OK. Leave a TODO clearly explaining the limitations and hinting that in the future we can do something along the lines of what I suggested.

python/revng/pypeline/storage/storage_provider.py
aleclearmind: > then there is no reasonable way to build the URL without a protocol

What's a "URL without a protocol"? :thinking:

python/revng/pypeline/storage/storage_provider.py
aleclearmind: > Exceptions are not meant to be caused by user actions, but by developers' errors

We usually `assert` in that case.
Exceptions are useful when something is somehow "recoverable". Using them to report information to the user is not entirely crazy, as long as they are formatted appropriately.

> This one indeed should not be an exception, but rather a helpful error message for the user.

I assume you fixed this then?

python/revng/internal/cli/revng2.py
```diff
@@ -12,42 +12,66 @@
...
-        "pipeline": "revng.pypeline.cli.pipeline:pipeline",
-        "project": "revng.pypeline.cli.project:project",
-    },
-)
-def cli():
-    pass
+@click.group()
+def quick():
+    """Quick commands (japanese toilet)"""
+    # WIP
```
aleclearmind: `WIP` comments cannot be merged. Does this pass `orc check-branch`?

python/revng/internal/pipeline.yml
```diff
@@ -0,0 +1,5 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+# placeholder for the revng pipeline
```
aleclearmind: Always start comments with capital.

python/revng/pypeline/cli/project/artifact.py
```diff
@@ -166,12 +170,14 @@ def run_analysis_command(
...
-            res_path,
-        )
+        if result_path is not None:
+            logger.debug('Writing result to: "%s"', result_path)
+            dump_container(
+                res_container,
+                result_path,
+            )
+        else:
+            print(dumps_container(res_container))
```
aleclearmind: `print` adds a newline, is that what we want?

python/revng/pypeline/container.py
```diff
@@ -150,38 +152,61 @@ def contains_all(self, obj: ObjectSet) -> bool:
...
-    obj_id_type = get_singleton(ObjectID)  # type: ignore[type-abstract]
     with open(path, "r", encoding="utf-8") as f:
-        container.deserialize(
-            {obj_id_type.deserialize(k): bytes.fromhex(v) for k, v in json.load(f).items()}
-        )
-    return container
+        return loads_container(container_type, f.read())
+
+
+def dumps_container(
```
aleclearmind: 1. Why is this not a method of container?
2. The `dumps` and `loads` convention is rather poor, let's not follow it.
   `.to_string()` and `.from_string()`?
   Same for `loads_container` and friends.

python/revng/pypeline/cli/utils.py
```diff
@@ -91,67 +81,61 @@ def __init__(
...
+        except ValueError:
+            # urlparse can raise ValueError on rare malformed inputs
+            self.fail(f'"{value}" could not be parsed as a URL.', param, ctx)
+
+        # Get all the registered providers
+        allowed_schemes = {
+            factory.scheme().lower() for factory in get_registry(StorageProviderFactory).values()
+        }
+        # Check that the scheme is supported
+        if parsed_url.scheme.lower() not in allowed_schemes:
```
aleclearmind: I'd drop `.lower()` everywhere. Postel's law is evil.
Let's be strict.

python/revng/pypeline/cli/utils.py
```diff
@@ -205,13 +189,20 @@ def build_help_text(
...
 ) -> str:
     """
     Build a standardized help text for a command.
     """
-    help_text = prologue
+    help_text: str = prologue
     help_text += "\n\n\b\nArguments:"
-    help_text += "\n - [R] MODEL : Path - The path to the model file to use."
+    if model_help:
+        help_text += "\n - [R] MODEL : Path - The path to the model file to use."
```
aleclearmind: I'd drop the `[R]` convention, it's not obvious.
If you really want, add a piece of text stating that this file will not be changed.

python/revng/pypeline/main.py
```diff
@@ -63,34 +65,127 @@ def import_pipebox(module_path: str, is_complete: bool) -> object:
...
     expose_value=False,
 )
-def cli():
+@click.pass_context
+def pype(ctx):
     pass
 
 
+pype.add_command(pipeline)
+pype.add_command(project)
```
aleclearmind, line 152: Do we really need to have these here? `:/`

python/revng/pypeline/storage/local_provider.py
```diff
@@ -147,7 +151,58 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+
+        model_path = directory / model_name
+        logger.info('Model "%s" found at "%s"', model_name, model_path)
+        # Compute the hash of the model path as a tentative unique identifier for the project
+        # TODO: we are relying on the *absolute* model path, which means that if the
+        # user moves the project around, it will be treated as a different project
+        # and caches will be recomputed, and most importantly, if someone deletes
+        # the project and creates a new one at the same path, it will reuse the
+        # old cache.
+        db_name = sha256(str(model_path)) + ".sqlite"
```
aleclearmind: Rename `sha256` to something more generic like `crypto_hash` or `strong_hash`.

python/revng/pypeline/cli/project/daemon.py
```diff
@@ -0,0 +1,101 @@
...
+    "--host",
+    type=str,
+    default="127.0.0.1",
+    help="The interface to bind to.",
+    show_default=True,
+)
+@click.option(
+    "--port",
+    type=int,
+    default=5000,
```
aleclearmind: Is this the port we use on the old daemon too? Let's use the original one.

python/revng/pypeline/cli/project/daemon.py
```diff
@@ -0,0 +1,101 @@
...
+
+    with open(ctx.obj["pipeline_path"], "r", encoding="utf-8") as f:
+        pipeline_yaml = yaml.safe_load(f.read())
+
+    app.state.daemon = Daemon(
+        version=revng.__version__,
+        pipeline_yaml=pipeline_yaml,
+        pipeline=ctx.obj["pipeline"],
+        debug=debug,
+        lock=lock_ty(),
```
aleclearmind: `_ty`

python/revng/pypeline/daemon/app.py
```diff
@@ -0,0 +1,152 @@
...
+        content={"error": str(exc), "traceback": traceback.format_exc()}, status_code=500
+    )
+
+
+def get_project_id(headers: dict[str, str]) -> str | None:
+    """Extract project ID from headers, return none if missing"""
+    return headers.get("x-projectid")
+
+
+# WIP how to handle invalidations
```
aleclearmind: No `WIP` comments, unless you want to discuss them.

Please review all the code.

python/revng/pypeline/daemon/lock/__init__.py
aleclearmind, line 1: I'm really skeptical about merging this file.

1. There's 4 new exceptions that no one is catching.
2. It is unclear to me what guarantees do we have that the lock renewal task is run within a certain time frame.
3. The code of `local_lock.py` is hard to read. No comments, no blank lines.
4. It is unclear to me why certain responsibilities are delegated to implementors of `Lock`. For instance, why can't we check the epoch in `Lock` itself (or in `LockManager`)?
   Similarly, are we sure that the whole expiration business cannot be centralized?
5. There is a whole discussion about network latency and and drifting clocks... but the timeout is by default 60 seconds AFAIU. It doesn't make much sense to me.
6. The whole usage of `metadata` is obscure to me.

Overall, this is not a trivial problem, and, as a reviewer, it's not clear what problems we are trying to solve. I understand that this was initially geared towards handling HHS as well, but, as it is, it's more of liability to maintain than helping us solve the problems we will have then.

We really needed a REP for this in order for me to be able to review this. But it's not in scope.

I'd like to minimize the scope of the locking system to merge now to a reduced set of problems:

1. Prevent doing two things at the same time
2. Reject wrong epochs

Anything concerning time (expiration, timeouts, network latency, disruptions in the space-time continuum) is out of scope. Lock renewals are out of scope.

I'd also be of the opinion that we don't need an abstract class until we have multiple implementations.

What's here has value, but it needs to be discussed and it needs a REP clearly stating what problems we're trying to solve.

python/revng/pypeline/daemon/lock/local_lock.py
```diff
@@ -0,0 +1,69 @@
...
+
+class LocalLock(Lock):
+    """
+    An in-memory lock useful if there is only one daemon running.
+    """
+
+    def __init__(self, lock_duration: int = 60):
+        self._epochs: dict[ProjectID, Epoch] = defaultdict(lambda: Epoch(0))
+        self._locks: dict[ProjectID, LockInfo] = {}
+        self._locks_lock = asyncio.Lock()  # Async lock to protect access to self._locks
```
aleclearmind: No same line comments

python/revng/pypeline/daemon/notification_broker/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+        try:
+            while self.is_active:
+                # Wait for a message in the queue
+                message = await self.message_queue.get()
+                if not self.is_active:
+                    break
+
+                try:
+                    await self.stream.write(message.encode())
+                except Exception as e:
```
aleclearmind: Catching `Exception` is illegal. What specific exceptions do you expect to have here?

python/revng/pypeline/daemon/notification_broker/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+                    logger.error(
+                        f"Error sending message to stream for project {self.project_id}: {e}"
+                    )
+                    break
+        except Exception as e:
```
aleclearmind: Catching `Exception` is illegal. What specific exceptions do you expect to have here?

python/revng/pypeline/daemon/notification_broker/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+        self.is_active = False
+
+
+class NotificationPublisher(ABC):
+    """NotificationPublisher is just the transmitting part of the notification
+    system and sends the notifications to the NotificationBroker.
+
+    This is separate from the NotificationBroker because the NotificationBroker
+    might, and probably will, be a separate service on a different machine, but
+    we still need the transmission part to be able to send notifications.
```
aleclearmind: I don't understand, can you make an (even hypothetical) example of `NotificationBroker` implementation able to work with distinct `NotificationPublisher`s?

The fact that the broker is on a different machine doesn't seem to justify to me the fac that we also need separate classes for the publisher and the broker.

python/revng/pypeline/daemon/notification_broker/local_broker.py
```diff
@@ -0,0 +1,91 @@
...
+        self.broker = broker
+
+    async def notify(self, project_id: ProjectID, message: str | bytes):
+        await self.broker.notify(project_id, message)
+
+
+class LocalNotificationBroker(NotificationBroker):
+    def __init__(self):
+        self._lock = asyncio.Lock()
+        self.subscribers: dict[ProjectID, set[NotificationSubscriber]] = defaultdict(set)
```
aleclearmind: I think it'd be safer to wrap `self.subscribers` in an object that exposes all the methods of the underlying object, but asserts the lock has to be taken before invoking them.

Maybe such a primitive could used elsewhere. Maybe it already exist.

```python
lel = Locked(x)

lel.lol() # Aborts

with lel.lock():
  lel.lol() # Works
```
Might be problematic with typing though.

python/revng/pypeline/daemon/notification_broker/local_broker.py
```diff
@@ -0,0 +1,91 @@
...
+                if len(self.subscribers[project_id]) == 0:
+                    del self.subscribers[project_id]
+
+        subscriber.close()
+        logger.debug(f"Stream unsubscribed from project {subscriber.project_id}")
+
+    def get_publisher(self) -> NotificationPublisher:
+        return LocalNotificationPublisher(self)
+
+    async def notify(self, project_id: ProjectID, message: str | bytes):
```
aleclearmind: Any reason to accept bytes?
If not, let's stick with `str`.

python/revng/pypeline/storage/storage_provider.py
```diff
@@ -128,9 +128,9 @@ def storage_provider_factory_factory(url: str) -> StorageProviderFactory:
     """Create a storage provider factory from an url."""
     scheme = urlparse(url).scheme
     factories = get_registry(StorageProviderFactory)  # type: ignore [type-abstract]
-    for factory_ty in factories.values():
-        if factory_ty.scheme() == scheme:
-            return factory_ty(url)
+    for factory_type in factories.values():
+        if factory_type.scheme() == scheme:
+            return factory_type(url)
     available_schemes = ", ".join(factories.keys())
```
aleclearmind, line 152: This changeset probably belongs to some other commit.

python/revng/pypeline/storage/local_provider.py
```diff
@@ -172,8 +172,8 @@ def get(
             return self.providers[project_id]
 
         # Figure out how the model should be name
-        model_ty = get_singleton(Model)  # type: ignore [type-abstract]
-        model_name = model_ty.model_name()
+        model_type = get_singleton(Model)  # type: ignore [type-abstract]
+        model_name = model_type.model_name()
```
aleclearmind: This changeset probably belongs to some other commit.

python/revng/pypeline/daemon/utils.py
```diff
@@ -0,0 +1,105 @@
...
+from typing import Any, Optional
+
+from revng.pypeline.model import ReadOnlyModel
+from revng.pypeline.object import Kind, ObjectID, ObjectSet
+from revng.pypeline.pipeline import Pipeline
+from revng.pypeline.storage.storage_provider import StorageProvider
+from revng.pypeline.utils.registry import get_singleton
+
+
+class BasicException(Exception):
```
aleclearmind: `DaemonException`?

tests/pypeline/pipebox.py
```diff
@@ -22,6 +22,10 @@
 T = TypeVar("T")
 
 
+def initialize(args: list[str]) -> None:
+    print("Pipebox initialized with args:", args)
```
aleclearmind: Never write to stdout informational messages

python/revng/pypeline/cli/utils.py
zommiommy: Consider that in the `pipeline` part of commands, I add [R] to files that will be read and [W] to files that will be written (and maybe read).

Like:
```
$ pype -C tests/pypeline pipeline run-pipe ToHigherKindPipe --help
Usage: pype pipeline run-pipe ToHigherKindPipe [OPTIONS] MODEL SOURCE
                                               DESTINATION

  Take the root object from the input container and adds all its children to
  the output container.

  Arguments:
   - [R] MODEL : Path - The path to the model file to use.
   - [R] SOURCE : RootDictContainer - the source container
   - [W] DESTINATION : ChildDictContainer - the destination container

Options:
  --list                          List the available objects for each
                                  argument.
  -s, --static-configuration TEXT
                                  Static configuration for the pipe
                                  "ToHigherKindPipe".
  -c, --configuration TEXT        Configuration for the pipe
                                  "ToHigherKindPipe".
  --destination-objects /CHILD1,/CHILD2,...
                                  The objects to require from container
                                  DESTINATION as a comma-separated list of
                                  IDs. If not passed, all objects will be
                                  requested.
  --help                          Show this message and exit.
```

python/revng/pypeline/container.py
zommiommy: 1. I can move them there; the reason is that there is no reason for a subclass to override them. Sure, we can put `@final`, but having them as functions enforces this.
2. Yeah `load` and `loads` is not great but it's common, so should I use `to_string`, `from_string`, `to_file`, and `from_file` ?

python/revng/pypeline/cli/utils.py
aleclearmind: 1. I understand but I don't like this syntax very much now that I see it. Maybe we should just write "this is an output file" or "this is an input file" at the end of the corresponding line.
   Also, out of `run-pipe` is much less meaningful, in particular in `revng artifact`.
2. >  add [R] to files that will be read and [W] to files that will be written (and maybe read).
   
   Side questions: how would a RW container appear in `run-pipe`?
   Do we get two positional arguments (input and output)?
   This is what [REP21](https://pad.rev.ng/s/NCYbJqNMM#) says.

python/revng/pypeline/container.py
aleclearmind: 1. Let's favor ergonomics more than trying to defend against people overriding things they shouldn't.
2. Yeah, we use that convention elsewhere.

python/revng/pypeline/main.py
zommiommy: Yes and no.

The alternative is that each command imports its "parent" and instead of doing `click.command()` does `pype.command()` or `project.command()`, but it means that the "deeper" file has to import the parent, which makes it easy to make circular imports.

A related topic is `PypeGroup` and `PypeCommand`; these are needed to add `--` to the usage line. 
With the current approach, I have to specify `cls=PypeGroup` and `cls=PypeCommand`. If, instead, we decide to invert the logic, I can specify [`group_class`](https://github.com/pallets/click/blob/309ce9178707e1efaf994f191d062edbdffd5ce6/src/click/core.py#L1530) and `command_class` on just the top-level group, and all subgroups will use them by default.

python/revng/pypeline/daemon/lock/__init__.py
zommiommy: Ok, if you agree, I'll remove all locking files from this PR. And we shall proceed with no locking until we discuss it.

python/revng/pypeline/daemon/notification_broker/__init__.py
zommiommy: 1. The doc is outdated; now the Broker and Publishers are merged into just the Broker.
2. My reasoning was that, if the pub/sub service is a completely different service, the only thing needed by the daemon is to just send notifications. In this case, the user would subscribe to talking directly (or through nginx) to the pub/sub service, and wait for notifications.

python/revng/pypeline/cli/project/daemon.py
zommiommy: Ok, will switch to port 8000

python/revng/pypeline/cli/utils.py
zommiommy: 1. 
> Maybe we should just write "this is an output file" or "this is an input file" at the end of the corresponding line.

I think that this would make it harder to figure out what's an input and what's an output. 

Maybe instead of just `Arguments` we can have `Input arguments` and `Output arguments`.

> Also, out of run-pipe is much less meaningful, in particular in revng artifact.

I agree.

2. Yup
```
$ pype -C tests/pypeline pipeline run-pipe InPlacePipe --help     
Usage: pype pipeline run-pipe InPlacePipe [OPTIONS] MODEL ARG_INPUT ARG_OUTPUT

  Modifies the input container in place.

  Arguments:
   - [R] MODEL : Path - The path to the model file to use.
   - [R] ARG_INPUT : ChildDictContainer - the input container the pipe will modify
   - [W] ARG_OUTPUT : ChildDictContainer - Like above, but for output.

Options:
  --list                          List the available objects for each
                                  argument.
  -s, --static-configuration TEXT
                                  Static configuration for the pipe
                                  "InPlacePipe".
  -c, --configuration TEXT        Configuration for the pipe "InPlacePipe".
  --arg-objects /CHILD1,/CHILD2,...
                                  The objects to require from container ARG as
                                  a comma-separated list of IDs. If not
                                  passed, all objects will be requested.
  --help                          Show this message and exit.
```

python/revng/pypeline/daemon/lock/__init__.py
aleclearmind: I'd like to reduce the current implementation to just handling the two situations I mentioned above: wrong epoch and preventing multiple requests from overlapping.

This a simple and reasonable scope that enables us to run things locally safely (where we don't really have timing issues, unless there's a biiiig problem).

python/revng/pypeline/daemon/notification_broker/__init__.py
aleclearmind: > The doc is outdated; now the Broker and Publishers are merged into just the Broker.

1. Uh? I still see two separate classes. Something is not pushed?

> My reasoning was that, if the pub/sub service is a completely different service, the only thing needed by the daemon is to just send notifications. In this case, the user would subscribe to talking directly (or through nginx) to the pub/sub service, and wait for notifications.

I see. Still, these are all things we can introduce later. Introducing code that's unused is complexity we have to maintain. Also, without proper design we're likely to miss some requirements.
Let's keep it simple and support what we need and/or have discussed a design for, as of today.

python/revng/pypeline/cli/utils.py
aleclearmind: > I think that this would make it harder to figure out what's an input and what's an output.

Right now we append `_INPUT` and `_OUTPUT` only if it's `READ_WRITE`:

```python
        for arg in pipe_ty.signature():
            if arg.access == TaskArgumentAccess.READ_WRITE:
                run_pipe_command = click.argument(
                    f"{arg.name}-input",
                    type=click.Path(exists=True, dir_okay=False, readable=True),
                )(run_pipe_command)
                run_pipe_command = click.argument(
                    f"{arg.name}-output",
                    type=click.Path(dir_okay=False, writable=True),
                )(run_pipe_command)
            elif arg.access == TaskArgumentAccess.READ:
                run_pipe_command = click.argument(
                    arg.name,
                    type=click.Path(exists=True, dir_okay=False, readable=True),
                )(run_pipe_command)
            elif arg.access == TaskArgumentAccess.WRITE:
                run_pipe_command = click.argument(
                    arg.name,
                    type=click.Path(dir_okay=False, writable=True),
                )(run_pipe_command)
```

Maybe we could just add those suffixes unconditionally, drop `[R]`/`[W]` and call it a day.

More uniformity and makes it obvious by the name it's an input vs output.

python/revng/pypeline/daemon/notification_broker/local_broker.py
aleclearmind: If this takes too much time, just leave a TODO.

python/revng/pypeline/cli/utils.py
zommiommy: Ok will do

python/revng/pypeline/daemon/notification_broker/local_broker.py
zommiommy: I don't think anything like that already exists.
The most similar thing we can safely do is:
```python
from threading import Lock

class Locked[T]:
    def __init__(self, value: T):
        self._value = value
        self.lock = Lock()
        
    @property
    def value(self) -> T:
        if not self.lock.locked():
            raise RuntimeError("Trying to acces a locked value without locking")
        return self._value
        
    def __enter__(self):
        self.lock.acquire()
        
    def __exit__(self, exc_type, exc_value, exc_traceback):
        self.lock.release()
        
l = Locked(10)

l.value # raises an exception

with l:
    l.value # can be done
```

python/revng/pypeline/main.py
zommiommy: I experimented with this. While inverting the chain makes it cleaner and can work, I think it's a bit awkward, and idk how to fix it.

For example, suppose we have a top-level command with a subcommand.

The folder structure would be:
```
project/
     __init__.py
     sub_command.py
```

`project/__init__.py`
```python
import click

@click.group()
def project():
    pass
```

`project/sub_command.py`
```python
from . import project

@project.command()
def subcommnad():
    pass
```

This is how I would structure it. But this has the problem that if `sub_command` is not imported by anyone it won't register the subcommand.

Therefore the logical place where to put it is in `__init__.py` like:

```python
import click

@click.group()
def project():
    pass

from . import sub_command as _
``` 

But note that the import must go AFTER the project definition, or we'd get a cyclical import.

I think this is worse than manually calling `add_command`, or we need to think a different folder / modules structure.

python/revng/pypeline/daemon/notification_broker/local_broker.py
zommiommy: The problem with this is that it fails in this case:
```python
from time import sleep
l = Locked(10)

def thread1():
    with l:
        sleep(30)

def thread2():
     sleep(1)
     l.value
```

Meaning that it only cares that someone locked it, not that the current thread acquired the lock.

zommiommy: When we acquire the lock, we can try to save the `pid`, `tid`, `thread_name`, and so on to fingerprint the thread, and then check it on the get, but we need a second lock to guard against mutations to these infos.

While I'd love some primitive like this, I don't think it can be easily done in python.

aleclearmind: Right. Grok suggests to have the context manager yield the inner object: since that would be the only way to obtain the object (apart from accessing manually `._value`, which we can ignore), it should be safe.

```python
T = TypeVar('T')

class Locked(Generic[T]):
    def __init__(self, obj: T) -> None:
        self._value: T = obj
        self._lock = ...

    @contextmanager
    def __call__(self) -> Any:
        """
        Context manager that acquires the lock and yields the protected object.
        """
        with self._lock:
            yield self._value
```

What do you think?

python/revng/pypeline/daemon/daemon.py
```diff
@@ -0,0 +1,340 @@
...
+    code: int
+    """The HTTP status code of the response."""
+    body: Any
+    """
+    The response body of the request.
+    """
+    headers: dict[str, str] = field(default_factory=dict)
+    """The headers of the response."""
+    notifications: list[Any] = field(default_factory=list)
+    """List of notifications to send through the NotificationPublisher"""
```
aleclearmind: `NotificationPublishers` no longer exists.

python/revng/pypeline/daemon/daemon.py
```diff
@@ -0,0 +1,340 @@
...
+    ):
+        self.version = version
+        self.pipeline_yaml = pipeline_yaml
+        self.pipeline = pipeline
+        self.debug = debug
+        self.storage_provider_factory = storage_provider_factory_factory(storage_provider_url)
+        self.cache_dir = cache_dir
+
+        self.epochs: dict[ProjectID, int] = defaultdict(lambda: 0)
+        self.lock = asyncio.Lock()
```
aleclearmind: This is kind of a scary lock.

Can you precisely elaborate what does it protect so that we have a crystal clear rule on when we need to acquire and when we don't?

I see you don't acquire it when accessing `request` or `get_singleton(Kind)`, but what's the general rule here?

Is it when we access something in `self`?

It'd be nice to hide all the objects protected by this lock into a clearly defined object that we can wrap into `Locked` (as we're discussing in another thread).

python/revng/pypeline/daemon/notification_broker/local_broker.py
zommiommy: That's definitely better.

The only problem left is this case:
```python

l = Locked(10)

with l as value:
    # DO stuff

# we can access value after the lock was released
print(value)
``` 

But we can decide that it's good enough and use it.

aleclearmind: I think that's rather adversarial and bad even if we were using `with open(...) as f`.

I think we can accept that, but let's add a comment.

An alternative would be to yield a `Nullable[T]` object that set to `None` upon `__exit__` from the context manager, but there would still be ways to circumvent this, if one really wants to be a nasty kid.

I think the implementation suggested above brings home 99% of the result, i.e., prevents accidental misuse of something protected by a lock.

zommiommy: K will use it. I don't think it is adversarial. This can happen if one mistakenly uses the wrong indentation on otherwise correct code.

python/revng/pypeline/daemon/daemon.py
zommiommy: It has three goals:
- protect access to `self.epochs`
- protects the creation of the storage provider, as we don't have any assumption of concurrency safety on it yet, and `set_model` and `get_model` might cause race conditions.
- stop multiple people from running multiple analyses or computing artifacts at the same time.

python/revng/pypeline/daemon/notification_broker/local_broker.py
zommiommy: Another problem is that we can't modify it.

```python
l = Locked(10)

with l() as value:
    value = 12

with l() as value:
    print(value)
```

This will still print 10, because `value` is an immutable integer.

aleclearmind: I agree it's a mistake that can happen, but it's broader problem that should IMO be addressed at linting level (i.e., don't use object defined in a `with` block out of the `with` block).

This said, if you think this is a real risk you can go with `Nullable[T]` with a `.value()` method. It comes at some readability cost, but it wouldn't be absurd.

python/revng/pypeline/daemon/daemon.py
aleclearmind: I understand the "high level" goals, but can we translate into a rule like "you shouldn't access A, B and C unless you have lock"?

This way we can move A, B and C into a separate object protected by `Locked`.

python/revng/pypeline/daemon/notification_broker/local_broker.py
aleclearmind: I understand, but it seems to me that your concerns are more related to Python (no references, weird variable scopes). Either we come up with a solution, or this is still better than hoping the developer remember to acquire the lock.

python/revng/pypeline/daemon/daemon.py
zommiommy: Our axioms are:
- While artifacts are being computed, or an analysis is running, no one can modify the model.
- The epoch must be modified iff the model changed, and thus `storage_provider.set_model` is called.
- You must not access `self.epochs` without a lock because someone else might be modifying it.

Therefore, in order to compute artifacts and run analysis, you need to lock **the entire lifetime of** `storage_provider` and the epochs object.

So we could split it into 3 locks, one that guards `self.epochs`, one that guards `self.storage_provider_factory`, and one that guards `self.pipeline`. But there would be 3 **nested locks**, and for `self.storage_provider_factory`, we must not release the lock until the produced `storage_provider` is dropped.

aleclearmind: I'm fine with a single lock, I'm just saying: let's outline all the resources protected by the lock into a `Locked` object.

Is it fair to say that all the fields of `Daemon` are read-only (post-`__init__`) *except* for `storage_provider_factory` and `epochs`?

If so, we can push them into a `Locked` object and call it a day.

zommiommy: Yes, they are read-only.

Will do, even if it's not a proper solution, because this way the lock is global, and so two different projects cannot run in parallel.

python/revng/pypeline/daemon/notification_broker/local_broker.py
zommiommy: done

python/revng/pypeline/daemon/daemon.py
zommiommy: done

python/revng/pypeline/container.py
```diff
@@ -151,62 +151,56 @@ def contains_all(self, obj: ObjectSet) -> bool:
...
-    """
-    with open(path, "w", encoding="utf-8") as f:
-        f.write(dumps_container(container))
+    @final
+    def to_string(self) -> str:
+        """
+        Dump a container into a serialized format.
+        """
+        return json.dumps(
+            {k.serialize(): bytes(v).hex() for k, v in self.serialize(self.objects()).items()},
```
mrjackv: In `from_string` base64 is used, whereas here `hex` is used, please squash [this commit](https://github.com/revng/revng/pull/536/commits/9f689c5a7214b04e2d238b5601d931b2cf8730ce) into this

python/revng/pypeline/container.py
```diff
@@ -151,62 +151,56 @@ def contains_all(self, obj: ObjectSet) -> bool:
...
+        )
+
+    @final
+    def to_file(self, path: str | Path) -> None:
+        """
+        Dump a container into a serialized format.
+        This is used to **save** cached objects from this container.
+        """
+        with open(path, "w", encoding="utf-8") as f:
+            f.write(self.to_string())
```
mrjackv: This is a bit of a niche optimization, but `to_file` could use `json.dump`, which avoids writing the intermediate string altogether. Leave a TODO comment.

python/revng/pypeline/cli/project/daemon.py
```diff
@@ -0,0 +1,88 @@
...
+import yaml
+
+import revng
+from revng.pypeline.daemon.app import app
+from revng.pypeline.daemon.daemon import Daemon
+
+logger = logging.getLogger(__name__)
+
+
+@click.command()
```
mrjackv, line 16: Add a `--production` option, should do the same as the `revng daemon` counterpart

python/revng/pypeline/cli/project/daemon.py
```diff
@@ -0,0 +1,88 @@
...
+    is_flag=True,
+    help="Enable debug mode.",
+)
+@click.option(
+    "--reload",
+    is_flag=True,
+    help="Enable auto-reload on file changes.",
+)
+# TODO: expose all other uvicorn parameters from
+# https://github.com/Kludex/uvicorn/blob/master/uvicorn/main.py
```
mrjackv: I'm not sure we want to expose all the CLI options of uvicorn here, that means every time an option is changed in uvicorn it needs to be changed here. I'd spend a little bit of time to see if there is an easy way to expose an `--uvicorn-args` option that works similarly to the old `revng daemon` (e.g. running main and manually composing the `argv`). Then the important options (e.g. `--port` or `--bind`) can be added as shortcuts

python/revng/pypeline/cli/project/daemon.py
```diff
@@ -0,0 +1,88 @@
...
+@click.pass_context
+def run_daemon(ctx, host, port, debug, reload, workers):
+    """Start the HTTP daemon."""
+
+    with open(ctx.obj["pipeline_path"], "r", encoding="utf-8") as f:
+        pipeline_yaml = yaml.safe_load(f.read())
+
+    app.state.daemon = Daemon(
```
mrjackv: Setting a sub-variable of a global variable is kind of hackish. I'd make a `make_starlette(daemon: Daemon)` function in `app.py` which returns a `Starlette` instance.

python/revng/pypeline/daemon/app.py
```diff
@@ -0,0 +1,150 @@
...
+    Route("/api/epoch", epoch_endpoint, methods=["GET"]),
+    Route("/api/pipeline", pipeline_endpoint, methods=["GET"]),
+    Route("/api/model", model_endpoint, methods=["GET"]),
+    Route("/api/artifact", artifact_endpoint, methods=["POST"]),
+    Route("/api/analysis", analysis_endpoint, methods=["POST"]),
+    WebSocketRoute("/api/subscribe", invalidation_websocket),
+]
+
+# Create the Starlette application
+app = Starlette(
```
mrjackv: Add `CORSMiddleware` and `GZipMiddleware` to the starlette instance (see `revng/internal/daemon/__init__.py`)

python/revng/pypeline/daemon/daemon.py
```diff
@@ -64,30 +61,31 @@ def __init__(
...
     async def get_epoch(self, request) -> Response:
-        async with self.project_data() as (epochs, _):
-            return Response(code=200, body={"epoch": epochs[request["project_id"]]})
+        project_id = request["project_id"]
+        token = request.get("token")
+        storage_provider_context = self.storage_provider_factory.get(
+            project_id=project_id,
+            token=token,
+            cache_dir=self.cache_dir,
+        )
```
mrjackv, line 73: Factor this out in `self.get_storage_provider_context(request)`

python/revng/pypeline/daemon/daemon.py
```diff
@@ -64,30 +61,31 @@ def __init__(
...
+        async with storage_provider_context as storage_provider:
+            return Response(code=200, body={"epoch": storage_provider.get_model()[1]})
```
mrjackv: `get_model()` can be quite expensive (e.g. a non-trivial model is in the 100s of kilobyte range), transfering all of it to then just return an int is quite wasteful. I'd have a separate `get_epoch` in the storage provider that just returns the int

python/revng/pypeline/container.py
```diff
@@ -152,39 +152,81 @@ def contains_all(self, obj: ObjectSet) -> bool:
...
+
+        file: TextIO = path_or_file
         json.dump(
-            {
-                k.serialize(): bytes(v).hex()
-                for k, v in container.serialize(container.objects()).items()
-            },
-            f,
+            self.to_dict(),
+            file,
```
mrjackv: Put this in the `else:` branch, so that you don't have to cast

python/revng/pypeline/container.py
```diff
@@ -152,39 +152,81 @@ def contains_all(self, obj: ObjectSet) -> bool:
...
+                return cls.from_file(f)
+
+        # Read the file
+        file: TextIO = path_or_file
+        return cls.from_dict(json.load(file))
+
+    @final
+    def to_dict(self) -> dict[str, str]:
+        return {
+            k.serialize(): bytes_to_string(v) for k, v in self.serialize(self.objects()).items()
```
mrjackv: You forgot `self.is_text()`

python/revng/pypeline/utils/__init__.py
```diff
@@ -2,14 +2,96 @@
...
+
+
+def bytes_to_string(data: Buffer, is_text: bool = False) -> str:
+    """Standard way to convert bytes to string in the codebase."""
+    if is_text:
+        return bytes(data).decode("utf-8")
+    return base64.b64encode(data).decode("utf-8")
+
+
+def string_to_bytes(data: str, is_text: bool = False) -> bytes:
```
mrjackv: Drop the default `= False`, this is an important argument, it should be explicitly specified in all call sites.
Also, make it `if: ... else: ...` so the two branches are aligned

aleclearmind:
Merged in 060ec57. Thanks.

PR #522, opened 2025-09-18
Pypeline pipes

mrjackv:
[Examples branch](https://github.com/mrjackv/revng/tree/feature/pypeline-pipes-examples)


include/revng/Canonicalize/SimplifySwitchPass.h
```diff
@@ -0,0 +1,44 @@
...
+private:
+  llvm::Pass &Pass;
+  const class Model &Model;
+  const BytesContainer &Binary;
+
+public:
+  static constexpr llvm::StringRef Name = "simplify-switch";
+  static constexpr revng::pypeline::PipeArgumentDocumentation
+    ArgumentsDocumentation[]{ { "Input", "" }, { "Output", "" } };
+  using ContainerTypes = std::tuple<const BytesContainer, LLVMRootContainer>;
```
aleclearmind: Let's avoid `std::tuple`, let's use something that does not have the semantic of instantiating the types, [like this](https://godbolt.org/z/jM5cdovv9).

mrjackv: Done

aleclearmind:
Undraft

include/revng/PipeboxCommon/Concepts.h
```diff
@@ -89,35 +89,54 @@ concept IsAnalysis = requires(T &A) {
...
                                              const revng::pypeline::Request &,
                                              const revng::pypeline::Request &,
                                              llvm::StringRef,
                                              Args...)> {
-  using ContainerTypes = std::tuple<std::remove_reference_t<Args>...>;
+  using ContainerTypes = TypeList<std::remove_reference_t<Args>...>;
   static constexpr size_t Size = sizeof...(Args);
 };
 
+template<typename T>
```
aleclearmind, line 113: Describe the difference between a "pipe" (i.e., a long-lived object that should *not* have state) and a "pipe run" (i.e., an object whose alive while running the pipe and therefore you can store state that's specific to that run).

include/revng/PipeboxCommon/Concepts.h
```diff
@@ -89,35 +89,54 @@ concept IsAnalysis = requires(T &A) {
 namespace detail {
 
 template<typename T>
-struct PipeRunTraits {};
+concept HasContainerTypes = SpecializationOf<typename T::ContainerTypes,
+                                             TypeList>;
```
aleclearmind, line 97: `Has`? `IsContainerList` maybe?

include/revng/PipeboxCommon/Helpers/PipeRunner.h
```diff
@@ -21,39 +21,34 @@ namespace revng::pypeline::helpers {
...
-                                                   llvm::StringRef,
-                                                   ContainersT...),
                         const Model &TheModel,
                         const pypeline::Request &Incoming,
                         const pypeline::Request &Outgoing,
                         llvm::StringRef Configuration,
                         ListType Containers) {
-    revng_assert(Incoming.size() == sizeof...(ContainersT));
-    revng_assert(Outgoing.size() == sizeof...(ContainersT));
+    using Traits = PipeRunTraits<T>;
```
aleclearmind, line 36: It's a "trait", not "trait*s*".
Fix everywhere.

include/revng/PipeboxCommon/Concepts.h
```diff
@@ -89,35 +89,54 @@ concept IsAnalysis = requires(T &A) {
...
+  requires(not HasContainerTypes<C>) and (IsContainerReference<Args> and ...)
+struct PipeRunTraitsHelper<
   revng::pypeline::ObjectDependencies (C::*)(const Model &,
                                              const revng::pypeline::Request &,
                                              const revng::pypeline::Request &,
                                              llvm::StringRef,
                                              Args...)> {
-  using ContainerTypes = std::tuple<std::remove_reference_t<Args>...>;
+  using ContainerTypes = TypeList<std::remove_reference_t<Args>...>;
   static constexpr size_t Size = sizeof...(Args);
```
aleclearmind: Rename to `ContainerCount`. But maybe these should be `ArgumentTypes` and `ArgumentCount`.

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
```diff
@@ -0,0 +1,380 @@
...
+#include "llvm/Pass.h"
+
+#include "revng/PipeboxCommon/Concepts.h"
+#include "revng/PipeboxCommon/LLVMContainer.h"
+#include "revng/PipeboxCommon/Model.h"
+#include "revng/Support/FunctionTags.h"
+
+namespace detail {
+
+class ObjectDependeciesHelper {
```
aleclearmind: Typo

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
aleclearmind, line 1: Split this file in one file per `class`. In particular, there's not point in pulling in LLVM headers for pipes that have nothing to do with LLVM IR.

include/revng/HeadersGeneration/ModelToHeaderPipe.h
```diff
@@ -0,0 +1,27 @@
...
+public:
+  static constexpr llvm::StringRef Name = "model-to-header";
+  static constexpr revng::pypeline::PipeArgumentDocumentation
+    ArgumentsDocumentation[]{ { "Buffer", "" } };
+
+  static void run(const Model &TheModel,
+                  llvm::StringRef StaticConfig,
+                  llvm::StringRef DynamicConfig,
+                  BytesContainer &Buffer);
+};
```
aleclearmind: Add a `static_assert` below each `class` you expect to respect a certain trait.

include/revng/HeadersGeneration/ModelToHeaderPipe.h
```diff
@@ -0,0 +1,27 @@
...
+#include "revng/PipeboxCommon/MapContainer.h"
+#include "revng/PipeboxCommon/Model.h"
+
+namespace revng::pypeline::pipes {
+
+class ModelToHeader {
+public:
+  static constexpr llvm::StringRef Name = "model-to-header";
+  static constexpr revng::pypeline::PipeArgumentDocumentation
+    ArgumentsDocumentation[]{ { "Buffer", "" } };
```
aleclearmind: Is this a "pipe" or a "pipe run"? Why do we still have an array?
Let's ditch arrays entirely in favor of `TypeList<SomethingArgument<...>>`.

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
```diff
@@ -0,0 +1,380 @@
...
+template<typename T, typename T2>
+struct SingleOutputPipeTraits {};
+
+template<typename T, typename... Args>
+struct SingleOutputPipeTraits<
+  T,
+  void (*)(const Model &, llvm::StringRef, llvm::StringRef, Args...)> {
+  using ContainerTypes = TypeList<Args...>;
+  static constexpr std::array
+    ArgumentsDocumentation = std::to_array(T::ArgumentsDocumentation);
```
aleclearmind: We're doing the same thing in too many different ways. Let's ditch `std::array` and `[]` in favor of a `TypeList<Something>`.

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
```diff
@@ -0,0 +1,380 @@
...
+    revng_assert(Index < Dependencies.size());
+    revng_assert(Outgoing[Index].size() == 1);
+    commitAllFor(Index);
+  }
+
+  void pushReadFields() { revng::Tracking::push(Binary); }
+
+  void popReadFields() { revng::Tracking::pop(Binary); }
+
+  revng::pypeline::ObjectDependencies &&takeDependencies() {
```
aleclearmind: I think it's common to use the term "extract" instead of "taking" here.

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
```diff
@@ -0,0 +1,380 @@
...
+    revng::Tracking::clearAndResume(Binary);
+  }
+  ~ObjectDependeciesHelper() { revng_assert(DependenciesTaken); }
+
+  ObjectDependeciesHelper(const ObjectDependeciesHelper &) = delete;
+  ObjectDependeciesHelper &operator=(const ObjectDependeciesHelper &) = delete;
+  ObjectDependeciesHelper(ObjectDependeciesHelper &&) = delete;
+  ObjectDependeciesHelper &operator=(ObjectDependeciesHelper &&) = delete;
+
+  void commit(const ObjectID &Object, size_t Index) {
```
aleclearmind: `Index` -> `ArgumentIndex`?

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
```diff
@@ -0,0 +1,380 @@
...
+
+  void commitUniqueTarget(size_t Index) {
+    revng_assert(Index < Dependencies.size());
+    revng_assert(Outgoing[Index].size() == 1);
+    commitAllFor(Index);
+  }
+
+  void pushReadFields() { revng::Tracking::push(Binary); }
+
+  void popReadFields() { revng::Tracking::pop(Binary); }
```
aleclearmind: I see a common pattern that could be made safer with a small helper class:

```
pushReadFields();
...
commit(Object);
popReadFields();
```

To:

```c++
class Committer {
public:
  Committer(Object) : Object(Object) { pushReadFields(); }
  ~Committer() { commit(Object); popReadFields(); }
};

for () {
  Committer C = ODH.committerFor(Object);
  ...
}
```

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
```diff
@@ -0,0 +1,380 @@
...
+template<typename T,
+         typename Base = SingleOutputPipeBase<T, WrappedPipeTraits<T>>>
+class LLVMFunctionPassPipe : public Base {
+private:
+  using LLVMRootContainer = revng::pypeline::LLVMRootContainer;
+  static_assert(std::is_same_v<typename Base::OutputContainerType,
+                               LLVMRootContainer>);
+  using ContainerTypesRef = detail::TupleWithRef<typename Base::ContainerTypes>;
+
+  class Pass : public llvm::ModulePass {
```
aleclearmind: On a second thought, I'd avoid having it as a nested class.

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
```diff
@@ -0,0 +1,380 @@
...
+  size_t Result = 0;
+  compile_time::forEach<T>([&Result]<typename A, size_t I>() {
+    if constexpr (not std::is_const_v<A>)
+      Result += 1;
+  });
+  return Result;
+}
+
+template<StrictSpecializationOf<TypeList> T>
+constexpr size_t writableContainerIndex() {
```
aleclearmind: 1. Rename to `uniqueWritableContainerIndx`
2. Assert there's actually only one

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
```diff
@@ -0,0 +1,380 @@
...
+  });
+  return Result;
+}
+
+} // namespace detail
+
+template<typename T>
+using SOPTraits = detail::SingleOutputPipeTraits<T, decltype(&T::run)>;
+
+template<typename T, typename C>
```
aleclearmind: What is `T`? I'd like a `concept` constraint on it, I'm losing track among the various abstraction levels here.

I also see that actual types you pass in `C` are typically parametric in `T`. This makes me think that you can add something like `C::inner_type` (`inner_type` is likely a bad name) and reduce the two template arguments to a single one.

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
```diff
@@ -0,0 +1,380 @@
...
+protected:
+  static constexpr size_t ContainersSize = std::tuple_size_v<ContainerTypes>;
+  static_assert(detail::writableContainersCount<ContainerTypes>() == 1);
+  static constexpr size_t
+    OutputContainerIndex = detail::writableContainerIndex<ContainerTypes>();
+  using OutputContainerType = std::tuple_element_t<OutputContainerIndex,
+                                                   ContainerTypes>;
+};
+
+template<typename T, typename Base = SingleOutputPipeBase<T, SOPTraits<T>>>
```
aleclearmind: Why is `Base` a template argument? If we never customize it, hardcode it.
This is getting quite complex, let's not have degrees of freedom we don't need.

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
```diff
@@ -0,0 +1,380 @@
...
+class SingleOutputPipe : public Base {
+public:
+  template<typename... Args>
+  revng::pypeline::ObjectDependencies
+  run(const Model &Model,
+      const revng::pypeline::Request &Incoming,
+      const revng::pypeline::Request &Outgoing,
+      llvm::StringRef Configuration,
+      Args &...Containers) {
```
aleclearmind: 1. Constraints on `Args`?
2. Can we propagate up until down here the expected list of arguments and assert they actually match?

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
```diff
@@ -0,0 +1,380 @@
...
+    revng_assert(Outgoing[this->OutputContainerIndex].size() == 1);
```
aleclearmind: Review all the uses of `operator[]` and see if you can use `.at` instead somewhere.

Actually I'm not even user how can you use `[]` on a `const` object.

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
```diff
@@ -0,0 +1,380 @@
...
+
+    detail::ObjectDependeciesHelper ODH(Model, Outgoing, this->ContainersSize);
+    T::run(Model, this->StaticConfiguration, Configuration, Containers...);
+    ODH.commitUniqueTarget(this->OutputContainerIndex);
+    return ODH.takeDependencies();
+  }
+};
+
+template<typename T>
+struct WrappedPipeTraits {
```
aleclearmind: 1. Do not use "Wrapped"! It carries no meaning.
   AFAIU this is offering some common features to all the *pipe run* `class`es so that they can be consumed by the template classes that turn them into an actual pipe.
2. If `T` is actually a "pipe run", add a `concept` constraint.
   Overall, I'd expect to have a `concept` constraint on *every* template argument of this file. It's quite hard to follow otherwise.

Please correct me if I'm misunderstanding something.

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
mrjackv: I was following `llvm::Expected` example with `takeError()`

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
mrjackv: To avoid writing it twice, e.g.
```c++
template<typename T>
class Foo : SingleOutputPipeBase<T, SOPTraits<T>> {
private:
  using Base = SingleOutputPipeBase<T, SOPTraits<T>>;
};
```

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
mrjackv: 1. Yes, but it's all pipe run classes except `SingleOutputPipe` where the arguments are derived from the `static void run(...)` function
2. Ok

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
mrjackv: 1. Ok
2. It's done in `SingleOutputPipeBase` with `static_assert(detail::writableContainersCount<ContainerTypes>() == 1);`, I cannot used `revng_assert` in the body because it's not constexpr.

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
mrjackv: Not having it as a inner class means that a whole bunch of things (e.g. `T::Analyses`, some `Base::` attributes) need to be explicitly passed, which would make the class more verbose.

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
aleclearmind: Makes sense.

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
aleclearmind: Mmmh, can't you forward `T` and `Base`? Then the code would be exactly the same

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
aleclearmind: A runtime check is fine too.

```c++
  bool Found = false;
  compile_time::forEach<T>([&Result, &Found]<typename A, size_t I>() {
    if constexpr (not std::is_const_v<A>) {
      revng_assert(not Found);
      Found = true;
      Result = I;
    }
  });

File: include/revng/HeadersGeneration/ModelToHeaderPipe.h (hunk 8 above)
Comment (mrjackv) [reply include/revng/HeadersGeneration/ModelToHeaderPipe.h]: > Is this a "pipe" or a "pipe run"?

It's a pipe run

> Why do we still have an array?

Because, unlike the other pipe runs, the container types in this one are deduced from the `static void run` method, so only the documentation is required.

File: include/revng/PipeboxCommon/Helpers/PipeWrappers.h (hunk 16 above)
Comment (aleclearmind) [reply include/revng/PipeboxCommon/Helpers/PipeWrappers.h]: Mmmh, I think this gives the illusion of something you can customize.
I'd ask you to assert that `Base` is actually what you want, but that would defeat the purpose.

Let's just repeat `Base`.

File: include/revng/PipeboxCommon/Helpers/PipeWrappers.h (hunk 19 above)
Comment (aleclearmind) [reply include/revng/PipeboxCommon/Helpers/PipeWrappers.h]: 1. I'm not sure I understand, I'll come back to this after there more concepts around.

File: include/revng/PipeboxCommon/Helpers/PipeRunner.h (hunk 4 above)
Comment (mrjackv) [reply include/revng/PipeboxCommon/Helpers/PipeRunner.h:36]: I've always seen the plural, e.g. the `llvm::yaml::*Traits` classes for serialization, `llvm::GraphTraits`, `KeyedObjectTraits` etc.

File: include/revng/PipeboxCommon/Concepts.h (hunk 3 above)
Comment (mrjackv) [reply include/revng/PipeboxCommon/Concepts.h:97]: IMHO `Has` makes sense, it's asserting that `T` has a using of `ContainerTypes` and that it's a specialization of `TypeList`

Comment (aleclearmind) [reply include/revng/PipeboxCommon/Concepts.h:97]: Ah "has" in the sense of "provides".
I need to take another look.

File: include/revng/PipeboxCommon/Helpers/PipeRunner.h (hunk 4 above)
Comment (aleclearmind) [reply include/revng/PipeboxCommon/Helpers/PipeRunner.h:36]: Right!

File: include/revng/PipeboxCommon/Helpers/PipeWrappers.h (hunk 13 above)
Comment (mrjackv) [reply include/revng/PipeboxCommon/Helpers/PipeWrappers.h]: I can forward things, sure, but I think it's misleading.
`LLVMFunctionPassPipe<T>::Pass` makes it very clear that it's _the pass of the pipe_
`detail::PassImpl<T, Base>` it's not so clear, because it lacks the connection to the `LLVMFunctionPassPipe` class

File: include/revng/HeadersGeneration/ModelToHeaderPipe.h (hunk 7 above)
Comment (mrjackv) [reply include/revng/HeadersGeneration/ModelToHeaderPipe.h]: I've added concept constraints to all the pipe wrappers, that will automatically assert if conditions are not met

File: include/revng/PipeboxCommon/Concepts.h (hunk 5 above)
Comment (mrjackv) [reply include/revng/PipeboxCommon/Concepts.h]: Done

File: include/revng/PipeboxCommon/Helpers/PipeWrappers.h (hunk 6 above)
Comment (mrjackv) [reply include/revng/PipeboxCommon/Helpers/PipeWrappers.h]: Fixed

File: include/revng/PipeboxCommon/Helpers/PipeWrappers.h
Comment (mrjackv) [reply include/revng/PipeboxCommon/Helpers/PipeWrappers.h:1]: Done

File: include/revng/PipeboxCommon/Helpers/PipeWrappers.h (hunk 9 above)
Comment (mrjackv) [reply include/revng/PipeboxCommon/Helpers/PipeWrappers.h]: Reworked everything to use `TypeList`

File: include/revng/PipeboxCommon/Helpers/PipeWrappers.h (hunk 11 above)
Comment (mrjackv) [reply include/revng/PipeboxCommon/Helpers/PipeWrappers.h]: Fixed

File: include/revng/PipeboxCommon/Helpers/PipeWrappers.h (hunk 12 above)
Comment (mrjackv) [reply include/revng/PipeboxCommon/Helpers/PipeWrappers.h]: Implemented

File: include/revng/PipeboxCommon/Helpers/PipeWrappers.h (hunk 14 above)
Comment (mrjackv) [reply include/revng/PipeboxCommon/Helpers/PipeWrappers.h]: It seems quite wasteful, there's already the `static_assert` that guarantees, at compile time, that there cannot be two non-const containers present, checking it again at runtime is unnecessarily wasteful.

File: include/revng/PipeboxCommon/Helpers/PipeWrappers.h (hunk 15 above)
Comment (mrjackv) [reply include/revng/PipeboxCommon/Helpers/PipeWrappers.h]: Added concepts around, let me know if more are needed

File: include/revng/PipeboxCommon/Helpers/PipeWrappers.h (hunk 16 above)
Comment (mrjackv) [reply include/revng/PipeboxCommon/Helpers/PipeWrappers.h]: Done

File: include/revng/PipeboxCommon/Helpers/PipeWrappers.h (hunk 18 above)
Comment (mrjackv) [reply include/revng/PipeboxCommon/Helpers/PipeWrappers.h]: Dropped all uses of `[]`, replaced with `.at`

File: include/revng/PipeboxCommon/Helpers/PipeWrappers.h (hunk 17 above)
Comment (mrjackv) [reply include/revng/PipeboxCommon/Helpers/PipeWrappers.h]: Done on both

File: include/revng/PipeboxCommon/Concepts.h (hunk 2 above)
Comment (mrjackv) [reply include/revng/PipeboxCommon/Concepts.h:113]: Do you mean adding doxygen around? Where should I put it?
Do note that the `PipeRun` concept here is different than the one in `PipeRunPipes`. In here it refers to the `Pipe.run(...)` method, under `PipeRunPipes` to the class of short-lived classes that instantiated when the pipe is run.

File: include/revng/PipeboxCommon/Helpers/PipeWrappers.h (hunk 14 above)
Comment (aleclearmind) [reply include/revng/PipeboxCommon/Helpers/PipeWrappers.h]: > there's already the static_assert

The other check is not here.

> checking it again at runtime is unnecessarily wasteful

It won't have any run-time cost, the generated code will be:

```c++
bool Found = false;
if (not not false)
  revng_abort(...);
Result = I;
Found = true;
```

Which will be optimzed to just `Result = I;`.

In case there's multiple matches, it will crash instead and performance there don't matter.

include/revng/PipeboxCommon/Helpers/PipeWrappers/SingleOutputPipe.h
```diff
@@ -0,0 +1,59 @@
...
+
+#include "revng/PipeboxCommon/Concepts.h"
+#include "revng/PipeboxCommon/Helpers/PipeWrappers/Base.h"
+#include "revng/PipeboxCommon/Helpers/PipeWrappers/Helpers.h"
+#include "revng/PipeboxCommon/Model.h"
+
+namespace detail {
+
+template<typename T, typename T2>
+struct SingleOutputPipeTrait {};
```
aleclearmind: Traits

include/revng/PipeboxCommon/Helpers/PipeWrappers.h
mrjackv: Added a `requires` to `writableContainerIndex`, as discussed

include/revng/PipeboxCommon/Helpers/PipeRunPipes/Base.h
```diff
@@ -0,0 +1,56 @@
...
+    if constexpr (not std::is_const_v<A>) {
+      Result = I;
+    }
+  });
+  return Result;
+}
+
+} // namespace detail
+
+template<typename T>
```
aleclearmind, line 39: Concept

include/revng/PipeboxCommon/Helpers/PipeRunPipes/Helpers.h
```diff
@@ -0,0 +1,213 @@
...
+  requires StrictSpecializationOf<typename T::Arguments, TypeList>;
+  requires checkArguments<typename T::Arguments>();
+};
+
+template<typename T>
+concept HasStaticRun = std::is_function_v<
+  std::remove_pointer_t<decltype(&T::run)>>;
+
+//
+// PipeRunContainerTypes
```
aleclearmind, line 178: Docs on the role of this

include/revng/PipeboxCommon/Helpers/PipeWrappers/Helpers.h
```diff
@@ -0,0 +1,213 @@
...
+
+//
+// PipeRunContainerTypes
+//
+
+template<typename T>
+struct PipeRunContainerTypesImpl {};
+
+template<HasArguments T>
+  requires(not HasStaticRun<T>)
```
aleclearmind: I'd split this into two new concepts with a more semantically meaningful name, e.g., `IsSingleObjectPipeRun` and `IsMultipleObjectsPipeRun`?

Then you can implement those as you like.

include/revng/PipeboxCommon/Helpers/PipeWrappers/Helpers.h
```diff
@@ -0,0 +1,213 @@
...
+
+template<typename T>
+using AddReference = decltype(addReference(std::declval<T>()));
+
+template<typename T>
+using ConstructorContainerArguments = AddReference<PipeRunContainerTypes<T>>;
+
+template<typename T, typename... Args>
+constexpr bool
+  checkPipeRun = ::compare<typename T::ContainerTypes, TypeList<Args...>>;
```
aleclearmind: This is used out of this file, it should probably not be in the `detail` namespace.

include/revng/PipeboxCommon/Helpers/PipeWrappers/LLVMFunctionPassPipe.h
```diff
@@ -0,0 +1,161 @@
...
+
+private:
+  static inline llvm::RegisterPass<Pass> X{ "", "", true, true };
+
+public:
+  using ArgumentsDocumentation = T::Arguments;
+
+public:
+  template<typename... Args>
+    requires detail::checkPipeRun<Base, Args...>
```
aleclearmind: You can reduce coupling by passing in `Base::ContainerTypes` instead of `Base`.
In fact, if you can replace this with `std::is_same_v<Base::ContainerTypes, TypeList<Args...>>`.

No need for helpers.

include/revng/PipeboxCommon/Helpers/PipeRunPipes/FunctionPipe.h
```diff
@@ -0,0 +1,60 @@
...
+    concat<detail::function_pipe::FixedArguments,
+           detail::ConstructorContainerArguments<T>>>();
+  { PipeRun.runOnFunction(Function) } -> std::same_as<void>;
+};
+
+template<IsFunctionPipe T>
+class FunctionPipe : public SingleOutputPipeBase<T> {
+private:
+  using Base = SingleOutputPipeBase<T>;
+  static_assert(Base::OutputContainerType::Kind == Kinds::Function);
```
aleclearmind, line 26: You don't need `Base::` here.
In fact, you can drop the using altogether (if you convert `checkPipeRun` to accept `ContainterTypes`.
Same elsewhere.

include/revng/PipeboxCommon/Helpers/PipeWrappers/FunctionPipe.h
```diff
@@ -0,0 +1,60 @@
...
+public:
+  template<typename... Args>
+    requires detail::checkPipeRun<Base, Args...>
+  revng::pypeline::ObjectDependencies
+  run(const Model &Model,
+      const revng::pypeline::Request &Incoming,
+      const revng::pypeline::Request &Outgoing,
+      llvm::StringRef Configuration,
+      Args &...Containers) {
+    detail::ObjectDependenciesHelper ODH(Model, Outgoing, this->ContainersSize);
```
aleclearmind: 1. Drop `this->`.
2. Review all the stuff in the `detail` namespace: if it's in `detail`, it should *not* be used out of the file where it's defined.

Same elsewhere.

include/revng/PipeboxCommon/Helpers/PipeRunPipes/Helpers.h
aleclearmind, line 1: This directory should not be called `PipeWrappers`. These are not "wrapping pipes". They are wrapping "pipe runners".

What about `PipeRunPipes`? It's not the greatest but still. Or maybe `PipeRunToPipe`.
`PipeRunWrappers`? I really dislike "wrappers". Wrapping into what?

include/revng/PipeboxCommon/Helpers/PipeWrappers/TypeDefinitionPipe.h
```diff
@@ -0,0 +1,63 @@
+#pragma once
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "revng/PipeboxCommon/Helpers/PipeWrappers/Base.h"
+#include "revng/PipeboxCommon/Helpers/PipeWrappers/Helpers.h"
+
+namespace detail::type_definition_pipe {
```
aleclearmind: Use `camelCase` for namespaces.

include/revng/PipeboxCommon/Helpers/PipeWrappers/Helpers.h
```diff
@@ -0,0 +1,213 @@
...
+  requires HasArguments<T>;
+};
+
+template<typename T, typename TL, size_t... I>
+concept CheckConstructorImpl = requires {
+  { T{ std::declval<std::tuple_element_t<I, TL>>()... } } -> std::same_as<T>;
+};
+
+template<typename T, SpecializationOf<TypeList> TL>
+constexpr bool checkConstructor() {
```
aleclearmind: `hasConstructor<T, int, long>()`

include/revng/PipeboxCommon/Helpers/PipeWrappers/TypeDefinitionPipe.h
```diff
@@ -0,0 +1,63 @@
...
+
+using FixedArguments = TypeList<const Model &, // Model
+                                llvm::StringRef, // StaticConfiguration
+                                llvm::StringRef>; // Configuration
```
aleclearmind: Inline in the only place where this is used.
Same elsewhere.

include/revng/PipeboxCommon/Helpers/PipeWrappers/LLVMFunctionPassPipe.h
```diff
@@ -0,0 +1,161 @@
...
+
+      const model::Binary &Binary = *Model.get().get();
+      for (const ObjectID *Object : Outgoing.at(Base::OutputContainerIndex)) {
+        auto Committer = ODH.getCommitterFor(*Object,
+                                             Base::OutputContainerIndex);
+
+        const MetaAddress &Entry = std::get<MetaAddress>(Object->key());
+        const model::Function &Function = Binary.Functions().at(Entry);
+        auto Iter = AddressToFunction.find(Function.Entry());
+        revng_assert(Iter != AddressToFunction.end());
```
aleclearmind: Just use `.at`.

include/revng/PipeboxCommon/Helpers/PipeWrappers/LLVMFunctionPassPipe.h
```diff
@@ -0,0 +1,161 @@
...
+        Instance.runOnFunction(Function, *Iter->second);
+      }
+
+      return true;
+    }
+  };
+
+private:
+  static inline llvm::RegisterPass<Pass> X{ "", "", true, true };
```
aleclearmind: Can we give it a name?

include/revng/ADT/TypeList.h
```diff
@@ -0,0 +1,49 @@
...
+         and (std::is_same_v<Args1, Args2> and ...);
+}
+
+} // namespace detail
+
+template<SpecializationOf<TypeList> T1, SpecializationOf<TypeList> T2>
+using concat = decltype(detail::concat(std::declval<T1>(), std::declval<T2>()));
+
+template<SpecializationOf<TypeList> T1, SpecializationOf<TypeList> T2>
+constexpr bool compare = detail::compare(T1{}, T2{});
```
aleclearmind: This is seems like a complex version of `std::is_same_v<T1, T2>`.

include/revng/PipeboxCommon/LLVMContainer.h
```diff
@@ -0,0 +1,80 @@
...
+  llvm::LLVMContext Context;
+  std::unique_ptr<llvm::Module> Module;
+
+public:
+  LLVMRootContainer() {
+    Module = std::make_unique<llvm::Module>("revng.module", Context);
+  }
+
+public:
+  std::set<ObjectID> objects() const { return std::set{ ObjectID() }; }
```
aleclearmind: Mmmh, is this a placeholder?

include/revng/PipeboxCommon/LLVMContainer.h
```diff
@@ -0,0 +1,80 @@
...
+
+  void deserialize(const std::map<const ObjectID *, llvm::ArrayRef<const char>>
```
aleclearmind, line 42: Wherever we can, we prefer references to pointers, unless you want them to be nullable.

include/revng/PipeboxCommon/LLVMContainer.h
```diff
@@ -0,0 +1,80 @@
...
+    for (const auto &[Key, Value] : Data) {
+      revng_assert(Key->kind() == Kind);
+
+      llvm::SmallVector<char> DecompressedData = zstdDecompress(Value);
+      llvm::MemoryBufferRef Ref{
+        { DecompressedData.data(), DecompressedData.size() }, "input"
+      };
+
+      auto NewModule = llvm::cantFail(llvm::parseBitcodeFile(Ref, Context));
+      Module = std::move(NewModule);
```
aleclearmind:
```suggestion
      Module = llvm::cantFail(llvm::parseBitcodeFile(Ref, Context));
```

include/revng/PipeboxCommon/LLVMContainer.h
```diff
@@ -0,0 +1,80 @@
...
+    llvm::raw_svector_ostream OS(Result[*Objects[0]].data());
+    ZstdCompressedOstream CompressedOS(OS, 3);
+    llvm::WriteBitcodeToFile(*Module, CompressedOS);
+    CompressedOS.flush();
+    return Result;
+  }
+
+  bool verify() const {
+    // NOLINTNEXTLINE
+    return llvm::verifyModule(*Module);
```
aleclearmind: Use `revng::forceVerify`.

include/revng/PipeboxCommon/MapContainer.h
```diff
@@ -0,0 +1,77 @@
...
+
+#include "revng/PipeboxCommon/Common.h"
+#include "revng/PipeboxCommon/ObjectID.h"
+
+namespace revng::pypeline {
+
+namespace detail {
+
+template<Kind TheKind, ConstexprString TheName>
+class MapContainer {
```
aleclearmind: This name focuses on the fact that this is a map, but all containers are map.
What's special about this is that the *value* of the map is binary data.

I'd call it `BinaryDataContainer` or `RawDataContainer` or `RawContainer`.

include/revng/PipeboxCommon/RawContainer.h
```diff
@@ -0,0 +1,77 @@
...
+public:
+  std::set<ObjectID> objects() const {
+    return std::views::keys(Map) | revng::to<std::set<ObjectID>>();
+  }
+
+  void deserialize(const std::map<const ObjectID *, llvm::ArrayRef<const char>>
+                     Data) {
+    for (const auto &[Key, Value] : Data) {
+      revng_assert(Key->kind() == Kind);
+      Map[*Key] = Buffer{ Value };
```
aleclearmind, line 33: Copying here is quite wasteful.

Leave a TODO about check if we can pass the ownership to `deserialize` so we don't copy.

Same for serialize, if the container is going to be destroyed, we could pass the ownership.

include/revng/PipeboxCommon/MapContainer.h
```diff
@@ -0,0 +1,77 @@
...
+    std::map<ObjectID, Buffer> Result;
+    for (const ObjectID *Object : Objects)
+      Result[*Object] = Map.at(*Object);
+    return Result;
+  }
+
+  bool verify() const { return true; }
+
+public:
+  bool contains(ObjectID Key) const { return Map.contains(Key); }
```
aleclearmind: Take a `const ObjectID &` everywhere.

include/revng/PipeboxCommon/Helpers/PipeRunPipes/FunctionPipe.h
mrjackv: No I cannot, `OutputContainerType` and friends are [dependent names](https://www.en.cppreference.com/w/cpp/language/dependent_name.html) so they cannot be used without `Base::` or `this->`

include/revng/PipeboxCommon/Helpers/PipeWrappers/FunctionPipe.h
mrjackv: 1. See the comment on the thread above

include/revng/PipeboxCommon/Helpers/PipeWrappers/TypeDefinitionPipe.h
mrjackv: Sure, but having the comments on the side (IMHO) helps. Otherwise at a glance it's not obvious what the types represent

include/revng/PipeboxCommon/LLVMContainer.h
mrjackv: Containers (`std::vector` and `std::map`) don't work well with references
[Godbolt](https://godbolt.org/z/cc9ohEqsa)

include/revng/PipeboxCommon/Helpers/PipeWrappers/TypeDefinitionPipe.h
aleclearmind: Yeah, but we try to avoid inline comments.

I'd add a more explicit comment where they are used. This reduces locality making it a bit harder to read.

No biggie though, up to you.

include/revng/PipeboxCommon/LLVMContainer.h
aleclearmind: Right.

include/revng/PipeboxCommon/Helpers/PipeWrappers/Helpers.h
mrjackv: Will rename to `hasConstructor`
However using a parameter pack makes things unnecessarily difficult, `ContainerTypes` is already a TypeList, using `concat` is quite easier than going through the trouble of unpacking it

aleclearmind: Ah OK, my main suggestion was renaming. I'm fine with using `TypeList`.

include/revng/PipeboxCommon/Helpers/PipeWrappers/LLVMFunctionPassPipe.h
mrjackv: Are you referring to the fact that the variable is named `X` or to the second argument of the `RegisterPass` constructor?

include/revng/PipeboxCommon/Helpers/PipeWrappers/SingleOutputPipe.h
mrjackv: Added the `s`

include/revng/PipeboxCommon/Helpers/PipeRunPipes/Base.h
mrjackv: Added

include/revng/PipeboxCommon/Helpers/PipeWrappers/Helpers.h
mrjackv: Added the two concepts that you proposed

include/revng/PipeboxCommon/Helpers/PipeWrappers/Helpers.h
mrjackv: This function in particular was dropped

include/revng/PipeboxCommon/Helpers/PipeWrappers/LLVMFunctionPassPipe.h
mrjackv: Reworked to use `std::is_same_v`

include/revng/PipeboxCommon/Helpers/PipeWrappers/FunctionPipe.h
mrjackv: 2. all the concepts that are defined in `Helpers.h` and used elsewhere are not in the `detail` namespace

include/revng/PipeboxCommon/Helpers/PipeRunPipes/Helpers.h
mrjackv: Renamed to `PipeRunPipes`

include/revng/PipeboxCommon/Helpers/PipeWrappers/TypeDefinitionPipe.h
mrjackv: These namespaces have been dropped

include/revng/PipeboxCommon/Helpers/PipeWrappers/Helpers.h
mrjackv: Renamed

include/revng/PipeboxCommon/Helpers/PipeWrappers/TypeDefinitionPipe.h
mrjackv: I've inlined the `TypeList` and added a comment that explains what the types represent as arguments

include/revng/PipeboxCommon/Helpers/PipeWrappers/LLVMFunctionPassPipe.h
mrjackv: Done

include/revng/ADT/TypeList.h
mrjackv: Yes, I must have had a brain fart when I did that, I've dropped it

include/revng/PipeboxCommon/LLVMContainer.h
mrjackv: Yes, I've forgot to actually implement it properly, it now uses `Module.empty()`

include/revng/PipeboxCommon/LLVMContainer.h
mrjackv: Done

include/revng/PipeboxCommon/LLVMContainer.h
mrjackv: Done, although `abort`-ing instead of returning a bool doesn't make me the happiest

include/revng/PipeboxCommon/MapContainer.h
mrjackv: Renamed to `RawContainer`

include/revng/PipeboxCommon/RawContainer.h
mrjackv: Added the `TODO`

include/revng/PipeboxCommon/MapContainer.h
mrjackv: Done

include/revng/PipeboxCommon/Helpers/PipeWrappers/LLVMFunctionPassPipe.h
aleclearmind: `X` is fine, I was referring to the pass.

include/revng/PipeboxCommon/LLVMContainer.h
aleclearmind: I understand, but then you should return an `llvm::Error` wrapping all the messages of the second argument of `verifyModule`.

Right now we're suppressing them (very very bad!).

include/revng/PipeboxCommon/Helpers/PipeRunPipes/Helpers.h
mrjackv: Done

include/revng/PipeboxCommon/Helpers/PipeWrappers/LLVMFunctionPassPipe.h
mrjackv: Done

aleclearmind:
Merged in 6d1fbdc. Thanks.

PR #523, opened 2025-09-22
Fix DAGify head election

andrealinux1:
In this PR, we introduce an additional criterion for election the `GenericRegion`s when multiple `Region`s are nested and share a candidate entry.

In addition, we introduce some failsafe checks for:
- Ensuring that the `isDAG` check also looks at disconnected part of the `ScopeGraph`.
- An additional `hasUnreachableBlocks` function, which checks that the every block is reachable from the entry, in the `ScopeGraph`.


lib/RestructureCFG/DAGifyPass.cpp
```diff
@@ -185,10 +185,13 @@ class DAGifyPassImpl {
...
 
     // Verify that the output `ScopeGraph` is acyclic, after `DAGify` has
-    // processed the input, but only when the `VerifyLog` is enabled
+    // processed the input, but only when the `VerifyLog` is enabled.
+    // In addition, we also verify that the `ScopeGraph` has not blocks
+    // disconnected from the entry block.
     if (VerifyLog.isEnabled()) {
       Scope<Function *> ScopeGraph(&F);
       revng_assert(isDAG(ScopeGraph));
+      revng_assert(not hasUnreachableBlocks(&F));
```
pfez, line 176: Test this first, `isDAG` second.

lib/RestructureCFG/ScopeGraphUtils.cpp
```diff
@@ -367,3 +369,25 @@ void simplifyTerminator(BasicBlock *BB, const BasicBlock *PlaceHolderTarget) {
...
+  // that we did not see during the DFS visit, it means that we have a block
+  // disconnected from the entry
+  for (BasicBlock *BB : nodes(ScopeGraph)) {
+    if (not DFSNodes.contains(BB)) {
+      return true;
+    }
+  }
+
+  return false;
+}
```
pfez: Can't this go in `GraphAlgorithms.h`?

include/revng/Support/GraphAlgorithms.h
```diff
@@ -396,15 +397,3 @@ nodesBetweenReverse(GraphT Source, GraphT Destination) {
...
-bool isDAG(GraphT Graph) {
-  for (llvm::scc_iterator<GraphT> I = llvm::scc_begin(Graph),
-                                  IE = llvm::scc_end(Graph);
-       I != IE;
-       ++I) {
-    if (I.hasCycle())
-      return false;
-  }
-  return true;
-}
```
pfez, line 404: Wait why are you moving this away from `GraphAlgorithms.h`?

include/revng/Support/GraphAlgorithms.h
```diff
@@ -243,21 +243,27 @@ template<class GraphT, class GT = llvm::GraphTraits<GraphT>>
 llvm::SmallSetVector<revng::detail::EdgeDescriptor<typename GT::NodeRef>, 4>
 getBackedges(GraphT Block) {
   llvm::SmallPtrSet<typename GT::NodeRef, 4> EmptySet;
-  return getBackedgesImpl<revng::detail::FilterSet::BlackList>(Block, EmptySet);
+  return getBackedgesImpl<revng::detail::FilterSet::BlackList,
+                          GraphT,
+                          GT>(Block, EmptySet);
```
pfez, line 249: Oh nice catch. So we were computing the backedges on the wrong graph because of the missing template arguments?

andrealinux1: Yes, this did not trigger a visible bug yet, but I saw it in an intermediate state where I was trying to replace the use of the `scc_iterator` with this `getBackedges` helper in order to check for loops in the `ScopeGraph`.

In the end I abandoned that solution, but still since this is in `GraphAlgorithms.h` I think its worth fixing rn.

lib/RestructureCFG/DAGifyPass.cpp
andrealinux1: Right, these checks are now performed in inverse order https://github.com/revng/revng/blob/940b7ef05ac031f0066c6046740a604aa5112c2f/lib/RestructureCFG/DAGifyPass.cpp#L192

lib/RestructureCFG/ScopeGraphUtils.cpp
andrealinux1: Yes, I moved it in `GraphAlgorithms.h` https://github.com/revng/revng/blob/940b7ef05ac031f0066c6046740a604aa5112c2f/include/revng/Support/GraphAlgorithms.h#L443

include/revng/Support/GraphAlgorithms.h
andrealinux1: Because I didn't think hard enough, and I though that a change we introduced in the new formulation of `isDAG`, in order to consider potential loops disconnected from the entry, made this not possible without specializing it directly on a `ScopeGraph`.

This is now templated, I'll open a comment with a couple of observations on why I made the choices on the templated arguments.

include/revng/Support/GraphAlgorithms.h
```diff
@@ -391,14 +398,66 @@ nodesBetweenReverse(GraphT Source, GraphT Destination) {
   return nodesBetweenImpl<GraphT, Inverse<GraphT>>(Source, Destination);
 }
 
-template<class GraphT>
+/// Helper function which checks that the `ScopeGraph` is a
+template<class GraphT, class NodeT>
```
andrealinux1, line 402: So, I'll try to give you a brief summary of the changes introduced here.

We now want to check for loops also in portions of the `Graph` disconnected from the entry block.

To do this, I basically instantiate multiple `scc_iterator`s, in the worse case one for each node in the graph (all disconnected node, and we have a loop in the last node we visit). This is costly of course, but this is a failsafe check that we use in `debug-log=verify`, so I think it is accetable.

However, the subtle problem, that made me add another template argument, the `NodeT` one, is due to the following:
- We accept an argument of type `GraphT` (in our sole use case at the moment, this is `Scope<Function *>`, a `ScopeGraph` object).
- However, we want to instantiate multiple `scc_iterator<>` on the node objects, in our case `Scope<BasicBlock *>`. And we cannot automatically obtain that object from the `GraphTraits`, because `NodeRef`, is actually just `BasicBlock *` (and we use it, for example, as a template argument for the `SmallSet` used to store the nodes that we still need to visit).
- Therefore, I added the additional `NodeT` template argument, so we can specify it and use it in our `scc_iterator`.
- I decided not to default the argument, specifically to `class NodeT = llvm::GraphTraits<GraphTraits>::NodeRef`, because I think it would be too bug prone. Indeed, in a first iteration of this, I didn't think carefully, and I omitted the `NodeT` template argument, deducing `NodeRef` in the body and instantiating the `scc_iterator<NodeRef>`, and this lead to a subtle bug where the `scc`s where computed not on the `ScopeGraph`, but on the plain CFG, because it was run on `BasicBlock *` and not on `Scope<BasicBlock *>`.
- For the reason above, I think it is better to make the user specify both the template argument, so he can take care of verifying that he is doing the correct^TM thing.

include/revng/Support/GraphAlgorithms.h
```diff
@@ -398,15 +398,42 @@ nodesBetweenReverse(GraphT Source, GraphT Destination) {
...
-      return false;
+
+  using NodeRef = llvm::GraphTraits<GraphT>::NodeRef;
+
+  // We start by collecting all the nodes in the graph. We then remove from this
+  // set, at each iteration, each node composing the `scc` that we find. We do
+  // this until no node remains in the set, meaning that we visited all the
+  // nodes in the graph, also eventual nodes, disconnected from the entry, which
+  // may form a loop.
+  llvm::SmallSet<NodeRef, 4> RemainingNodes;
```
pfez: The graph is not being mutated.
Just use a set of `VisiteNodes`. 
Then instead of `while (not RemainingNodes.empty())` just iterate on `nodes(Graph)` and to the SCC visit inside, always adding `I*` to `VistedNodes`.
In this way you only do a single visit of the graph instead of 2.

andrealinux1: Using the additional `RemainingNodes` was not to avoid invalidation due to graph mutation, but was to reduce the number of `scc` instantiation in the inside loop body.

The idea was that, assuming that a lot of the nodes are inside a `scc` which is visited "early" during the outer iteration, we highly reduce the number of `scc` instantiation due to the fact that we run out "fast" of the nodes in `RemainingNodes`.

However this is a debug/failsafe function, so I agree we can simplify it just iterating on the `nodes(Graph)` as you suggested.

However, if we just iterate on the `nodes(Graph)`, I'm not entirely sure why you also want to introduce a `VisitedNodes` set? You want to use it to _skip_ over nodes during the main iteration?

pfez: See here https://github.com/revng/revng/blob/develop/lib/DataLayoutAnalysis/Middleend/CollapseSCC.cpp#L42

andrealinux1: Yeah, so we want to use the `VisitedNodes` as a skiplist, gotcha, integrated the changes here https://github.com/revng/revng/blob/855a9309ec0a5bda6f3236bd98f3000a6d7cb36d/include/revng/Support/GraphAlgorithms.h#L412

PR #528, opened 2025-09-26
Simplify TTG schema format and handling

Cvelth:


scripts/tuple_tree_generator/templates/tuple_tree_gen.py.tpl
```diff
@@ -53,8 +52,16 @@ class DiffYamlDumper(YamlDumper):
     pass
 
 
-def random_id():
-    return random.randint(2 ** 10 + 1, 2 ** 64 - 1)
+def next_id():
+    if not hasattr(next_id, 'next_index'):
+        next_id.next_index = 0
```
mrjackv: Stuffing the index inside the function object is very awkward in both python and typescript, IMHO either go with a global or make a simple class and declare a global instance

scripts/tuple_tree_generator/schema/struct.py
```diff
@@ -13,6 +13,8 @@
 from .upcastable import UpcastableDefinition
 
 # TODO: we should return the fields in a stable order (in all_fields, all_optional_fields, etc)
+# **ivan**: are they not already stable? :thinking-face:
+# The `fields` is always a list, and don't have multiple inheritance, what am I missing?
```
aleclearmind: I tend to agree.

scripts/tuple_tree_generator/templates/struct_impl.cpp.tpl
```diff
@@ -219,6 +219,7 @@ void revng::Tracking::stop(const /*= base_namespace =*/::/*= struct.name =*/ &LH
 
 /** endif **/
 
+// **ivan**: the following never generates anything. What was it supposed to check?
 /**- for field in struct.fields **/
 /**- if struct | is_struct_field **/
 static_assert(not (TupleTreeCompatible</*= field | field_type =*/> and KeyedObjectContainerCompatible</*= field | field_type =*/>));
```
aleclearmind: I think it should have been `if field | is_struct_field`...?

scripts/tuple_tree_generator/schema/schema.py
```diff
@@ -56,13 +56,15 @@ def get_definition_for(self, type_name):
 
     def struct_definitions(self) -> List[StructDefinition]:
         toposorter: TopologicalSorter = TopologicalSorter()
-        for struct in self.definitions.values():
+        for struct in sorted(self.definitions.values(), key=lambda d: d.name or ""):
```
aleclearmind: Here you do `or ""` but assert on it below. Drop `or ""`. Same below.

scripts/tuple_tree_generator/schema/struct.py
```diff
@@ -113,21 +97,23 @@ def __init__(
...
         self.resolved_element_type = None
 
     def resolve_references(self, schema):
         self.resolved_element_type = schema.get_definition_for(self.element_type)
-        if self.upcastable:
+        if (
+            isinstance(self.resolved_element_type, StructDefinition)
+            and self.resolved_element_type.abstract
+        ):
             self.resolved_element_type = UpcastableDefinition(self.resolved_element_type)
```
aleclearmind, line 105:
```python
        if self.resolved_element_type.abstract:
            assert isinstance(self.resolved_element_type, StructDefinition)
            self.resolved_element_type = UpcastableDefinition(self.resolved_element_type)
```

Same above.

scripts/tuple_tree_generator/schema/struct.py
```diff
@@ -66,23 +66,6 @@ def resolve_references(self, schema):
         raise NotImplementedError()
 
 
-class UpcastableStructField(StructField):
```
aleclearmind, line 67: `:O`

scripts/tuple_tree_generator/generators/typescript.py
```diff
@@ -293,7 +293,7 @@ def is_optional(field: StructField):
         return field.optional or isinstance(field.resolved_type, ReferenceDefinition)
 
     @staticmethod
-    def is_upcastable(field: StructField):
+    def is_directly_upcastable(field: StructField):
```
aleclearmind: What do you mean by "directly"? :thinking:

tests/unit/TupleTreeGenerator/schema.yml
```diff
@@ -18,9 +18,7 @@ definitions:
           elementType: uint64_t
           type: SortedVector
       - name: ReferenceField
-        reference:
-          pointeeType: uint64_t
-          rootType: TestClass
+        reference-to: uint64_t
```
aleclearmind: We use `camelCase`

scripts/tuple_tree_generator/templates/tuple_tree_gen.py.tpl
aleclearmind: What about

```python

def next_id()
    counter = 0
    def impl():
        counter = counter + 1
        return counter - 1
    return impl

# ...

= field(default_factory=next_id())
```

include/revng/Model/model-schema.yml
```diff
@@ -134,7 +134,7 @@ definitions:
         doc: |-
           A unique numeric identifier for this type.
         type: uint64_t
-        is_guid: true
+        is_global_index: true
```
aleclearmind: To make this more obvious, maybe we can borrow terminology from DBs: `autoincrement`.

scripts/tuple_tree_generator/templates/enum.h.tpl
```diff
@@ -58,6 +58,11 @@ inline Values fromName(llvm::StringRef Name) {
     return Invalid;
   }
 }
+
+inline bool verify(/*= base_namespace =*/::/*= enum.name =*/::Values Value) {
```
aleclearmind: Mmmmh. Why do we need this?

I think we could have:

```c++
bool isValid(/*= base_namespace =*/::/*= enum.name =*/::Values Value) {
  revng_assert(Value != Values::Invalid);
  return Value != Values::Invalid;
}
```

aleclearmind:
NOTETOSELF: resume reviewing from `Model: make TTG verification more strict` (included).

scripts/tuple_tree_generator/templates/tuple_tree_gen.py.tpl
mrjackv: That also works, @Cvelth your pick

Cvelth: I prefer global approach, mostly because at some point we'll definitely want to do the thing we do C++-side of adding values to this `counter` if given binary already has types with IDs greater than it (currently, adding new types from python/typescript to an existing binary is probably pretty sad). It's one thing to abuse function-object / global object. And a totally different beast to reach inside a factory!

scripts/tuple_tree_generator/templates/struct_impl.cpp.tpl
Cvelth: Right, obvious in hind-sight. My bad on the noise.

scripts/tuple_tree_generator/schema/schema.py
Cvelth: We get a type-checking error if lambda passed to the `key` returns `str | None`. `or ""` is the easiest way to force it to always return struct.

And, as you mentioned, we assert on the empty string below anyway.

scripts/tuple_tree_generator/schema/struct.py
Cvelth: No, the order is important, as only `StructDefinition` has `abstract`.

scripts/tuple_tree_generator/schema/struct.py
Cvelth: Yeah, this one is my bad.

One of the previous times I touched this - this part slipped through into the final iteration even though it wasn't used anymore :D

scripts/tuple_tree_generator/generators/typescript.py
Cvelth: No strong reason actually, I just needed a distinct name at a time to avoid collision with something else 🤔

We can rename it back now, I think. Or, maybe we can come up with a better name?

---------------

That aside, "`MyField` is directly upcastable" as in "you can directly call `MyField.upcast()`".

tests/unit/TupleTreeGenerator/schema.yml
Cvelth: Actually, this confuses me a bit. Didn't we decide to use `kebab-case` everywhere it's supported?

(no big deal either way, this string in highly internal, so who cares)

scripts/tuple_tree_generator/templates/tuple_tree_gen.py.tpl
aleclearmind: Can't we do what C++ is doing right away?

scripts/tuple_tree_generator/schema/schema.py
aleclearmind: You could have

```python
from typing import TypeVar

T = TypeVar('T')

def notNone(x: T | None) -> T:
    assert x is not None
    return x
````

File: scripts/tuple_tree_generator/generators/typescript.py (hunk 7 above)
Comment (aleclearmind) [reply scripts/tuple_tree_generator/generators/typescript.py]: If there's not reason, I'd keep the old name.

I'm not sure what a not *directly* upcastable type would look like.

File: tests/unit/TupleTreeGenerator/schema.yml (hunk 8 above)
Comment (aleclearmind) [reply tests/unit/TupleTreeGenerator/schema.yml]: Yeah, but either all or nothing. I'd stick with camelCase for now, you can add a `TODO` somewhere about switching to `kebab-case`.

File: scripts/tuple_tree_generator/templates/enum.h.tpl (hunk 10 above)
Comment (Cvelth) [reply scripts/tuple_tree_generator/templates/enum.h.tpl]: > Why do we need this?

Consider that no non-key enum can now fail deserialization when missing or set to invalid.
Even a model like
```yaml
Architecture: Invalid
```
will deserialize successfully with the `Architecture` field set to `Invalid`.

This will still fail thanks to LLVM, but that's beside the point:
```yaml
Architecture: 42
```

And, putting `model::Architecture::verify(MyArchitecture)` (or `isValid` as you are suggesting, no strong feelings on my end) feels a lot cleaner than `MyArchitecture == model::Architecture::Invalid` (which is leaking the implementation details of it being an enum)!

Bonus point, when we finally switch to `class`-like enums we discussed at some point in the past (the ones that force you to write `model::Architecture::AArch64` while allowing fields), this can just become `MyArchitecture.isValid()` - win-win :tada:

(speaking about that, should I add that to my cold-backlog for the next time I happen to be in-between tasks?)

---------------------

That being said, `<= Count` part is just me being paranoid. There's no way that can come from _our_ deserializer, but we do plan allowing user-written pipes, don't we? So there's no such thing as too much safety.

I can probably be an assert (I'm assuming that's what you meant, `Value != Values::Invalid` is just a mis-copy-paste), but to me this _kind of_ feels like a piece of user input, so...

---------------------

P. S. Looking into this made me realize that we have a weird enum deserializer implementation (probably since before TTG because what it is now). I did a quick attempt at reducing the complexity there, but it doesn't seem trivial to simplify. I feel like that's one more of the areas TTG can greatly improve the amount of spaghetti we need. Should I add that to the cold backlog too?

By "it", I mean to generate all the traits directly using TTG (instead of it pointing to wrappers). We can probably also reduce the amount of custom `IO` objects everywhere to do something TTG _already_ generates (think `getNameFromYAMLEnumScalar` and friends)!

scripts/tuple_tree_generator/templates/tuple_tree_gen.py.tpl
Cvelth: We definitely should. The thing is, that means new type creation must depend on the binary.

It was easy for me to do that c++-side because I'm familiar with that part of the codebase.

Passing binary up in python/typescript without clearly understanding which part does what sounds like asking for dependency cycles to appear.

scripts/tuple_tree_generator/schema/schema.py
Cvelth: I could, but that's way more complexity :P

Actually, now that I think about it, why can _name_ be `None` again? I can't see any benefit in omitting it when specifying the schema.

For context, we're talking about this name:
```yaml
  - name: Binary               # <! this one!
    type: struct

    doc: <omitted>

    fields:
      - name: Architecture
        doc: The architecture for this binary.
        type: Architecture
```

Maybe we should instead just force it to be a full on string?

scripts/tuple_tree_generator/generators/typescript.py
Cvelth: Sounds good! My bad on letting this slip through.

tests/unit/TupleTreeGenerator/schema.yml
Cvelth: Looks like it's only `elementType` and this new one 😲.

Do you see a reason not to do it right away?

--------------------

P. S. My first instinct was to look at the `metaschema` to figure out how much we'd need to rename and it looks like it's outdated actually! Why do we need it again if it's not used? (see the new `(DANGER)` commit).

scripts/tuple_tree_generator/templates/enum.h.tpl
aleclearmind: > putting model::Architecture::verify(MyArchitecture)...

"Putting" where? In the verification routines?

That part is quite aware of the fact that it's dealing with an `enum`.
If you want an helper called `isValid`, works, but `verify` doesn't sound right, there are situations where `Invalid` passes verification (e.g., `DefaultABI` can be `Invalid`).

> kind of feels like a piece of user input

There (should be) no way for the user to do that, let's `assert`. If there is, it's a bug, not erroneous user input.

> Should I add that to the cold backlog too?

Not entirely clear what the problem is here. You can backlog this as super-low-prio. This is just a bridge task, let's not pull in more stuff, we have more important stuff to work on, lots of it.

scripts/tuple_tree_generator/templates/tuple_tree_gen.py.tpl
aleclearmind: I'm a bit confused. I feel like we should either do nothing or do what C++ does.

If starting from zero is not right (unless it's a new model), we should not do that. It gives the false hope of automatically managing this.

I'm also fine with doing nothing, if this is too complex, but doing something correct some of the times feels like a bad idea.

scripts/tuple_tree_generator/schema/schema.py
aleclearmind: > I could, but that's way more complexity :P

Feels like a function we could reuse elsewhere.

> Maybe we should instead just force it to be a full on string?

:+1:

tests/unit/TupleTreeGenerator/schema.yml
aleclearmind: > Looks like it's only elementType and this new one 😲.

"it's" what? `rootType` is camel case, `pointeeType` is camel case. `reference-to` should be as well.

> Do you see a reason not to do it right away?

Do what specifically? Leave a TODO, this PR is already big enough, we have to move on.

> P. S. My first instinct was to look at the metaschema to figure out how much we'd need to rename and it looks like it's outdated actually! Why do we need it again if it's not used? (see the new (DANGER) commit).

I've no idea.

scripts/tuple_tree_generator/templates/enum.h.tpl
Cvelth: > "Putting" where? In the verification routines?

Mostly, yes. But also random sanity asserts (I didn't scout the codebase for places where we already do `!= Invalid`, but I plan to use this for all the future sanity asserts :) )

> If you want an helper called `isValid`, works

:tada:

> There (should be) no way for the user to do that

```cpp
Binary.Architecture() = model::Architecture::Values(-1);
```
?

It's not unlikely if we include "pipe-writer" among "user"s.

> Not entirely clear what the problem is here.

Mostly me getting sad every time I have to touch `YAMLTraits.h`. But yeah, I agree that this is not important in the grand scheme of things.

tests/unit/TupleTreeGenerator/schema.yml
Cvelth: > `rootType` is camel case, `pointeeType` is camel case.

 `rootType` and `pointeeType` are gone now! `reference-to` is the one that replaces them!
 
With those gone, we have exactly two camel-case identifiers in the TTG spec: `elementType` (for sequence fields) and `referenceTo`.

> Do what specifically?

Replace all the mentions of `elementType` with `element-type` in TTG. From my point of view, it's the exact same amount of effort to go `elementType -> element-type` than it is to go `reference-to -> referenceTo`. Might as well do the "correct" one.

> I've no idea.

So should we drop the metaschema? Add an angry sounding TODO-comment? Leave it be and pretend it correct?

scripts/tuple_tree_generator/schema/schema.py
Cvelth: > Feels like a function we could reuse elsewhere.

I don't know, I didn't look if we have one already :P (maybe I should have, but that's beside the point) So it's likely if I (or someone else) runs into the same situation again, the instinct would be to just add it anew. Which will lead to more clutter.

That being said, it's hopefully not needed in this case, `name` should always be set - so it's an empty discussion anyway.

scripts/tuple_tree_generator/templates/enum.h.tpl
aleclearmind: > It's not unlikely if we include "pipe-writer" among "user"s.

Pipe writers are not users, they are devs and deserve asserts.
If it can be done from Python instead...

Cvelth: Yes, and, unless I'm mistaken, the only way to go from python model to c++ model is through serialization-deserialization cycle, right? In which case it is indeed impossible.

tests/unit/TupleTreeGenerator/schema.yml
aleclearmind: It feels a bit wrong to me, I'd stick with with `camelCase` for similarity with JSON Schema's convention.

scripts/tuple_tree_generator/templates/enum.h.tpl
aleclearmind: Yes, correct.

lib/ABI/Definition.cpp
```diff
@@ -44,7 +44,10 @@ static bool isVectorRegister(model::Register::Values Register) {
 ///
 /// \return `true` if the ABI is valid, `false` otherwise.
 static bool verifyReturnValueLocation(const abi::Definition &D) {
-  if (D.ReturnValueLocationRegister() == model::Register::Invalid) {
+  if (not model::Register::verify(D.ReturnValueLocationRegister())) {
+    if (D.ReturnValueLocationRegister() != model::Register::Invalid)
+      return false; // The register value is severely broken (>= Count).
```
aleclearmind: This is confusing. Assert on `>= Count` within verify and drop the second `if`.

share/revng/test/tests/analysis/CollectCFG/x86_64/try-catch-ehframe.S.cfg.yml
```diff
@@ -25,5 +25,5 @@
     - End: "6"
       ID: "5"
       Successors:
-        - "-Destination": ""
```
aleclearmind, line 28: Maybe we now can drop the feature of `revng model compare` to handle the `-` prefix?

scripts/tuple_tree_generator/generators/cpp.py
```diff
@@ -22,9 +22,11 @@ def __init__(
         schema: Schema,
         base_namespace: str,
         emit_tracking: bool,
+        string_types=[],
         user_include_path: Optional[str] = None,
     ):
         self.schema = schema
+        self.string_types = string_types
         self.emit_tracking = emit_tracking
```
aleclearmind, line 30: These changes are hard to review, I don't see any usage of this.
The commit does not specify the reason for this change nor there is docs.

As I mention other times, these changes would be better suited to be in the same commit of their first user.

Anyway, from what I see, this seems to assume there's a C++ method `fromString`.
What if we ditch all of this and just assume there's a trait in C++ you *have to* implement?
We could use this to handle `std::string` (which would return itself), `MetaAddress` and `enums`.

I'd be very unhappy if we had to add yet another argument to the already obscure CMake machinery. Let's strive to prune complexity.

include/revng/Model/model-schema.yml
```diff
@@ -1446,10 +1453,9 @@ definitions:
         doc: |-
           Sets a recommended comment line width to improve their readability.
 
-          The default value is `80`.
-
           Set to `-1` for unlimited line size.
         type: uint64_t
+        default: 80
```
aleclearmind: Aaaah *massages his own nipples*

include/revng/Model/model-schema.yml
```diff
@@ -189,6 +190,7 @@ definitions:
       - name: Index
         doc: The argument index.
         type: uint64_t
+        default: -1
```
aleclearmind, line 193: Worth it? What's you reasoning?

include/revng/EarlyFunctionAnalysis/efa-schema.yml
```diff
@@ -22,6 +22,7 @@ definitions:
       - name: Name
         type: string
         doc: Optional name for debugging purposes
+        default: ""
```
aleclearmind, line 25: Can we omit this?

include/revng/EarlyFunctionAnalysis/efa-schema.yml
```diff
@@ -37,14 +38,17 @@ definitions:
     fields:
       - name: ID
         type: BasicBlockID
+        default: ":Invalid"
```
aleclearmind, line 41: Can we omit this?
What happens if we do?

include/revng/Model/NamingConfiguration.h
```diff
@@ -10,138 +10,6 @@ class model::NamingConfiguration
...
-    if (UnnamedFunctionPrefix().empty())
-      return "function_";
-    else
-      return UnnamedFunctionPrefix();
-  }
-
-  llvm::StringRef unnamedTypeDefinitionPrefix() const {
-    if (UnnamedTypeDefinitionPrefix().empty())
-      return "";
-    else
```
aleclearmind, line 40: Aaaaaaaaaaaaaaaaaaaaaaaaaaaaaah *intensely massages his own nipples*

share/doc/revng/user-manual/tutorial/running-analyses.md
```diff
@@ -25,13 +25,15 @@ $ grep -A7 'Segments:' example.yml
...
-    StartOffset:     0
     FileSize:        2528
     IsReadable:      true
-    IsWriteable:     false
     IsExecutable:    true
+    Relocations:
+      - Address:         "0x401d88:Generic64"
 ```
 
+**ivan**: no clue where `Relocations` came from. Do you have an idea? Or should I go looking for it blindly?
```
aleclearmind: I don't understand your question?

I guess the input binary has a relocation that was not popping up before, for some reason?

python/revng/internal/cli/_commands/generate_migrations/metaschema.yml
```diff
@@ -98,26 +98,30 @@ definitions:
...
-      const:
-        type: boolean
-        default: false
-    required:
-      - name
-      - type
-      - doc
+  # DANGER!!!!!
+  # This commit doesn't break anything, how was metaschema supposed to be used?!!
+  # DANGER!!!!!
```
aleclearmind: @mrjackv: @Cvelth says that commenting random stuff in here has no effect.
Can you comment on this?

lib/ABI/Definition.cpp
Cvelth: Yep, that was what we decided already in the other thread!

share/revng/test/tests/analysis/CollectCFG/x86_64/try-catch-ehframe.S.cfg.yml
Cvelth: Should we? Do you really think we will never find use for it again? Dropping is easy - re-introducing, not so much.

scripts/tuple_tree_generator/generators/cpp.py
Cvelth: For some context, both python and typescript "backends" already had this in place. It was only c++ one that was lagging behind.

> `fromString` -> trait

Sure, no complains on that, BUT what do we do about other backends? They have similar assumptions in place. Do we just let them be for now?

include/revng/Model/model-schema.yml
Cvelth: > What's you reasoning?

We discussed this heavily during the PR dropping random IDs. We decided to start ordering from 0 instead of 1 because that is more consistent with most _reasonable_ programming languages :D

But the problem then is that we have to way to differentiate between a valid newly constructed type that is yet to get an ID and the very first type in the binary.

This way we can assert on this `-1` being then when "inserting" a type into the binary to make sure each type is only inserted once, as the insertion is what assign the _real_ index.

Cvelth: But that whole thing was postponed until the `default: -1` is possible. Since it's a whole lot messier otherwise!

include/revng/EarlyFunctionAnalysis/efa-schema.yml
Cvelth: Of course. That's just me being extra explicit.

include/revng/EarlyFunctionAnalysis/efa-schema.yml
Cvelth: It will call `fromString` for `BasicBlockID` with `""` as the input. Which, I assume, will lead to the exact same result.

That being said, this is cleaner!

Cvelth: Ah, wait! No! That's what would happen c++-side. For python and typescript, `BasicBlockID` is just a string - so it will stay as is (`""`).

share/doc/revng/user-manual/tutorial/running-analyses.md
Cvelth: Pretty much yeah. What I'm asking is 

```
why now?! I don't feel like i touched anything that would have introduced a seemingly random relocation
```

share/revng/test/tests/analysis/CollectCFG/x86_64/try-catch-ehframe.S.cfg.yml
aleclearmind: Unused features are bad.

This was introduced to tackle the problem of optional fields, if we no longer have them, we don't need it.

Also, we have the `git` history.

scripts/tuple_tree_generator/generators/cpp.py
aleclearmind: Uff, if we have it elsewhere OK...
Leave a TODO about the trait I guess.

include/revng/Model/model-schema.yml
aleclearmind: OK, but what's the practical implication of leaving `default: 0`? :thinking:

include/revng/EarlyFunctionAnalysis/efa-schema.yml
aleclearmind: Is it just a string? We have `MetaAddress` in Python/TypeScript, don't we have `BasicBlockID`?

share/doc/revng/user-manual/tutorial/running-analyses.md
aleclearmind: I do not know the answer, let's see what the CI thinks `:P`

share/revng/test/tests/analysis/CollectCFG/x86_64/try-catch-ehframe.S.cfg.yml
Cvelth: Ok, sure, we can try.

That being said, quick search (`"-\w+":`) shows that we still have this pattern in the repo, so I'd need to take a deeper look whether those are necessary or just there for extra explicitness.

include/revng/Model/model-schema.yml
Cvelth: None. Except for the manual assert checking for this exact thing:
[`95ab93a` (#528)](https://github.com/revng/revng/pull/528/commits/95ab93a170e6a1897ea0447cb30245eebd7320b1#diff-9fc9a9c46221bd88ccba85350b3e949fd3d4e71c11c862b1bfa17d3e342c3997R87)

include/revng/EarlyFunctionAnalysis/efa-schema.yml
Cvelth: I didn't look to deeply, but inside _TTG_ they are all threated like strings.

They might have more explicit representation later, when running _generated_ code. But this part focuses on generation-time. And at generation-time those are strings.

share/revng/test/tests/analysis/CollectCFG/x86_64/try-catch-ehframe.S.cfg.yml
aleclearmind: Yeah, I see

```
$ grep '"-' $(find -name "*.model.yml" )
./share/revng/test/tests/analysis/DetectABI/x86_64/call-site-register-used-on-one-path.S.model.yml:    "-Arguments": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/callee-saved-register-in-callee.S.model.yml:    "-Arguments": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/callee-saved-register-in-callee.S.model.yml:    "-ReturnValues": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/callee-saved-register.S.model.yml:    "-Arguments": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/callee-saved-register.S.model.yml:    "-ReturnValues": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/clobbered-register-in-callee.S.model.yml:    "-PreservedRegisters": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/clobbered-register-in-callee.S.model.yml:    "-Arguments": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/clobbered-register-in-callee.S.model.yml:    "-ReturnValues": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/dead-register.S.model.yml:    "-Arguments": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/dead-register.S.model.yml:    "-Arguments": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/draof.S.model.yml:    "-Arguments": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/raofc.S.model.yml:    "-Arguments": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/register-clobbered-on-all-paths.S.model.yml:    "-Arguments": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/register-clobbered-on-all-paths.S.model.yml:    "-ReturnValues": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/register-clobbered-on-all-paths.S.model.yml:    "-PreservedRegisters": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/return-value-and-clobbered-register-in-callee.S.model.yml:    "-PreservedRegisters": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/return-value-and-clobbered-register-in-callee.S.model.yml:    "-ReturnValues": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/scratch-register.S.model.yml:    "-Arguments": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/scratch-register.S.model.yml:    "-ReturnValues": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/scratch-register.S.model.yml:    "-PreservedRegisters": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/stack-pointer-is-not-an-argument.S.model.yml:    "-Arguments": 0
./share/revng/test/tests/analysis/DetectABI/x86_64/write-used-only-one-path.S.model.yml:    "-ReturnValues": 0
```

include/revng/Model/model-schema.yml
aleclearmind: I'm not sure about the benefit of this thing. I'm tempted to just drop it.

Your call.

scripts/tuple_tree_generator/schema/struct.py
Cvelth: Dropped the TODO.

scripts/tuple_tree_generator/templates/struct_impl.cpp.tpl
Cvelth: Wait, this is not as obvious as it seemed. What is it even trying to check? `not TupleTreeCompatible and not KeyedObjectContainerCompatible`? Why do we want that for our fields? Is this from some time long ago when we only allowed "simple" fields or something?

We have some struct-fields that currently fail this check. Should we just drop it?

scripts/tuple_tree_generator/generators/typescript.py
Cvelth: Done.

tests/unit/TupleTreeGenerator/schema.yml
Cvelth: Done.

scripts/tuple_tree_generator/templates/enum.h.tpl
Cvelth: Done.

lib/ABI/Definition.cpp
Cvelth: Done.

share/revng/test/tests/analysis/CollectCFG/x86_64/try-catch-ehframe.S.cfg.yml
Cvelth: Yep, dropping some of these would reduce test quality, IMO.

Closing the thread.

scripts/tuple_tree_generator/generators/cpp.py
Cvelth: Added a comment to `TupleTreeGenerator.cmake`:

```cmake
# TODO: `STRING_TYPES` can use a bit more love. Currently it's up to the given
# TTG-backend to handle them correctly. We should set more rigid rules related
# to it, for example, c++ backend should probably use a dedicated trait.
```

Cvelth: P. S. I hate how whatever thing we use to format cmake breaks the indentation and make this TODO look like this :(

include/revng/Model/model-schema.yml
Cvelth: > Your call.

Well, I really like the idea of having a dedicated "newly-constructed-and-invalid-until-inserted" state. So, I'd prefer we keep this.

share/doc/revng/user-manual/tutorial/running-analyses.md
Cvelth: (I dropped the `**ivan**` comment)

scripts/tuple_tree_generator/schema/schema.py
Cvelth: Ok, this is not as simple. Turns out we use the same `Definition` base for everything. Including, for example, sequences and references - and those don't have a name.

Which means we do have to make it optional, even if it's guaranteed to be present for all the "root" definitions (like the ones iterated here).

-------------

As such, I went with our suggested `notNull` (except I renamed it into `assertNotNone` for extra explicitness).

scripts/tuple_tree_generator/templates/tuple_tree_gen.py.tpl
Cvelth: To unblock this PR, I moved everything related to this into a new one:
https://github.com/revng/revng/pull/532

include/revng/Model/model-schema.yml
Cvelth: Done (as a part of a different PR):
https://github.com/revng/revng/pull/532

scripts/tuple_tree_generator/schema/schema.py
Cvelth: P. S. Actually, check-conventions failed on that name, the final one is `assert_not_none`.

python/revng/internal/cli/_commands/generate_migrations/metaschema.yml
mrjackv: Ok, I did some investigating, in essence this is what happened:
1. Before the model schema merge, `extract_yaml.py` (the tool that took the schema bits from `.h` files) would run `jsonschema.validate`
2. After the tool was dropped, the `metaschema.yml` has been moved to the `generate-migrations` command, because that's the only place where it's used
3. In fact, running `revng model generate-migrations ...` does generate an error on this branch

So basically, the problem is that the validation is never triggered during build time.
So I think the correct course of actions is something along the lines of:
1. Moving the `metaschema.yml` somewhere else (e.g. `share/`)
2. Have either one of this:
   1. The `scripts/tuple-tree-generate.py` checks the metaschema
   2. Somewhere in `TupleTreeGenerator.cmake` a python script is run to check this
   3. `check-conventions` gets a new pass that checks this
   
   (IMHO I prefer 1, but any of these 3 works)

3. Still have generate-migrations check it with the metaschema in the new location

@aleclearmind @Cvelth

Cvelth: Sounds good to me! I'll take care of this then.

Thanks for looking into it!

scripts/tuple_tree_generator/schema/schema.py
aleclearmind: A function called `assert` is not expected to return something, I'd stick with `not_none`. We have `notNull` in C++ `:P`

scripts/tuple_tree_generator/generators/cpp.py
aleclearmind: If you really want to....
```cmake
# cmake-format: off
# ...
# cmake-format: on
```

scripts/tuple_tree_generator/schema/schema.py
Cvelth: Sure!

python/revng/internal/cli/_commands/generate_migrations/metaschema.yml
aleclearmind: What's the situation here?

scripts/tuple_tree_generator/generators/cpp.py
Cvelth: Yeah, no, not worth it for just a todo.

scripts/tuple_tree_generator/templates/struct_impl.cpp.tpl
aleclearmind: I think we wanted to keep "keyed" objects for vectors.

> We have some struct-fields that currently fail this check.

Can you name one?

> Should we just drop it?

We probably can.

python/revng/internal/cli/_commands/generate_migrations/metaschema.yml
Cvelth: Looking into it right now. This made me realize that this PR does in fact need migration! My bad on not noticing that sooner.

scripts/tuple_tree_generator/templates/struct_impl.cpp.tpl
Cvelth: > Can you name one?

For example, this would assert every time there is a TTG-generated struct used as a field of another TTG-generated struct.
Think
```cpp
struct A;
struct B { A a; };
```

We don't have any in the model currently, but, for example, yield uses that.

Cvelth: Either way, it's hard to determine whether we need something like this without knowing what in particular is it designed to check. And I fail to grasp what it's about 🤔.

aleclearmind: I don't know to be honest. We can drop it.

Cvelth: I already dropped it from the currently pushed branch (I have a certain distaste for code that both doesn't work and makes it really hard to understand what it's trying to do :P).

Feel free to close this unless, I don't know, the reason behind this piece of code strikes you in a dream /j

scripts/tuple_tree_generator/schema/schema.py
Cvelth: Done.

scripts/tuple_tree_generator/generators/cpp.py
Cvelth: (feel free to close this if the comment above looks good enough to you)

python/revng/internal/cli/_commands/generate_migrations/metaschema.yml
Cvelth: Done, please take a look at the new commits:
- `migrations: do not duplicate the template`
- `TTG: move metaschema.yml back to the TTG`
- `TTG: verify given schema`

P. S. Also, turns out migration wasn't necessary after all 😅, my bad on that.

CMakeLists.txt
```diff
@@ -261,6 +261,11 @@ add_custom_command(
 add_custom_target(clang-format-dot-file ALL
                   DEPENDS "${CMAKE_BINARY_DIR}/share/revng/.clang-format")
 
+# Install the TTG metaschema so that model migration tools can access it.
+file(MAKE_DIRECTORY "${CMAKE_BINARY_DIR}/share/revng/tuple-tree-generator")
+install(FILES "${CMAKE_SOURCE_DIR}/scripts/tuple_tree_generator/metaschema.yml"
+        DESTINATION share/revng/tuple-tree-generator/)
```
aleclearmind: This shouldn't be necessary. The whole `share/` directory is installed as is from the build directory.

IIRC we 1) copy `$SOURCE_DIR/share` to `$BUILD_DIR/share` and then install all of `$BUILD_DIR/share`. Please report if this is not the case.

I'd suggest to move `scripts/tuple_tree_generator/metaschema.yml` to `share/revng/tuple-tree-generator/metaschema.yml` so this all becomes automatic.

scripts/tuple-tree-generate.py
```diff
@@ -190,6 +191,11 @@ def main():
     with open(args.schema, encoding="utf-8") as f:
         raw_schema = yaml.safe_load(f)
 
+    metaschema_path = Path(__file__).parent / "tuple_tree_generator" / "metaschema.yml"
+    with open(metaschema_path, encoding="utf-8") as f:
+        metaschema = yaml.safe_load(f)
+        jsonschema.validate(instance=raw_schema, schema=metaschema)
```
aleclearmind: When you use a `Path`, you can do `.open` or even `.read_text`. No biggie though.

CMakeLists.txt
Cvelth: That doesn't really work. The thing, we need `scripts/tuple_tree_generator/metaschema.yml` at build-time (or more like at ttg-generation-time)! Way before installation. So, IMO, it's a lot easier to provide it there so that the one from sources is used right away. And then we can install the copy for migrations (which is run from an installed copy of revng-internal cli).

scripts/tuple-tree-generate.py
Cvelth: I actually used `.read_text` elsewhere today! I love that thing, but yaml deserializer reads the file on its own (and I hope they use something more efficient, like mmapping internally :P)

TIL about `.open` though!

As for here, I did this simply because that's exactly what is done right above this to open `raw_schema`. And, IMO, it's better to use mediocre but consistent code as opposed to a mix and match of different styles!

CMakeLists.txt
aleclearmind: I'm confused: how is moving the file from `scripts/tuple_tree_generator` to `share/revng/tuple-tree-generator` *in the source directory* changing any of this?

It just saves you from specifying these lines (`file(MAKE_DIRECTORY` + `install(FILES`).

Cvelth: > how is moving the file

Currently, tuple-tree-generator python scripts are self-contained in their subdirectory. Moving the file out, would mean teaching it about the file structure _outside_ its root.

Consider that these scripts do not have a dependency on the revng cli infrastructure (and they cannot really be, they are pretty much the first thing that runs when revng is built), so they are on their own.

Moving this file outside would mean encoding something like `../../share/revng/` in the source, which, IMO, smells a bit.

---------------

> It just saves you from specifying these lines

Well, we still end up having to specify the path difference _somewhere_. And, IMO, it's much better to do so in the _build system_ configuration rather than in an obscure rarely touched script.

Cvelth: TL;DR: we need this file available at two points: at the very start of build logic. And then at "runtime" for a subcommand.

IMO, placing it in a place that makes sense for the more obscure use-case + an explicit copy (installation) when building makes the most sense.

Cvelth: I guess another option is to place this file wherever and then manually pass it as an argument to TTG, which wouldn't be too bad. But I still expect that to touch way more cmake than just a `file(MAKE_DIRECTORY + install` pair.

------------

Also, `file(MAKE_DIRECTORY` is not even necessary I guess, since that's the only file in that directory, we could probably get away with calling the file something like `tuple-tree-metaschema.yml` and placing it into `share/revng`?

aleclearmind: It's very suspicious to have a verbatim copy of a file in the source directory in a different location once installed.

`../share/revng` is not much worse than

```python
metaschema_path = Path(__file__).parent / "tuple_tree_generator" / "metaschema.yml"
```

Let's keep it simple and move it to `share/revng`.
This is a build time tool, we don't need to strive for perfect elegance.

Cvelth: Well, `Path(__file__).parent` is an orthogonal problem. This path still needs to be script-relative and not working-directory-relative.

But sure, as you said, it indeed doesn't need to be perfect.
```python
    metaschema_path = Path(__file__).parent / "../share/revng/tuple-tree-generator/metaschema.yml"
```
works.

(Done)

PR #530, opened 2025-10-06
Fix Pypeline invalidation

mrjackv:


scripts/tuple_tree_generator/generators/cpp.py
```diff
@@ -299,3 +299,13 @@ def render_docstring(docstr: str):
         if not rendered_docstring.endswith("\n"):
             rendered_docstring = rendered_docstring + "\n"
         return rendered_docstring
+
+    @staticmethod
+    def enum_underlying_type(enum: EnumDefinition):
+        # The 2 accounts for `Invalid` and `Count`
+        total_members = len(enum.members) + 2
+        for size in (8, 16, 32, 64):
+            if total_members <= 2**size:
```
aleclearmind, line 352: Unformatted?

include/revng/PipeboxCommon/Helpers/AnalysisRunner.h
```diff
@@ -23,9 +23,6 @@ namespace revng::pypeline::helpers {
 template<typename ContainerListUnwrapper>
```
aleclearmind: 1. ~~Add a `concept` for `typename ContainerListUnwrapper`.~~ It can be dropped.
2. The implementations of `ContainerListUnwrapper` should `assert` in case the Ith element is not there. As of now, at least the native one is using `ArrayRef::operator[]` which does not assert. I'm not sure about `nanobind::list::operator[]`.
3. `ContainerListUnwrapper::unwrap` does not *unwrap* anything, it gets the Ith element.
   It feels very much like `std::get<I>(...)`.
   In fact, I think that `PipeRunner` and `AnalysisRunner` can drop the `typename ContainerListUnwrapper` template argument and just use a trait. You can implement `std::get` for `ArrayRef` and for `nanobind::list`, or you can make a similar new trait (`revng::dynamic_get<I>`?).
   The type of the list can be inferred from the `Containers` argument.
4. Please stop using the term *wrap* or *unwrap* altogether. Please never use them again. Every time it means "do stuff" and it's confusing. If you can't come up with a better name, leave a WIP comment and we can discuss during code review.

include/revng/ADT/CompilationTime.h
```diff
@@ -96,6 +96,22 @@ constexpr std::optional<size_t> select(CallableType &&Callable) {
                         std::forward<CallableType>(Callable));
 }
 
+/// Calls \ref Callable with unpacked sequence of \ref IterationCount.
+template<size_t IterationCount, typename CallableType>
+constexpr auto unpack(CallableType &&Callable) {
```
aleclearmind: What is this unpacking? Very confusing name out of context.

`callWithIndexSequence`?

include/revng/PipeboxCommon/ObjectID.h
```diff
@@ -170,6 +182,43 @@ class ObjectID {
     }
   }
 
+  std::vector<uint8_t> serializeBytes() const {
```
aleclearmind: `toBytes`

include/revng/PipeboxCommon/ObjectID.h
```diff
@@ -170,6 +182,43 @@ class ObjectID {
...
+    const Kind &TheKind = kind();
+    std::vector<uint8_t> Result;
+
+    if (TheKind == Kinds::Binary) {
+      return {};
+    } else if (TheKind == Kinds::Function) {
+      Result.push_back(1);
```
aleclearmind: No magic numbers. Make an `enum`.

include/revng/PipeboxCommon/ObjectID.h
```diff
@@ -170,6 +182,43 @@ class ObjectID {
     }
   }
 
+  std::vector<uint8_t> serializeBytes() const {
+    const Kind &TheKind = kind();
+    std::vector<uint8_t> Result;
+
+    if (TheKind == Kinds::Binary) {
```
aleclearmind, line 197: Here you are making a deep comparison. I'd use a comparison between pointers. They are singletons.

include/revng/Support/IntegerSerialization.h
```diff
@@ -0,0 +1,85 @@
...
+
+template<isUintOrEnum... Args>
+inline std::vector<uint8_t> serializeIntsToBytes(const Args &...Values) {
+  std::vector<uint8_t> Result(serializedIntsSize<Args...>);
+  uint8_t *Ptr = Result.data();
+  (detail::writeNext<Args>(Ptr, Values), ...);
+  return Result;
+}
+
+template<SpecializationOf<std::tuple> T>
```
aleclearmind: You could use `TupleSizeCompatible`.

include/revng/ADT/Concepts.h
```diff
@@ -125,3 +125,11 @@ template<typename T, typename... Types>
 inline constexpr bool anyOf() {
   return (std::is_same_v<T, Types> || ...);
 }
+
+template<typename T>
+concept isUint = anyOf<T, uint8_t, uint16_t, uint32_t, uint64_t>();
```
aleclearmind: 1. Capitalize all `concept`s.
2. I'd go with `std::is_integral_v`. Why not handle signed types as well?

include/revng/Support/IntegerSerialization.h
```diff
@@ -0,0 +1,85 @@
...
+
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/Support/Endian.h"
+
+#include "revng/ADT/CompilationTime.h"
+#include "revng/ADT/Concepts.h"
+
+namespace detail {
+
+inline constexpr auto IntEndianness = llvm::support::little;
```
aleclearmind: 1. This is an *opinionated* integer serialization library. For instance, it serializes things in little endian. Add documentation about the format we're making up, assume someone has to implement a compatible parser.
   Don't assume this is obvious.
2. Expand all `Int` to `Integer`

include/revng/Support/IntegerSerialization.h
```diff
@@ -0,0 +1,85 @@
...
+inline constexpr auto Unaligned = llvm::support::unaligned;
+
+template<SpecializationOf<std::tuple> T>
+inline constexpr size_t computeSize() {
```
aleclearmind: `packedSize`?

include/revng/Support/IntegerSerialization.h
```diff
@@ -0,0 +1,85 @@
...
+  size_t Result = 0;
+  compile_time::repeat<std::tuple_size_v<T>>([&Result]<size_t I>() {
```
aleclearmind, line 21: Can you use `std::visit`?

include/revng/Support/IntegerSerialization.h
```diff
@@ -0,0 +1,85 @@
...
+
+template<isUintOrEnum... Args>
+struct SerializedIntsSize<Args...> {
+  static constexpr size_t Size = computeSize<std::tuple<Args...>>();
+};
+
+template<SpecializationOf<std::tuple> T>
+struct SerializedIntsSize<T> {
+  static constexpr size_t Size = computeSize<T>();
+};
```
aleclearmind, line 40: I think these can be turned into a

```c++
template<typename T>
constexpr size_t SerializedIntSize;


template<isUintOrEnum... Args>
constexpr size_t SerializedIntSize<Args> = ...;
```

include/revng/Support/IntegerSerialization.h
```diff
@@ -0,0 +1,85 @@
...
+namespace detail {
+
+inline constexpr auto IntEndianness = llvm::support::little;
+inline constexpr auto Unaligned = llvm::support::unaligned;
+
+template<SpecializationOf<std::tuple> T>
+inline constexpr size_t computeSize() {
+  size_t Result = 0;
+  compile_time::repeat<std::tuple_size_v<T>>([&Result]<size_t I>() {
+    Result += sizeof(std::tuple_element_t<I, T>);
```
aleclearmind: I think you're assuming you `tuple` is composed by primitive types, but you don't assert it anywhere.
What if it's another tuple?

You could recur instead of using `sizeof`. Otherwise, just assert.

Please document what is the scope of this library. What do you want to handle? How generic do you want to be?

include/revng/Support/IntegerSerialization.h
```diff
@@ -0,0 +1,85 @@
...
+inline constexpr size_t
+  serializedIntsSize = detail::SerializedIntsSize<T...>::Size;
+
+template<isUintOrEnum... Args>
+inline std::vector<uint8_t> serializeIntsToBytes(const Args &...Values) {
+  std::vector<uint8_t> Result(serializedIntsSize<Args...>);
+  uint8_t *Ptr = Result.data();
+  (detail::writeNext<Args>(Ptr, Values), ...);
+  return Result;
+}
```
aleclearmind: 1. I'm a bit confused here. I understand you want to return a `vector<uint8_t>` (as opposed to use a `raw_ostream`, which [s also supported](https://llvm.org/doxygen/EndianStream_8h_source.html#l00028)), but why follow this approach of reserve + use a pointer as a cursor? You could use `push_back`.
   This would make more sense if you wanted to return a `std::array` (given that you can compute the size of the array statically).
2. I think using a `SmallVector<uint8_t>` here would be nice.

include/revng/Support/IntegerSerialization.h
```diff
@@ -0,0 +1,85 @@
...
+template<isUintOrEnum T>
+void writeNext(uint8_t *&Ptr, const T &Value) {
+  llvm::support::endian::write<T, Unaligned>(Ptr, Value, IntEndianness);
+  Ptr += sizeof(T);
+}
+
+template<isUintOrEnum T>
+T readNext(const uint8_t *&Ptr) {
+  return llvm::support::endian::readNext<T, Unaligned>(Ptr, IntEndianness);
+}
```
aleclearmind: The fact that `writeNext` increments `Ptr` but `readNext` doesn't is veeeery suspicious.

include/revng/Support/IntegerSerialization.h
```diff
@@ -0,0 +1,85 @@
...
+template<isUintOrEnum... ArgsT>
+inline void
+deserializeIntsFromBytes(llvm::ArrayRef<uint8_t> Data, ArgsT &...Args) {
+  revng_assert(Data.size() == serializedIntsSize<ArgsT...>);
+  const uint8_t *Ptr = Data.data();
+  ((Args = detail::readNext<ArgsT>(Ptr)), ...);
+}
+
+template<SpecializationOf<std::tuple> T>
+inline void deserializeIntsFromBytes(llvm::ArrayRef<uint8_t> Data, T &Value) {
```
aleclearmind: I'd just call these `fromBytes` and `toBytes`.

lib/PipeboxCommon/Python/_pipebox.cpp
```diff
@@ -51,6 +51,18 @@ NB_MODULE(_pipebox, m) {
     .def_static("root", &ObjectID::root)
     .def("serialize", &ObjectID::serialize)
     .def_static("deserialize", &ObjectID::deserialize)
+    .def("serialize_bytes",
```
aleclearmind: `to_bytes` and `from_bytes`

python/revng/pypeline/object.py
```diff
@@ -225,6 +230,17 @@ def deserialize(cls, obj: str) -> ObjectID:
         """Convert a string representation back to an object"""
         raise NotImplementedError()
 
+    @abstractmethod
+    def serialize_bytes(self) -> bytes:
+        """Convert the object to a bytes representation that can be later be deserialized"""
+        raise NotImplementedError()
```
aleclearmind, line 247: This is an understatement.

We need to put a bigass explanation about the expectations we have.
If you don't want to specify the full format (byte discriminator + data + byte discriminator + ...), we have at very least to explain the fact that these need to be lexicographically sortable, and explaining what we mean by that.

include/revng/Support/IntegerSerialization.h
mrjackv: the `readNext` in llvm automatically increments the pointer, the `writeNext` counterpart [has been introduced in llvm 19.1](https://github.com/llvm/llvm-project/commit/51d85b7b107e9508b90cec274f3f122444fe7ee6) so I had to use `write` which does not increment it and do it manually.

scripts/tuple_tree_generator/generators/cpp.py
mrjackv: No, I've run `check-branch` on this
Edit: I've also re-run check-conventions just to make sure, it's properly formatted.

include/revng/ADT/Concepts.h
mrjackv: 1. Ok
2. `std::is_integral` also includes other types which are not desirable, such as `bool`, `char*_t`, `char` and `wchar_t`. Handling signed types can be added, but I wanted to be strict on the types supported. If need be a `IsInt` and `IsIntOrEnum` can be used

include/revng/Support/IntegerSerialization.h
mrjackv: `std::visit` only works for `std::variant`, not for `std::tuple`

include/revng/Support/IntegerSerialization.h
mrjackv: 1. I think `std::array` is the way to go, it's the one that makes the most sense. Using ostreams makes little sense since on the caller side it would always be paired with a `llvm::raw_sv_stream`
2. See (1)

include/revng/Support/IntegerSerialization.h
aleclearmind: OK, but add a comment about this fact.

include/revng/ADT/Concepts.h
aleclearmind: 2. > other types which are not desirable

   Why are the undesirable?
   `uint8_t` could just be a `typedef` to `char`.

include/revng/Support/IntegerSerialization.h
aleclearmind: Right, `std::apply`.

include/revng/PipeboxCommon/Helpers/AnalysisRunner.h
mrjackv: 2. I'll check, in case I'll add an assertion
3. The `ContainerListUnwrapper` carries both the list type and the function that "unwraps". This allows to assert that, e.g. in the python case I'm actually passing a `nanobind::list` and not `ArrayRef` by accident. If you want to say that you prefer calling the `unwrap` function `get`, then I'm fine with that.
4. Ok

include/revng/PipeboxCommon/ObjectID.h
mrjackv: `Kind` is just an enum, either way it's a just an `int` comparison
Also, due to the objects being shared with python, it's hard to make the assumptions that the instances are singleton.

include/revng/ADT/CompilationTime.h
mrjackv: > What is this unpacking?

This will be used either with `std::tuple` or parameter packs, in these contexts the need of `size_t ...I` is because you want to individually unpack the elements and do something with them, hence the name

include/revng/PipeboxCommon/ObjectID.h
aleclearmind: Ah OK, my bad.

include/revng/PipeboxCommon/Helpers/AnalysisRunner.h
aleclearmind: 3. I don't think `AnalysisRunner` should try to enforce the type of the arguments.
   The caller is giving it the list of containers and specifying, through another mean, the type it must have. It's redundant, there's no ambiguity (except for the implicit cast from `std::vector` to `ArrayRef`, but that can be worked around easily).
   The only thing that matters for `AnalysisRunner` is to be able to get the I-th element.
   But I realize that, for the Python side, we're also casting, so `std::get` is not suitable.
   Still, there's no need to have a template argument, you can make a new a function `extractContainerFromList` that's specialized for `ArrayRef<Container *>` and `nanobind::list`.

include/revng/ADT/CompilationTime.h
aleclearmind: The first function is not tuple specific. Also, I'm not sure how one can say that you can "unpack" a tuple into a series of indixes. One can "unpack" a tuple into its members.

```c++
unpack<3>([] <size_t... I> () { });
// vs
callWithIndexSequence<3>([] <size_t... I> () { });
```

Also, add an example usage.

include/revng/Support/IntegerSerialization.h
mrjackv: Ah wait, `std::apply` works on an __instance__ of a tuple, here the computation is done on the __type__ of a tuple

include/revng/Support/IntegerSerialization.h
mrjackv: This:
```
template<typename T>
constexpr size_t SerializedIntSize;
```
Does not work, it needs to be initialized. Whereas the `struct` version will throw an compile-time error because the `Size` member is not present.

include/revng/Support/IntegerSerialization.h
mrjackv: > but you don't assert it anywhere

It's asserted as part of the concept in `writeNext` and `readNext` on the individual elements

include/revng/ADT/CompilationTime.h
mrjackv: Renamed and added an example

include/revng/PipeboxCommon/ObjectID.h
mrjackv: Dropped the magic numbers

include/revng/Support/IntegerSerialization.h
mrjackv: Replaced with `TupleSizeCompatible`

include/revng/Support/IntegerSerialization.h
mrjackv: 2. Done

include/revng/Support/IntegerSerialization.h
mrjackv: Changed to return an `std::array`

include/revng/Support/IntegerSerialization.h
mrjackv: Done

include/revng/Support/IntegerSerialization.h
mrjackv: Renamed

lib/PipeboxCommon/Python/_pipebox.cpp

python/revng/pypeline/object.py
mrjackv: Changed the docstring to document the expected the binary format

include/revng/PipeboxCommon/ObjectID.h
mrjackv: Done

include/revng/Support/IntegerSerialization.h
mrjackv: Added another `static_assert` in `computeSize` that checks that the element type is one that we expect

include/revng/Support/IntegerSerialization.h
aleclearmind: Mmmh, this works:

```c++
template<typename T>
constexpr int SerializedIntSize;

template<>
constexpr int SerializedIntSize<int> = 44;

int lol = SerializedIntSize<int>;
// int lol2 = SerializedIntSize<float>;
```

mrjackv: The example you showed does not compile for the same reason: [link](https://godbolt.org/z/cq8oaf363)

include/revng/PipeboxCommon/Helpers/AnalysisRunner.h
mrjackv: I've had a stab at it, but it's really hard to make all the parts work as expected and keep the functions segregated in their respected subdirectories:
* The extractors live `Native/Helpers.h` and `Python/Helpers.h`, so they are not visible from the parent `AnalysisRunner` and `PipeRunner`
* I'm getting compile errors `error: call to function 'extractContainerFromList' that is neither visible in the template definition nor found by argument-dependent lookup`, I think is due to the fact that headers are ordered in the "wrong" way

This could be solved by moving the two `extractContainerFromList` into a header in `PipeboxCommon/Helpers` but that would break all the isolation given by the `Native` and `Python` subdirectories

include/revng/ADT/Concepts.h
mrjackv: I've dropped the `isUint` and `isUintOrEnum`, added `IsIntegralOrEnum` and used it

include/revng/Support/IntegerSerialization.h
mrjackv: I've renamed the public variable to `packedSize`, I think the helper function should still be called `computeSize` since it follows the semantics of a function

include/revng/Support/IntegerSerialization.h
mrjackv: 1. Added a comment describing how data is (de)serialized

include/revng/Support/IntegerSerialization.h
aleclearmind: Mh, I tested it, but not with clang. Oh well.

include/revng/PipeboxCommon/Helpers/AnalysisRunner.h
aleclearmind: I can't help you without further details, but I don't think header ordering matters here.

The instantiation is triggered from a `.cpp` file where all the includes should be available. Are you including the correct headers files in the `.cpp`?

https://godbolt.org/z/jedWG8dEY

include/revng/Support/IntegerSerialization.h
aleclearmind: I'm not a fan of `compute...` functions. Each function computes something.

I think a "compute" prefix is appropriate only if your computation is a first class citizen in your context (like, a calculator? `:P`).

include/revng/PipeboxCommon/Helpers/AnalysisRunner.h
mrjackv: I've pushed the tentative change for you to take a look: [link](https://github.com/mrjackv/revng/commit/89ed07ebd99cb8dff6071b89c29e7e66d3e297ed)

I'm 100% sure that when the templates get instantiated in the `.cpp` files the proper function is present, but the compile errors are there nonetheless.

include/revng/Support/IntegerSerialization.h
mrjackv: `packedTupleSize`?

include/revng/PipeboxCommon/Helpers/AnalysisRunner.h
aleclearmind: I'm taking a look, please let me have the name of the failing `ninja` target next time, so I can compile less stuff.

mrjackv:
```
ninja libmockPipebox.so
```

aleclearmind: I came up with this, you might be able to minimize it a bit.

https://github.com/aleclearmind/revng/commit/126ed30557beb30828d1b3f77555c831993b41c2

`ninja mockPipebox` works

mrjackv: Pushed the changes, they can be found in the commit `PipeboxCommon: rework {Analysis,Pipe}Runner`

include/revng/Support/IntegerSerialization.h
aleclearmind: Works

mrjackv: Done

python/revng/pypeline/storage/util.py
```diff
@@ -2,4 +2,25 @@
...
 
+from revng.pypeline.object import Kind
+from revng.pypeline.utils.registry import get_singleton
+
+__all__ = ("_REVNG_VERSION_PLACEHOLDER", "_KIND_MAXDEPTH", "check_kind_structure")
+
 _REVNG_VERSION_PLACEHOLDER = "1.0.0"
+
+# Number of bytes that the biggest ObjectID will contain when serialized to bytes
+_KIND_MAXDEPTH = 17
```
aleclearmind: 1. How did you come up with this number?
2. Why is this called "depth"? This is not a "depth".

python/revng/pypeline/storage/util.py
```diff
@@ -2,4 +2,25 @@
...
+
+
+def check_kind_structure():
+    kind_type = get_singleton(Kind)
+    kind_root = kind_type.root()
+    kinds = kind_type.kinds()
+    # Check that we can store all Kinds in a byte
+    assert len(kinds) <= 2**8
+
+    # Check that each byte does not exceed the mask size
```
aleclearmind: Mask?

python/revng/pypeline/storage/util.py
```diff
@@ -2,4 +2,25 @@
...
+    for kind in kinds:
+        assert kind.byte_size() <= (_KIND_MAXDEPTH - 1)
```
aleclearmind: `< _KIND_MAXDEPTH`

python/revng/pypeline/storage/util.py
```diff
@@ -2,4 +2,25 @@
...
+
+__all__ = ("_REVNG_VERSION_PLACEHOLDER", "_KIND_MAXDEPTH", "check_kind_structure")
+
 _REVNG_VERSION_PLACEHOLDER = "1.0.0"
+
+# Number of bytes that the biggest ObjectID will contain when serialized to bytes
+_KIND_MAXDEPTH = 17
+
+
+def check_kind_structure():
```
aleclearmind, line 25: Add documentation on why we need this, stating this is a temporary hack to avoid having to implement blablabla.

python/revng/pypeline/storage/memory.py
```diff
@@ -87,27 +88,51 @@ def put(
...
-                    if entry.object_id in objects:
-                        del objects[entry.object_id]
+    @staticmethod
+    def _is_related(target: ObjectID, to: ObjectID) -> bool:
+        test_object: ObjectID | None = target
+        while test_object is not None:
+            if test_object == to:
+                return True
+            test_object = test_object.parent()
+        return False
```
aleclearmind: 1. This function should also test viceversa, instead of pushing this responsibility on the caller.
3. This should likely be a `ObjectID` method.

python/revng/pypeline/storage/memory.py
```diff
@@ -87,27 +88,51 @@ def put(
...
+
+    def invalidate(self, invalidation_list: ModelPathSet):
+        invalidated: dict[ContainerLocation, set[ObjectID]] = defaultdict(set)
+
+        entries = set()
```
aleclearmind: 1. Typing
2. Rename to `objects_to_delete`?

python/revng/pypeline/storage/memory.py
```diff
@@ -87,27 +88,51 @@ def put(
...
+    @staticmethod
+    def _is_related(target: ObjectID, to: ObjectID) -> bool:
+        test_object: ObjectID | None = target
+        while test_object is not None:
+            if test_object == to:
+                return True
+            test_object = test_object.parent()
+        return False
+
+    def invalidate(self, invalidation_list: ModelPathSet):
```
aleclearmind: Better comment this function

python/revng/pypeline/storage/sqlite3.py
```diff
@@ -16,7 +16,9 @@
 from revng.pypeline.utils.registry import get_singleton
 
 from .storage_provider import ContainerLocation, ProjectMetadata, SavePointsRange, StorageProvider
-from .util import _REVNG_VERSION_PLACEHOLDER
+from .util import _KIND_MAXDEPTH, _REVNG_VERSION_PLACEHOLDER, check_kind_structure
+
+_KIND_MASK = f"x'{"ff" * _KIND_MAXDEPTH}'"
```
aleclearmind: Explain how the `object_id` `BLOB` now works and the reason.

python/revng/pypeline/storage/sqlite3.py
```diff
@@ -74,14 +76,23 @@
...
     WHERE dependencies.model_path IN ({model_paths})
-          AND objects.container_id = dependencies.container_id
+          AND (
+            (
+              objects.object_id >= dependencies.object_id
+              AND objects.object_id <= CAST((dependencies.object_id || {kind_mask}) AS BLOB)
+            ) OR (
+              dependencies.object_id != x'' AND objects.object_id = x''
+            )
+          )
```
aleclearmind, line 117: Comment with examples. This is not obvious.

tests/pypeline/test_all.py
```diff
@@ -30,6 +32,11 @@
 from revng.pypeline.task.requests import Requests
 from revng.pypeline.task.savepoint import SavePoint
 
+# Patch away the check for the kind structure, this is needed because, unlike
+# the actual implementation, we have the GRANDCHILD kind which fails the check
+setattr(_storage_memory_module, "check_kind_structure", lambda: None)
+setattr(_storage_sqlite3_module, "check_kind_structure", lambda: None)
```
aleclearmind: Mmmmmh, how can we test it if it's broken? :thinking:

This doesn't sound right.

tests/pypeline/test_all.py
```diff
@@ -30,6 +32,11 @@
 from revng.pypeline.task.requests import Requests
 from revng.pypeline.task.savepoint import SavePoint
 
+# Patch away the check for the kind structure, this is needed because, unlike
+# the actual implementation, we have the GRANDCHILD kind which fails the check
+setattr(_storage_memory_module, "check_kind_structure", lambda: None)
```
aleclearmind: I don't think this has to be a limitation of the in-memory module.

Can we make this hack specific to the Sqlite storage backend?

python/revng/pypeline/storage/storage_provider.py
```diff
@@ -32,6 +32,9 @@ class ContainerLocation:
     configuration_id: ConfigurationId
 
 
+Invalidated = dict[ContainerLocation, set[ObjectID]]
```
aleclearmind: I'd just call this `ObjectInContainerSet`.
If you really want to ties this to invalidation, call this `InvalidedObjects`.

tests/pypeline/test_all.py
```diff
@@ -653,3 +653,58 @@ def test_schedule_serdes(model):
     )
     schedule_str = schedule.serialize()
     pipeline.deserialize_schedule(schedule_str)
+
+
+def test_storage_invalidation(storage_provider: StorageProvider):
```
aleclearmind, line 651: Is this tested against all storage providers?

python/revng/pypeline/storage/util.py
mrjackv: 1. 1 byte for the Kind type + 16 bytes for the `MetaAddress`
2. It's depth in the sense that it's the maximum length in bytes of the longest sequence, see `MaxObjectSize` in REP49

python/revng/pypeline/storage/util.py
mrjackv: This makes it clearer, the kind types can be encoded in one byte, if we had more than 256, this would be `assert kind.byte_size() <= (_KIND_MAXDEPTH - 2)`

tests/pypeline/test_all.py
mrjackv: It's a limitation of the `check_kind_structure` function
In theory the function could actually walk the kind tree and determine that the one in testing is conformant, but that would be extra code with very little tradeoff.

tests/pypeline/test_all.py
mrjackv: It's broken in the sense that the test implementation has 3 ranks, `root -> child -> grandchild`, the `check_kind_structure` does a simplified check that it's just `root -> <everything>`.

To clarify:
* More proper check:
   1. Compute the longest serialized path for ObjectID
   2. Check that `_KIND_MAXDEPTH` is less than that
* Current simplified check:
   1. Check that we only have 2 ranks (root and children)
   2. Check that the children are smaller than `_KIND_MAXDEPTH - 1`

tests/pypeline/test_all.py
mrjackv: It's tested against both `InMemoryStorageProvider` and `Sqlite3StorageProvider`

python/revng/pypeline/storage/util.py
aleclearmind: 1. Document it
2. OK, this is influenced by the maximum depth, but it's the maximum *size* of the serialized object id, isn't it? Let's make the name reflect this.

python/revng/pypeline/storage/util.py
aleclearmind: I'm really confused by your reasoning around this.

Please make a `_MAX_SIZE` variable and cleanly explain how you compute it.

tests/pypeline/test_all.py
mrjackv: nvm, this writing it out made me realize it's easier than I expected, will push it fixed without the hack on tests

python/revng/pypeline/storage/util.py
mrjackv: 1. Done
2. Changed to `_OBJECTID_MAXSIZE`

python/revng/pypeline/storage/util.py
mrjackv: Done

python/revng/pypeline/storage/util.py
mrjackv: Done

python/revng/pypeline/storage/memory.py
mrjackv: 1. Done
2. Moved to `ObjectID`

python/revng/pypeline/storage/memory.py
mrjackv: 1. Done
2. Done

tests/pypeline/test_all.py
mrjackv: Pushed the fixed `check_kind_structure`

python/revng/pypeline/storage/util.py
mrjackv: This comment has been dropped and better ones have been put in its place

tests/pypeline/test_all.py
mrjackv: The `setattr` has been dropped

python/revng/pypeline/storage/storage_provider.py
mrjackv: Renamed to `InvalidedObjects`

python/revng/pypeline/storage/sqlite3.py
mrjackv: Added a big comment explaining what's going on

python/revng/pypeline/storage/sqlite3.py
mrjackv: Done

python/revng/pypeline/storage/memory.py
mrjackv: Done

python/revng/pypeline/storage/storage_provider.py
aleclearmind: Hopefully `InvalidatedObjects`

python/revng/pypeline/storage/util.py
```diff
@@ -2,4 +2,53 @@
...
+        if kind == kind_root:
+            kind_sizes[kind] = 0
+        else:
+            kind_sizes[kind] = kind_sizes[kind.parent()] + 1 + kind.byte_size()
+
+    # Check that the maximum kind size does not exceed the expected maximum size
+    assert max(kind_sizes.values()) <= _OBJECTID_MAXSIZE
+
+
+def check_object_id(object_id: ObjectID):
```
aleclearmind: This has too much of a generic name for being in `util`. I understand you want to reuse it for the HHS backend, but as of today it makes it look like `rank() >= 2` is not supported anywehere.

I'd keep it close to sqlite3 storage and move it to a more appropriate place later on.

Otherwise, find a more suitable name, like `check_object_id_supported_by_sql`.

mrjackv: Renamed to `check_object_id_supported_by_sql` and added a comment explaining what the function does

python/revng/pypeline/storage/storage_provider.py
mrjackv: Yes, I typo'd only in the previous message

PR #531, opened 2025-10-08
Add input container for pypeline

mrjackv:


python/revng/pypeline/cli/pipeline/run_pipe.py
```diff
@@ -160,16 +160,17 @@ def run_pipe_command(
         # Load the containers with args form the command line
         containers = []
         for arg in pipe.arguments:
+            arg_name = arg.name.replace("-", "_")
```
aleclearmind: Factor out this normalization into a function

lib/PipeboxCommon/Python/_pipebox.cpp
```diff
@@ -86,7 +86,11 @@ NB_MODULE(_pipebox, m) {
            return Handle.diff(*nanobind::cast<Model *>(Other));
          })
     .def("clone", &Model::clone)
-    .def("serialize", &Model::serialize)
+    .def("serialize",
+         [](Model &Handle) {
+           llvm::SmallVector<char, 0> Buffer = Handle.serialize();
+           return nanobind::bytes(Buffer.data(), Buffer.size());
```
aleclearmind, line 111: Is this copying?

include/revng/PipeboxCommon/Model.h
```diff
@@ -26,6 +26,9 @@ class Model {
   Model clone() const { return *this; }
 
   std::set<ObjectID> children(const ObjectID &Obj, Kind Kind) const {
+    if (Obj.kind() == Kind)
+      return { Obj };
```
aleclearmind: I think we should assert the kind is different from the current one and it's a child.

I'm not a child of myself `:P`

include/revng/ADT/CUniquePtr.h
```diff
@@ -0,0 +1,53 @@
...
+    this->Ptr = decltype(this->Ptr)(Ptr, &close);
+    return *this;
+  }
+
+  element_type &operator*() { return *Ptr; }
+  pointer operator->() { return &*Ptr; }
+  pointer get() { return &*Ptr; }
+
+private:
+  static void close(element_type *Ptr) {
```
aleclearmind: `destroy`?

include/revng/ADT/CUniquePtr.h
```diff
@@ -0,0 +1,53 @@
...
+  static_assert(std::tuple_size_v<typename FT::Arguments> == 1,
+                "Destructor must take a single argument");
+  static_assert(std::is_void_v<typename FT::ReturnType>
+                  or not std::is_same_v<decltype(OkValue), std::monostate>,
+                "Must specify an OkValue if destructor is non-void");
+
+public:
+  using pointer = std::tuple_element_t<0, typename FT::Arguments>;
+  static_assert(std::is_pointer_v<pointer>);
+  using element_type = std::remove_pointer_t<pointer>;
```
aleclearmind: Snake-case is not compatible with our conventions.

include/revng/ADT/CUniquePtr.h
```diff
@@ -0,0 +1,53 @@
...
+
+private:
+  std::unique_ptr<element_type, void (*)(pointer)> Ptr;
```
aleclearmind: No abbreviations

include/revng/ADT/CUniquePtr.h
```diff
@@ -0,0 +1,53 @@
...
+
+public:
+  CUniquePtr() : Ptr(nullptr, &close) {}
+  CUniquePtr(element_type *Ptr) : Ptr(Ptr, &close) {}
+  CUniquePtr &operator=(element_type *Ptr) {
+    this->Ptr = decltype(this->Ptr)(Ptr, &close);
+    return *this;
+  }
```
aleclearmind: `unique_ptr` does not have this method. I'd prefer not to have it.

include/revng/ADT/CompilationTime.h
```diff
@@ -160,4 +161,31 @@ struct ArrayTraits<std::array<T, N>> {
...
+struct FunctionTraits<ReturnT (&)(Args...)> {
+  using ReturnType = ReturnT;
+  using Arguments = TypeList<Args...>;
+};
+
+template<typename ReturnT, typename... Args>
+struct FunctionTraits<ReturnT (*)(Args...)> {
+  using ReturnType = ReturnT;
+  using Arguments = TypeList<Args...>;
+};
```
aleclearmind, line 191: These feel quite useful! I'd keep them out of the `detail` namespace.

include/revng/ADT/CUniquePtr.h
```diff
@@ -0,0 +1,53 @@
...
+//
+
+#include "revng/ADT/CompilationTime.h"
+
+/// Helper class that wraps a std::unique_ptr for C-like objects that have a
+/// `free` function. Uses a template parameter for the destructor so that the
+/// pointer type can be automatically inferred. A lot of C destructors return a
+/// return code, to accommodate that, specify the `OkValue` template parameter;
+/// this will be asserted to be equal at destruction.
+template<auto &Destructor, auto OkValue = std::monostate{}>
```
aleclearmind: 1. `OkValue`: If you want a "nope" value I'd go for `void`.
2. `Destructor`: what exactly is `auto` here in practice? o_O Typically I'd put a `std::integral_constant`, but if this works it's pretty cool.

include/revng/ADT/CUniquePtr.h
```diff
@@ -0,0 +1,53 @@
...
+class CUniquePtr {
```
aleclearmind: This `class` basically just customizes the destructor by adding an assertion.

Can we do this by just using a custom "deleter" in a regular `unique_ptr`?

I'd prefer not to have `unique_ptr`-like classes, if we can avoid them.


```c++
template<auto &Deleter, int Expected>
void myDeleter(void *Ptr) {
  assert(Deleter(Ptr) == Expected);
}

template<auto &Deleter, int Expected>
using MyUniquePtr = std::unique_ptr<getTypeFromDeleter<Deleter>, myDeleter<Deleter, Expected>>;
```

include/revng/Support/TarIO.h
```diff
@@ -0,0 +1,68 @@
...
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "llvm/Support/MemoryBuffer.h"
+
+#include "revng/ADT/CUniquePtr.h"
+#include "revng/Support/Generator.h"
+
+#include "archive.h"
```
aleclearmind: Is this a system header? We need to put it in the right spot!

include/revng/Support/TarIO.h
```diff
@@ -0,0 +1,68 @@
...
+  Plain,
+  Gzip,
+};
+
+class TarReader {
+private:
+  CUniquePtr<archive_read_free, ARCHIVE_OK> Archive;
+
+public:
+  TarReader(llvm::ArrayRef<const char> Ref, TarFormat Format = TarFormat::Gzip);
```
aleclearmind: Avoid default arguments

include/revng/Support/Tar.h
aleclearmind, line 1: Just `Tar.h`

lib/Support/TarIO.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+  RC = archive_read_support_format_tar(Archive.get());
+  revng_assert(RC == ARCHIVE_OK);
+
+  RC = archive_read_open_memory(Archive.get(), Ref.data(), Ref.size());
+  revng_assert(RC == ARCHIVE_OK);
+}
+
+cppcoro::generator<TarReader::Entry> TarReader::entries() {
+  archive_entry *Entry;
```
aleclearmind: *Always* initialize scalars

lib/Support/Tar.cpp
```diff
@@ -0,0 +1,117 @@
...
+  // possible or PAX if necessary
+  RC = archive_write_set_format_pax_restricted(Archive.get());
+  revng_assert(RC == ARCHIVE_OK);
+
+  RC = archive_write_open2(Archive.get(),
+                           this,
+                           NULL,
+                           &archiveWrite,
+                           NULL,
+                           NULL);
```
aleclearmind, line 84: Document what's going on, it's not obvious.

State that we do this because we want to be in control of the stream writing the data.
I assume that `archive` won't open any files but just call `archiveWrite`, correct? If so, state it.
State that `archive_write_open2` will invoke `archiveWrite` passing `this`.

include/revng/ADT/CUniquePtr.h
mrjackv: I know, but this is to make it look like the `using`s of `std::unique_ptr`

image

aleclearmind: I know, still, this is not worth an exception, unless you're trying to respect some contract/concept.

include/revng/ADT/CUniquePtr.h
mrjackv: 1. `void` does not work, it needs to be an object, you cannot do `void{}`
2. `Destructor` is the destruction function, like `int archive_read_free(struct archive *)`, so you do:
    ```c++
    CUniquePtr<archive_read_free, ARCHIVE_OK> Ptr(archive_read_new());
    ```

include/revng/ADT/CompilationTime.h
mrjackv: The `using` below exposes it outside of the `detail::` namespace

include/revng/Support/TarIO.h
mrjackv: This is from libarchive

lib/PipeboxCommon/Python/_pipebox.cpp
mrjackv: Yes, but there's no way around it, if you return `bytes` then it needs to be copied from C

aleclearmind: Can't we return... a `Buffer`?

include/revng/ADT/CUniquePtr.h
aleclearmind: 1. Mmh, OK.
2. `auto &` feels magic, but pretty good!

include/revng/Support/TarIO.h
aleclearmind: We need to add it to `.clang-format`.

lib/PipeboxCommon/Python/_pipebox.cpp
mrjackv: That's what was done previously, but then it's annoying for basically any downstream user, e.g.
```
model_yaml = yaml.safe_load(bytes(model.serialize()))
```

Basically using `Buffer` just delays the copy, unless you have something on the other side that's `memoryview` aware (e.g. **not** yaml) then it doesn't copy

aleclearmind: I understand, I see in the CPython docs there's no "move contructor".

Add a comment about the fact that we're copying.

> model_yaml = yaml.safe_load(bytes(model.serialize()))

`yaml.safe_load` takes a `stream` as far as I can tell.

```
Help on function safe_load in module yaml:

safe_load(stream)
    Parse the first YAML document in a stream
    and produce the corresponding Python object.

    Resolve only basic YAML tags. This is known
    to be safe for untrusted input.
```

Maybe we can use use `BytesIO` with a `memoryview`?
We don't need to do this now but, if viable, it's worth to leave a comment about this could be done without a copy.

include/revng/Model/model-schema.yml
```diff
@@ -743,6 +751,13 @@ definitions:
         upcastable: true
         optional: true
 
+      - name: Binary
+        doc: The binary from where the segment is loaded from
```
aleclearmind: 1. Add trailing dot.
2. Put this as the first field.

include/revng/Model/model-schema.yml
```diff
@@ -1859,3 +1874,31 @@ definitions:
         doc: |-
           Print all the addresses exactly how disassembler emitted them
           in PC-relative mode.
+
+  - name: BinaryIdentifier
```
aleclearmind: 1. I'm OK with this for now, but please backlog renaming `model::Binary` to `model::Project` and turning `BinaryIdentifier` into just `Binary`. It's time to do this.
2. Move this before the definition of segment. Let's strive to keep things in some "reading order".

include/revng/Model/model-schema.yml
```diff
@@ -108,6 +108,14 @@ definitions:
           elementType: TypeDefinition
         optional: true
 
+      - name: Binaries
```
aleclearmind: I'd put this before `Segments`.

include/revng/TupleTree/TupleTree.h
```diff
@@ -128,10 +128,12 @@ class TupleTree {
...
-    auto Visitor = [&Map](TTR &Reference) {
-      auto It = Map.find(Reference);
-      if (It != Map.end())
-        Reference = It->second;
+    auto Visitor = [&Map]<typename TTRA>(TTRA &Reference) {
+      if constexpr (std::is_same_v<TTRA, TTR>) {
+        auto It = Map.find(Reference);
+        if (It != Map.end())
+          Reference = It->second;
+      }
```
aleclearmind, line 138: What's going on here?

lib/Model/Verification.cpp
```diff
@@ -90,6 +90,19 @@ bool Segment::verify(VerifyHelper &VH) const {
...
 
+bool BinaryIdentifier::verify(VerifyHelper &VH) const {
+  auto Guard = VH.suspendTracking(*this);
+
+  auto IsLowerHex = [](const char &C) {
+    return ('0' <= C and C <= '9') or ('a' <= C and C <= 'f');
+  };
+
+  if (Hash().size() != 64 or not llvm::all_of(Hash(), IsLowerHex))
+    return VH.fail("Invalid hash", Hash());
```
aleclearmind: "Hash has invalid format. A series of 64 lowercase hex digits representing $HASH_ALGORITHM is expected."

lib/Model/Verification.cpp
```diff
@@ -891,6 +904,18 @@ bool Binary::verify(VerifyHelper &VH) const {
     return llvm::any_of(ExecutableSegments, ContainsAddress);
   };
 
+  // Verify that there's (0, 1) binaries present
+  // TODO: remove this once multi-binary is implemented
+  if (Binaries().size() > 1)
+    return VH.fail("Binaries must either container 0 or 1 elements",
```
aleclearmind: Typo

lib/Model/Visits.cpp
```diff
@@ -7,6 +7,11 @@
 #include "revng/Model/Binary.h"
 #include "revng/TupleTree/VisitsImpl.h"
 
+//
+// To optimize compile times, each referenced type needs to provide
+// specializations for the `getByPath` function.
```
aleclearmind:
```
// To optimize compile times, we explicitly instantiate the template specializations of the relevant types, here, by hand.
```

python/revng/model/migrations/versions/v7.py
```diff
@@ -0,0 +1,10 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+from revng.model.migrations import MigrationBase
+
+
+class Migration(MigrationBase):
+    def migrate(self, model):
+        pass
```
aleclearmind, line 10: We can't just leave this like this.

We need to properly define the current semantics of `Binaries`.

AFAIU, as of now, having none is allowed but the pypeline will fail at `Lift` time, correct?

Then, once we ditch the old pipeline, this will become mandatory and then we'll have to decided we want to do with the migration.

We need to document this somewhere, probably in the doc of `Binaries`.

python/revng/pypeline/cli/pipeline/run_pipe.py
```diff
@@ -11,13 +12,26 @@
...
 from revng.pypeline.object import ObjectSet
+from revng.pypeline.storage.file_storage import FileRequest, FileStorage
 from revng.pypeline.task.pipe import Pipe
 from revng.pypeline.task.task import TaskArgumentAccess
 from revng.pypeline.utils.registry import get_registry, get_singleton
 
 logger = logging.getLogger(__name__)
 
 
+class DirectoryFileStorage(FileStorage):
```
aleclearmind: 1. Add a comment about what this is.
2. I'd call this `SimpleDirectoryFileStorage` or just `SimpleFileStorage`.

python/revng/pypeline/storage/file_storage.py
```diff
@@ -0,0 +1,19 @@
...
+class FileRequest:
+    hash: str  # noqa: A003
+    name: str | None = field(default=None)
+    size: int = field(default=-1)
+
+
+class FileStorage(ABC):
+    @abstractmethod
+    def get_files(self, requests: list[FileRequest]) -> dict[str, bytes]:
+        """Request files, return a dict <hash> -> <contents>"""
```
aleclearmind: More like `FileRetriever` or `FileProvider`?

python/revng/pypeline/cli/pipeline/run_pipe.py
```diff
@@ -139,8 +153,17 @@ def build_pipe_command(
         default=False,
         help="List the available objects for each argument.",
     )
+    @click.option(
+        "--file-storage",
+        type=click.Path(exists=True, file_okay=False, path_type=Path),
+        default=None,
```
aleclearmind: What about using `cwd`?

python/revng/pypeline/storage/file_storage.py
```diff
@@ -0,0 +1,19 @@
...
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass, field
+
+
+@dataclass
+class FileRequest:
+    hash: str  # noqa: A003
+    name: str | None = field(default=None)
+    size: int = field(default=-1)
```
aleclearmind: Why go for `-1` as opposed to `None`?

python/revng/pypeline/storage/file_provider.py
```diff
@@ -0,0 +1,19 @@
...
+@dataclass
+class FileRequest:
+    hash: str  # noqa: A003
+    name: str | None = field(default=None)
+    size: int = field(default=-1)
+
+
+class FileStorage(ABC):
+    @abstractmethod
+    def get_files(self, requests: list[FileRequest]) -> dict[str, bytes]:
```
aleclearmind, line 22: Documentation!

1. Are all the fields of `FileRequest` mandatory?
2. What is the key of the the returned dictionary? I give you a list and you return a `dict[str, ...]`?

python/revng/pypeline/storage/file_provider.py
```diff
@@ -0,0 +1,19 @@
...
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass, field
+
+
+@dataclass
+class FileRequest:
+    hash: str  # noqa: A003
```
aleclearmind, line 12: What hash is this? Docs!

python/revng/pypeline/storage/memory.py
```diff
@@ -133,3 +136,22 @@ def prune_objects(self):
         """
         self.storage.clear()
         self.dependencies.clear()
+
+    def put_files_in_file_storage(self, files: list[FileStorageEntry]) -> list[str]:
```
aleclearmind: `put_in_file_storage` or `put_file_in_storage`. Same for the getter.

python/revng/pypeline/storage/memory.py
```diff
@@ -133,3 +136,22 @@ def prune_objects(self):
...
+        result = []
+        for file in files:
+            contents = None
+            if file.contents is not None:
+                contents = file.contents
+            elif file.path is not None:
+                contents = file.path.read_bytes()
+
+            assert contents is not None
+            hash_ = hashlib.sha256(contents).hexdigest()
```
aleclearmind: Factor *all* uses of the hash so we can change it easily.

python/revng/pypeline/storage/null.py
```diff
@@ -80,3 +82,9 @@ def prune_objects(self):
         """
         Prunes all the objects (except metadata) from storage
         """
+
+    def put_files_in_file_storage(self, files: list[FileStorageEntry]) -> list[str]:
+        raise ValueError("Unsupported")
```
aleclearmind: `pass`. I'd just fail upon an attempt to get it.

python/revng/pypeline/storage/storage_provider.py
```diff
@@ -71,6 +74,16 @@ def __len__(self) -> int:
         return self.end - self.start + 1
 
 
+@dataclass
+class FileStorageEntry:
```
aleclearmind, line 81: Docs!

include/revng/PipeboxCommon/BinariesContainer.h
```diff
@@ -0,0 +1,89 @@
...
+//
+
+#include "llvm/Support/SHA256.h"
+
+#include "revng/PipeboxCommon/Common.h"
+#include "revng/Support/TarIO.h"
+
+namespace revng::pypeline {
+
+class BinariesContainer {
```
aleclearmind, line 23: 1. Does this `class` need to respect some `concept`?
   `static_assert` at the end.
2. Docs! At very least we need to document this appears as a `tar`.
   Also, document the structure within the `.tar`.
3. Let's say no to [tarbombs](https://www.linfo.org/tarbomb.html): all files in the `.tar` should be in a directory.

include/revng/PipeboxCommon/BinariesContainer.h
```diff
@@ -0,0 +1,89 @@
...
+
+public:
+  std::set<ObjectID> objects() const {
+    if (Files.empty())
+      return std::set<ObjectID>{};
+    else
+      return std::set{ ObjectID() };
+  }
+
+  void deserialize(const std::map<const ObjectID *, llvm::ArrayRef<const char>>
```
aleclearmind: 1. `ArrayRef` data is implicitly `const`.
2. Isn't `deserialize` now a `static` method?

python/revng/internal/pipebox.py
```diff
@@ -2,7 +2,73 @@
...
+        files = file_storage.get_files(requests)
+        buffer = BytesIO()
+        with tarfile.open(mode="w", fileobj=buffer) as tar:
+            for request in requests:
+                file = files[request.hash]
+
+                info = tarfile.TarInfo()
+                info.size = len(file)
+                info.name = request.hash
+                info.mode = 0o644
```
aleclearmind, line 67: Is the `TarWriter` using the same permissions?

Here we're creating a new conventional format for a `.tar` archive, for which we already have multiple implementations. We need to document very strictly and very thoroughly how it looks like.

include/revng/TupleTree/TupleTree.h
mrjackv: Before this commit it was implicitly assumed that the visitor would visit references of the same type, since the only reference present in the model was `TupleTreeReference<TypeDefinition, Binary>`, now that there's also `TupleTreeReference<BinaryIdentifier, Binary>` type of the visited reference and the replacement reference differ and therefore the lambda needs the `TTRA` type

include/revng/PipeboxCommon/BinariesContainer.h
mrjackv: 1. I'll check but I remember already going down this rabbit hole
2. `Model.deserialize` is static, `Container.deserialize` isn't

include/revng/PipeboxCommon/BinariesContainer.h
mrjackv: 1. Ok
2. Ok
3. What prefix do you have in mind? Mind that this container isn't user-facing

include/revng/TupleTree/TupleTree.h
aleclearmind: Add a comment like `Only consider references of the types we're interested in`

include/revng/PipeboxCommon/BinariesContainer.h
aleclearmind: 3. Still. `binaries/`? :man_shrugging:

include/revng/PipeboxCommon/BinariesContainer.h
aleclearmind: 1. OK, add a comment if there's a proper reason.
2. Mh, maybe we should at least leave a TODO about making it static.

lib/PipeboxCommon/Python/_pipebox.cpp
mrjackv: > yaml.safe_load takes a stream as far as I can tell.

Yes, but there are specializations for `str` and `bytes`: [link](https://github.com/yaml/pyyaml/blob/d51d8a138f7230834fc6e95635ff09ebd329185f/lib/yaml/reader.py#L72)

> Maybe we can use use BytesIO with a memoryview?

`BytesIO` works with just `bytes` objects and nothing else. That said if we really needed we could have some kind of class that conforms to `io.BufferedReader` and wraps `Buffer`

python/revng/pypeline/storage/file_provider.py
mrjackv: 1. No, just `hash`, the others are "nice to have"
2. The `hash`,  as stated in the docstring

lib/PipeboxCommon/Python/_pipebox.cpp
aleclearmind: OK, let's leave a TODO about that.

python/revng/pypeline/storage/file_provider.py
aleclearmind: 1. Docs

include/revng/Support/TarIO.h
mrjackv: Is this enough or are you thinking of a more general rule?

```
      { Regex: '^"archive\.h"$', Priority: -8 },
```

aleclearmind: We also have `archive_entry.h` AFAICT.

Let's put this after `boost` but before LLVM.

python/revng/pypeline/cli/pipeline/run_pipe.py
mrjackv: Done

lib/PipeboxCommon/Python/_pipebox.cpp
mrjackv: Added a `TODO` explaining the situation

include/revng/ADT/CUniquePtr.h
mrjackv: Went with a plain `std::unique_ptr` with a custom destructor

include/revng/ADT/CUniquePtr.h
mrjackv: This has been dropped

include/revng/ADT/CUniquePtr.h
mrjackv: This has been dropped

include/revng/ADT/CUniquePtr.h
mrjackv: It worked pretty well, now `CUniquePtr` is a `using`

include/revng/Support/TarIO.h
mrjackv: Done

include/revng/Support/TarIO.h
mrjackv: Dropped the default value

include/revng/Support/Tar.h
mrjackv: Done

lib/Support/TarIO.cpp
mrjackv: Fixed this and a couple more

lib/Support/Tar.cpp
mrjackv: Added a comment explaining what's going on

include/revng/ADT/CUniquePtr.h
mrjackv: This code has been dropped

include/revng/PipeboxCommon/Model.h
mrjackv: It was to work around how the python code was using `Model.children`. I've fixed the python code.

include/revng/Model/model-schema.yml
mrjackv: Done both

include/revng/Model/model-schema.yml
mrjackv: 1. Ok, added to my `TODO`
2. Done

include/revng/Model/model-schema.yml
mrjackv: Done

include/revng/TupleTree/TupleTree.h
mrjackv: Comment added

lib/Model/Verification.cpp
mrjackv: Done

lib/Model/Verification.cpp
mrjackv: Fixed

lib/Model/Visits.cpp
mrjackv: Fixed the comment

python/revng/pypeline/cli/pipeline/run_pipe.py
mrjackv: 1. Done
2. Renamed to `SimpleFileProvider` since the base class is now `FileProvider`

python/revng/pypeline/storage/file_storage.py
mrjackv: Went with `FileProvider`

python/revng/pypeline/cli/pipeline/run_pipe.py
mrjackv: Sure

python/revng/pypeline/storage/file_storage.py
mrjackv: No real reason, changed to `None`

python/revng/pypeline/storage/file_provider.py
mrjackv: 1. Added

python/revng/pypeline/storage/file_provider.py
mrjackv: Done

python/revng/pypeline/storage/memory.py
mrjackv: Went with `put_files_in_storage` and `get_files_from_storage`

python/revng/pypeline/storage/memory.py
mrjackv: Done

python/revng/pypeline/storage/null.py
mrjackv: Had to do a little bit more because the result is supposed to be the list of hashes, but no biggie

python/revng/pypeline/storage/storage_provider.py
mrjackv: Added docs describing what each field is needed for

python/revng/model/migrations/versions/v7.py
mrjackv: > AFAIU, as of now, having none is allowed but the pypeline will fail at Lift time, correct?

Not in this PR, but in the next PR where I'm implementing the first half of the pipeline the `Lift` pipe will check that segments map to a binary (in `checkPrecondition`)

> Then, once we ditch the old pipeline, this will become mandatory and then we'll have to decided we want to do with the migration.

Yes

> We need to document this somewhere, probably in the doc of Binaries.

So, something along the lines of "A model is valid without any `Binaries`, however do note that doing so will result in any pipe that needs to read the input binary to throw an error" in the docs?

include/revng/PipeboxCommon/BinariesContainer.h
mrjackv: 1. Done
2. Done
3. Done

include/revng/PipeboxCommon/BinariesContainer.h
mrjackv: 1. Checked, removed the `const`
2. Thought about it for a bit, I think keeping it non-static is useful for two reasons:
   1. It allows a container to be composed incrementally (e.g. if there are memory problems, a container can be loaded in "chunks")
   2. In simple cases, a pipe (even in python) can load its results through `deserialize` without having to expose additional methods (e.g. `ImportBinaries` uses `deserialize` to load the tar into the container)

python/revng/internal/pipebox.py
mrjackv: > Is the TarWriter using the same permissions?

Yes

> We need to document very strictly and very thoroughly how it looks like.

I've added some docs onto `BinariesContainer`, let me know if that's enough or it needs to be documented even more (and where)

python/revng/model/migrations/versions/v7.py
aleclearmind: > So, something along the lines of "A model is valid without any Binaries, however do note that doing so will result in any pipe that needs to read the input binary to throw an error" in the docs?

I would be more explicit about the fact that we're in a transitional phase and that this is only used in the new experimental pipeline. However, one day this will become mandatory.

include/revng/PipeboxCommon/BinariesContainer.h
```diff
@@ -0,0 +1,99 @@
...
+#include "revng/PipeboxCommon/Common.h"
+#include "revng/PipeboxCommon/Concepts.h"
+#include "revng/Support/Tar.h"
+
+namespace revng::pypeline {
+
+/// Container that stores binaries for later consumption of Pipes. The files
+/// are stored in memory. When (de)serializing the format is a plain tar
+/// (without compression) where each entry of the tar archive is named after
+/// its hash, with "binaries/" prefixed. Each file has 644 permission.
```
aleclearmind: What hash?

I'd make a bullet list of points with the characteristics.
Maybe add an example.

mrjackv: Done

python/revng/model/migrations/versions/v7.py
mrjackv: Added a long `doc` in `Binaries` explaining the current situation and the outlook, let me know if it explains everything

PR #533, opened 2025-10-13
Clift jump and loop overhaul, beautification pipe

vasama-work:


include/revng/Clift/CliftOpInterfacesJump.td
```diff
@@ -0,0 +1,32 @@
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#ifndef MLIR_CLIFT_OP_INTERFACES_JUMP
+#define MLIR_CLIFT_OP_INTERFACES_JUMP
+
+include "revng/Clift/CliftOpInterfacesLabel.td"
+
+def Clift_JumpStatementOpInterface
```
pfez, line 10: This should be a trait.

include/revng/Clift/CliftOpTraits.h
```diff
@@ -17,34 +17,6 @@ LogicalResult verifyNoFallthroughTrait(Operation *Op);
 
 } // namespace impl
 
-template<typename UseType>
-struct OneUseOfType {
```
pfez, line 21: Why did this go away?

In this way we're not verifying those properties anymore.

I understand we can't say that `MakeLabelOp` now only have one use that is a `AssignLabelOp`, but we can still say that it has at most a single use with the new `LabelAssignmentOpInterface`. That's still important.

include/revng/Clift/CliftOpInterfacesJump.td
```diff
@@ -0,0 +1,32 @@
...
+  : Clift_OpInterface<"JumpStatementOpInterface",
+                      [Clift_StatementOpInterface]> {
+
+  let methods = [
+    InterfaceMethod<
+      /*desc=*/"Returns the target label of this jump operation.",
+      "mlir::Value", "getLabel", (ins),
+      /*methodBody=*/[{}],
+      /*defaultImplementation=*/ [{}]
```
pfez, line 19: The default implementation should return the only label-typed operand.

include/revng/Clift/CliftOpInterfacesLabel.td
```diff
@@ -0,0 +1,30 @@
...
+include "revng/Clift/CliftOpInterfacesBasic.td"
+
+def Clift_LabelAssignmentOpInterface
+  : Clift_OpInterface<"LabelAssignmentOpInterface",
+                      [Clift_StatementOpInterface]> {
+
+  let methods = [
+    InterfaceMethod<
+      /*desc=*/"Returns the number of jump labels assigned by this operation.",
+      "unsigned", "getLabelAssignmentCount", (ins),
```
pfez: `getAssignedLabelCount`

include/revng/Clift/CliftOpInterfacesLabel.td
```diff
@@ -0,0 +1,30 @@
...
+      /*methodBody=*/[{}],
+      /*defaultImplementation=*/ [{}]
+    >,
+    InterfaceMethod<
+      /*desc=*/"Returns the specified jump label assigned by this operation.",
+      "mlir::Value", "getLabelAssignment", (ins "unsigned":$Index),
```
pfez: `getAssignedLabel`

include/revng/Clift/CliftOpInterfacesLabel.td
```diff
@@ -0,0 +1,30 @@
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#ifndef MLIR_CLIFT_OP_INTERFACES_LABEL
+#define MLIR_CLIFT_OP_INTERFACES_LABEL
+
+include "revng/Clift/CliftOpInterfacesBasic.td"
+
+def Clift_LabelAssignmentOpInterface
```
pfez, line 10: I think this should also have an API that returns a range of `MakeLabelOp`: `getAssignedLabels`.

include/revng/Clift/Clift.td
```diff
@@ -284,6 +295,14 @@ def Clift_AssignLabelOp : Clift_StatementOp<"assign_label"> {
 
   let extraClassDeclaration = [{
     [[nodiscard]] MakeLabelOp getLabelOp();
+
+    unsigned getLabelAssignmentCount() {
+      return 1;
+    }
+
+    mlir::Value getLabelAssignment(unsigned Index) {
```
pfez: Something is very fishy if someone passes a nonzero `Index` here.
Assert it's zero.

lib/Clift/CliftOpInterfaces.cpp
```diff
@@ -9,10 +9,24 @@
...
+#include "revng/Clift/CliftOpInterfacesJump.cpp.inc"
+
 } // namespace mlir
 
-bool mlir::clift::isLvalueExpression(mlir::Value Value) {
+namespace clift = mlir::clift;
+using namespace clift;
+
+LabelAssignmentOpInterface clift::impl::getLabelAssignment(mlir::Value Label) {
+  return Label.getDefiningOp<MakeLabelOp>().getAssignment();
```
pfez, line 28: So this crashes if `Label` is not a `MakeLabelOp`? Ok

include/revng/Clift/CliftOpTraits.td
```diff
@@ -7,10 +7,6 @@
 
 include "mlir/IR/OpBase.td"
 
-class Clift_OneUseOfType<string op> :
```
pfez, line 10: Same as another previous comment. Why did this go away?

include/revng/Clift/CliftOpInterfaces.td
```diff
@@ -5,45 +5,9 @@
...
-      /*desc=*/"Returns true if the operation represents an lvalue expression.",
-      "bool", "isLvalueExpression", (ins),
-      /*methodBody=*/[{}],
-      /*defaultImplementation=*/ [{
-        return false;
-      }]
-    >,
-  ];
-}
+include "revng/Clift/CliftOpInterfacesJump.td"
```
pfez: Why are you only including the `CliftOpInterfacesJump.td` and not the `CliftOpInterfacesLabel.td`?

include/revng/Clift/CliftOpHelpers.h
```diff
@@ -0,0 +1,162 @@
...
+    return BlockPosition{ Op->getBlock(), Op->getIterator() };
+  }
+
+  static BlockPosition getNext(mlir::Operation *Op) {
+    return BlockPosition{ Op->getBlock(), std::next(Op->getIterator()) };
+  }
+
+  static BlockPosition getEnd(mlir::Region &R) {
+    return { &R.front(), R.front().end() };
+  }
```
pfez, line 31: This assumes `R` only has a single block. Assert it.

pfez: Depending on the use cases, it could also be useful to have `getEnd(mlir::Block *B)`.

include/revng/Clift/CliftOpHelpers.h
```diff
@@ -0,0 +1,162 @@
...
+
+  if constexpr (std::is_same_v<Operation, mlir::Operation *>) {
+    return Op;
+  } else {
+    return mlir::dyn_cast<Operation>(Op);
+  }
+}
+
+template<typename OpT = mlir::Operation *, typename PredicateT>
+OpT getLeadingOp(mlir::Region &Region, PredicateT &&Predicate) {
```
pfez: Rename this to `getFirstOpIf`.

include/revng/Clift/CliftOpHelpers.h
```diff
@@ -0,0 +1,162 @@
...
+  return {};
+}
+
+template<typename OpT = mlir::Operation *>
+OpT getLeadingOp(mlir::Region &Region) {
+  return getLeadingOp<OpT>(Region, [](OpT) { return true; });
+}
+
+template<typename OpT = mlir::Operation *, typename PredicateT>
+OpT getTrailingOp(mlir::Region &Region, PredicateT &&Predicate) {
```
pfez: Rename this to `getLastOpIf`

include/revng/Clift/CliftOpHelpers.h
```diff
@@ -0,0 +1,162 @@
...
+}
+
+inline YieldOp getYieldOp(mlir::Region &R) {
+  return getTrailingOp<YieldOp>(R);
+}
+
+template<typename... ArgsT>
+StatementOpInterface getTrailingStatement(mlir::Region &R, ArgsT &&...Args) {
+  return getTrailingOp<StatementOpInterface>(R, std::forward<ArgsT>(Args)...);
+}
```
pfez: I don't think we need a variadic version of this.

include/revng/Clift/CliftOpHelpers.h
```diff
@@ -0,0 +1,162 @@
...
+
+inline StatementOpInterface getTrailingJumpOp(mlir::Region &R) {
+  return getTrailingOp<StatementOpInterface>(R, [](auto Op) {
+    return Op->template hasTrait<mlir::OpTrait::clift::NoFallthrough>();
+  });
+}
```
pfez: This is more like `getLastNoFallthroughStatement`.

include/revng/Clift/CliftOpInterfacesJump.td
vasama-work: Why do you consider a trait more appropriate? I think interface makes sense here, so that it's possible to treat these operations generically.

include/revng/Clift/CliftOpTraits.h
vasama-work: Oh, I just moved the verification into `MakeLabelOp::verify`.

include/revng/Clift/CliftOpInterfacesJump.td
vasama-work: It's better not to provide a default implementation here, because all implementing operations already have a `getLabel()` member function which gets used to implement this. An explicit default here would never get used.

lib/Clift/CliftOpInterfaces.cpp
vasama-work: If *not defined by* a `MakeLabelOp`, but yes.

include/revng/Clift/CliftOpTraits.td
vasama-work: The answer is the same. :)

include/revng/Clift/CliftOpInterfaces.td
vasama-work: The others are included transitively.

include/revng/Clift/CliftOpHelpers.h
vasama-work: Perhaps. Currently we never allow multiple blocks in any region.

include/revng/Clift/CliftOpHelpers.h
vasama-work: I can introduce two overloads `getLastStatement` and `getLastStatementIf` instead.

include/revng/Clift/CliftOpHelpers.h
vasama-work: I added the assert. Also added `getBegin(mlir::Region &)`.

include/revng/Clift/CliftOpInterfacesLabel.td
vasama-work: Since that range type would be fairly complicated, I think if and when it's needed, it'll be best to do it as a free function in `CliftOpHelpers.h`, say.

include/revng/Clift/CliftOpInterfaces.td
```diff
@@ -8,6 +8,6 @@
 // The MLIR tablegen tool emits declarations in alphabetical order, causing
 // incorrect emission of interface hierarchies.
 
-include "revng/Clift/CliftOpInterfacesJump.td"
+include "revng/Clift/CliftOpInterfacesFlow.td"
```
pfez: I'm puzzled by the `Flow` naming.

So you want to put here interfaces that have to do with "flow" in what sense?
From what I see, it seems to be control flow, but also e.g. `goto`s deal with control flow.
So perhaps "structured control flow"?

I'd just merge the various `CliftOpInterfacesJump` and `CliftOpInterfacesFlow` files into `CliftOpInterfacesControlFlow`.

pfez: Perhaps even the `CliftOpInterfacesLabel` could be merged in there.

In this way all the interfaces that have to do with control flow in any way are all in the same place.

include/revng/Clift/CliftOpInterfacesJump.td
pfez: I think that it would be better for an operation to only override the default if it does something different than what all the "regular" operation with that interface do. In this case they all do the same thing. So I'd drop the `getLabel` implementations in the individual ops and let all of them just use the default.

include/revng/Clift/CliftOpInterfacesJump.td
pfez: You should be able to treat these operations generically anyway, as explained [here](https://mlir.llvm.org/docs/Traits/#using-a-trait). Of course that's not truly generic, but it achieves the goals I have in mind.

In general, traits are simpler and less powerful. All traits can also be implemented as interfaces. But if a feature is simple and can be implemented using a trait I think we should prefer it.

lib/Clift/CliftOpInterfaces.cpp
pfez: Nevermind I can't read.

include/revng/Clift/CliftOpInterfacesJump.td
vasama-work: > To expose this information opaquely to transformations and analyses, [interfaces](https://mlir.llvm.org/docs/Interfaces/) may be used.

By converting it to a trait we lose the opaque `getLabel()` and the ability to use `isa`, `dyn_cast`, etc. So no, the same thing cannot be implemented as a trait. Speaking of, I think the `getLabel()` discussion below is relevant...

include/revng/Clift/CliftOpInterfacesJump.td
vasama-work: Okay. Let me clarify. The `getLabel()` functions of the individual ops are not going away. They are present because each of those operations has an operand named `label`. I see where you're going, and your intuition for how to structure this would be fine for a regular C++ virtual interface, but the fact is that things work a little differently in MLIR.

include/revng/Clift/CliftOpInterfaces.td
vasama-work: Ah. This is the same issue we had with the type interfaces. I would be perfectly happy to define all the operation interfaces in the same file, being that there are not so many of them, but that does not work. You see tablegen emits the C++ code for the interfaces in *alphabetical* order, meaning that in many cases a derived interface is emitted before its parent. The definitions are split up into multiple files in order to avoid this.

share/revng/pipelines/revng-pipelines.yml
```diff
@@ -692,7 +692,7 @@ Branches:
               - materialize-trivial-goto
           - Type: llvm-to-clift
             UsedContainers: [functions.bc.zstd, functions.mlir]
-          - Type: clift-legalization
+          - Type: clift-beautification
```
pfez: Let's call this `clift-optimization`.
In general let's not use `beautification` at all anymore. Let's use "optimization" everywhere instead.

include/revng/Clift/Clift.h
```diff
@@ -23,6 +23,9 @@
 
 namespace mlir::clift::impl {
 
+inline constexpr unsigned BreakLabelFlag = 1;
+inline constexpr unsigned ContinueLabelFlag = 2;
```
pfez: If we're going to use this as a flag, which is what we do, let's use `1 << 1` instead of `2`.

share/revng/pipelines/revng-pipelines.yml
pfez: And I mean this across all the clift-related codebase. We can just ignore the legacy stuff. It will go away when we drop it.

include/revng/Clift/Clift.td
```diff
@@ -248,15 +248,42 @@ def Clift_GlobalVariableOp : Clift_Op<"global",
...
+  : Clift_BasicStatementOp<mnemonic, Clift_LoopOpInterface, traits> {
+
+  let arguments = (ins LoopLabelMaskAttr:$label_mask,
+                       Variadic<Clift_LabelType>:$labels);
+
+  let extraClassDeclaration = [{
+    unsigned getLabelAssignmentCount() {
+      // Add up the total number of set flags [0, 2].
+      unsigned M = getLabelMask();
+      return (M & 1) + (M >> 1);
```
pfez: Don't we have `std::popcount`?

include/revng/Clift/Clift.td
```diff
@@ -248,15 +248,42 @@ def Clift_GlobalVariableOp : Clift_Op<"global",
...
+// the break and continue label assignments respectively.
+def LoopLabelMaskAttr : TypedUnsignedIntegerAttrBase<UI<2>,
+                                                     "uint8_t",
+                                                     "Loop label mask">;
+
+class Clift_LoopStatementOp<string mnemonic, list<Trait> traits = []>
+  : Clift_BasicStatementOp<mnemonic, Clift_LoopOpInterface, traits> {
+
+  let arguments = (ins LoopLabelMaskAttr:$label_mask,
+                       Variadic<Clift_LabelType>:$labels);
```
pfez, line 275: Can't we get away without `LoopLabelMaskAttr`?
I assume this was designed not to pass `undef` labels as arguments.

I don't like having that explicitly because it can go out of sync, unless we have a `verify` for and I haven't seen it yet.

Instead of the attribute, we could have 2 mandatory arguments, and implicitly assume the first is always the `continue` and the second the `break`.
Then, when either is not defined, which you can easily detect when parsing textual assembly, we just leave it "null".
And when you print you should have no problems in just skipping the "null" label arguments, being them `continue` or `break`.
Not sure MLIR supports having "null" mandatory arguments.

Otherwise, I have an half baked idea of a more radical change I'm just throwing out there.
We could turn `break` and `continue` into 2 regions of the loop (instead of arguments).
Then each region either has an optional argument (when e.g. the `continue` label is implicit vs explicit), or we can even make the regions themselves optional, while keeping the label argument of each region mandatory.
Having optional arguments on mandatory regions is probably more comfortable if you want to build a loop, and add break/continue labels later I guess, or remove labels from existing loops, which is something we plan to do often.

What do you think?

include/revng/Clift/Clift.td
```diff
@@ -442,12 +513,17 @@ def Clift_SwitchOp : Clift_BasicStatementOp<"switch", Clift_BranchOpInterface> {
...
 
-def Clift_WhileOp : Clift_StatementOp<"while"> {
+def Clift_WhileOp : Clift_LoopStatementOp<"while", [NoRegionArguments]> {
   let regions = (region Clift_ExpressionRegion:$condition,
                         Clift_StatementRegion:$body);
 
+  let builders = [
+    OpBuilder<(ins)>
+  ];
+
```
pfez, line 511: These builders you have for all loops now, am I correct in assuming they are for building a loop with explicit `break`/`continue` labels?

lib/Clift/Clift.cpp
```diff
@@ -477,14 +490,113 @@ mlir::LogicalResult GlobalVariableOp::verify() {
...
+  if (Parser.parseOptionalKeyword("continue").succeeded()) {
+    UnresolvedOperand Operand;
+    if (Parser.parseOperand(Operand).failed())
+      return mlir::failure();
+    Labels.push_back(Operand);
+    Mask |= clift::impl::ContinueLabelFlag;
+  }
+
+  LabelMask = makeLoopLabelMask(Parser.getContext(), Mask);
+
```
pfez, line 529: Note to self, to come back to this after we've settled the discussion on the label mask attribute.

lib/Clift/Clift.cpp
```diff
@@ -477,14 +490,113 @@ mlir::LogicalResult GlobalVariableOp::verify() {
...
+  }
+
+  if (not Loop->isAncestor(getOperation())) {
+    return emitOpError() << getOperationName()
+                         << " must target a nesting loop label.";
+  }
+
+  return mlir::success();
+}
+
```
pfez, line 598: Nice :)

lib/CliftEmitC/CBackend.cpp
```diff
@@ -994,17 +1040,21 @@ class CliftToCEmitter : CEmitter {
...
 
   static bool mayElideBraces(mlir::Region &R) {
     mlir::Operation *OnlyOp = getOnlyOperation(R);
     return OnlyOp != nullptr and mayElideBraces(OnlyOp);
   }
 
-  RecursiveCoroutine<void> emitImplicitBlockStatement(mlir::Region &R,
-                                                      bool EmitBlock) {
+  RecursiveCoroutine<bool>
+  emitImplicitBlockStatement(mlir::Region &R, bool EmitBlock, auto EmitRegion) {
```
pfez, line 1071: So `EmitRegion` is more like `EmitAtBlockEnd`?

include/revng/Clift/Clift.td
vasama-work: With the new mask attribute and the associated operands, it is no longer possible to create a loop operation without specifying the mask and operands as it was before. These builders just give you a loop with no assigned labels.

lib/CliftEmitC/CBackend.cpp
vasama-work: Not as currently defined. That would probably solve this problem too, but I'm not keen on overly specific APIs when the more general alternative is no more complicated or even simpler.

include/revng/Clift/Clift.td
vasama-work: The only way we can have optional operands is if the set of operands is variadic. The mask then tells you what those variadic operands mean. I suspect your concerns would be solved by adding more functions to mutate the loop labels safely. I think there's also a way to disable the default builder so one cannot create a loop with explicit mask and operands.

I think you'll have to elaborate on this half baked idea of yours. It's not clear to me. The loop op must use the labels it wants to assign. Region arguments are introducers of SSA values, not users. We could use two regions and those regions could contain `assign_label` ops for the loop labels, but that would just be horribly complicated and probably more likely to go out of sync.

vasama-work: You're right that we're currently missing verification for this. I'll add verification for checking that the mask's bitcount matches the number of operands.

include/revng/Clift/Clift.td
vasama-work: Yeah, we do. I was kind of being paranoid about only counting the low two bits, but I think I wrote this before I changed the type of the attribute from a 64-bit integer to a 2-bit integer.

include/revng/Clift/Clift.td
vasama-work: I added that verification as a trait, and moved the label access interface to the trait as well. Moving those functions will be important later due to a limitation of MLIR, in order to avoid duplication when more member functions are added to the different concrete loop operations (each `extraClassDeclaration` completely replaces that of the parent).

(In the optimisation branch there are functions related to discarded and boolean-tested expressions.)

include/revng/Clift/CliftOpInterfaces.td
vasama-work: As an alternative, we could patch LLVM, and then just define all the interfaces in a single file. That might be as simple as swapping a `std::map` for `MapVector`, might not - especially because the `MapVector` interface in our current version of LLVM does not support the full standard associative container interface.

pfez: > You see tablegen emits the C++ code for the interfaces in alphabetical order, meaning that in many cases a derived interface is emitted before its parent. The definitions are split up into multiple files in order to avoid this.

Right. Let's not patch LLVM. This explanation is enough.

include/revng/Clift/Clift.td
pfez: Ok so let's go with `popcount`.

include/revng/Clift/CliftOpInterfaces.td
pfez: I'm still puzzled by the `Flow` naming. `ControlFlow` is better I think. Not great, but given that you have to chop up the interfaces in a bunch of files we have to accept some level of compromise.

include/revng/Clift/CliftOpTraits.h
```diff
@@ -25,6 +26,26 @@ class NoFallthrough : public OpTrait::TraitBase<ConcreteType, NoFallthrough> {
   }
 };
 
+template<typename ConcreteType>
+class AssignsLoopLabels
```
pfez, line 30: Nice!

lib/Clift/CliftOpTraits.cpp
```diff
@@ -19,3 +19,29 @@ LogicalResult traits_impl::verifyNoFallthroughTrait(Operation *const Op) {
   }
   return mlir::success();
 }
+
+LogicalResult traits_impl::verifyAssignsLoopLabelsTrait(Operation *const Op) {
```
pfez, line 23: Remember to add tests for this.

include/revng/Clift/CliftOpTraits.h
pfez: Right. I thought we could have some use case for `OneUseOfType` for other stuff, but actually we currently don't and I don't see other obvious cases coming up in the future.

Goodbye `OneUseOfType`.

include/revng/Clift/Clift.td
pfez: > I suspect your concerns would be solved by adding more functions to mutate the loop labels safely.

This!

> I think there's also a way to disable the default builder so one cannot create a loop with explicit mask and operands.

And this!

Basically, at the structural level I understand the need for the attribute. But programmatically I want to disable as much as possible the access to the attribute itself. The only way to affect the attribute should be methods (or builders) that automatically and correctly infer the attribute from the arguments.

> I think you'll have to elaborate on this half baked idea of yours. It's not clear to me. 

It wasn't entirely clear to me either. I was just throwing that around to see it could go somewhere without violating constraints imposed by MLIR. But it's not necessary, the other changes you've proposed achieve the same results without having to come up with weird constructs.

lib/CliftEmitC/CBackend.cpp
pfez: Fine let's keep it this way.

lib/Clift/CliftOpTraits.cpp
vasama-work: Done.

include/revng/Clift/Clift.td
vasama-work: Done. The loop op interface now has `setBreakLabel` and `setContinueLabel`. The default builders are gone.

include/revng/Clift/CliftOpInterfacesControlFlow.td
```diff
@@ -42,6 +42,14 @@ def Clift_LoopOpInterface
         return mlir::clift::impl::getLoopLabel($_op, 0);
       }]
     >,
+    InterfaceMethod<
+      /*desc=*/"Replaces the loop break label (if any).",
+      "void", "setBreakLabel", (ins "mlir::Value":$Label),
```
pfez, line 47: We're also going to need `clearBreakLabel` and `clearContinueLabel` when enabling unlabeled `break`/`continue`.

include/revng/Clift/Clift.td
```diff
@@ -273,6 +273,8 @@ class Clift_LoopStatementOp<string mnemonic, list<Trait> traits = []>
 
   let arguments = (ins LoopLabelMaskAttr:$label_mask,
                        Variadic<Clift_LabelType>:$labels);
+
+  let skipDefaultBuilders = 1;
```
pfez, line 280: Ok, But add a comment to explain *why* we're opting out of these.

include/revng/Clift/CliftOpInterfacesControlFlow.td
vasama-work: Ah. I made it so that specifying a null value removes the operand.

pfez: Fine

PR #536, opened 2025-10-28
Implement first half of the pipeline in pypeline

mrjackv:


python/revng/pypeline/cli/pipeline/run_pipe.py
```diff
@@ -158,10 +158,7 @@ def run_pipe_command(
         logger.debug("and kwargs: %s", kwargs)
 
         # Create the pipe
-        pipe = pipe_type(
-            name=pipe_name,
-            static_configuration=static_configuration,
-        )
+        pipe = pipe_type(static_configuration)
```
aleclearmind, line 159: I guess the goal was to give name to things that can have wildly different meanings depending on the static configuration, e.g., `LLVMPipe`.

But I'm no necessarily against this.

python/revng/pypeline/container.py
```diff
@@ -188,7 +188,10 @@ def to_string(self) -> str:
         Dump a container into a serialized format.
         """
         return json.dumps(
-            {k.serialize(): bytes(v).hex() for k, v in self.serialize(self.objects()).items()},
+            {
+                k.serialize(): bytes(v).decode() if self.is_text() else base64.b64encode(v).decode()
```
aleclearmind: Why are you not specifying `utf-8` here?

python/revng/pypeline/schema.yml
```diff
@@ -76,10 +76,8 @@ $defs:
         description: The name of the pipe.
         example: my_pipe
       configuration:
-        type: string
+        type: [string, object]
         description: The static configuration for the pipe, if any.
```
aleclearmind: Maybe add some docs about what's going to happen to this field if it's not a `string`.

python/revng/pypeline/pipeline_parser.py
```diff
@@ -55,9 +55,16 @@ def parse_pipe(
...
         bindings.append(container_decls[arg])
 
-    configuration = task.get("configuration", "")
+    configuration = task.get("configuration")
+    if configuration is None:
+        configuration_string = ""
+    elif isinstance(configuration, str):
+        configuration_string = configuration
+    else:
+        configuration_string = yaml.safe_dump(configuration)
```
aleclearmind, line 65: Shall we use JSON? This serialization is almost never exposed to humans, it'd make sense to relax pipes to have JSON parser instead of a YAML parser. What do you think=

python/revng/pypeline/pipeline.py
```diff
@@ -57,16 +57,16 @@ class Pipeline(Generic[C]):
...
     """
 
-    __slots__ = ("declarations", "root", "artifacts", "analyses")
+    __slots__ = ("declarations", "roots", "artifacts", "analyses")
 
     def __init__(
         self,
         declarations: Set[ContainerDeclaration],
-        root: PipelineNode,
+        roots: list[PipelineNode],
```
aleclearmind: I'm not sure this is beneficial.
We lose the property of having a tree (now we have a forest) for what in return?

We can leave the pipeline as it is and have the code building the pipeline from the YAML adding a null pipe (possibly only if needed).

What do you think?

python/revng/pypeline/task/pipe.py
```diff
@@ -83,15 +83,7 @@ def prerequisites_for(
...
                 continue
 
-            for ridx, object_list in enumerate(requests):
-                # The requests must be for the writeable containers,
-                # otherwise it's not possible to satisfy them
-                if self.arguments[ridx].access == TaskArgumentAccess.READ:
-                    assert len(object_list) == 0, (
-                        f"Expected an empty request for {self.arguments[ridx].name}, "
-                        f"but got {object_list}."
-                    )
```
aleclearmind, line 97: I'm confused by this assertion.
But also the previous check confuses me.

> We must fill the readable containers

What does that mean?
Are you the author of this code?

The original code was just:

```python
    @final
    def prerequisites_for(
        self, model: ReadOnlyModel, requests: Requests, configuration_id: ConfigurationId
    ) -> Requests:
        result = Requests()

        # WIP (?): ensure that return an empty set of requests does not prevent it from running

        # Cross-contaminate inputs and outputs
        for input_declaration in self.inputs:
            for _, object_list in requests.items():
                result.update(
                    input_declaration,
                    model.move_to_kind(object_list, input_declaration.container_type.kind),
                )

        return result
```

include/revng/PipeboxCommon/Helpers/PipeRunPipes/Base.h
```diff
@@ -34,6 +34,21 @@ constexpr size_t writableContainerIndex() {
...
+
+template<typename T>
+struct CheckPreconditionMixin {};
+
+template<HasPipeRunCheckPrecondition T>
+struct CheckPreconditionMixin<T> {
+  llvm::Error checkPrecondition(const Model &Model) const {
+    return T::checkPrecondition(Model);
+  }
+};
```
aleclearmind, line 59: This are used in a single place. Inline them by simply adding

```c++
if constexpr (HasPipeRunCheckPrecondition<T>) {
  T::checkPrecondition(Model)
}
```

include/revng/PipeboxCommon/Helpers/Registrars.h
```diff
@@ -111,6 +111,14 @@ struct RegisterPipe {
              "incoming"_a,
              "outgoing"_a,
              "configuration"_a);
+
+      if constexpr (HasCheckPrecondition<T>) {
```
aleclearmind, line 144: I'm not sure why we're going the extra mile to make the `checkPrecondition` method optional in `SingleOutputPipeBase`.

It's not something the user will write, let's just always implement it and, in case the wrapper pipe run does not offer it, let's just return `true`.

lib/Support/CMakeLists.txt
```diff
@@ -35,6 +36,26 @@ if(NOT LibArchive_FOUND)
...
+  Passes
+  ScalarOpts
+  Support
+  Target
+  TransformUtils
+  Vectorize
+  X86AsmParser
+  X86CodeGen
+  X86Desc
+  X86Info)
```
aleclearmind, line 61: Shouldn't we have corresponding change removing these libs from `librevngPipes`?

include/revng/PipeboxCommon/Common.h
```diff
@@ -25,14 +25,33 @@ using ModelPath = std::string;
 using ObjectDependencies = std::vector<
   std::vector<std::pair<ObjectID, ModelPath>>>;
 
-template<ConstexprString N, ConstexprString HT>
-struct PipeArgumentDocumentation {
+enum class Access {
+  // The requested container will only be read (request objects as a dependency)
```
aleclearmind: 1. Use Doxygen
2. Add hints about:
   1. Even if you pick `Read`, the argument (in exceptional cases) can be non-`const`.
   2. You can have *at most* one argument that is `Write`/`ReadWrite`. We will commit the targets on it.

lib/Support/IRHelpers.cpp
```diff
@@ -709,3 +711,15 @@ cloneFiltered(const llvm::Module &Module,
 
   return Cloned;
 }
+
+std::unique_ptr<llvm::Module> copyModule(const llvm::Module &Module,
```
aleclearmind: 1. `cloneInto` or `cloneIntoContext`
2. Can you compare your implementation with `llvm::orc::cloneToNewContext` and see if we can steal some idea?

include/revng/PipeboxCommon/Helpers/PipeRunPipes/FunctionPipe.h
```diff
@@ -40,11 +42,16 @@ class FunctionPipe : public SingleOutputPipeBase<T> {
...
+    auto &RequestedFunctions = Outgoing.at(this->OutputContainerIndex);
 
-    for (const ObjectID *Object : Outgoing.at(this->OutputContainerIndex)) {
+    llvm::Task FunctionTask(RequestedFunctions.size(),
+                            "Running pipe " + this->Name + " on functions");
+    for (const ObjectID *Object : RequestedFunctions) {
       auto Committer = ODH.getCommitterFor(*Object, this->OutputContainerIndex);
 
       const MetaAddress &Entry = std::get<MetaAddress>(Object->key());
+      FunctionTask.advance(Entry.toString(), true);
```
aleclearmind: Try to keep `.advance` as early as possible in the loop.

include/revng/Lift/Lift.h
```diff
@@ -172,3 +175,22 @@ class LiftPass : public llvm::ModulePass {
...
   bool runOnModule(llvm::Module &M) override;
 };
+
+namespace revng::pypeline::pipes {
+
+class Lift {
+public:
+  static constexpr llvm::StringRef Name = "Lift";
+  using Arguments = TypeList<PipeArgument<"Input", "">,
+                             PipeArgument<"Output", "", Access::Write>>;
```
aleclearmind: 1. Do not leave documentation empty. Never!
2. Add a comment on why we use `Access::Write` as opposed to `Auto`. Do this everywhere we don't use `Auto`.

include/revng/Model/RawBinaryView.h
```diff
@@ -176,6 +176,25 @@ class RawBinaryView {
...
+public:
+  // Factored-out checkPrecondition, should be added to each pipe that uses
+  // this class
+  static llvm::Error checkPrecondition(const model::Binary &Binary) {
+    llvm::SmallVector<llvm::Error, 0> Errors;
+    if (Binary.Binaries().size() != 1)
+      Errors.push_back(revng::createError("Binaries should contain exactly 1 "
+                                          "entry"));
+
+    for (const model::Segment &Segment : Binary.Segments()) {
```
aleclearmind: Add a comment about the fact that eventually this will be part of `Binary::verify`.

python/revng/internal/pipeline.yml
```diff
@@ -5,9 +5,13 @@
...
+  - name: llvm-root
+    type: LLVMRootContainer
 
 branches:
   isolate:
     tasks:
       - pipe: ImportFiles
         arguments: [binaries-container]
+      - pipe: Lift
+        arguments: [binaries-container, llvm-root]
```
aleclearmind: I love how this branch grows the pipeline little by little.

lib/Lift/LiftCheckPrecondition.h
aleclearmind, line 1: Can we try not to have this header file?

include/revng/Pipebox/LLVMPipe.h
```diff
@@ -0,0 +1,82 @@
...
+#include "revng/PipeboxCommon/LLVMContainer.h"
+#include "revng/PipeboxCommon/Model.h"
+
+namespace revng::pypeline::pipes {
+
+namespace detail {
+
+class LLVMPipeBase {
+public:
+  using Arguments = TypeList<PipeArgument<"container", "">>;
```
aleclearmind: In `Lift` arguments names are capitalized. Let's pick one convention and stick to it.

include/revng/Pipebox/LLVMPipe.h
```diff
@@ -0,0 +1,82 @@
...
+
+  struct Configuration {
+    std::vector<std::string> Passes;
+    static Configuration parse(llvm::StringRef Input);
+  };
+
+private:
+  llvm::legacy::PassManager PassManager;
```
aleclearmind: This might be a thread-safety liability.
I'd replace this with a `vector<PassInfo *>` and rebuild the pass manager at every run.

include/revng/Pipebox/LLVMPipe.h
```diff
@@ -0,0 +1,82 @@
...
+
+    revng_assert(Outgoing[0].size() == 1);
+    revng_assert(Outgoing[0][0]->kind() == Kinds::Binary);
+
+    getPassManager().run(Container.getModule());
+    return {};
+  }
+};
+
+class LLVMPipe : public detail::LLVMPipeBase {
```
aleclearmind: Maybe a better name would be `LLVMPassManagerPipe` or `PureLLVMPassManagerPipe` or `PureLLVMPassesPipe`.
Otherwise by the name it's not obvious that you're supposed to only run "pure LLVM" passes in here and therefore you don't need to commit anything.

Also, add some docs about the fact that these do not commit anything.

include/revng/Pipebox/TupleTreeContainer.h
```diff
@@ -0,0 +1,56 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "revng/PipeboxCommon/Common.h"
+#include "revng/PipeboxCommon/ObjectID.h"
+
+namespace revng::pypeline {
+
+template<TupleTreeCompatible T, Kind TheKind, ConstexprString TheName>
+class TupleTreeContainer {
```
aleclearmind, line 13: Very cool

lib/FunctionIsolation/IsolateFunctions.cpp
```diff
@@ -587,71 +548,52 @@ void IsolateFunctionsImpl::run() {
...
-  // Obtain the set of requested targets
-  pipeline::ExecutionContext &Context = *LECP.get();
-  const pipeline::TargetsList &RequestedTargets = LECP.getRequestedTargets();
-
-  Task IsolateTask(RequestedTargets.size(), "Isolating functions");
-  for (const pipeline::Target &Target : RequestedTargets) {
-    IsolateTask.advance(Target.toString(), true);
-    Context.getContext().pushReadFields();
+  T->advance("Create dynamic functions", true);
+}
```
aleclearmind: Uh?

lib/FunctionIsolation/IsolateFunctions.cpp
```diff
@@ -549,13 +510,13 @@ void IsolateFunctionsImpl::run() {
   //
   // TODO: we can (and should) push processing of dynamic functions into the
   //       loop emitting individual local functions, and make it lazy
-  T.advance("Create dynamic functions", true);
-  Task DynamicFunctionsTask(Binary.ImportedDynamicFunctions().size(),
-                            "Dynamic functions creation");
+  T->advance("Create dynamic functions", true);
+  DynamicFunctionsTask.emplace(Binary.ImportedDynamicFunctions().size(),
```
aleclearmind: I don't see why this should become a `class` member.

lib/FunctionIsolation/IsolateFunctions.cpp
```diff
@@ -207,24 +209,24 @@ class IsolateFunctionsImpl {
   std::map<MetaAddress, Function *> IsolatedFunctionsMap;
   std::map<StringRef, Function *> DynamicFunctionsMap;
 
-  ControlFlowGraphCache *Cache = nullptr;
+  CFGGetterType CFGGetter;
   FunctionType *IsolatedFunctionType = nullptr;
-  pipeline::LoadExecutionContextPass &LECP;
+
+  std::optional<Task> T;
```
aleclearmind: Drop `T`. If you reorganize the function layout, you oftentimes need to reorganize the tasks as well.

If you want, you can have a task for prologue, runOnFunction and epilogue, and then, if you want to, you can have subtasks for each of the above.

Cross-function `Task`s it not a great pattern

lib/FunctionIsolation/IsolateFunctions.cpp
```diff
@@ -749,17 +691,86 @@ bool IF::runOnModule(Module &TheModule) {
...
+
+  auto CFGGetter =
+    [&CFG](const MetaAddress &Address) -> const efa::ControlFlowGraph & {
+    return *CFG.getElement(ObjectID(Address));
+  };
+
+  Impl = std::make_unique<IFI>(Module.getFunction("root"),
+                               GCBI,
+                               *Model.get().get(),
+                               CFGGetter);
```
aleclearmind, line 750: Outline all of this into a `static` method returning an `IFI` so you can avoid the `unique_ptr` and you can do

```c++
Isolate::Isolate(const class Model &Model,
                 llvm::StringRef Config,
                 llvm::StringRef DynamicConfig,
                 const CFGMap &CFG,
                 LLVMRootContainer &Root,
                 LLVMFunctionContainer &Output) :
  Root(Root), Output(Output), GCBI(*Model.get().get()), Impl(myStaticMethod()) {}
```

If you really want to keep things this way, switch from `unique_ptr` to `optional`.

lib/EarlyFunctionAnalysis/AttachDebugInfo.cpp
```diff
@@ -321,3 +345,35 @@ struct AttachDebugInfoToABIEnforcedPipe {
...
+      revng_assert(LLVMFunction == nullptr);
+      LLVMFunction = &Function;
+    }
+  }
+  revng_assert(LLVMFunction != nullptr);
+
+  ::AttachDebugInfo Impl(Binary, Module, GCBI, CFGGetter);
+  Impl.prologue();
+  Impl.runOnFunction(TheFunction, *LLVMFunction);
+  Impl.epilogue();
```
aleclearmind, line 377: Add a `TODO` about the fact that, once we ditch the new pipeline, `Impl` should be inlined in here.

lib/FunctionIsolation/IsolateFunctions.cpp
```diff
@@ -749,17 +691,86 @@ bool IF::runOnModule(Module &TheModule) {
...
+  Impl->prologue();
```
aleclearmind, line 751: Add a `TODO` about the fact that, once we ditch the new pipeline, `Impl` should be inlined in here. Same elsewhere.

python/revng/internal/pipeline.yml
```diff
@@ -44,6 +44,12 @@ branches:
           - name: isolate
             container: llvm-functions
 
+  enforce-abi:
+    from: isolate
```
aleclearmind, line 75: Why a new branch?

python/revng/pypeline/container.py
mrjackv: `bytes.decode` defaults to `utf-8` by defalt, the signature from the docs is:
```
bytes.decode(encoding='utf-8', errors='strict')
```

I can add it if you want to be extra pedantic

python/revng/pypeline/pipeline_parser.py
mrjackv: I'd stick with YAML for the simple reason that "on the other side" we have LLVM, and handling things with YAMLTraits is way easier that having to deal with JSON. Also YAML is quite common, so I don't expect downstream users having some problem using an alternative YAML library (in the C++ context, they would have needed a JSON one anyways) to parse the configuration

python/revng/pypeline/pipeline.py
mrjackv: Having a null pipe seems really a hack to me. What's the benefit in representing the pipeline as a tree? All the functions rely on `walk_pipeline` anyways so it's not like there's some intrinsic property of a tree that is being lost

python/revng/pypeline/container.py
aleclearmind: As you wish, but then do the same for `str.encode`.

python/revng/pypeline/pipeline_parser.py
aleclearmind: `YAMLTraits` should handle JSON just fine.

python/revng/pypeline/task/pipe.py
mrjackv: >  Are you the author of this code?

No (I think you wrote this)

> What does that mean?

Basically, when running the chain of `prerequisites_for`, this assertion would be hit if there were two consumers of the same container. Take for example:

* Containers C1, C2, C3 (for the sake of example, let's assume they are all kind=binary)
* Pipes:
  * Pipe P1: writes C1
  * Pipe P2: reads C1, writes C2
  * Pipe P3: reads C1, writes C3
 
In the following order `C1 -> C2 -> C3`

So, when calculating dependencies in `C3:/binary`, this happens:
* P3.prerequisites_for is called with `{C3: '/binary'}`, returns `{C1: '/binary'}`
* P2.prerequisites_for is called with `{C1: '/binary'}`, the assertion is hit because P2 does not produce C1, however in this case it just needs to forward it back. Hypotetically it returns `{C1: '/binary'}`
* Hypotetically, P1.prerequisites_for is called with `{C1: '/binary'}`, it returns the empty set because it produces C1

include/revng/PipeboxCommon/Helpers/PipeRunPipes/Base.h
mrjackv: The idea here is that, by using the mixin and doing `if constexpr (HasCheckPrecondition<T>)` in `Registrars.h`, then the Python/C++ barrier is never crossed and pipeline just calls `Pipe.check_precondition` which is a no-op

python/revng/pypeline/pipeline.py
aleclearmind: I find it conceptually easier to think about a tree. It's a matter of preference and simplicity.

When we dump the pipeline to Graphviz what do we do? Do we pretend it's a tree or do we dump it as is?

include/revng/PipeboxCommon/Helpers/Registrars.h
mrjackv: See answer in the previous comment

lib/Support/CMakeLists.txt
mrjackv: ~~I'll check~~
Fixed

include/revng/Lift/Lift.h
mrjackv: 1. I'm not entirely sure what to write, other than obvious things like `The input binaries that will be lifted` and `The output lifted llvm module`
2. `Write` is quite normal IMHO, it just means that a pipe produces the objects in the container for the first time. (I'm totally OK on the other hand in documenting the spots where we have `Access::Read`)

lib/Lift/LiftCheckPrecondition.h
mrjackv: It's just to not having to duplicate code between `Lift.cpp` and `LiftPipe.cpp`. It's a temporary measure

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: I'm not sure what your doubt here is.
This code is exactly the same (in terms of functionality) as what's currently on `develop`

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: `Task T` is a parent of `DynamicFunctionsTask` and the `Task` object created by the pipewrapper
If `DynamicFunctionsTask` becomes a local variable to the functions it induces a dangling pointer in `Task T` pointing to the now-deleted `DynamicFunctionsTask`

include/revng/PipeboxCommon/Helpers/PipeRunPipes/Base.h
aleclearmind: My feeling is that it's not worth the extra-complexity, but I'll leave the decision to you.

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: The `Task T` wraps the entire execution of the pipeline, I did this to maintain as close of a compatibility as possible to how things are done/happening currently in the pipeline

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: The `unique_ptr` is used because in `IsolateFunctions.h` I forward-declare `IsolateFunctionsImpl` but I do not define its body, so the only way for it to work is via `unique_ptr`

python/revng/internal/pipeline.yml
mrjackv: There's a few things that branch off `isolate`:
* hexdump
* enforce-abi
* emit-cfg
* render-svg-call-graph
* render-svg-call-graph-slice
* process-assembly

include/revng/Lift/Lift.h
aleclearmind: 1. Works. Always capitalize LLVM though.
2. OK.

lib/Lift/LiftCheckPrecondition.h
aleclearmind: Why not `Lift.h`?

lib/Support/IRHelpers.cpp
mrjackv: 2. They do exactly as we're doing now: [link](https://github.com/llvm/llvm-project/blob/llvmorg-16.0.0/llvm/lib/ExecutionEngine/Orc/ThreadSafeModule.cpp#L44)

lib/Lift/LiftCheckPrecondition.h
mrjackv: Works

lib/FunctionIsolation/IsolateFunctions.cpp
aleclearmind: This is confusing and doesn't bring much if not compatibility with the old pipeline, but there's no reason to be compatible.

> If you want, you can have a task for prologue, runOnFunction and epilogue, and then, if you want to, you can have subtasks for each of the above.

Reorganize the code like this, or, more in general, make sure `llvm::Task` are always function local variables and not passed around.

include/revng/FunctionIsolation/IsolateFunctions.h
```diff
@@ -21,3 +22,35 @@ class IsolateFunctions : public llvm::ModulePass {
 
   void getAnalysisUsage(llvm::AnalysisUsage &AU) const override;
 };
+
+class IsolateFunctionsImpl;
+
+namespace revng::pypeline::pipes {
```
aleclearmind: These are not *pipes* right? These are *pipe runs*, are they?

lib/FunctionIsolation/IsolateFunctions.cpp
aleclearmind: `IsolateFunctionsImpl` and `Isolate` will eventually become the same thing, so I'm not sure how beneficial is to give them different visibility in headers.

Anyway, if you want to keep things this way, for now, add a TODO.

python/revng/pypeline/cli/pipeline/run_pipe.py
mrjackv: Pipes that use static configuration are rare already, I'd KISS and re-introduce this feature if needed

include/revng/PipeboxCommon/Helpers/PipeRunPipes/Base.h
mrjackv: I'll keep it as-is, the longest branch of the pipeline is somewhere around 20-30 nodes, I'd keep the transition from python to C++ to a minimum

python/revng/pypeline/container.py
mrjackv: Used the no-argument version for both

python/revng/pypeline/schema.yml
mrjackv: Done

python/revng/pypeline/pipeline_parser.py
mrjackv: To be consistent with the rest of the codebase which uses exclusively YAML I'm sticking with it

python/revng/pypeline/pipeline.py
mrjackv: Half-reworked this change, now it uses a "fake" root node with a dummy pipe

lib/Support/IRHelpers.cpp
mrjackv: 1. Renamed to `cloneIntoContext`

include/revng/PipeboxCommon/Helpers/PipeRunPipes/FunctionPipe.h
mrjackv: Fixed

lib/Lift/LiftCheckPrecondition.h
mrjackv: Done

include/revng/Pipebox/LLVMPipe.h
mrjackv: Renamed to `Container`

include/revng/Pipebox/LLVMPipe.h
mrjackv: Done

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: Dropped

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: Dropped all uses of `std::optional<llvm::Task>`

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: Dropped all uses of `std::optional<llvm::Task>`

include/revng/FunctionIsolation/IsolateFunctions.h
mrjackv: Renamed the namespace to be `revng::pypeline::piperuns`

include/revng/PipeboxCommon/Common.h
mrjackv: 1. Done
2. Done both parts

include/revng/Lift/Lift.h
mrjackv: 1. Done
2. Done

include/revng/Model/RawBinaryView.h
mrjackv: Done

include/revng/Pipebox/LLVMPipe.h
mrjackv: Renamed to `PureLLVMPassesPipe`

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: Added a TODO

lib/EarlyFunctionAnalysis/AttachDebugInfo.cpp
mrjackv: Done

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: Done

python/revng/pypeline/task/requests.py
```diff
@@ -104,3 +104,7 @@ def __repr__(self):
 
     def __eq__(self, other) -> bool:
         return self.requests == other.requests
+
+    def reduce(self) -> Requests:
```
aleclearmind: `minimize`?

lib/Recompile/CompileModulePipe.cpp
```diff
@@ -200,6 +184,29 @@ static void compileModuleRunImpl(const Context &Context,
   revng::verify(M);
   PM.run(*M);
   revng::verify(M);
+}
+
+static void compileModuleRunImpl(const Context &Context,
```
aleclearmind, line 193: Having two functions with the same name is confusing :thinking:

lib/Recompile/LinkForTranslationPipe.cpp
```diff
@@ -34,3 +34,52 @@ void LinkForTranslation::run(ExecutionContext &EC,
...
+  return llvm::Error::success();
+}
+
+static void writeToFile(llvm::StringRef Path, llvm::StringRef Buffer) {
+  std::error_code EC;
+  llvm::raw_fd_ostream OS(Path, EC);
+  revng_assert(not EC);
+
+  OS << Buffer;
+  OS.flush();
```
aleclearmind: Not necessary

include/revng/PipeboxCommon/Helpers/PipeRunPipes/FunctionPipe.h
```diff
@@ -38,17 +40,27 @@ class FunctionPipe : public SingleOutputPipeBase<T> {
       llvm::StringRef Configuration,
       Args &...Containers) {
     ObjectDependenciesHelper ODH(Model, Outgoing, this->ContainerCount);
+
+    llvm::Task PipeRunTask(3, "Running " + this->Name);
+    PipeRunTask.advance(this->Name + ": prologue");
```
aleclearmind: In `advance` don't repeat things that are already in the name of the task.

include/revng/PipeboxCommon/Helpers/PipeRunPipes/FunctionPipe.h
```diff
@@ -38,17 +40,27 @@ class FunctionPipe : public SingleOutputPipeBase<T> {
       llvm::StringRef Configuration,
       Args &...Containers) {
     ObjectDependenciesHelper ODH(Model, Outgoing, this->ContainerCount);
+
+    llvm::Task PipeRunTask(3, "Running " + this->Name);
```
aleclearmind: As a convention, we call `Task`s just `T`.

python/revng/pypeline/object.py
```diff
@@ -325,6 +326,35 @@ def __eq__(self, other: object) -> bool:
             return False
         return self.objects == other.objects
 
+    @classmethod
+    def _from_iterable(cls, it):
+        # This functions is called by various collections.abc.Set functions to
+        # compose the container for merge operations (e.g. `&`). These are
+        # overridden below. Throw an error in the offchance that we forgot to
+        # implement one of them.
+        raise NotImplementedError
```
aleclearmind: Aren't you supposed to invoke `NotImplementedError()`? :thinking:

python/revng/pypeline/task/requests.py
```diff
@@ -76,6 +76,12 @@ def __iter__(self):
     def __getitem__(self, key: ContainerDeclaration) -> ObjectSet:
         return self.requests[key]
 
+    def get_or_empty(self, key: ContainerDeclaration) -> ObjectSet:
```
aleclearmind: Turn this into `get(self, key, default=None, /)` (`dict.get`).

python/revng/pypeline/task/savepoint.py
```diff
@@ -84,31 +84,28 @@ def run(
...
 
-        # Cache the containers present for this configuration
         for decl in self.to_save:
-            if decl not in incoming:
-                continue
-            if len(incoming[decl]) == 0:
-                continue
-            storage_provider.put(
-                ContainerLocation(
+            # Cache the containers present for this configuration
```
aleclearmind: I'm not entirely sure what's going on here.

Can you expand on the commit message?

mrjackv: I'll fix the commit message, but basically, if a savepoint had in the schedule:
```
incoming:
- /binary
outgoing:
- /binary
```

before this commit it would serialize to storage `/binary` and then deserialize it from storage 

Edit: added clarification in the commit message

python/revng/pypeline/object.py
mrjackv: I'll fix it for consistency, but they are equivalent, from the python docs:

> Otherwise, `raise` evaluates the first expression as the exception object. It must be either a subclass or an instance of `BaseException`. If it is a class, the exception instance will be obtained when needed by instantiating the class with no arguments.

Edit: fixed

lib/Recompile/CompileModulePipe.cpp
mrjackv: Eventually the one with `Context` will be dropped with the old pipeline, so there will be only one. I can add a TODO if it makes it more explicit.

include/revng/PipeboxCommon/Helpers/PipeRunPipes/FunctionPipe.h
mrjackv: `T` already refers to the pipe type

lib/Recompile/CompileModulePipe.cpp
aleclearmind: I don't know, maybe add a comment like "this is used by the old pipeline" and "this is used by both the new and the old pipeline".

include/revng/PipeboxCommon/Helpers/PipeRunPipes/FunctionPipe.h
aleclearmind: `T1`

python/revng/pypeline/task/requests.py
mrjackv: Renamed to `minimize`

lib/Recompile/CompileModulePipe.cpp
mrjackv: Added comments

lib/Recompile/LinkForTranslationPipe.cpp
mrjackv: Dropped

include/revng/PipeboxCommon/Helpers/PipeRunPipes/FunctionPipe.h
mrjackv: Fixed

include/revng/PipeboxCommon/Helpers/PipeRunPipes/FunctionPipe.h
mrjackv: Renamed to `T1`

python/revng/pypeline/task/requests.py
mrjackv: Done, split off in a standalone commit

share/revng/test/tests/pypeline-comparison/compare.sh
```diff
@@ -0,0 +1,108 @@
...
+
+function files_in_dir() {
+    find "$1" -type f -printf '%f\n'
+}
+
+# Comparison between multiple artifact objects (e.g. function, type-definition)
+function compare() {
+    local OLD_EXTRACTED NEW_EXTRACTED OLD_FILES PREFIX="$1"
+    OLD_EXTRACTED=$(mktemp -d --tmpdir="$WORKDIR")
+    NEW_EXTRACTED=$(mktemp -d --tmpdir="$WORKDIR")
```
aleclearmind: I think we can avoid `mktemp` here and just do `mkdir` but OK.

mrjackv: Replaced with `mkdir`

PR #537, opened 2025-10-30
RecursiveCoroutine improvements

vasama-work:


include/revng/ADT/UniqueCoroutineHandle.h
```diff
@@ -0,0 +1,78 @@
...
+#include <coroutine>
+
+template<typename PromiseT = void>
+class UniqueCoroutineHandle {
+  using HandleType = std::coroutine_handle<PromiseT>;
+
+  HandleType Handle;
+
+public:
+  constexpr UniqueCoroutineHandle() noexcept : Handle(nullptr) {}
```
pfez, line 16: Explicitly `delete` the copy constructor and assignment.

include/revng/ADT/UniqueCoroutineHandle.h
```diff
@@ -0,0 +1,78 @@
...
+    if (Handle)
+      Handle.destroy();
+  }
+
+  friend constexpr void swap(UniqueCoroutineHandle &LHS,
+                             UniqueCoroutineHandle &RHS) noexcept {
+    HandleType LHSHandle = LHS.Handle;
+    LHS.Handle = RHS.Handle;
+    RHS.Handle = LHSHandle;
+  }
```
pfez, line 54: Seems like `swap` is never used. Did you mean to use it in the assignment operator?

include/revng/ADT/RecursiveCoroutine-coroutine.h
```diff
@@ -37,21 +41,20 @@ struct ReturnBase {
...
+  void return_void() {}
+  void take() {}
 };
 
 template<typename ReturnT>
 struct RecursivePromise : public ReturnBase<ReturnT> {
 
-  using promise_type = RecursivePromise<ReturnT>;
-  using coro_handle = std::coroutine_handle<promise_type>;
+  using CoroHandleType = std::coroutine_handle<RecursivePromise>;
```
pfez: Just `CoroHandle`. There's no point in the `Type` suffix.

include/revng/ADT/RecursiveCoroutine-coroutine.h
```diff
@@ -66,190 +69,204 @@ struct RecursivePromise : public ReturnBase<ReturnT> {
...
 
-  RecursivePromise &operator=(RecursivePromise &&Other) {
-    if (this != &Other) {
-      this->AwaiterContinuation = Other.AwaiterContinuation;
-      Other.AwaiterContinuation = {};
-    }
-    return *this;
-  }
-  RecursivePromise(RecursivePromise &&Other) { *this = std::move(Other); }
-
```
pfez, line 125: What happened to these?

I guess you're relying on the fact that the defaulted move constructor and assignment are generated by the compiler.
It's fine to use the defaulted move operators, but I want them explicitly defaulted. See [rule of 5](https://en.cppreference.com/w/cpp/language/rule_of_three.html)

include/revng/ADT/RecursiveCoroutine-coroutine.h
```diff
@@ -66,190 +69,204 @@ struct RecursivePromise : public ReturnBase<ReturnT> {
...
   RecursiveCoroutine(const RecursiveCoroutine &) = delete;
-
-  // Movable, but we clean up the OwnedHandle, because we don't want it to be
-  // destroyed twice in the destructor
-  RecursiveCoroutine &operator=(RecursiveCoroutine &&Other) {
-    this->OwnedHandle = Other.OwnedHandle;
-    Other.OwnedHandle = {};
-    return *this;
-  }
-  RecursiveCoroutine(RecursiveCoroutine &&Other) { *this = std::move(Other); }
```
pfez, line 211: Again, rule of 5 please.

tests/unit/RecursiveCoroutine2.cpp
```diff
@@ -0,0 +1,103 @@
+/// \file RecursiveCoroutine2.cpp
+/// Tests `RecursiveCoroutine`
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
```
pfez, line 5: Please document all the tests. Describe exactly what you're trying to test in each one.

include/revng/ADT/UniqueCoroutineHandle.h
vasama-work: They are are implicitly deleted due to the user-defined move constructor and assignment. Do you still want them *explicitly* deleted? It makes no practical difference of any kind.

include/revng/ADT/UniqueCoroutineHandle.h
vasama-work: This is a generally extremely useful primitive when working with coroutines. So I didn't write it from scratch here, but rather copied it from my own personal library and changed the naming convention and formatting. The swap is just there for completeness. It's not currently used.

include/revng/ADT/RecursiveCoroutine-coroutine.h
vasama-work: No move constructor or assignment operator are generated due to the presence of the user-declared copy constructor and assignment operator. There is no need for the promise to be movable. These functions would never be used.

include/revng/ADT/RecursiveCoroutine-coroutine.h
vasama-work: Rule of 5 is about declaring *both* the move constructor and move assignment, if move is desired. As opposed to declaring say 4 or 6 special member functions.

Anyway, no move constructor or assignment operator are generated due to the presence of the user-declared copy constructor and assignment operator. There is no reason for this class to be movable. You cannot move a function call either.

include/revng/ADT/UniqueCoroutineHandle.h
pfez: Yes I want them explicitly deleted.

include/revng/ADT/RecursiveCoroutine-coroutine.h
pfez: Oh nice. These don't need to be moveable.

Anyway, if they don't need to be moveable, then explicitly `delete` the move constructor and assignment.

include/revng/ADT/RecursiveCoroutine-coroutine.h
pfez: Same as above. If this doesn't need to be moveable, explicitly delete the move constructor and move assignment.

pfez:
Merged in 9e665256130c5a2e78da7d3073c8a6c64a4aca91

Thanks!

PR #538, opened 2025-10-30
Clift loop detection

vasama-work:


tools/clift-opt/Main.cpp
```diff
@@ -26,6 +27,7 @@ int main(int Argc, char *Argv[]) {
   Registry.insert<mlir::LLVM::LLVMDialect>();
   Registry.insert<mlir::clift::CliftDialect>();
 
+  mlir::registerTransformsPasses();
```
pfez, line 26: What do you plan to use this for?

include/revng/Clift/Clift.td
```diff
@@ -85,6 +85,11 @@ class Clift_AllTypesMatch<list<string> names>
                               "mlir::clift::removeConst($_self.getType())",
                               "type, ignoring qualifiers">;
 
+class Clift_StaticType<Type constraint, string cppClassName>
+  : Type<constraint.predicate,
+         constraint.description,
+         cppClassName>;
```
pfez: What's a "static type" in this context?

lib/Clift/Clift.cpp
```diff
@@ -495,9 +495,19 @@ static mlir::IntegerAttr makeLoopLabelMask(mlir::MLIRContext *Context,
...
 }
 
-static void
-buildLoop(OpBuilder &Builder, OperationState &State, unsigned RegionCount) {
-  State.addAttribute("label_mask", makeLoopLabelMask(Builder.getContext(), 0));
+static void buildLoop(OpBuilder &Builder,
+                      OperationState &State,
+                      unsigned RegionCount,
+                      LoopOpInterface OtherLoop = {}) {
+  constexpr llvm::StringRef AttrName = "label_mask";
```
pfez: Can we have the `"label_mask"` string spelled out explicitly only in a single place?
Right now you have three instances here and another one in `CliftOpTraits.cpp`.
Please deduplicate.

pfez: Also `"label_mask"` it's a bit weird. This is more like a `"continue_break_mask"` IIUC. Can you come up with a string that is more self-explanatory even without knowing the internals?

lib/Clift/Clift.cpp
```diff
@@ -495,9 +495,19 @@ static mlir::IntegerAttr makeLoopLabelMask(mlir::MLIRContext *Context,
...
+
+  if (OtherLoop) {
+    State.addOperands(OtherLoop->getOperands());
+    State.addAttribute("label_mask", OtherLoop->getAttr("label_mask"));
```
pfez: Why do we need the `OtherLoop` here? 
IIUC this is some kind of copy-builder that is going to be used in transforms that change a loop of one kind into a loop of another kind.
Please clarify all of this in a comment.

lib/CliftTransforms/Legalization.cpp
```diff
@@ -26,7 +27,8 @@ static mlir::OpOperand &getOnlyUse(mlir::Value Value) {
 
 static void modifyResultType(mlir::PatternRewriter &Rewriter,
```
pfez, line 35: Please document this.
When is the caller supposed to set `PreserveExpressionType` to `true`/`false`?
What happens if the callers sets it to a wrong value?

include/revng/Clift/CliftOpHelpers.h
```diff
@@ -275,4 +275,19 @@ OpT getOnlyUser(mlir::Value Value) {
   return nullptr;
 }
 
+//===-------------------------- Expression usage --------------------------===//
+
+/// Returns true if the value category of the value affects the behaviour of any
+/// of its users.
+bool isValueCategorySignificant(mlir::Value Value);
```
pfez: Please add examples of this too.

lib/CliftTransforms/LoopDetection.cpp
```diff
@@ -0,0 +1,211 @@
...
+  // Maps each label assignment to the backwards gotos targeting it.
+  llvm::MapVector<AssignLabelOp, llvm::SmallVector<GotoOp, 2>> Labels;
+
+  Function->walk([&](mlir::Operation *Op) {
+    if (auto Label = mlir::dyn_cast<AssignLabelOp>(Op)) {
+      Labels.insert({ Label, {} });
+    } else if (auto Goto = mlir::dyn_cast<GotoOp>(Op)) {
+      mlir::Operation *Assignment = Goto.getLabelAssignmentOp();
+      if (auto A = mlir::dyn_cast<AssignLabelOp>(Assignment)) {
+        if (auto Iterator = Labels.find(A); Iterator != Labels.end()) {
```
pfez, line 244: My understanding is that `walk` traverses `Function` in syntactical order.

What if we visit a `Goto` targeting a label before seeing the associated `AssignLabelOp`. We would not enter this `if`.

I think you can overcome this by also matching `MakeLabelOp`, and inserting the associated `AssignLabelOp` in the map when you visit the `MakeLabelOp`. In this way when you visit a `GotoOp` you're sure you've seen the `MakeLabelOp` (because of dominance) and everything works.

lib/CliftTransforms/LoopDetection.cpp
```diff
@@ -0,0 +1,211 @@
...
+          if (isBackwardGoto(Goto, Iterator->first))
+            Iterator->second.push_back(Goto);
+        }
+      }
+    }
+  });
+
+  for (const auto &[Label, Gotos] : std::views::reverse(Labels)) {
```
pfez, line 256: AFAIU, `Labels` are collected in syntax order, so using `reverse` here guarantees that we're matching loops in a way that processes first those loops whose label appears last in syntax order.

This is pretty important for the inner functioning of the whole thing. So please explain this explicitly in a comment.

lib/CliftTransforms/LoopDetection.cpp
pfez: Perhaps this is the case of some examples of loops that weren't matched that you've seen around.

Please add a unit test for this.

lib/CliftTransforms/LoopDetection.cpp
```diff
@@ -0,0 +1,211 @@
...
+                                 PrimitiveKind::SignedKind,
+                                 /*Size=*/1);
+  auto True = Builder.create<ImmediateOp>(LoopLoc, Bool, 1);
+
+  Builder.create<YieldOp>(LoopLoc, True);
+
+  Loop.getBody().push_back(BodyBlock);
+
+  if (BoundingGotoIsInLabelScope) {
+    // A loop-goto at the end of the loop body can be erased.
```
pfez: Please call them `continue-goto`, it makes it easier to grasp.

lib/CliftTransforms/LoopDetection.cpp
```diff
@@ -0,0 +1,211 @@
...
+  return Builder.create<MakeLabelOp>(Location);
+}
+
+static void createLoop(FunctionOp Function,
+                       AssignLabelOp LoopLabel,
+                       llvm::ArrayRef<GotoOp> Gotos) {
+  mlir::Location LoopLoc = mlir::UnknownLoc::get(Function->getContext());
+
+  mlir::Block *LabelBlock = LoopLabel->getBlock();
+  bool BoundingGotoIsInLabelScope = Gotos.back()->getBlock() == LabelBlock;
```
pfez, line 117: Add a comment explaining why we know that `Goto.back()` is the bounding goto.

lib/CliftTransforms/LoopDetection.cpp
```diff
@@ -0,0 +1,211 @@
...
+  Loop.getBody().push_back(BodyBlock);
+
+  if (BoundingGotoIsInLabelScope) {
+    // A loop-goto at the end of the loop body can be erased.
+    revng_assert(Gotos.back()->getBlock() == BodyBlock);
+    revng_assert(std::next(Gotos.back()->getIterator()) == BodyBlock->end());
+    Gotos.back()->erase();
+
+    // The last loop-goto has been erased and is no longer needed:
+    Gotos = Gotos.drop_back(1);
```
pfez, line 150: Is this just an optimization or is there something that breaks, even down the pipeline, if you don't do this?

I get that if you always do what's in the `else` you'll always emit the `break` and you'll end up with situations where at the end of the body of a loop you have a `continue` followed by an unreachable `break.
But despite that being ugly, would that break something?

If it's just an optimization say it in a comment.

lib/CliftTransforms/LoopDetection.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+static void createLoop(FunctionOp Function,
+                       AssignLabelOp LoopLabel,
+                       llvm::ArrayRef<GotoOp> Gotos) {
+  mlir::Location LoopLoc = mlir::UnknownLoc::get(Function->getContext());
+
+  mlir::Block *LabelBlock = LoopLabel->getBlock();
+  bool BoundingGotoIsInLabelScope = Gotos.back()->getBlock() == LabelBlock;
+
+  auto BodyBegin = std::next(LoopLabel->getIterator());
```
pfez, line 119: Wait, the `LoopLabel` should stay inside the loop body according to the REP.
I realize that at this point the semantic is equivalent.
But do you have any reason to leave it out?

lib/CliftTransforms/LoopDetection.cpp
```diff
@@ -0,0 +1,211 @@
...
+    Loop.setContinueLabel(ContinueLabel);
+
+    for (GotoOp Goto : Gotos) {
+      Builder.setInsertionPoint(Goto);
+      Builder.create<ContinueToOp>(Goto.getLoc(), ContinueLabel);
+      Goto->erase();
+    }
+  }
+
+  if (auto BreakAssignment = getNextOp<AssignLabelOp>(Loop)) {
```
pfez, line 180: This part here got me thinking, because it looked like a leftover of the old days when you had explicit `AssignLabelOp` for the break. But looking at it more closely it seems fine.

IIUC you just look for `goto`s inside the body of the loop and turn them into breaks that break to the break label. That's fine.

I'm just wondering if we ever want to elect a different successor for the loop. Like if we have many goto's inside the loop jumping to something very far away, and just a few actually breaking to the break.
In that case we'd like to do something better than just picking the syntactical goto, but it would make all the rest very messy. Any thought about this?

lib/CliftTransforms/LoopDetection.cpp
```diff
@@ -0,0 +1,211 @@
...
+      if (not Loop->isAncestor(Goto))
+        continue;
+      if (not Loop.getContinueLabel())
+        Loop.setContinueLabel(createLabel(Builder, Function, LoopLoc));
+
+      Builder.setInsertionPoint(Goto);
+      Builder.create<ContinueToOp>(Goto.getLoc(), Loop.getContinueLabel());
+      Goto->erase();
+    }
+  }
```
pfez, line 224: You've just handled `Goto`s to continue above. It's not clear to me what do you expect to match here?

Seems like a leftover of when you had explicit `AssignLabelOp` for the continue.

tools/clift-opt/Main.cpp
vasama-work: So far I've only used `--canonicalize`. Not sure what else is in there off the top of my head.

lib/CliftTransforms/LoopDetection.cpp
vasama-work: This label could be targeted by gotos from outside of the loop. It's better to leave it outside as that does not interfere with any later transforms. Any goto inside the loop will be converted to a continue-to and will no longer target this label.

lib/CliftTransforms/LoopDetection.cpp
vasama-work: Sounds to me like something better done in LLVM IR. And even if it was done in Clift, that would be later.

lib/CliftTransforms/LoopDetection.cpp
vasama-work:
```c
loop:
  if (C)
    goto interesting_label;
  do_something();
interesting_label:
  goto loop;
```
`interesting_label` is what is being targeted here. I don't know if we're going to ever see this pattern, but if we do, here is what we would end up with otherwise:
```c
loop:
  while (1) {
    if (C)
      goto interesting_label;
    do_something();
  interesting_label:
  }
```

lib/CliftTransforms/LoopDetection.cpp
vasama-work: This is by design. A goto jumping forward (to an unvisited label) would not form a loop.

lib/Clift/Clift.cpp
vasama-work: Just so we're on the same page, this string is controlled by the tablegen definition of `Clift_LoopStatementOp` which specifies these arguments:
```
  let arguments = (ins LoopLabelMaskAttr:$label_mask,
                       Variadic<Clift_LabelType>:$labels);
```
This name also controls the loop statement interface (e.g. `getLabelMask()`). I think `getLabelMask` is a perfectly fine name for this, but we can discuss alternatives too. But it is a very rarely used, internal implementation detail anyway.

As for using a named constant: I can do that, but in this case it's a little bit like making a macro for `getLabelMask()`.

vasama-work: > Can you come up with a string that is more self-explanatory even without knowing the internals?

No matter what, if you interact with this thing in any way, you have to understand the internals. Otherwise, there is nothing useful you can do with it.

lib/CliftTransforms/LoopDetection.cpp
vasama-work: I don't like `continue-goto`, because at this point there is no loop and continue has no meaning. Perhaps `looping-goto` would be better than `loop-goto`? I can also define the term before its use.

lib/CliftTransforms/LoopDetection.cpp
pfez: :+1:

lib/CliftTransforms/LoopDetection.cpp
pfez: Ok nice.

lib/CliftTransforms/LoopDetection.cpp
pfez: Sure. On IR we are in fact already doing something like that.
So basically if there were opportunities for doing this we're already doing that on LLVM as of now.
Whatever loop successor ends up here has already been selected with that CFG-based logic.

I'm closing this thread.

lib/CliftTransforms/LoopDetection.cpp
pfez: You've added comments. Nice

lib/CliftTransforms/LoopDetection.cpp
pfez: Done

lib/CliftTransforms/Legalization.cpp
pfez: Done.

lib/CliftTransforms/LoopDetection.cpp
pfez: Nice.

lib/Clift/Clift.cpp
pfez: Done

lib/Clift/Clift.cpp
pfez: No strong feelings about the `"label_mask"` string. I was just thinking out loud if we could come up with something better. Let's leave it like it is.

On the other hand I would have loved to see, in C++, something like `LoopOpInterface::getLabelMaskName()` instead of a plain string `"label_mask"`. But we have no `getLabelMaskName` in `LoopOpInterface` as of now. And even if we did, I don't think we could use that for generating `label_mask` here:

```
  let arguments = (ins LoopLabelMaskAttr:$label_mask, Variadic<Clift_LabelType>:$labels);
```

lib/CliftTransforms/LoopDetection.cpp
pfez: Doesn't seem tragic. In principle I think we could leave this and other following loop and label optimizations should get rid of it. But I understand the effort of keeping the output of this pass as nice as possible.

In any case, as usual, document this.

include/revng/Clift/Clift.td
vasama-work: For most type constraints, the stored type is `clift::ValueType`. But in this case we want to use a stored (or static if you will) type of `clift::FunctionType` instead.

lib/Clift/Clift.cpp
vasama-work: So you're okay with the status quo? Or I could add a `getLabelMaskAttrName()` in `AssignsLoopLabels`.

include/revng/Clift/Clift.td
vasama-work: Looks like I'll have to revert this change though. Aside from a function type, this type could also be a typedef.

lib/CliftTransforms/LoopDetection.cpp
vasama-work: To be clear, this is a continue-like jump and we don't have any later detection of breaks or continues, so that goto will never be turned into a continue. Label merging later would merge the label assignment into the loop's continue label, after which the goto targeting a loop label within the loop would become invalid and verify would fail.

vasama-work:
I added `MakeLabelOp` canonicalization in loop detection.

include/revng/Clift/Clift.td
vasama-work: I made some changes to allow typedefs but still get a nice C++ interface.

include/revng/Clift/CliftOpHelpers.h
vasama-work: I added an example.

lib/Clift/Clift.cpp
pfez: > Or I could add a `getLabelMaskAttrName()` in `AssignsLoopLabels`

This. Thank you.

lib/CliftTransforms/LoopDetection.cpp
pfez: This last point is pretty important.
It means that you can't just remove that part of the code and expect other optimizations to take care of the results, because you would generate invalid clift.
Add this to the comment section you've already inserted above that piece of code.

include/revng/Clift/Clift.td
pfez: Nice.

include/revng/Clift/CliftOpHelpers.h
pfez: :thinking: ok.

More in general, until now we've kept the idea of value categories outside of everything related to clift.
And in this PR there isn't AFAICT anything that uses that.
So please, rip out `isValueCategorySignificant` and `reliesOnOperandValueCategory` in a separate commit, and drop it from this PR.
We can then discuss that at ease in the context of what we really need it for, and if it's really needed we can include it in a future PR along with the logic that actively uses this.

vasama-work: Sure. You're right it's not used here. It's used in other parts of REP 52 which are not yet in this PR.

vasama-work: FWIW, we do already care about value categories in Clift, and have for a long time. The listed operands verify that their operands have lvalue category.

vasama-work: I removed it.

lib/CliftTransforms/LoopDetection.cpp
vasama-work: I added further explanation.

lib/Clift/Clift.cpp
vasama-work: I added a `LoopLabelMaskAttrName` constant in `mlir::clift::impl` instead.

include/revng/Clift/CliftOpHelpers.h
pfez: Thanks.

PR #543, opened 2025-11-12
Pypeline: add tar support for containers

zommiommy:
Now containers have .to_bytes and .from_bytes, and all to/from methods accept a container_format argument. Moreover, pype project artifact now accepts --format to specify in which format to serialize the resulting container, and has a --tar shortcut.

### DIFF
--- a/python/revng/pypeline/cli/utils.py
+++ b/python/revng/pypeline/cli/utils.py
@@ -127,11 +127,6 @@ class EagerParsedPath(click.Path):
         if ctx is not None:
             if ctx.obj is None:
                 ctx.obj = {}
-            if self.name in ctx.obj:
-                raise ValueError(
-                    f'Argument "{self.name}" already set in context, '
-                    "this is likely a bug in the code."
-                )
             # Store the parsed value in the context object
             ctx.obj[self.name] = res
             # And also store the path to the parsed value
--- a/tests/pypeline/CMakeLists.txt
+++ b/tests/pypeline/CMakeLists.txt
@@ -38,6 +38,22 @@ set_tests_properties(
   PROPERTIES LABELS "pypeline" ENVIRONMENT
              "PYTHONPATH=${CMAKE_BINARY_DIR}/${PYTHON_INSTALL_PATH}")
 
+revng_add_test(
+  NAME
+  test_pype
+  COMMAND
+  "${Python_EXECUTABLE}"
+  -m
+  pytest
+  -p
+  no:cacheprovider
+  -v
+  --rootdir
+  "${CMAKE_CURRENT_SOURCE_DIR}"
+  "${CMAKE_CURRENT_SOURCE_DIR}/test_pype.py")
+set_tests_properties(
+  test_pype PROPERTIES LABELS "pypeline" ENVIRONMENT
+                       "PYTHONPATH=${CMAKE_BINARY_DIR}/${PYTHON_INSTALL_PATH}")
 #
 # Generate libmockPipeboxImpl and libmockPipebox
 #
--- a/tests/pypeline/test_pype.py
+++ b/tests/pypeline/test_pype.py
@@ -1,0 +1,183 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+import functools
+import shutil
+import sys
+import tarfile
+from pathlib import Path
+
+import pytest
+import yaml
+from click.testing import CliRunner
+
+from revng.pypeline.main import pype
+
+ROOT = Path(__file__).resolve().parent
+
+ENV_FILES = ["pipeline.yml", "pipebox.py", "model.yml"]
+
+
+@pytest.fixture
+def runner() -> CliRunner:
+    return CliRunner()
+
+
+@pytest.fixture(params=["yaml", "tar"])
+def container_format(request):
+    return request.param
+
+
+def setup_env(test_func):
+    """Decorator to setup an isolated filesystem for CLI tests.
+    This won't pass the arguments to the pipebox."""
+
+    @functools.wraps(test_func)
+    def wrapped(runner: CliRunner, *args, **kwargs):
+        # This is just a chdir to a tmp folder
+        with runner.isolated_filesystem():
+            # Create the necessary files inside the new filesystem
+            for file in ENV_FILES:
+                shutil.copyfile(ROOT / file, file)
+            # Run the test
+            test_func(runner, *args, **kwargs)
+
+    return wrapped
+
+
+def run_partial(runner: CliRunner, **kwargs):
+    """Convenience partial function to pre-fill common args"""
+
+    def run(*args: str):
+        result = runner.invoke(cli=pype, args=args, **kwargs)
+        if result.exit_code != 0:
+            print(result.output, file=sys.stderr)
+        assert result.exit_code == 0
+        return result
+
+    return run
+
+
+def is_tar_file(path: Path) -> bool:
+    return tarfile.is_tarfile(path)
+
+
+def is_yaml_file(path: Path) -> bool:
+    try:
+        with open(path, "r") as f:
+            yaml.safe_load(f)
+        return True
+    except (yaml.YAMLError, FileNotFoundError):
+        return False
+
+
+def check_file_format(path: Path, expected_format: str):
+    if expected_format == "yaml":
+        assert is_yaml_file(path), f"File {path} is not a valid yaml file"
+    elif expected_format == "tar":
+        assert is_tar_file(path), f"File {path} is not a valid tar file"
+    else:
+        raise ValueError(f"Unknown format: {expected_format}")
+
+
+@setup_env
+def test_pipeline(runner: CliRunner, container_format: str):
+    run = run_partial(
+        runner,
+        obj={
+            "pipebox_args": ["first pipebox arg", "second pipebox arg"],
+        },
+    )
+
+    # First call some pipes to manually generate some containers
+    run(
+        "pipeline", "run-pipe", "GeneratorPipe", f"--format={container_format}", "model.yml", "out1"
+    )
+    check_file_format(Path("out1"), container_format)
+
+    run(
+        "pipeline",
+        "run-pipe",
+        "InPlacePipe",
+        f"--format={container_format}",
+        "model.yml",
+        "out1",
+        "out2",
+    )
+    check_file_format(Path("out2"), container_format)
+
+    run(
+        "pipeline",
+        "run-pipe",
+        "ToLowerKindPipe",
+        f"--format={container_format}",
+        "model.yml",
+        "out2",
+        "out3",
+    )
+    check_file_format(Path("out3"), container_format)
+
+    # Call a couple of analyses
+    run(
+        "pipeline",
+        "run-analysis",
+        "NullAnalysis",
+        "model.yml",
+        "out1",
+    )
+
+    run(
+        "pipeline",
+        "run-analysis",
+        "PurgeAllAnalysis",
+        "model.yml",
+        "out1",
+    )
+
+    run(
+        "pipeline",
+        "run-analysis",
+        "AddStuffAnalysis",
+        "model.yml",
+        "out1",
+    )
+
+
+@setup_env
+def test_project(runner: CliRunner, container_format: str):
+    run = run_partial(
+        runner,
+        obj={
+            "pipebox_args": ["first pipebox arg", "second pipebox arg"],
+        },
+    )
+
+    # Compute artifacts
+    run(
+        "project",
+        "artifact",
+        "ChildArtifact",
+        f"--format={container_format}",
+        "-o",
+        "child_artifact",
+    )
+    check_file_format(Path("child_artifact"), container_format)
+
+    run(
+        "project",
+        "artifact",
+        "RootArtifact",
+        f"--format={container_format}",
+        "-o",
+        "root_artifact",
+    )
+    check_file_format(Path("root_artifact"), container_format)
+
+    # Run analysis
+    run("project", "analyze", "NullAnalysis")
+    run("project", "analyze", "blackhole")
+    run("project", "analyze", "init_analysis")
+
+    # Run an analysis list
+    run("project", "analyze", "all_analyses")


python/revng/pypeline/container.py
```diff
@@ -4,18 +4,23 @@
...
 from pathlib import Path
-from typing import Annotated, Dict, Generator, TextIO, Tuple, Type, final
+from typing import Annotated, Generator, Literal, Tuple, Type, final
 
 from .object import Kind, ObjectID, ObjectSet
 from .utils import bytes_to_string, is_mime_type_text, string_to_bytes
 from .utils.cabc import ABC, abstractmethod
 from .utils.registry import get_singleton
 
+ContainerFormat = Literal["tar", "json"]
```
aleclearmind: I'd switch to an `Enum`, add `auto` and assert we don't have `auto` where we don't want it.

python/revng/pypeline/container.py
```diff
@@ -154,84 +159,159 @@ def contains_all(self, obj: ObjectSet) -> bool:
...
-        return cls.from_file(StringIO(data))
+        match container_format:
+            case "json":
+                return cls.from_bytes(data.encode("utf-8"), container_format="json")
+            case "tar":
+                return cls.from_bytes(base64.b64decode(data), container_format="tar")
+        raise ValueError(f"Unknown container format: {container_format}")
+
+    @classmethod
+    def container_format_from_bytes(cls, data: bytes) -> ContainerFormat:
```
aleclearmind: `s/container_format_from_bytes/detect_format/`

python/revng/pypeline/container.py
```diff
@@ -154,84 +159,159 @@ def contains_all(self, obj: ObjectSet) -> bool:
...
+        is_tar = tarfile.is_tarfile(io.BytesIO(data))
+        if is_tar:
+            return "tar"
+        elif data.startswith(b"{"):
```
aleclearmind: Apparently the JSON grammar does not allow for white space, but I'd still strip initial whitespace.

python/revng/pypeline/container.py
```diff
@@ -154,84 +159,159 @@ def contains_all(self, obj: ObjectSet) -> bool:
...
+            return "json"
```
aleclearmind: @mrjackv What do we think of `json` vs YAML here?

python/revng/pypeline/cli/pipeline/run_pipe.py
```diff
@@ -141,18 +143,29 @@ def build_pipe_command(
...
         file_storage: Path,
+        container_format: ContainerFormat,
+        is_tar: bool,
+        is_json: bool,
         **kwargs,
     ) -> None:
+        if is_tar:
+            container_format = "tar"
+        if is_json:
+            container_format = "json"
```
aleclearmind: We should give an error if there's both `--tar` and `--json`.

python/revng/pypeline/cli/utils.py
```diff
@@ -315,3 +315,31 @@ def compute_objects(
     required=False,
     help="The token to pass to the storage provider.",
 )
+
+
+container_format_option = click.option(
+    "--format",
+    "container_format",
+    type=click.Choice(get_args(ContainerFormat) + ("auto",)),
```
aleclearmind: 1. Does this affect the input containers?
   I think we should always auto-detect the format of input container and not offer an option to customize it.
2. `auto` is not affected by whether the container's content is textual or not (`is_text`). Correct?
3. What do we do if the output is `stdout`? AFAIU the default is `auto`. Hopefully we don't fail to simply emit on stdout without specifying `--format`.
   It is a very common use case to do `revng2 artifact decompile | otherstuff`.

tests/pypeline/pipeline.yml
```diff
@@ -15,6 +15,11 @@ containers:
   - name: root_destination
     type: RootDictContainer
 
+analyses_lists:
+  - name: "all_analyses"
+    analyses: ["init_analysis", "blackhole", "NullAnalysis", "init_analysis"]
```
aleclearmind: Expand into one per line

python/revng/pypeline/pipeline.py
```diff
@@ -96,6 +103,31 @@ def __init__(
...
+                        f"Analysis {analysis_name} in analyses list {analysis_list.name} is not "
+                        "defined in the pipeline"
+                    )
+            if analysis_list.name in self.analyses_lists:
+                raise ValueError(
+                    f"Analyses list alias {analysis_list.name} is defined multiple times in the "
+                    "pipeline"
+                )
+            # Add the analysis list
+            self.analyses_lists[analysis_list.name] = analysis_list
```
aleclearmind: Make sure you always use *analysis list* or *analysis lists*.
Don't use *analyses list* or *analyses lists*.

python/revng/pypeline/analysis.py
```diff
@@ -61,3 +61,12 @@ class AnalysisBinding:
     analysis: Analysis
     bindings: tuple[ContainerDeclaration, ...]
     node: PipelineNode
+
+
+@dataclass(frozen=True, slots=True)
+class AnalysisList:
```
aleclearmind, line 83: During your talk I saw that when we run an analysis list you can still request objects in containers. Correct?

The semantics of that is unclear to me. Is it like a collection of all the `--$ARGUMENT-objects` arguments of all the analyses we're going to run?

python/revng/pypeline/cli/project/analyze.py
```diff
@@ -85,6 +89,154 @@ def _build_analysis_command(self, analysis_name: str, pipeline: Pipeline):
 
         return run_analysis_command
 
+    def _build_analysis_list_command(self, analysis_list_name: str, pipeline: Pipeline):
```
aleclearmind, line 92: @mrjackv Can you review this commit?

python/revng/pypeline/pipeline.py
```diff
@@ -47,6 +47,15 @@ def is_cacheable(self) -> bool:
         """An artifact is cacheable if it's backed by a savepoint."""
         return isinstance(self.node.task, SavePoint)
 
+    def to_web_dict(self) -> dict:
```
aleclearmind: I don't see anything web specific here. Do you?

I'd drop `_web` everywhere.

python/revng/pypeline/daemon/daemon.py
```diff
@@ -45,6 +50,41 @@ def to_dict(self):
         return result
 
 
+def get_web_pipeline(version: str, pipeline: Pipeline) -> dict[str, Any]:
```
aleclearmind, line 52: @mrjackv Can you review this commit?

tests/pypeline/CMakeLists.txt
```diff
@@ -38,6 +38,22 @@ set_tests_properties(
...
+  COMMAND
+  "${Python_EXECUTABLE}"
+  -m
+  pytest
+  -p
+  no:cacheprovider
+  -v
+  --rootdir
+  "${CMAKE_CURRENT_SOURCE_DIR}"
+  "${CMAKE_CURRENT_SOURCE_DIR}/test_pype.py")
```
aleclearmind, line 53: Is this formatting from our linter?

tests/pypeline/test_pype.py
aleclearmind, line 1: @mrjackv Can you review this?

python/revng/pypeline/container.py
mrjackv: There's `StrEnum` that should work better

python/revng/pypeline/container.py
mrjackv: In what sense? Technically if we expect there might be YAML in input the only reliable way to check for it is do `json.loads` and see if it throws an error

aleclearmind: What I meant is: we're pretty committed to YAML. Shall we purge JSON entirely from the CLI in favor of YAML?

We need a general rule on how to adopt JSON vs YAML.
I guess for the daemon it makes sense to go JSON, but maybe for the CLI we should go to YAML. What's your take?

python/revng/pypeline/daemon/daemon.py
```diff
@@ -45,6 +50,41 @@ def to_dict(self):
...
+        "artifacts": [artifact.to_web_dict() for artifact in pipeline.artifacts.values()],
+        "analyses": [analysis.to_web_dict() for analysis in pipeline.analyses.values()],
+        "analyses_lists": [
+            analysis_list.to_web_dict() for analysis_list in pipeline.analyses_lists.values()
+        ],
+    }
+
+    # Ensure that it respects the agreed schema
+    validator = jsonschema.Draft7Validator(schema)
+    validator.validate(web_pipeline)
```
mrjackv, line 82: Checking every time that the result conforms to the schema is kind of expensive. I'd add a test in `test_daemon` instead that checks this.

python/revng/pypeline/daemon/utils.py
```diff
@@ -74,7 +75,8 @@ def compute_artifact(
     model: ReadOnlyModel,
     artifact_name: str,
     artifact_data: dict,
-) -> dict[str, Any]:
+    container_format: ContainerFormat = "json",
```
mrjackv: I think this belongs to another commit

python/revng/pypeline/daemon/daemon.py
```diff
@@ -243,40 +251,49 @@ async def analyze(self, request) -> Response:
...
-                model=ReadOnlyModel(model),
-                analysis_name=analysis,
-                requests=requests,
-                analysis_configuration=configuration,
-                pipeline_configuration=pipeline_configuration,
-                storage_provider=storage_provider,
-            )
+            current_model = ReadOnlyModel(model)
+            total_invalidated: InvalidatedObjects = {}
+            for analysis_name in analysis_list:
```
mrjackv: Having both cli and daemon manually "unwrap" the analysis list is IMHO the wrong way to handle this. I'd propose adding a `run_analysis_list` to `Pipeline` instead.

python/revng/pypeline/container.py
mrjackv: Hmmmm, also for the CLI I'd stick with JSON. There's a lot of people that are very used to doing things with `jq`. I'm not so sure how commonly used `yq` is. Also it makes things very uniform w.r.t. the daemon so for example, the scripting code will become more uniform.

python/revng/pypeline/cli/project/analyze.py
mrjackv: Done

python/revng/pypeline/daemon/daemon.py
mrjackv: Done

tests/pypeline/test_pype.py
mrjackv: Done

python/revng/pypeline/cli/utils.py
zommiommy: 1. no, only outputs
2. yes it's not affected
3. on stdout it defaults to json

python/revng/pypeline/daemon/daemon.py
zommiommy: It checks it exactly once during the daemon __init__

python/revng/pypeline/cli/utils.py
aleclearmind: 3. What's the exact logic? I've seen code throwing an exception in case the extension doesn't match.
   Are we special casing stdout? :thinking:

python/revng/pypeline/daemon/daemon.py
mrjackv: Ah my bad, looks good

python/revng/pypeline/container.py
aleclearmind: OK.

I guess the rule should be: if a human is likely to write and maintain (e.g., commit to a repo) a file, we'll use YAML, otherwise if it's mostly for "automated" manipulation, we can go with JSON.

What do you think?

mrjackv: As a rule it works, but I'd propose something along these lines:

> If a end-user (e.g. someone using the CLI, someone curl-ing the daemon) is receiving/sending the data, then the input/output should be JSON. If it is an internal file (e.g. pipeline configuration) that is developer-facing then it should be YAML.

aleclearmind: Eh, but the model is going to be YAML and you'll want to use it from the CLI. `:/`

mrjackv: Hmmmm, you're right, yeah, the rule you propose is better :laughing:

python/revng/pypeline/analysis.py
zommiommy: Yeah it didn't make sense, now you can only compute all objects. A related problem is the analysis configurations. I'd do it by name, but one can put in the list the same analysis multiple times and that's annoying.

aleclearmind: The same analysis multiple times in a list? Let's ban that.

python/revng/pypeline/cli/utils.py
zommiommy: https://github.com/zommiommy/revng/blob/f04bf32353962fcbef051584877373438be8df03/python/revng/pypeline/cli/project/artifact.py#L158-L162

python/revng/pypeline/analysis.py
zommiommy: Ok, then I we can have args like `--<analysis-name>--configuration` to configure each analysis

python/revng/pypeline/cli/utils.py
aleclearmind: I'm not very happy about this special casing of stdout.

I'm starting to think the whole `auto` thing is misguided and we should just always emit YAML unless the user explicitly asks for `tar`.

I'd be surprised if I go from `revng2 artifact asd -o asd.tar` to `revng2 artifact asd | tar -xf -` and it wouldn't work.

Let's drop `auto` entirely (for output).

tests/pypeline/CMakeLists.txt
zommiommy: I guess? I just copied the formatting from `tests/pipeline/CMakeLists.txt` and hoped in `revng check-convetions --force-format`

python/revng/pypeline/analysis.py
aleclearmind: OK, that would work.

python/revng/pypeline/cli/utils.py
zommiommy: done, but using yaml as default now the CLI args look like this:
```
 --yaml                          Shortcut for --format=yaml.
  --tar                           Shortcut for --format=tar.
  --format [tar|yaml]             Format to use for the output container,
                                  either on stdout or in the result path.
                                  [default: yaml]
```

So the `--yaml` shortcut won't be ever used. Should I remove it?

aleclearmind: It would only make sense if we can get click to be an alias for `--format=yaml` so that if you have `--yaml --tar --yaml` you get YAML. But AFAIU it's not how it works, they are separate flags and one of them takes over.

In any case, it's not very important, you can drop it. I wouldn't obsess over it.

PR #545, opened 2025-11-18
Clift optimisation

vasama-work:


include/revng/CliftTransforms/PatternRewriter.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
```
pfez, line 4: Putting this here for lack of a better place.
Can you add in the commit message the references to the upstream commits that introduce these APIs?

share/revng/compile-flags.cfg
```diff
@@ -33,6 +33,11 @@
 # very likely to ever be able to re-enable this warning.
 -Wno-address-of-packed-member
 
+# Pointer arithmetic involving large but non-negative constant values triggers
+# this diagnostic. Ultimately there is no good way to ensure that decompiled
+# code does not contain such pointer arithmetic, so the warning is disabled.
+-Wno-array-bounds
```
pfez, line 42: Ok, but please add a few examples of the offending snippets to the comment.

lib/CliftPipes/OptimizationPipe.cpp
```diff
@@ -38,9 +38,20 @@ class OptimizationPipe {
     mlir::PassManager PM(CliftContainer.getContext(),
                          clift::FunctionOp::getOperationName());
 
-    PM.addPass(mlir::createCanonicalizerPass());
+    PM.addPass(clift::createTrivialReturnEliminationPass());
```
pfez, line 53: What was the reason again to run this before the others?

share/revng/include/primitive-types.h
```diff
@@ -215,3 +215,22 @@ extern uintmax_t undef_value(void);
...
+//
+// bit_cast
+//
+
+#if defined(__GNUC__) && !defined(__clang__) && !defined(__cplusplus)
+// GCC implements __builtin_bit_cast, but it is only available in C++ mode.
+// Since we've just detected GCC, we can use a few GNU extensions to implement
+// behaviour equivalent to __builtin_bit_cast using memcpy (__builtin_memcpy).
+#define bit_cast(T, ...)                                   \
+  (__extension__({                                         \
```
pfez, line 230: :thinking: interesting, I didn't know about `__extension__`.

share/revng/include/primitive-types.h
```diff
@@ -215,3 +215,22 @@ extern uintmax_t undef_value(void);
...
+
+//
+// bit_cast
+//
+
+#if defined(__GNUC__) && !defined(__clang__) && !defined(__cplusplus)
+// GCC implements __builtin_bit_cast, but it is only available in C++ mode.
+// Since we've just detected GCC, we can use a few GNU extensions to implement
+// behaviour equivalent to __builtin_bit_cast using memcpy (__builtin_memcpy).
+#define bit_cast(T, ...)                                   \
```
pfez, line 229: Why do we need a variadic macro?

share/revng/include/primitive-types.h
```diff
@@ -215,3 +215,22 @@ extern uintmax_t undef_value(void);
 
 #define break_to goto
 #define continue_to goto
+
+//
+// bit_cast
+//
+
+#if defined(__GNUC__) && !defined(__clang__) && !defined(__cplusplus)
```
pfez: Not sure we have to care about `__cplusplus`, our headers are meant to be compiled as C anyway.

Is all the rest of the header guaranteed to compile properly as C++ already? If not, checking for `__cplusplus` is pointless, and only gives the false illusion that we expect the rest of the header to compile fine as C++.

Also, even if the rest of the header compiles fine with C++ already, that's "by accident", in the sense that we've never made any commitment to guaranteeing that the header compiles as C++.
And I'm not ready to make that commitment.

If you really want to still check for `__cplusplus`, that's fine, but document right above that we're doing this for compatibility and *not* as a guarantee.

share/revng/include/primitive-types.h
vasama-work: I'm not sure if we *need* it... I'm not *sure* there isn't some valid expression we could emit from Clift here that contains a comma without parentheses... Generally I find it's just better to make the thing variadic in these cases.

share/revng/include/primitive-types.h
vasama-work: I can take it out too.

lib/CliftPipes/OptimizationPipe.cpp
vasama-work: We want to get rid of trivial jumps early on, in order to enable other transformations. For jumps with explicit targets, we do this based on the label during iterative statement rewriting. The reason we do it the way we do based on labels appearing in a specific region is so that changes affecting other operations nested directly or indirectly within that region cause the pattern to be retried. For returns that is less feasible, because there is no way to iterate over the set of returns in a function like you can iterate over the set of jumps targeting a label. On the other hand, we expect that trivial returns can be eliminated early without having to do other statement rewrites. If in the future this turns out not to be true, one alternative is to convert void returns into gotos before statement rewriting.

share/revng/include/primitive-types.h
vasama-work: Yeah. Handy for these sorts of things. Suppresses warnings about the use of extensions.

share/revng/compile-flags.cfg
vasama-work: I updated the comment with a simple example.

include/revng/CliftTransforms/PatternRewriter.h
vasama-work: I updated the commit message.

include/revng/Clift/Clift.td
```diff
@@ -351,6 +351,24 @@ def Clift_AssignLabelOp
   }];
 }
 
+def Clift_BlockStatementOp
```
pfez, line 354: Adding this here for lack of better place.

Remember to add tests for this. It's already noted in the commit message, but I'm noting this down here as well.

share/revng/include/primitive-types.h
pfez: Ok, add a comment to explain that the variadic macro is just used for forward compatibility, in the case that at some point we want to `bit_cast` an expression that contains a comma.

share/revng/include/primitive-types.h
pfez: Yeah, take it out please.

lib/CliftTransforms/Legalization.cpp
```diff
@@ -103,12 +111,25 @@ struct PointerResizePattern : mlir::OpRewritePattern<OpT> {
     return mlir::success();
   }
 
+  mlir::LogicalResult replaceIntegerOperand(mlir::PatternRewriter &Rewriter,
+                                            clift::ExpressionOpInterface Op,
+                                            unsigned Index = 0) const {
+    mlir::OpOperand &Operand = Op->getOpOperand(Index);
+
+    auto OldType = clift::getPrimitiveIntegerType(Operand.get().getType());
+    if (not OldType or OldType.getSize() == TargetPointerSize)
```
pfez, line 120: This says `replaceIntegerOperand` but then goes on checking if the operand is a pointer-sized integer.

`replacePointerSizedIntegerOperand` is probably a better name.

lib/CliftTransforms/Legalization.cpp
```diff
@@ -88,6 +88,14 @@ struct PointerResizePattern : mlir::OpRewritePattern<OpT> {
                                    TargetPointerSize);
   }
 
+  clift::PrimitiveType
+  makeTargetIntegerType(mlir::PatternRewriter &Rewriter,
+                        clift::PrimitiveType OldIntegerType) const {
+    return clift::PrimitiveType::get(Rewriter.getContext(),
+                                     OldIntegerType.getKind(),
+                                     TargetPointerSize);
```
pfez, line 96: Same here. `makeTargetPointerSizedIntegerType`

lib/CliftTransforms/Legalization.cpp
```diff
@@ -149,14 +173,31 @@ using ResizePtrSubPattern = ResizePointerArithmeticPattern<clift::PtrSubOp>;
 struct ResizePtrDiffPattern : PointerResizePattern<clift::PtrDiffOp> {
   using PointerResizePattern::PointerResizePattern;
 
+  mlir::LogicalResult replaceIntegerResult(mlir::PatternRewriter &Rewriter,
```
pfez, line 176: And here `s/Integer/PointerSizedInteger/`

lib/CliftPipes/OptimizationPipe.cpp
pfez: All of this is valuable information that should be documented here.

lib/CliftTransforms/TrivialReturnElimination.cpp
```diff
@@ -0,0 +1,67 @@
...
+using PassBase = clift::impl::CliftTrivialReturnEliminationBase<T>;
+
+struct TrivialReturnEliminationPass : PassBase<TrivialReturnEliminationPass> {
+  void runOnOperation() override {
+    mlir::MLIRContext *Context = &getContext();
+
+    mlir::RewritePatternSet Patterns(Context);
+    Patterns.add<TrivialReturnEliminationPattern>(Context);
+
+    // TODO: Use walkAndApplyPatterns
```
pfez, line 56: There's plenty of this `TODO`s around the codebase. I assume `walkAndApplyPatterns` is an API available upstream that we don't have available in our current LLVM version? If so, feel free to close this thread.

include/revng/CliftTransforms/Passes.td
```diff
@@ -11,6 +11,13 @@ class Clift_Pass<string arg, string operation> : Pass<arg, operation> {
   let dependentDialects = ["clift::CliftDialect"];
 }
 
+def CliftTrivialReturnElimination : Clift_Pass<"trivial-return-elimination",
```
pfez, line 15: We've got no unit test for this. Unless it's in another commit that I haven't seen yet.

It's easy to come up with simple but non-trivial cases to unit test, both for testing that the optimization expectedly succeeds in some cases, and expectedly fails in others.

lib/CliftTransforms/LabelMerging.cpp
```diff
@@ -0,0 +1,101 @@
...
+  mlir::LogicalResult
+  matchAndRewrite(LoopOpInterface Loop,
+                  mlir::PatternRewriter &Rewriter) const override {
+    mlir::LogicalResult Result = mlir::failure();
+
+    auto ReplaceLabel = [&](AssignLabelOp Assignment, mlir::Value NewLabel) {
+      Rewriter.replaceAllUsesWith(Assignment.getLabel(), NewLabel);
+      Rewriter.eraseOp(Assignment);
+      Result = mlir::success();
+    };
```
pfez: Explicit captures please.

lib/CliftTransforms/LabelMerging.cpp
```diff
@@ -0,0 +1,101 @@
...
+struct AssignLabelMergingPattern : mlir::OpRewritePattern<AssignLabelOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  mlir::LogicalResult
+  matchAndRewrite(AssignLabelOp AssignLabel,
+                  mlir::PatternRewriter &Rewriter) const override {
+    mlir::Block::iterator Pos = std::next(AssignLabel->getIterator());
+
+    if (Pos == AssignLabel->getBlock()->end())
+      return mlir::failure();
```
pfez, line 34: :thinking:
If `AssignLabel` is the last operation of a block, I think the correct thing to do would be to look at the fallthrough.
You've got `getFallthroughTarget` in `RewriteHelpers.cpp` but that skips the `AssignLabelOp`s. Perhaps you need one here that doesn't skip them?

lib/CliftTransforms/LabelMerging.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+    auto NextAssignLabel = mlir::dyn_cast<AssignLabelOp>(&*Pos);
+    if (not NextAssignLabel)
+      return mlir::failure();
```
pfez, line 38: Looking at how you use this here, something with the same API as `getNextOp<AssignLabelOp>` would be nice, but with the semantic of `getFallthroughTarget` that does not skip over `AssignLabelOp`s could do the trick?

lib/CliftTransforms/LabelMerging.cpp
pfez: But now that I think about it, `getFallthroughTarget` just stops at the end `BlockPosition`, it doesn't actually fall through to the parent scope.

To clarify, I'm thinking about a scenario like:

```c
if (condition) {
  do_stuff();
  label_0:;
}
label_1:
do_more();
```

I'd like to be able to merge `label_0` into `label_1` and obtain

```c
if (condition)
  do_stuff()
label_1:
do_more();
```

tests/unit/mlir_lit_tests/statement-rewrites/label-merging-assign-assign.mlir
```diff
@@ -0,0 +1,55 @@
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+// RUN: %revngcliftopt %s --label-merging | FileCheck %s
```
pfez, line 5: There are no tests to test that we're merging also loop labels properly.

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
```diff
@@ -0,0 +1,166 @@
...
+  Then,
+  Else,
+};
+
+static bool hasSingleNoFallthrough(mlir::Region &R) {
+  return static_cast<bool>(getLastNoFallthroughStatement(R));
+}
+
+// Computes an approximation of the size of a statement region in C.
+static double approximateRegionWeight(mlir::Region &R) {
```
pfez: Don't use `double`. `uint64_t` is fine.

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
```diff
@@ -0,0 +1,166 @@
...
+    // Apply boolean negations:
+    {
+      mlir::RewritePatternSet Patterns(Context);
+      populateWithBooleanNegationPatterns(Patterns);
+
+      // TODO: Use walkAndApplyPatterns
+      if (mlir::applyPatternsAndFoldGreedily(Function, std::move(Patterns))
+            .failed())
+        signalPassFailure();
+    }
```
pfez: Please document why we need this.

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
```diff
@@ -0,0 +1,166 @@
...
+  if (ThenHasDirectNoFallthrough != ElseHasDirectNoFallthrough)
+    return static_cast<HoistingTarget>(ElseHasDirectNoFallthrough);
+
+  // Otherwise, the sizes of both branches are approximated, and a decision is
+  // made by comparing those approximations.
+  double ThenWeight = approximateRegionWeight(If.getThen());
+  double ElseWeight = approximateRegionWeight(If.getElse());
+
+  if (ThenWeight == ElseWeight)
+    return std::nullopt;
```
pfez: Hoist the `Else` if their score is equal.

This will likely require a few tests to be changed.

lib/CliftTransforms/Legalization.cpp
vasama-work: That's getting quite verbose. And it's already scoped within `PointerResizePattern`. We don't need to explain every piece of minutiae in naming. At some point it becomes counter-productive.

lib/CliftTransforms/LabelMerging.cpp
vasama-work: That's reasonable. However, I want to reduce the complexity of any given transform. I think this calls for a rewrite I had before, which I called "label motion", which would move labels as far back as possible. In this case `label_0` would be moved out of the branch.

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
vasama-work: I used `double` because this is very much a fuzzy approximation, and as we improve the heuristics here in the future, I suspect we might prefer floating point. And using floats in the first place acts as documentation telling the users that they cannot expect an integer value.

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
vasama-work: I think that would be worse. Out of these options, I would rather see the former, because there's no way we can decide that one is better than the other:
```c
if (C)
  goto L1;
else
  goto L2;

if (C)
  goto L1;
goto L2;
```

lib/CliftTransforms/LabelMerging.cpp
vasama-work: Is this related to the previous thread?

lib/CliftPipes/OptimizationPipe.cpp
vasama-work: I added a comment.

share/revng/include/primitive-types.h
vasama-work: I added a comment explaining that.

lib/CliftTransforms/ImmediateRadixDeduction.cpp
```diff
@@ -0,0 +1,125 @@
...
+    if (isBitwiseOperator(User))
+      return 0x10;
+  }
+
+  return deduceBestIntegerRadix(Immediate.getValue());
+}
+
+// For each integer immediate, deduce the best radix and if it is not 10 (the
+// default), attach it in a discardable integer attribute named "clift.radix".
+struct ImmediateRadixDeductionPattern : mlir::OpRewritePattern<ImmediateOp> {
```
pfez, line 83: This `RewritePattern` is not handling how we emit case labels.

I'd like to have an `OpRewritePatter<SwitchOp>` that applies the same heuristic at all the case labels, then picks the radix used by the majority of them and attaches a `clift.radix` attribute to the `SwitchOp` itself. The attribute is finally used by the backend to print the literals of each switch case in the same way.

The `RewritePattern`s for `ImmediateOp`s and `SwitchOp`s must be separate of course, but I'd like them both to run as part of the same pass.

lib/CliftTransforms/Statements.cpp
```diff
@@ -0,0 +1,530 @@
...
+      invertBooleanExpression(Rewriter, If.getLoc(), If.getCondition());
+    }
+
+    // The region opposite the break-region, if not empty, must be inlined into
+    // the loop body.
+    mlir::Region &OtherRegion = ThenBreak ? If.getElse() : If.getThen();
+    if (not OtherRegion.empty()) {
+      mlir::Block *Body = &While.getBody().front();
+      inlineBlockBefore(Rewriter, &OtherRegion.front(), Body, Body->end());
+    }
```
pfez: Wait, doesn't this change semantics?

Before inlining the `OtherRegion` is executed conditionally to not breaking out of the loop. IIUC (and tests seem to confirm my understanding), after inlining `OtherRegion` is executed unconditionally, even if you then end up breaking out of the loop.

I think this shows nicely in the `do-while-conversion-with-hoisted-else.mlir` test.
There if `%0 = clift.imm 20` had side effects we'd be breaking semantics.

lib/CliftTransforms/Statements.cpp
```diff
@@ -0,0 +1,530 @@
...
+
+  void initialize() { setDebugName("trivial-jump-elimination"); }
+
+  using TrivialJumpVector = llvm::SmallVector<JumpStatementOpInterface>;
+
+  /// Collects trivial jumps targeting a specific label. \p LabelPosition
+  /// must point to the next non-label-assignment statement after the label
+  /// described by \p Label.
+  void findTrivialJumps(TrivialJumpVector &TrivialJumps,
+                        mlir::Value Label,
```
pfez, line 315: Shouldn't this be a `mlir::MakeLabelOp`?

lib/CliftTransforms/LabelMerging.cpp
pfez: > But now that I think about it, getFallthroughTarget just stops at the end BlockPosition, it doesn't actually fall through to the parent scope.

Ok, what I said here is false.

But the rest of the argument stands.

Let's talk about this "label motion". I honestly don't remember discussing about this. If we discarded it, what was the reason?

pfez: Off the top of my head. It seems in conflict with branch equalization.

lib/CliftTransforms/Statements.cpp
```diff
@@ -0,0 +1,530 @@
...
+  matchAndRewrite(mlir::Region &Region,
+                  mlir::PatternRewriter &Rewriter) const override {
+    revng_assert(Region.hasOneBlock());
+    mlir::Block *Outer = &Region.front();
+
+    auto Begin = Outer->begin();
+    auto End = Outer->end();
+
+    // Search backwards from the end of the region for a transformable branch
+    // operation. Only one branch is transformed per each pattern application.
```
pfez, line 234: What was the reason of this again?

lib/CliftTransforms/Statements.cpp
```diff
@@ -0,0 +1,530 @@
...
+        if (FallthroughRegion->empty())
+          FallthroughRegion->emplaceBlock();
+
+        // The selected statements are moved to the end of the fallthrough
+        // region.
+        mlir::Block *Inner = &FallthroughRegion->front();
+        Inner->getOperations().splice(Inner->end(),
+                                      Outer->getOperations(),
+                                      B,
+                                      E);
```
pfez, line 271: So this is basically
```c++
mlir::Block *ToMove = extractAsBlock(Outer, B, E);
inlineBlockBefore(Rewriter, ToMove, Inner, Inner->end());
```

A few comments.
1. `extractAsBlock` is currently in `LoopDetection.cpp` but it should go in `PatternRewriter.h`
2. Looking at `inlineBlockBefore` in `PatternRewriter.h`, why does it take a `PatternRewriter &` argument without using it?
3. Still on `inlineBlockBefore` in `PatternRewriter.h`, can't the `Dst` argument be computed by the `Pos` argument, making it redundant?

lib/CliftTransforms/Statements.cpp
```diff
@@ -0,0 +1,530 @@
...
+
+      // Find the unique fallthrough region of the branch operation, if any.
+      mlir::Region *FallthroughRegion = findUniqueFallthroughRegion(Branch);
+      if (FallthroughRegion == nullptr)
+        continue;
+
+      mlir::Block::iterator B = std::next(End);
+      mlir::Block::iterator E = Outer->end();
+
+      // It is not desirable to hoist any label assignments.
```
pfez: Why not? Expand the comment to explain.

lib/CliftTransforms/Statements.cpp
```diff
@@ -0,0 +1,530 @@
...
+      // Skip backwards over any trailing labels in the nesting region:
+      while (B != E and mlir::isa<AssignLabelOp>(&*std::prev(E)))
+        --E;
```
pfez, line 254: Also, I this misses some cases, like:
```mlir
// B points here
clift.assign_label %0
clift.anything_but_assign_label
clift.assign_label %1
// E initially points here
```
When you exit this while loop you have.
```mlir
// B points here
clift.assign_label %0
clift.anything_but_assign_label
// E points here
clift.assign_label %1
```

So you actually end up hoisting the first label assignment.

If you really want to avoid hoisting `clift.assign_label`, you should start from `B` and iterate forward until you find one.

But it becomes more important to understand why you don't want to hoist `AssignLabelOp`.

Also, what about other operations that assign labels, like loops. Would you want them to be hoisted or not?

lib/CliftTransforms/Statements.cpp
```diff
@@ -0,0 +1,530 @@
...
+                             OuterIf.getThen(),
+                             Rewriter)
+          .succeeded())
+      return mlir::success();
+
+    return mlir::failure();
+  }
+
+  mlir::LogicalResult
+  matchAndRewriteImpl1(IfOp OuterIf,
```
pfez: `matchAndRewriteImpl1` and `matchAndRewriteImpl2` are ugly names.
Let's do `matchAndRewriteOuter` and `matchAndRewriteInner`.

lib/CliftTransforms/Statements.cpp
```diff
@@ -0,0 +1,530 @@
...
+          Result = mlir::success();
+      }
+    }
+    return Result;
+  }
+};
+
+//===--------------------- Statement rewrite patterns ---------------------===//
+
+/// Rewrites nested if-statements into conditions with conjunctions.
```
pfez, line 57: So this is basically a very simple form rewriting nested if-statements to take advantage of short-circuiting.

We could think of doing something more advanced with the comma operator, where the inner if-statement is not required to be the only statement in its block but it's enough for it to be the last after a list of expression-statements.

Any particular reason why you haven't done that?

One thing that comes to mind is that you probably don't want to do that if the number of expression-statements before the inner if-statement is over a certain threshold. AFAICT the current implementation is equivalent to what I just described and just implicitly sets the threshold to 0, and for that reason is always guaranteed to never have to emit the comma operator.

tests/unit/mlir_lit_tests/statement-rewrites/trivial-jump-elimination.mlir
```diff
@@ -0,0 +1,85 @@
...
+// COM:      // CHECK: clift.expr
+// COM:      clift.expr {
+// COM:        clift.yield %arg0 : !int32_t
+// COM:      // CHECK: }
+// COM:      }
+// COM:
+// COM:      // CHECK-NOT: clift.goto
+// COM:      clift.goto %label_2
+// COM:    // CHECK: }
+// COM:    }
```
pfez: :thinking: what?

include/revng/CliftTransforms/BooleanNegations.td
```diff
@@ -0,0 +1,51 @@
...
+
+#ifndef MLIR_CLIFT_BOOLEAN_NEGATION_REWRITES
+#define MLIR_CLIFT_BOOLEAN_NEGATION_REWRITES
+
+include "mlir/IR/PatternBase.td"
+
+include "revng/Clift/Clift.td"
+include "revng/CliftTransforms/ExpressionHelpers.td"
+
+// !(x == y) -> x != y
```
pfez, line 13: It just occurred to me that we're not doing De Morgan.

I mean:
```c
!(!a && !b) -> a || b
!(!a || !b) -> a && b
```
Is that on purpose?
Wouldn't the real world examples you've look at benefit from that?
Or even if we don't do De Morgan explicitly that emerges by a composition of all the other transforms we do?

include/revng/CliftTransforms/BooleanNegations.td
```diff
@@ -0,0 +1,51 @@
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
```
pfez, line 2: In general all the transforms in this file feel like they should be part of canonicalization.

lib/CliftTransforms/LabelMerging.cpp
vasama-work: We didn't discuss it. I didn't see much need for it, so I didn't put it in the REP. Branch equalisation does not move trailing label assignments.

lib/CliftTransforms/Statements.cpp
vasama-work: Yeah, you're right. Good catch!

lib/CliftTransforms/Statements.cpp
vasama-work: You mean `clift::MakeLabelOp`. But why do you think so? The label value is enough.

include/revng/CliftTransforms/BooleanNegations.td
```diff
@@ -0,0 +1,51 @@
...
+  : Pat<(Clift_LogicalNotOp (Clift_CmpLeOp $lhs, $rhs)),
+        (Clift_CmpGtOp $lhs, $rhs)>;
+
+// !(x >= y) -> x < y
+def NotCmpGePattern
+  : Pat<(Clift_LogicalNotOp (Clift_CmpGeOp $lhs, $rhs)),
+        (Clift_CmpLtOp $lhs, $rhs)>;
+
+// !!x -> x
+//        x ? 1 : 0
```
pfez, line 44: A few comments on this.

1. We probably need also `!!!x -> !x`, and we need that to have "higher priority" than `!!x -> x ? 1 : 0`
2. This was the result of me insisting that we should get rid of double negation, and you dealing with the limits of patter rewriting. The result is fine, but I don't think it's enough.
  I think legalization should do something about this. Legalization already does context-sensitive stuff in boolean tested context. I think it should start converting `x ? 1 : 0` to just `x` in those contexts as well.

lib/CliftTransforms/Statements.cpp
vasama-work: It's region-based, because the pattern (wrt. the transformed branch operation) is context-dependent. 

There's no need to equalise multiple branches at once, and it's easier to reason about when it's just one at a time.

lib/CliftTransforms/Statements.cpp
vasama-work: Yes, it's equivalent, but we don't need to create a new block and then immediately destroy it. Better to just move the operations.

1. Maybe, but for now there are no other users. I can move it if you want. 🤷 
2. This is a member function of `PatternRewriter` (actually `RewriterBase`) in upstream.
3. I'm not sure. Probably not... Even if it can, I didn't design this API.

lib/CliftTransforms/Statements.cpp
vasama-work: I'm clarifying the comment, but I'll explain here too. I don't want to hoist *trailing* label assignments, because the last valid position for a label assignment can be considered most canonical. This also relates to the "label motion" transform that I meantioned wrt. label merging.

include/revng/CliftTransforms/Expressions.td
```diff
@@ -0,0 +1,130 @@
...
+
+// x != 0 -> x
+//           x ? 1 : 0
+// Note: The choice of replacement depends on whether the result is
+//       boolean-tested.
+def CmpNeWithZeroPattern
+  : Pat<(Clift_CmpNeOp:$res $arg, (Clift_ImmediateOp:$imm $value)),
+        (forceBooleanResult $res, $arg),
+        [(Constraint<CPred<"$0.getValue().isZero()">> $value)]>;
+
```
pfez, line 29: Why aren't these in `BooleanNegations.td`?

include/revng/CliftTransforms/Expressions.td
```diff
@@ -0,0 +1,130 @@
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
```
pfez, line 2: LLVM IR often uses `xor` with `0xFFFFFFFF` to mean bitwise negation, because bitwise negation isn't an operation in LLVM IR.

We need a rewrite pattern for this, because I want to see `~x` in C, not `x ^ 0xFFFFFFFF`.

include/revng/CliftTransforms/BooleanNegations.td
vasama-work: Just didn't think of it. Good catch.

> Or even if we don't do De Morgan explicitly that emerges by a composition of all the other transforms we do?

I don't think so.

We could add it to the REP, but... In what shape? Do we want to specifically only target the form you've shown, or:
```
!(x && y) -> !x || !y
!(x || y) -> !x && !y
```
Maybe even this?
```
!x && !y -> !(x || y)
!x || !y -> !(x && y)
```
Probably not the latter, but in isolation it can be hard to say which one is better.

include/revng/CliftTransforms/Expressions.td
vasama-work: `BooleanNegations.td` specifically deals with `!e`, because we insert a lot of boolean negations. We don't insert comparisons with zero.

include/revng/CliftTransforms/Expressions.td
vasama-work: Sure, that's easy enough.

tests/unit/mlir_lit_tests/statement-rewrites/trivial-jump-elimination.mlir
vasama-work: Mhm... The question is whether this jump is trivial and should be eliminated:
```c
L:
  while (C) {
    A;
    goto L;
  }
```
Until we actually see this, probably not...

include/revng/CliftTransforms/ExpressionHelpers.td
```diff
@@ -0,0 +1,48 @@
...
+//
+
+#ifndef MLIR_CLIFT_EXPRESSION_HELPERS
+#define MLIR_CLIFT_EXPRESSION_HELPERS
+
+include "mlir/IR/PatternBase.td"
+
+def ValuesAreEqual : Constraint<CPred<[{ $0 == $1 }]>>;
+
+def ValuesAreNotEqual : Constraint<CPred<[{ $0 == $1 }]>>;
```
pfez: `==`? I suppose you meant `!=`

lib/CliftTransforms/Statements.cpp
vasama-work: The current rewrite always obviously makes the code better. With commas it's not as obvious. We could do commas, but I would certainly want to think about it more. Do you feel that it's important to do it right now?

include/revng/CliftTransforms/ExpressionHelpers.td
vasama-work: Yes. Good catch.

include/revng/CliftTransforms/BooleanNegations.td
vasama-work: 1. That sounds very suspect to me. The problem is that we lack the ability to properly reason about boolean values. All of this is just adding more and more terrible hacks. I think if we want to keep these for now, we should introduce a lightweight boolean typing solution. I think we could do it in a way that doesn't require any model changes.
2. Legalisation is not the right place to do this. This has nothing to do with legalisation.

lib/CliftTransforms/Statements.cpp
pfez: No this was just hypothetical. Let's leave commas out of the picture.

tests/unit/mlir_lit_tests/statement-rewrites/trivial-jump-elimination.mlir
pfez: I see. Then drop the comments from the test please.

include/revng/CliftTransforms/Expressions.td
pfez: Ok add it then.

lib/CliftTransforms/Statements.cpp
pfez: 1. Yes
2. and 3. I get that you're getting this from upstream, but either you cherry-pick in LLVM (and we get whatever API comes from upstream), or you manually import part of the code (like you're doing here). In the latter case adapt the API to whatever makes sense for us.

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
pfez: With such a simple example it doesn't seem to make much of a difference.

But if the then and the else are big, and contain nested code, hoisting is always preferable.

Please hoist the else in case of score parity.

lib/CliftTransforms/Statements.cpp
vasama-work: I wanted to match the upstream API as closely as possible,
1. because the upstream API is exactly what we need, and
2. in order to make it very easy to upgrade in the future:

```cpp
function(Rewriter, Whatever)
// becomes
Rewriter.function(Whatever)
```

The rewriter being unused does raise some questions though. We really should be using it to notify the rewriter of updates to the relevant operations. I've changed the code to do that.

lib/CliftTransforms/Statements.cpp
vasama-work: Done.

include/revng/CliftTransforms/Expressions.td
vasama-work: Done.

lib/CliftTransforms/Statements.cpp
vasama-work: Done.

lib/CliftTransforms/ImmediateRadixDeduction.cpp
vasama-work: I added a rewrite pattern for deducing switch radices.

include/revng/Clift/Clift.td
vasama-work: I added a backend unit test for this.

lib/CliftTransforms/Expressions.cpp
```diff
@@ -24,6 +24,12 @@ using namespace clift;
 
 namespace {
 
+static bool areAllBitsSet(llvm::APInt Value, mlir::Type Type) {
+  unsigned BitWidth = getUnderlyingIntegerType(Type).getSize() * 8;
+  uint64_t MinusOne = static_cast<uint64_t>(-1) >> (64 - BitWidth);
+  return Value.getZExtValue() == MinusOne;
```
pfez: `llvm::APInt`s have `isAllOnes` and `getAllOnes` methods.

Can't you just:
```c++
return Value.isAllOnes();
```
?

include/revng/CliftTransforms/Expressions.td
```diff
@@ -0,0 +1,130 @@
...
+// Constraint: the cast kinds are the same.
+// Constraint: the cast kind is collapsible (implemented in .cpp file).
+def CollapsibleCastPattern
+  : Pat<(Clift_CastOp:$res (Clift_CastOp $arg, $inner_kind), $outer_kind),
+        (Clift_CastOp $arg, $inner_kind),
+        [(ValuesAreEqual $inner_kind, $outer_kind),
+         (Constraint<CPred<"isCollapsibleCastKind($0.getValue())">> $inner_kind)]>;
+
+// *(T*)p = x ->         *p = (T)x
+//            -> *(T*)&( *p = (T)x )
```
pfez, line 93: This rewrite is worse than the original form.
The problem is that you can't put a constraint on the fact that the assignment is discarded.
Just drop this rewrite from the PR and from the associated REP.

include/revng/CliftTransforms/Expressions.td
```diff
@@ -0,0 +1,130 @@
...
+        (replaceWithValue $arg)>;
+
+// (T)x -> x
+// Constraint: the type of x is equivalent to T.
+def RedundantCastPattern
+  : Pat<(Clift_CastOp:$res $arg, $kind),
+        (replaceWithValue $arg),
+        [(TypesAreEquivalent $res, $arg)]>;
+
+// (T)(U)x -> (T)x
```
pfez, line 83: We're missing a pattern that rewrites a widening cast followed by a truncation as a single direct cast.

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
pfez: `double` makes it seem like we put some thought into it, which isn't true.
Put an integer for now.

lib/CliftTransforms/ImmediateRadixDeduction.cpp
```diff
@@ -101,6 +101,42 @@ struct ImmediateRadixDeductionPattern : mlir::OpRewritePattern<ImmediateOp> {
...
+        BestRadix = Radix;
+        BestRadixCount = Count;
+      }
+    }
+
+    // TODO: One walkAndApplyPatterns is available, it will not be necessary to
+    //       add the attribute when the selected radix is 10.
+    auto RadixValue = llvm::APSInt(llvm::APInt(32, BestRadix));
+    auto RadixAttr = mlir::IntegerAttr::get(getContext(), RadixValue);
+    Switch->setAttr("clift.radix", RadixAttr);
```
pfez, line 134: We need the counterpart for this in `CBackend.cpp`

lib/CliftTransforms/Statements.cpp
pfez: I don't necessarily agree that we don't want to move that. But it's not urgent, and at least now I understand your reasoning. So it's fine for now, and we'll see from real world examples and C QA if we need this.

include/revng/CliftTransforms/BooleanNegations.td
pfez: 1. Nevermind. It did not occur to me that `!!!x -> !x` is already handled by this very same rewrite pattern, and that in that case it will trigger first in the innermost, which is always in boolean-tested context, so it will always pick the nice rewrite (not the one with the ternary).
2. Ok putting stuff in legalization is a reflex that I have to suppress :). In any case, I can't see the ugly `x ? 1 : 0`. It has to go away. You can't constain a match only on boolean-tested contexts? Fine. Then don't do it as a rewrite pattern. Let's just make a standalone pass that looks over all the boolean contexts and if the condition is a ternary with 1 and 0 promote it to `x`. Then schedule that after iterative expression rewriting, so we can get rid of the most obvious offenders.

include/revng/CliftTransforms/BooleanNegations.td
pfez: At least definitely the form I've shown, and definitely not the last form you've shown.

As for these forms
```
!(x && y) -> !x || !y
!(x || y) -> !x && !y
```
I have mixed feelings.
In particular, is pushing the `!` towards the leafs of the expression always beneficial?
On the surface it would seems so, because it would then enable stuff like `!(a==b)` to be turned into `a != b` if that is a leaf.
But on the other hand it could quickly increase the number of overall negations you need in the whole expression.
In principle, one could think of doing the opposite, e.g.: `!x || !y -> !(x && y)`, which would in turn enable other optimizations like `!(!x || !y) -> !!(x && y)` which is then turned into `x && y` by the other pattern (`!!x -> x`).

So overall, I think there isn't a clear cut answer on these forms, and for this reasons let's leave them out for now.

include/revng/CliftTransforms/BooleanNegations.td
pfez: 1. It just occurred to me that `!!x -> x` can always be done (instead of injecting the `?`) if `x` is a boolean expression or a comparison already. `forceBooleanResult` doesn't do that. It just checks in which context the expression is used, which is only a part of the story. The same should hold for the other rewrite that sometimes emits `x ? 1 : 0` which matches `x != 0`.

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
vasama-work: Done.

include/revng/CliftTransforms/BooleanNegations.td
vasama-work: Yes, but that doesn't seem terribly helpful. We have three classes of boolean-returning operations:
1. `!` itself; not helpful.
2. Comparisons; we already handle negations of these separately.
3. `&&` and `||`; this goes back to the De Morgan discussion above.

lib/CliftTransforms/Expressions.cpp
vasama-work: That will fail to match immediates narrower than 64 bits. `0x00000000FFFFFFFF` is not all ones. The immediate value is `uint64_t`.

include/revng/CliftTransforms/Expressions.td
vasama-work: As we discussed before, the Clifter doesn't emit chained assignments, so these expressions will never be rewritten into the second form in code coming from the Clifter. In practice this rewrite makes a big positive difference in the C output quality.

lib/CliftTransforms/ImmediateRadixDeduction.cpp
vasama-work: Yes, forgot about that. Good catch.

include/revng/CliftTransforms/BooleanNegations.td
vasama-work: Done.

include/revng/CliftTransforms/Expressions.td
vasama-work: I implemented this and tested on calc. It was never matched. It's also not as simple as eliminating the inner cast, because that could make the outer cast invalid. So I think until we need this, we should hold off on it.

include/revng/CliftTransforms/Expressions.td
vasama-work: I think just like lvalue-to-rvalue conversion and scalar-to-boolean conversion, this case really calls for an explicit discard operation. `clift.discard`, which takes anything and has type `void`. Then we could rewrite `discard(*(T*)p = x)` into `discard(*p = (T)x)`.

include/revng/CliftTransforms/BooleanNegations.td
vasama-work: I changed `forceBooleanResult` to check whether the rewritten expression is a boolean expression.

pfez: Makes sense.

That leaves a one-shot pass that takes care of `x ? 1 : 0` in boolean evaluated contexts after expression rewrites and we are done.

include/revng/CliftTransforms/Expressions.td
pfez: > I implemented this and tested on calc. It was never matched

Odd. But ok. Let's leave this out.

lib/CliftTransforms/LabelMerging.cpp
pfez: Yeah it was. Closing this now anyway.

include/revng/CliftTransforms/Passes.td
vasama-work: I added four tests.

lib/CliftTransforms/Statements.cpp
pfez: > The rewriter being unused does raise some questions though

Yeah that was the main smell.

> I've changed the code to do that.

Thanks.

lib/CliftTransforms/LabelMerging.cpp
pfez: Yeah ok, we don't want to move labels for now. Closing this.

lib/CliftTransforms/Statements.cpp
pfez: Don't we have a guarantee that `Label` is always a `MakeLabelOp`?

include/revng/CliftTransforms/BooleanNegations.td
pfez: > That leaves a one-shot pass that takes care of x ? 1 : 0 in boolean evaluated contexts after expression rewrites and we are done.

What about this?

lib/CliftTransforms/Expressions.cpp
pfez: The immediate value is always `uint64_t`?

Then `return Value.trunc(BitWidth).isAllOnes();` ?

lib/CliftTransforms/Statements.cpp
vasama-work: Yes, but this function only needs the label value.

include/revng/CliftTransforms/BooleanNegations.td
vasama-work: Looking at calc, this was never even matched, so it's going to be very rare. I think we shouldn't put any more effort into hacks for now.

lib/CliftTransforms/Expressions.cpp
vasama-work: Yeah, that works. I changed it.

include/revng/CliftTransforms/BooleanNegations.td
pfez: Fair enough. We have much bigger fish to fry. Closing this.

PR #546, opened 2025-11-18
pypeline: add advanced invalidation

mrjackv:


libexec/revng/check-conventions
```diff
@@ -636,7 +636,7 @@ class CheckCommitPass(Pass):
         if re.search(r"\bwip\b", subject, re.IGNORECASE):
             problems.append("Commit subject line contains 'WIP'")
 
-        if subject[0] in "([{":
+        if (p_index := "([{".find(subject[0])) >= 0 and subject[-1] == ")]}"[p_index]:
```
aleclearmind: 1. Obscure code.
2. Why are we doing this? Why are we checking the last character?

include/revng/PipeboxCommon/Helpers/Python/RunAnalysis.h
```diff
@@ -18,14 +18,17 @@ inline llvm::Error runAnalysis(T &Handle,
...
   Model *CppModel = nanobind::cast<Model *>(TheModel);
+  auto ContainerTuple = makeContainerTuple<T>(Containers);
 
-  return revng::pypeline::helpers::runAnalysis(Handle,
-                                               *CppModel,
-                                               Incoming,
-                                               Configuration,
-                                               Containers);
+  {
+    nanobind::gil_scoped_release X;
```
aleclearmind, line 25: This commit is not about "factoring out" something, is about introducing `gil_scoped_release`.

include/revng/PipeboxCommon/Helpers/Helpers.h
```diff
@@ -4,13 +4,22 @@
...
-#include <cstdlib>
+#include "revng/PipeboxCommon/Concepts.h"
 
 namespace revng::pypeline::helpers {
 
-template<typename C, size_t I, typename ListType>
-struct ExtractContainerFromList {
-  static C &get(ListType &Containers);
-};
+/// Helper fuVnction that simplifies unpacking pipe's container list into a
```
aleclearmind: Typo

include/revng/PipeboxCommon/Helpers/Helpers.h
```diff
@@ -4,13 +4,22 @@
...
+/// well-typed tuple of references, that can be later run with `runPipe` or
+/// `runAnalysis`.
+template<typename T, typename CallableT>
+  requires(IsPipe<T> or IsAnalysis<T>)
+auto makeContainerTuple(CallableT &&Callable) {
```
aleclearmind: 1. `makeContainerTuple` is not a great name, it should be more explicit about the fact that it builds a tuple *given a pipe* (or an analysis).
2. The fact that you use the same name for this and the other two functions makes the code obscure.
3. Try to rewrite this with more decoupled:

```c++
// Implement for vector<Container *> and for nanobind::list
template<typename ContainerType>
struct ExtractFromContainerTrait {
  template<size_t Index, typename ExpectedType>
  ExpectedType &get(ContainerType &C);
};

template<IsTupleOfReferences Tuple,
         typename ContainerType>
Tuple vectorToTuple(ContainerVector &Containers) {
  // I know you can't do this here, but with an helper function
  // where you expand the Is, it should be doable.
  return Tuple(ExtractFromContainerTrait::get<I, std::tuple_element<I, Tuple>>(Containers)...);
}
```

Advantages: a single `vectorToTuple` function, no lambdas, decouple from the analysis/pipes.
Then you'll need to make a function that given a pipe/analysis calls this with the appropriate tuple.

libexec/revng/check-conventions
mrjackv: 2. Otherwise check-conventions complains about the following subject `{Pipe,Analysis}Run: factor out container extract` because it begins with `{`

aleclearmind: I still don't understand why you're checking the last character.

I'm not sure what you're trying to cover here.
This should prevent from using WIP in the commit title, I'd just prevent from using WIP entirely there. If you want we can also ban "fixup".

`\b(wip|fixup)\b`?

mrjackv: > This should prevent from using WIP in the commit title

That's not the check in question, that one is the previous check which greps if there is `WIP` in the subject

> I still don't understand why you're checking the last character.

This check is intended for "temporary" commits, such as:
```
(do not push)
(End of commits from feature/XXXXX)
(SQUASH: XXXXX)
```

The check previously only checked the first character, where here I'm checking if the first character is `([{` **and** if there is a matching `)]}` at the end

aleclearmind: I'm not sure why we have such check. The "temporary" commits you describe is not a well-known convention in revng. I'd drop it.

mrjackv: Ok, I'll drop the check

mrjackv: Dropped

include/revng/PipeboxCommon/Helpers/Python/RunAnalysis.h
mrjackv: Reworked the commit message to reflect this

include/revng/PipeboxCommon/Helpers/Helpers.h
mrjackv: Fixed

include/revng/PipeboxCommon/Helpers/Helpers.h
mrjackv: Used `containerVectorToTuple` for the function name, recycled the `ExtractContainerFromList` struct that was previously used by `runPipe` and `runAnalysis`.

include/revng/PipeboxCommon/Helpers/Python/RunAnalysis.h
aleclearmind: The message now says "allow". Please be super-explicit that we're introducing the lock.

Something like `{Pipe,Analysis}Run: release Python GIL`.

Touching locks/multi-threading is something very delicate, put the right weight on it.

python/revng/pypeline/schema.yml
```diff
@@ -17,6 +17,18 @@ $defs:
...
+        items:
+          type: object
+          required:
+            - analysis
+          additionalProperties: false
+          properties:
+            analysis:
+              type: string
+              description: The name of the analysis to run.
+        minItems: 1
```
aleclearmind: I'd drop `minItems: 1`.

python/revng/pypeline/schema.yml
```diff
@@ -17,6 +17,18 @@ $defs:
         uniqueItems: true
         items:
           $ref: "#/$defs/container_decl"
+      analyses:
```
aleclearmind, line 22: Maybe we should call these "pure analyses" or "inputless analyses".

If we just say analyses one might think he can find all of them here.

python/revng/pypeline/schema.yml
```diff
@@ -17,6 +17,18 @@ $defs:
...
+        type: array
+        items:
+          type: object
+          required:
+            - analysis
+          additionalProperties: false
+          properties:
+            analysis:
+              type: string
+              description: The name of the analysis to run.
```
aleclearmind: Why not just a list of strings?

python/revng/internal/pipeline.yml
```diff
@@ -28,6 +28,12 @@ containers:
   - name: llvm-root-with-functions
     type: LLVMRootContainer
 
+analyses:
+  - analysis: ApplyDiff
+  - analysis: VerifyDiff
+  - analysis: SetModel
+  - analysis: VerifyModel
```
aleclearmind: kebab-case?

include/revng/PipeboxCommon/ModelAnalyses.h
```diff
@@ -0,0 +1,73 @@
+#pragma once
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "revng/PipeboxCommon/Model.h"
+
+struct ApplyDiff {
```
aleclearmind: Use a `class`. We use `struct` for stuff we want to have all fields public, typically simple data structures.

include/revng/PipeboxCommon/ModelManipulationAnalyses.h
aleclearmind, line 1: 1. Could we implement these analyses in Python? Not necessarily today.
2. `s/ModelAnalyses/BasicAnalyses/` or `ModelManipulationAnalyses`.

include/revng/PipeboxCommon/ModelAnalyses.h
```diff
@@ -0,0 +1,73 @@
...
+
+    if (not MaybeModel->verify())
+      return revng::createError("SetModel failed: model failed to verify");
+
+    Model.get() = *MaybeModel;
+    return llvm::Error::success();
+  }
+};
+
+struct VerifyModel {
```
aleclearmind: What do we use this one for?

include/revng/PipeboxCommon/ModelAnalyses.h
```diff
@@ -0,0 +1,73 @@
...
+      return MaybeDiff.takeError();
+
+    if (llvm::Error Error = MaybeDiff->apply(Model.get()); Error)
+      return Error;
+
+    return llvm::Error::success();
+  }
+};
+
+struct VerifyDiff {
```
aleclearmind: What do we use this one for?

include/revng/PipeboxCommon/Common.h
```diff
@@ -78,4 +74,22 @@ class Buffer {
...
+  std::vector<std::pair<ObjectID, ModelPath>>>;
+
+/// Type used to return advanced invalidation data from a pipe run. Will be fed
+/// back to the pipe's `invalidate` function to determine if there are
+/// additional objects to invalidate.
+using AdvancedInvalidation = std::vector<
+  std::vector<std::tuple<ObjectID, Buffer>>>;
+
+/// Type representing the output of a `Pipe::run` method
+using PipeOutput = std::tuple<ObjectDependencies, AdvancedInvalidation>;
```
aleclearmind: Switch to `struct`s, no point in using pairs like this.

include/revng/PipeboxCommon/Common.h
```diff
@@ -78,4 +74,22 @@ class Buffer {
...
 
+/// Type used to return the dependencies of objects produced by a pipe
+/// The first index maps to the i-th container of the pipe
+using ObjectDependencies = std::vector<
+  std::vector<std::pair<ObjectID, ModelPath>>>;
+
+/// Type used to return advanced invalidation data from a pipe run. Will be fed
+/// back to the pipe's `invalidate` function to determine if there are
+/// additional objects to invalidate.
+using AdvancedInvalidation = std::vector<
```
aleclearmind: Let's drop "advanced" entirely in favor of "custom". For instance, `CustomInvalidationData` (I know I don't like data, but here the content is opaque, so I don't dislike it too much).

Explain why `vector<vector<...>>`.

include/revng/PipeboxCommon/Helpers/PipeRunPipes/Base.h
```diff
@@ -58,6 +58,26 @@ struct CheckPreconditionMixin<T> {
...
+                                          &ID,
+                                        const ModelDiff &Diff) {
+  { T::invalidate(ID, Diff) } -> std::same_as<std::vector<std::set<ObjectID>>>;
+};
+
+template<typename T>
+struct InvalidateMixin {};
+
+template<HasPipeRunInvalidate T>
+struct InvalidateMixin<T> {
```
aleclearmind: This only seems to have a single usage, you can inline it and use `requires` instead.

```c++
template<typename T>
class Asd {
    int lol() requires false { return 42; }
};
```

python/revng/pypeline/model.py
```diff
@@ -15,6 +15,20 @@
 ModelPathSet = Set[ModelPath]
 
 
+class ModelDiff(ABC):
+    """
+    Abstract class representing a diff between two model classes
```
aleclearmind: two model instances

python/revng/internal/pipebox.py
```diff
@@ -57,9 +57,9 @@ def run(
         incoming: list[ObjectSet],
         outgoing: list[ObjectSet],
         configuration: Configuration,
-    ) -> PipeObjectDependencies:
+    ) -> PipeRunOutput:
```
aleclearmind: `PipeRunOutput` is too empty as a term.

What about just `PipeDependencies`?

python/revng/pypeline/storage/local_provider.py
```diff
@@ -504,3 +557,40 @@ def _compare_file(path: Path, request: FileRequest) -> bool:
...
+                            pipe_id,
+                            configuration_hash,
+                            index,
+                            object_id.to_bytes(),
+                            bytes(invalidation_blob),
+                        ),
+                    )
+            self._write_metadata(cursor)
+
+    def get_advanced_dependencies(
```
aleclearmind: This does not return dependencies but opaque data that enable you to compute dependencies.

`get_custom_invalidation_data`

python/revng/pypeline/pipeline.py
```diff
@@ -440,7 +441,39 @@ def run_analysis(
             incoming=[requests.get(decl) for decl in analysis_info.bindings],
             configuration=analysis_configuration,
         )
-        invalidated = storage_provider.invalidate(ReadOnlyModel(new_model).diff(model))
+
+        diff = model.diff(ReadOnlyModel(new_model))
+
+        adv_invalidated_objects: list[ObjectsToInvalidate] = []
```
aleclearmind: 1. No abbreviations
2. Outline the handling of custom invalidation in a function, makes the function more readable

python/revng/pypeline/pipeline.py
```diff
@@ -440,7 +441,39 @@ def run_analysis(
...
+        for node in self.walk_pipeline():
+            if not isinstance(node.task, Pipe):
+                continue
+
+            assert node.savepoint_range is not None
+
+            # Optimization: since few pipes actually implement advanced
+            # invalidation and it requires querying storage, check that the
+            # method has actually been overridden.
+            if node.task.__class__.invalidate is Pipe.invalidate:
```
aleclearmind: Hide this in `pipe.has_custom_invalidation`

python/revng/pypeline/pipeline_node.py
```diff
@@ -236,7 +237,21 @@ def run(
                         f"{index_deps}"
                     )
                 result.extend((self.bindings[index].name, obj, path) for obj, path in index_deps)
-            return result
+
+            if not all(len(x) == 0 for x in pipe_output.advanced_invalidation):
```
aleclearmind: Comment

python/revng/pypeline/task/pipe.py
```diff
@@ -2,15 +2,57 @@
...
+    A list representing the dependencies between the an object (in a certain container) produced
+    by a certain task and the model.
+    """,
+]
+
+
+@dataclass
+class PipeOutput:
+    dependencies: ObjectDependencies
+    advanced_invalidation: PipeAdvancedInvalidation = field(default_factory=list)
```
aleclearmind:
```
PipeOutput -> ScheduledTaskDependencies
PipeRunOutput -> PipeDependencies
```

include/revng/Lift/Lift.h
```diff
@@ -176,11 +176,17 @@ class LiftPass : public llvm::ModulePass {
   bool runOnModule(llvm::Module &M) override;
 };
 
-namespace detail {
+namespace lift::internal {
```
aleclearmind: Our namespaces, except for `detail`, always start with `revng::`.

include/revng/PipeboxCommon/Helpers/Python/RunAnalysis.h
mrjackv: Reworked the commit message to be extra clear on what's happening

lib/Lift/Lift.cpp
```diff
@@ -102,13 +102,28 @@ bool LiftPass::runOnModule(llvm::Module &M) {
...
+  static void inputOne(IO &IO, StringRef Key, JumpTargetMap &Data) {
+    MetaAddress Address = MetaAddress::fromString(Key);
+    IO.mapRequired(Key.str().c_str(), Data[Address]);
+  }
+
+  static void output(IO &IO, JumpTargetMap &Data) {
+    for (auto &[Key, Value] : Data)
+      IO.mapRequired(Key.toString().c_str(), Value);
+  }
+};
```
aleclearmind, line 122: We have `llvm::yaml::ScalarTraits` for `MetaAddress`, I don't think you need to handle this by hand with `CustomMappingTraits`.

aleclearmind:
@mrjackv please add a test for the advanced/custom invalidation feature in Python.

include/revng/PipeboxCommon/ModelAnalyses.h
mrjackv: The `Verify*` analyses are there in case someone that doesn't have access to revng directly (e.g. `revng-ui`, scripting) needs to verify that a model/diff is valid

include/revng/PipeboxCommon/ModelAnalyses.h
mrjackv: See answer on the previous comment

lib/Lift/Lift.cpp
mrjackv: The traits are not for `MetaAddress`, they are for the whole `std::map<MetaAddress, bool>` class. I can look to dispatch the `MetaAddress` logic to the `ScalarTraits` but it looks a lot of extra adapting for marginal gains

python/revng/pypeline/schema.yml
mrjackv: 1. Just in case in the future we need to add extra keys, e.g. `Docs` we don't have to convert the list of strings into a list of objects
2. It allows re-cycling the parsing logic for the non-container analyses

include/revng/PipeboxCommon/ModelAnalyses.h
aleclearmind: Can't they just try to apply it and report back the error in case of failure?

python/revng/internal/pipeline.yml
mrjackv: Pipes are so far `PascalCase`, I'd stick with that since it reflects the class name of the analysis as well.

include/revng/PipeboxCommon/ModelManipulationAnalyses.h
mrjackv: 1. Hypotetically yes, but it would be very expensive because the model would need to be serialized, re-parsed in python, the modifications made and then re-parsed back in C++
2. Ok

lib/Lift/Lift.cpp
```diff
@@ -100,3 +101,106 @@ bool LiftPass::runOnModule(llvm::Module &M) {
 
   return false;
 }
+
+using JumpTargetMap = std::map<MetaAddress, bool>;
```
aleclearmind, line 109: What's the semantics of the `bool`? Docs.

lib/Lift/Lift.cpp
aleclearmind: OK

python/revng/internal/pipeline.yml
aleclearmind: These become available on the CLI, and there we must use kebab-case.

include/revng/PipeboxCommon/ModelAnalyses.h
mrjackv: The semantic is different:
* `ApplyDiff`: I want to apply the diff and check the error if it doesn't apply (will modify the model if the diff is well-formed)
* `VerifyDiff`: would this diff hypotetically apply to this model (does not actually modify the model if the diff is well-formed)
* `SetModel`: I have a model file, please make it the model of the project
* `VerifyModel`: verify the provided model (basically `model::verify` but can be called from the daemon)

The semantic is of `ApplyDiff` and `SetModel` is to modify the project if everything is well-formed, this is not the same as `Verify{Diff,Model}` which do guarantee that no matter what the current model is not changed.

python/revng/pypeline/schema.yml
aleclearmind: This is (will be) a private schema, no need to think to backward compatibility.

To me it's needlessly cumbersome, if we add stuff we'll fix this, let's not do thing just in case.

include/revng/PipeboxCommon/ModelAnalyses.h
aleclearmind: Do we actually use these in the UI? If so, why?

mrjackv: We don't, but not having them also means that potential users of scripting do not have access to them.

include/revng/PipeboxCommon/Common.h
mrjackv: Done

include/revng/PipeboxCommon/Helpers/PipeRunPipes/Base.h
mrjackv: Done

python/revng/pypeline/model.py
mrjackv: Fixed

python/revng/internal/pipebox.py
mrjackv: Renamed

python/revng/pypeline/storage/local_provider.py
mrjackv: Renamed to `add_custom_invalidation_data` and `get_custom_invalidation_data`

python/revng/pypeline/pipeline.py
mrjackv: Done

python/revng/pypeline/task/pipe.py
mrjackv: Renamed

include/revng/Lift/Lift.h
mrjackv: Added `revng::`

lib/Lift/Lift.cpp
mrjackv: Added doxygen

python/revng/pypeline/schema.yml
mrjackv: Dropping it is marginal, it allows writing `analyses: []` and it's not consistent with the rest of the schema

python/revng/pypeline/schema.yml
mrjackv: Given what you said below:

> This is (will be) a private schema, no need to think to backward compatibility.

I think it makes more sense to leave it as `analyses`. There's already schema constraints/assertions in place that forbid this from having containers and the pipe/savepoint ones from not having any

python/revng/pypeline/schema.yml
mrjackv: Had to re-add the `name` key, closing since this comment is no longer valid.

python/revng/internal/pipeline.yml
mrjackv: Re-added the `name` key, analyses are now `kebab-case`.

include/revng/PipeboxCommon/ModelAnalyses.h
mrjackv: Done

include/revng/PipeboxCommon/ModelManipulationAnalyses.h
mrjackv: 2. Renamed to `ModelManipulationAnalyses`

python/revng/pypeline/pipeline.py
mrjackv: 1. Done
2. Done

python/revng/pypeline/pipeline_node.py
mrjackv: Added comments throughout

python/revng/internal/pipeline.yml
aleclearmind: The name of the analysis itself should be kebab-case.
Let's keep the number of names to the minimum.
`VerifyModel` is the name of the `class`. The analysis is called `verify-model` and this should be the `Name` field as well.

mrjackv: That means also having the python-generated nanbind class name be `kebab-case` which is not valid python
Addendum: also pipes currently have `PascalCase` names, and those will show up when doing `revng2 pipeline run-pipe`, should those also be `kebab-case`?

aleclearmind: That's a distinct problem that has to be resolved when we register a pipe/analysis, possibly with an additional string or by making `RegisterPipe` a macro.

The name of the analysis out of Python and C++ (i.e., the CLI) has to be kebab-case and it's not a property of the pipeline, it's just how things must be.

mrjackv: Analysis names now reflect the `Name` attribute in C++, which is `kebab-case`

include/revng/PipeboxCommon/Common.h
mrjackv: Done

mrjackv:
> @mrjackv please add a test for the advanced/custom invalidation feature in Python.

Done

python/revng/pypeline/schema.yml
aleclearmind: I'd drop all `minItems: 1`. Schemas are supposed to validate semantics and having zero analysis is perfectly fine. Schemas are not for enforcing linting rules.

include/revng/PipeboxCommon/ModelAnalyses.h
aleclearmind: We should eventually devise a better solution to this. This is an hack.

Add a comment about this.

tests/pypeline/pipebox.py
```diff
@@ -532,6 +533,65 @@ def run(
...
+
+    def invalidate(
+        self, invalidation_data: PipeCustomInvalidation, diff: ModelDiff
+    ) -> list[ObjectSet]:
+        assert len(invalidation_data) == 1
+        assert len(invalidation_data[0]) == 1
+        object_, data = invalidation_data[0][0]
+        assert object_ == MyObjectID(MyKind.ROOT)
+        assert data == b"foo"
+        return [ObjectSet(MyKind.ROOT, {object_})]
```
aleclearmind, line 581: Comment

tests/pypeline/pipebox.py
```diff
@@ -532,6 +533,65 @@ def run(
...
+        return (RootDictContainer,)
+
+    def run(
+        self,
+        model: Model,
+        containers: list[Container],
+        incoming: list[ObjectSet],
+        configuration: str,
+    ):
+        # This analysis does nothing
```
aleclearmind: Comment why

tests/pypeline/test_all.py
```diff
@@ -707,3 +708,40 @@ def add_object(save_start: int, save_end: int, object_id, path: str):
         ContainerLocation(2, container_id, configuration_id): {function1},
         ContainerLocation(3, container_id, configuration_id): {root},
     }
+
+
+def test_custom_invalidation(model, storage_provider: StorageProvider):
```
aleclearmind, line 713: Comment this function

mrjackv: Done

tests/pypeline/pipebox.py
mrjackv: Done

tests/pypeline/pipebox.py
mrjackv: Done

include/revng/PipeboxCommon/ModelAnalyses.h
mrjackv: Added

python/revng/pypeline/schema.yml
mrjackv: Dropped `minItems`

PR #549, opened 2025-12-02
GenericRegion: improve `Head` election

andrealinux1:
The `Head` election now works in a bottom-up fashion (child regions are processed before the parent ones), and ensures that the `Head` election process is coherent across multiple nested regions. This means that:
1) The late entries are coherent between the nested regions. If a node
   is selected as a late entry for a region, the parent region must
   follow this decision too.
2) If a node which is a `Head` candidate for a parent region, is
   selected as the `Head` for a child region, the same choice must be
   performed for the parent region too.

### DIFF
--- a/lib/RestructureCFG/GenericRegionInfo.cpp
+++ b/lib/RestructureCFG/GenericRegionInfo.cpp
@@ -6,10 +6,12 @@
 #include "llvm/ADT/GenericCycleImpl.h"
 #include "llvm/ADT/GenericCycleInfo.h"
 #include "llvm/ADT/GraphTraits.h"
+#include "llvm/ADT/SmallVector.h"
 #include "llvm/IR/CFG.h"
 #include "llvm/IR/Function.h"
 #include "llvm/IR/SSAContext.h"
 
+#include "revng/RestructureCFG/GenericRegion.h"
 #include "revng/RestructureCFG/GenericRegionInfo.h"
 #include "revng/RestructureCFG/ScopeGraphGraphTraits.h"
 #include "revng/Support/Debug.h"
@@ -121,6 +123,23 @@ computeShortesPath(GraphT F) {
   return ShortestPathFromEntry;
 }
 
+/// Helper static function which computes the `Head` candidates for a given
+/// region
+template<class NodeT>
+static llvm::SmallVector<NodeT>
+getHeadCandidates(GenericRegion<NodeT> &Region) {
+  llvm::SmallVector<NodeT> HeadCandidates;
+  for (NodeT Block : Region.blocks()) {
+    for (NodeT Predecessor : graph_predecessors(Block)) {
+      if (not Region.containsBlock(Predecessor)) {
+        HeadCandidates.push_back(Block);
+      }
+    }
+  }
+
+  return HeadCandidates;
+}
+
 template<class GraphT, class GT>
 void GenericRegionInfo<GraphT, GT>::electHead(GraphT F) {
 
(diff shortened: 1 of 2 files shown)


lib/RestructureCFG/GenericRegionInfo.cpp
```diff
@@ -135,47 +154,72 @@ void GenericRegionInfo<GraphT, GT>::electHead(GraphT F) {
...
-      for (NodeT Block : CurrentRegion->blocks()) {
-        for (NodeT Predecessor : graph_predecessors(Block)) {
-          if (not CurrentRegion->containsBlock(Predecessor)) {
-            HeadCandidates[Block]++;
+      for (NodeT Block : getHeadCandidates(*CurrentRegion)) {
+        HeadCandidates[Block]++;
+      }
+
+      // Criterion 1
+      for (auto ChildRegion : CurrentRegion->children()) {
```
pfez, line 206: Does `CurrentRegion->children()` return only the directly nested regions or also the "grandchildren"?

lib/RestructureCFG/GenericRegionInfo.cpp
```diff
@@ -135,47 +154,72 @@ void GenericRegionInfo<GraphT, GT>::electHead(GraphT F) {
...
+
+        for (NodeT Block : getHeadCandidates(*ChildRegion)) {
+          if (Block != ChildHead) {
+            HeadCandidates.erase(Block);
+          }
+        }
+      }
+
+      // Criterion 2
+      for (auto ChildRegion : CurrentRegion->children()) {
```
pfez, line 228: Is `CurrentRegion->children()` returning child regions in a deterministic way?

lib/RestructureCFG/GenericRegionInfo.cpp
```diff
@@ -121,6 +123,23 @@ computeShortesPath(GraphT F) {
   return ShortestPathFromEntry;
 }
 
+/// Helper static function which computes the `Head` candidates for a given
+/// region
+template<class NodeT>
+static llvm::SmallVector<NodeT>
+getHeadCandidates(GenericRegion<NodeT> &Region) {
+  llvm::SmallVector<NodeT> HeadCandidates;
+  for (NodeT Block : Region.blocks()) {
```
pfez, line 132: Is the order of `Region.blocks()` deterministic?
If not, this will affect the order of `HeadCandidates`, and could lead to non-determinism in picking a region's head.

andrealinux1: The underlying data structure containing the blocks is a `SmallSetVector`, therefore given an initial DFS visit, the order of the `Blocks` should be deterministic. We strived to obtain this result, and we even have unit tests for `GenericRegion` that check the order of the blocks. The only thing that could break this assumption is if the initial DFS used for computing the `GenericRegion` changes, but that would lead to other failures before in tests.

We have two options here I think:
1) Maintaining the assumption that the blocks order is deterministic, or
2) decouple this assumption from this part of the code. That would mean to make this part of the code resistant to changes in the blocks order (but imagine that we use the post order to assign an order, I think that if the "initial" DFS changes also the post order could change).

lib/RestructureCFG/GenericRegionInfo.cpp
andrealinux1: We use a `llvm::SmallSetVector` to store the `children` here to, so the reasoning becomes "do we insert the children of a region in a deterministic way?". Here what I investigated:
- Under the hood, we initialize the `GenericRegion`s starting from the `GenericCycle`s provided by lllvm.
- I took a look a the algorithm computing it here https://llvm.org/docs/doxygen/GenericCycleImpl_8h_source.html. My understanding is that even there the `children` are stored in a deterministic way (in a `std::vector`), and the population order is also deterministic (or better, it is deterministic given the usual deterministic `dfs`, as for the other comment below).
- The thing that unsettle me a bit is the fact that the unit tests of llvm, they use the `CHECK-DAG` directive when checking siblings cycles (the `CHECK-DAG` basically ignores the order of a set of things that you want to match), as it is done [at this line](https://github.com/llvm/llvm-project/blob/1ca763b76423a17a2101a4579b5d74bade4f0ce4/llvm/test/Analysis/CycleInfo/basic.ll#L81). This leaves me a bit puzzled, but I'm pretty sure that assigned a specific deterministic `dfs`, the children region too remain deterministic in their order.

lib/RestructureCFG/GenericRegionInfo.cpp
andrealinux1: The answer to your question is no, but I thought that taking into consideration the _grandchildren_ was not necessary for the following reasoning:

- I think that we can avoid iterating on all the grandchildren at each step, and rely on the fact that the _additive_ composition of the property does the work for us (sorry for the war crime usage of the _additive_ term in the presence of a mathematician:)).
- Let's split the reasoning following the two criteria we need to meet:
1) For the first criterion, imagine we have the nested regions A B C (nesting is outer->inner), if some node X is a late entry for C, either is is a late entry for B too (and this in turns means that for being a late entry, it must have been a candidate head before, but not choosen as the head),  or it is a late entry for C but not for B.
In the first scenario, since X is a late entry for B too, when we analyze A, we will find it as a late entry for B, which is a direct child of A, and so we are good.
If however X is a late entry for C but not for B, it means that we do not have a edge _crossing multiple layers_ from A to C (or X would be a late entry for B too). In that case X is contained in C, B, A by construction, but it is not a late entry for B, therefore not a candidate head for A. Therefore we do not care to look at it when analyzing the candidate heads for A.
2) For the second criterion, I think that is evident that, saying that we have nested regions A B C, if something is a candidate head (called node X) for C, either it is a candidate head for B too, and in this case it means that we have a sort of edge _crossing multiple layers_, or it is not a candidate head for B because it is in B but not in C, and in this case we do not have the _crossing multiple layers_ edge.
Now analyzing the perspective of A, if the candidate head X for C was a candidate head for B too, it means that we have the _multiple layers crossing_ edge, but this at the same time means that since the candidate head X is a candidate for B too, we will consider it in the reasoning phase of A (since B it is our child). In the case candidate head X is not a candidate for B, it means that we do not have a _multiple layers crossing_ edge, and the elected head for B could not be X. Therefore, when removing the candidates head of B from A, we will not encounter X anyway.

lib/RestructureCFG/GenericRegionInfo.cpp
pfez: That's ok. I think the guarantees of determinism we have from LLVM are enough.
Even if upstream isn't necessarily committed to preserving this determinism as a strong guarantee, they would need to do wildly convoluted changes just for the sake of making the visit non-deterministic, which would be quite some work for no good.

The answer is then "it's as deterministic as a DFS in upstream LLVM" that for all our goals and purposes means it's deterministic enough.

lib/RestructureCFG/GenericRegionInfo.cpp
pfez: See my other comment. This is deterministic enough. I'm closing this.

lib/RestructureCFG/GenericRegionInfo.cpp
pfez: 1. Ok that makes sense. And it's what I was aiming for wit my question. Please document this.
2. Definitely not clear. And the way I understand it doesn't make sense to me. 
  I would rephrase as.
  If we're looking at A, and one of A's candidates heads (X) is also in a direct children region (B), do we need to look at B's children for criterion 2?
  The answer is no because if there is one of B's children that "forces" B's head to X, it's enough to look at B.
  And if there isn't a children of B that forces B's head to X, we don't care about it at all, because we just need A to be coherent with B in this sense.
  Does this make sense to you? If so document this as well.
  
When I say "document" I mean with comments on each loop on on `->children()`.

Thanks.

andrealinux1: I added an explanation for each of the criterion [here](https://github.com/revng/revng/blob/b7017a3487914ac2d7834b1269f22b804b371579/lib/RestructureCFG/GenericRegionInfo.cpp#L190) and [here](https://github.com/revng/revng/blob/b7017a3487914ac2d7834b1269f22b804b371579/lib/RestructureCFG/GenericRegionInfo.cpp#L217)

PR #550, opened 2025-12-03
 Implement second half of the pipeline in pypeline

mrjackv:


python/revng/pypeline/cli/project/analyze.py
```diff
@@ -121,7 +121,7 @@ def _build_analysis_list_command(self, analysis_list_name: str, pipeline: Pipeli
             config = f'Configuration for the analysis list "{analysis_list_name}".'
             run_analysis_command = click.option(
                 f"--{normalize_flag(analysis_name)}-configuration",
-                f"{analysis_name}_configuration",
+                f"{normalize_pos_arg_name(analysis_name)}_configuration",
```
aleclearmind, line 124: The CLI arguments use `-`, not `_`, correct?

python/revng/pypeline/container.py
```diff
@@ -183,7 +183,7 @@ def contains_all(self, obj: ObjectSet) -> bool:
         assert (
             self.kind == obj.kind
         ), f"Container {self} has kind {self.kind}, but the object set has kind {obj.kind}."
-        return self.objects().issubset(obj)
+        return obj.issubset(self.objects())
```
aleclearmind, line 186: Ouch

include/revng/PipeboxCommon/Helpers/PipeRunPipes/LLVMFunctionPassPipe.h
```diff
@@ -1,229 +0,0 @@
...
-  using Base = SingleOutputPipeBase<T>;
-  static_assert(anyOf<typename Base::OutputContainerType,
-                      revng::pypeline::LLVMRootContainer,
-                      revng::pypeline::LLVMFunctionContainer>());
-  static constexpr bool
-    SingleModule = not std::is_same_v<typename Base::OutputContainerType,
-                                      revng::pypeline::LLVMFunctionContainer>;
-  using ContainerTypesRef = detail::TupleWithRef<typename Base::ContainerTypes>;
-
-  class Pass : public llvm::ModulePass {
```
aleclearmind, line 66: :crying_cat_face:

include/revng/Recompile/CompileModulePipe.h
```diff
@@ -64,19 +64,30 @@ using ObjectFileContainer = BytesContainer<"ObjectFileContainer",
...
+                                             "The compiled object file",
+                                             Access::Write>>;
+
+  CompileRootModule(const Model &TheModel,
+                    llvm::StringRef StaticConfig,
+                    llvm::StringRef DynamicConfig,
+                    LLVMRootContainer &Input,
+                    ObjectFileContainer &Output);
+
+  void run();
```
aleclearmind, line 90: The tracking starts at the constructor, right?
It's not just tracking what's done in `run`, correct?

include/revng/ADT/STLExtras.h
```diff
@@ -448,6 +448,9 @@ struct add_reference {
 template<typename T>
 using add_reference_t = add_reference<T>::type;
 
+template<typename T, typename V>
+using mirror_const = std::conditional_t<std::is_const_v<T>, const V, V>;
```
aleclearmind: 1. We have `ConstIf`.
2. Use snake_case only if we need to be compatible with the STL.

include/revng/Model/IRHelpers.h
```diff
@@ -63,3 +65,30 @@ inline llvm::IntegerType *getLLVMTypeForScalar(llvm::LLVMContext &Context,
...
 }
+
+/// Given an LLVM module with a single isolated function, find it and return a
+/// reference to it. This function will assert if there is any other amount (0,
+/// 2 -> inf) number of functions. If the Address is provided it will also
+/// assert that the found function has the expected metaaddress.
+template<typename T>
+  requires(std::same_as<T, llvm::Module> or std::same_as<T, const llvm::Module>)
+decltype(auto)
+getUniqueIsolatedFunction(T &Module, const MetaAddress *Address = nullptr) {
```
aleclearmind: 1. ```
   ConstOrNot<llvm::Module> auto &Module, const MetaAddress *Address = nullptr)
   ```
2. Do you actually use the default argument?
3. In general, I'd avoid sending in a `nullptr`, you can use `MetaAddress::invalid`. For data structures that have an invalid state, it's better to use it instead of `nullptr` or `std::nullopt`.

include/revng/PipeboxCommon/Helpers/PipeRunPipes/Base.h
```diff
@@ -58,14 +58,27 @@ concept HasPipeRunInvalidate = requires(const T &A,
 
 } // namespace detail
 
+template<typename T>
```
aleclearmind: Concept? :thinking:

lib/FunctionIsolation/EnforceABI.cpp
```diff
@@ -561,11 +561,9 @@ static pipeline::RegisterPipe<EnforceABIPipe> Y;
 
 namespace revng::pypeline::piperuns {
 
-void EnforceABI::runOnFunction(const model::Function &TheFunction) {
-  llvm::Module &Module = Output.getModule(ObjectID(TheFunction.Entry()));
-  llvm::Function *LLVMFunction = Module.getFunction(NameBuilder
-                                                      .llvmName(TheFunction));
-
+void EnforceABI::runOnFunction(const model::Function &Function,
+                               llvm::Function &LLVMFunction) {
```
aleclearmind: Very nice

include/revng/PipeboxCommon/Helpers/PipeRunPipes/Base.h
```diff
@@ -58,14 +58,27 @@ concept HasPipeRunInvalidate = requires(const T &A,
...
+
 template<typename T>
 concept SingleOutputPipeBaseCompatible = requires {
   requires HasName<T>;
   requires SpecializationOf<PipeRunContainerTypes<T>, TypeList>;
 };
 
 template<SingleOutputPipeBaseCompatible T>
-class SingleOutputPipeBase {
+class SingleOutputPipeBase : protected SingleOutputPipeOutputTraits<T> {
```
aleclearmind: Why `protected`?

include/revng/PipeboxCommon/Helpers/PipeRunPipes/LLVMFunctionPipeRun.h
```diff
@@ -0,0 +1,72 @@
...
+  requires IsBasePipeRun<T>;
+  { PipeRun.runOnFunction(Function, LLVMFunction) } -> std::same_as<void>;
+};
+
+// This class wraps a PipeRun that implements a `runOnFunction` with both the
+// `model::Function` and `llvm::Function`. Since there a few member objects
+// that are exposed by the PipeRun (e.g. `Arguments`) and a few optional
+// methods that might be exposed (e.g. `checkPrerequisites`) this class
+// inherits the PipeRun so that it doesn't have to "proxy" everything, which
+// would be required if the class was encapsulated into this one.
```
aleclearmind: 1. Use Doxygen.
2. Split in multiple paragraphs.

include/revng/PipeboxCommon/Helpers/PipeRunPipes/Base.h
```diff
@@ -58,14 +58,27 @@ concept HasPipeRunInvalidate = requires(const T &A,
 
 } // namespace detail
 
+template<typename T>
+struct SingleOutputPipeOutputTraits {
```
aleclearmind: 1. I don't understand what this is. Sometimes you inherit from it (weird inheriting from a trait, unless you're a trait as well), sometimes you just use it.
   Let's discuss this in person maybe.

include/revng/PipeboxCommon/Helpers/PipeRunPipes/LLVMFunctionPipeRun.h
```diff
@@ -0,0 +1,72 @@
...
+template<IsLLVMFunctionPipeRun T>
+class LLVMFunctionPipeRun : public T {
```
aleclearmind: 1. I'm not sure what this is. Does it respect `IsFunctionPipe`?
   If so, `static_assert` in its body.
2. (unrelated) The `IsFunctionPipe` concept should be called `IsFunctionPipeRun`, correct?
   Please review all concept names in this sense. Also, all those that respect this (and similar) concepts should be either have a `*PipeRun` suffix or (better) be in a namespace that makes it obvious that these are pipe runs.
3. `T` is not a pipe run strictly speaking (I don't think it respects the `IsFunctionPipe` concept). Add a note above the `IsLLVMFunctionPipeRun` concept about this. Same for the MLIR one.

lib/RemoveLiftingArtifacts/RemoveLiftingArtifacts.cpp
```diff
@@ -225,3 +227,14 @@ struct RemoveLiftingArtifactsPipe {
 };
 
 static pipeline::RegisterLLVMPass<RemoveLiftingArtifactsPipe> Y;
+
+namespace revng::pypeline::piperuns {
+
+void RemoveLiftingArtifacts::runOnFunction(const model::Function &Function,
```
aleclearmind: Please review all the pipe runs you implemented and add something like

```cpp
// TODO: merge ::RemoveLiftingArtifacts into RemoveLiftingArtifacts once we dismiss the old pipeline
```

Do this for all the situations where this is obvious.

include/revng/RemoveLiftingArtifacts/PromoteInitCSVToUndef.h
```diff
@@ -0,0 +1,33 @@
...
+
+class PromoteInitCSVToUndef {
+private:
+  const model::Binary &Binary;
+
+public:
+  static constexpr llvm::StringRef Name = "promote-init-csv-to-undef";
+  using Arguments = TypeList<PipeRunArgument<LLVMFunctionContainer,
+                                             "Module",
+                                             "function LLVM module(s)">>;
```
aleclearmind: I think we should hoist in some header the `PipeRunArgument` instantiations that are identical all over the place.

lib/PromoteStackPointer/InjectStackSizeProbesAtCallSitesPass.cpp
```diff
@@ -65,3 +71,15 @@ char InjectStackSizeProbesAtCallSitesPass::ID = 0;
...
 static RegisterMSSACS R("measure-stack-size-at-call-sites",
                         "Measure Stack Size At Call Sites Pass");
+
+namespace revng::pypeline::piperuns {
+
+void InjectStackSizeProbesAtCallSites::runOnFunction(const model::Function
+                                                       &Function) {
+  llvm::Module &Module = ModuleContainer.getModule(ObjectID(Function.Entry()));
+  GeneratedCodeBasicInfo GCBI(Binary);
+  GCBI.run(Module);
```
aleclearmind, line 83: Ah-ah, so this is the trick.

lib/Canonicalize/SimplifySwitchPass.cpp
```diff
@@ -279,3 +281,60 @@ void SimplifySwitch::run(pipeline::ExecutionContext &EC,
...
+SimplifySwitch::SimplifySwitch(const class Model &Model,
+                               llvm::StringRef Config,
+                               llvm::StringRef DynamicConfig,
+                               const BinariesContainer &BinariesContainer,
+                               LLVMFunctionContainer &ModuleContainer) :
+  ModuleContainer(ModuleContainer), Binary(*Model.get().get()) {
+  auto BinaryBuffer = BinariesContainer.getFile(0);
+  BinaryView.emplace(Binary,
+                     llvm::StringRef{ BinaryBuffer.data(),
+                                      BinaryBuffer.size() });
```
aleclearmind: Illegal usage of `std::optional`! Use an helper function and do `BinaryView(myHelperFunction(...))`.

lib/Canonicalize/SimplifySwitchPass.cpp
```diff
@@ -279,3 +281,60 @@ void SimplifySwitch::run(pipeline::ExecutionContext &EC,
...
+
+public:
+  static inline char ID = 0;
+
+  SimplifySwitchPass(const model::Binary &Binary, RawBinaryView &BinaryView) :
+    llvm::ModulePass(ID), Binary(Binary), BinaryView(BinaryView) {}
+
+  bool runOnModule(llvm::Module &M) override {
+    llvm::Function &Function = getUniqueIsolatedFunction(M);
+    auto &LVI = getAnalysis<LazyValueInfoWrapperPass>(Function).getLVI();
```
aleclearmind, line 304: So, in the end you didn't manage to instantiate LVI without going through the analysis route?
For the `DominatorTree` it's quite straightforward.

At very least explain why in a comment and/or leave a TODO.

python/revng/internal/pipeline.yml
```diff
@@ -198,6 +198,34 @@ branches:
             - hoist-struct-phis
       - pipe: legacy-segregate-stack-accesses
         arguments: [llvm-functions]
+      - pipe: pure-llvm-passes-pipe
+        arguments: [llvm-functions]
+        configuration:
+          Passes:
```
aleclearmind: Can we have a field named `Passes` in C++ but the YAML serializer use `passes`?

include/revng/RemoveLiftingArtifacts/MakeSegmentRef.h
```diff
@@ -0,0 +1,45 @@
...
+                    "function LLVM module(s)">>;
+
+  MakeSegmentRef(const class Model &Model,
+                 llvm::StringRef Config,
+                 llvm::StringRef DynamicConfig,
+                 const BinariesContainer &BinariesContainer,
+                 LLVMFunctionContainer &ModuleContainer) :
+    Binary(*Model.get().get()) {
+    llvm::ArrayRef<char> Buffer = BinariesContainer.getFile(0);
+    BinaryView.emplace(Binary, llvm::StringRef{ Buffer.data(), Buffer.size() });
```
aleclearmind: Illegal use of `std::optional`, please fix everywhere.

lib/RemoveLiftingArtifacts/MakeSegmentRefPass.cpp
```diff
@@ -32,13 +32,21 @@
...
 #include "revng/Support/OpaqueFunctionsPool.h"
 
 using namespace llvm;
 
+static cl::opt<bool> DisableFunctionPointersOpt("disable-function-pointers",
+                                                cl::desc("Disable emitting "
+                                                         "pointers to "
+                                                         "functions through "
+                                                         "segments"),
+                                                cl::init(false));
```
aleclearmind, line 55: Add a comment explaining why we do this and note that once we rewrite the `emit-segment-refs` this can go.

aleclearmind (commented):
NOTETOSELF: resume reviewing starting from `Add ParseBinaryAnalysis to pypeline`, included.

python/revng/pypeline/cli/project/analyze.py
mrjackv: Yes, but the variable where this is stored needs to be snake_case

include/revng/Recompile/CompileModulePipe.h
mrjackv: It's done from the constructor onward

include/revng/PipeboxCommon/Helpers/PipeRunPipes/Base.h
mrjackv: Basically the part that was taken out and made into `SingleOutputPipeOutputTraits` was the `protected` part that was previously present in this class

lib/Canonicalize/SimplifySwitchPass.cpp
mrjackv: > So, in the end you didn't manage to instantiate LVI without going through the analysis route?

There are a few dependencies and they nest quite deeply, at some point I thought something along the lines of "Ah I'll write something that automatically computes the dependencies for me" and that's where I've said "why should I do all of this when there `PassManager` already?"

> At very least explain why in a comment and/or leave a TODO

Will do

include/revng/Model/IRHelpers.h
mrjackv: 1. Ok
2. Yes, in `SimplifySwitchPass`
3. Ok, even though tecnically you could have an isolated function (in the IR sense) that has an invalid MA

include/revng/PipeboxCommon/Helpers/Registrars.h
```diff
@@ -194,4 +195,7 @@ template<typename T>
 using RegisterTypeDefinitionPipe = RegisterPipe<TypeDefinitionPipe<T>>;
 
 template<typename T>
-using RegisterLLVMFunctionPassPipe = RegisterPipe<LLVMFunctionPassPipe<T>>;
+using RegisterLLVMFunctionPipe = RegisterFunctionPipe<LLVMFunctionPipeRun<T>>;
```
aleclearmind: `RegisterFunctionPipeRun`

include/revng/PipeboxCommon/Helpers/Registrars.h
aleclearmind: Concept

include/revng/PipeboxCommon/Helpers/PipeRunPipes/LLVMFunctionPipeRun.h
aleclearmind: Rotate inheritance. Turn `LLVMFunctionPipeRun` into a mixin/CRTP so that what's currently `T` becomes actually a pipe run and we no longer have a "semi-pipe run".

include/revng/ADT/STLExtras.h
mrjackv: 1. Done
2. Done

include/revng/Model/IRHelpers.h
mrjackv: 1. Done
2. .
3. Switched to `MetaAddress::invalid()`

include/revng/PipeboxCommon/Helpers/PipeRunPipes/Base.h
mrjackv: This has been dropped

include/revng/PipeboxCommon/Helpers/PipeRunPipes/LLVMFunctionPipeRun.h
mrjackv: This has been dropped

lib/Canonicalize/SimplifySwitchPass.cpp
mrjackv: Done

include/revng/RemoveLiftingArtifacts/MakeSegmentRef.h
mrjackv: Fixed and checked

lib/RemoveLiftingArtifacts/MakeSegmentRefPass.cpp
mrjackv: Added

include/revng/PipeboxCommon/Helpers/Registrars.h
mrjackv: Added them

include/revng/PipeboxCommon/Helpers/Registrars.h
mrjackv: Done

include/revng/PipeboxCommon/Helpers/PipeRunPipes/LLVMFunctionPipeRun.h
mrjackv: Done

mrjackv: I've rotated the inheritance, now there's `LLVMFunctionMixin<Derived>` and `MLIRFunctionMixin<Derived>`

lib/RemoveLiftingArtifacts/RemoveLiftingArtifacts.cpp
mrjackv: Done

include/revng/RemoveLiftingArtifacts/PromoteInitCSVToUndef.h
mrjackv: Done

python/revng/internal/pipeline.yml
mrjackv: Done

lib/Canonicalize/SimplifySwitchPass.cpp
```diff
@@ -279,3 +282,60 @@ void SimplifySwitch::run(pipeline::ExecutionContext &EC,
...
+
+public:
+  void getAnalysisUsage(llvm::AnalysisUsage &AU) const override {
+    AU.addRequired<DominatorTreeWrapperPass>();
+    AU.addRequired<LazyValueInfoWrapperPass>();
+  }
+};
+
+using SSP = SimplifySwitchPass;
+static inline llvm::RegisterPass<SSP> SSPR{ "pypeline-simplify-switch",
```
aleclearmind: I *think* you can avoid registering this if you don't want to expose it.

lib/Canonicalize/SimplifySwitchPass.cpp
```diff
@@ -279,3 +282,60 @@ void SimplifySwitch::run(pipeline::ExecutionContext &EC,
...
+
+SimplifySwitch::SimplifySwitch(const class Model &Model,
+                               llvm::StringRef Config,
+                               llvm::StringRef DynamicConfig,
+                               const BinariesContainer &BinariesContainer,
+                               LLVMFunctionContainer &ModuleContainer) :
+  ModuleContainer(ModuleContainer),
+  Binary(*Model.get().get()),
+  BinaryView(makeBinaryView(Model, BinariesContainer)) {
+  PM.add(new DominatorTreeWrapperPass);
```
aleclearmind: 1. Do you *have to* schedule `DomintatorTree`? You're not scheduling `LazyValueInfo`.
2. Instantiate `DominatorTree` manually, do not depend on the analysis. Let's try to keep the number of analyses to a minimum, so that if we decide to drop LVI, we can easily demote this to a non-`Pass`.

include/revng/Canonicalize/SimplifySwitch.h
```diff
@@ -0,0 +1,50 @@
...
+
+public:
+  static constexpr llvm::StringRef Name = "simplify-switch";
+  using Arguments = TypeList<
+    PipeRunArgument<const BinariesContainer, "Binaries", "The input binaries">,
+    PipeRunArgument<LLVMFunctionContainer,
+                    "Module",
+                    "function LLVM module(s)">>;
+  using Analyses = TypeList<llvm::LazyValueInfoWrapperPass,
+                            llvm::DominatorTreeWrapperPass>;
```
aleclearmind: Drop?

include/revng/ADT/CompilationTime.h
```diff
@@ -200,4 +200,34 @@ template<auto &V>
...
 using FunctionTraits = detail::FunctionTraits<decltype(V)>;
 
+/// Helper function that converts a pack of booleans to their constexpr
+/// counterpart, this is useful to e.g. have a type depending on a command-line
+/// option. Do note that the compiler will expand this to all the possible
+/// combinations, so given N booleans there will be 2**N template expansions.
+template<typename CallableT, typename... BoolType>
+  requires(std::is_same_v<BoolType, bool> and ...)
+inline constexpr auto
+bools(CallableT &&Callable, bool Bool, BoolType... BoolRest) {
```
aleclearmind: 1. You are an evil person. But I like the way you think.
2. I'd call this `invokeAllCombinations` (or is it *permutations*?).

include/revng/Model/Importer/Binary/BinaryImporter.h
```diff
@@ -16,9 +16,16 @@ class ObjectFile;
...
-                         llvm::object::ObjectFile &BinaryHandle,
-                         const ImporterOptions &Options);
-llvm::Error importBinary(TupleTree<model::Binary> &Model,
-                         llvm::StringRef Path,
-                         const ImporterOptions &Options);
+llvm::Error
+importBinary(TupleTree<model::Binary> &Model,
+             llvm::object::ObjectFile &BinaryHandle,
+             const ImporterOptions &Options,
+             TupleTreeReference<model::BinaryIdentifier, model::Binary>
```
aleclearmind: Use `BinaryReference` everywhere you can.

include/revng/Model/Importer/Binary/ImportBinaryAnalysis.h
```diff
@@ -29,3 +31,17 @@ class ImportBinaryAnalysis {
 };
 
 } // namespace revng::pipes
+
+namespace revng::pypeline::analyses {
+
+class ParseBinaryAnalysis {
```
aleclearmind, line 40: Docs

lib/Model/Importer/Binary/ELFImporter.cpp
```diff
@@ -420,7 +420,15 @@ Error ELFImporter<T, HasAddend>::import(const ImporterOptions &Options) {
...
     // Import Dwarf
     DwarfImporter Importer(Model);
-    Importer.import(TheBinary.getFileName(), AdjustedOptions);
+    auto Buffer = MemoryBuffer::getMemBuffer(TheBinary.getData(), "", false);
+
+    std::string Filename;
+    if (Reference.isValid())
+      Filename = Reference.get()->Name();
+    else if (not TheBinary.getFileName().empty())
+      Filename = TheBinary.getFileName().str();
```
aleclearmind, line 435: What is going on here?

lib/Model/Importer/WellKnownModels.cpp
```diff
@@ -48,66 +49,85 @@ class ImportWellKnownModelsAnalysis {
...
-      auto Model = llvm::cantFail(TupleTree<model::Binary>::fromFile(Path));
-      auto NewWKM = std::make_unique<WellKnownModel>(std::move(Model), Model);
-      WellKnownModels.push_back(std::move(NewWKM));
-    }
+  void run(pipeline::ExecutionContext &Context);
+};
+
+} // namespace revng::pipes
+
+static void wellKnownModel(TupleTree<model::Binary> &Model) {
```
aleclearmind: This name means nothing to me

include/revng/EarlyFunctionAnalysis/CollectFunctionsFromCalleesPass.h
```diff
@@ -12,6 +12,14 @@
 #include "revng/Support/IRHelpers.h"
 #include "revng/Support/MetaAddress.h"
 
+namespace revng::internal {
```
aleclearmind: I don't like `revng::internal`. Why do you need it?

I feel like you're trying to circumvent some rule on the `details` namespace `:P`

include/revng/DataLayoutAnalysis/DLA.h
```diff
@@ -0,0 +1,22 @@
...
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "revng/PipeboxCommon/LLVMContainer.h"
+#include "revng/PipeboxCommon/Model.h"
+
+namespace revng::pypeline::analyses {
+
+class DLA {
```
aleclearmind: Do not use abbreviations. `AnalyzeDataLayout`.

lib/DataLayoutAnalysis/Frontend/DLACreateInterProceduralTypes.cpp
```diff
@@ -276,7 +276,10 @@ bool TSBuilder::createInterproceduralTypes(llvm::Module &M,
     TS.addEqualityLink(SegmentNode, SegmentRefNode);
 
     for (const Use &U : F.uses()) {
-      auto *Call = cast<CallInst>(U.getUser());
+      auto *Call = dyn_cast<CallInst>(U.getUser());
+      if (Call == nullptr)
+        continue;
```
aleclearmind: Nope, use the `callers` function.

include/revng/PipeboxCommon/CliftContainer.h
aleclearmind, line 1: @vasama-work can you review this please?

include/revng/Clift/Helpers.h
```diff
@@ -22,4 +22,32 @@ inline MetaAddress getMetaAddress(clift::FunctionOp F) {
   return MetaAddress::invalid();
 }
 
+inline decltype(auto)
```
aleclearmind: Mmmh, please be explicit (you can use `ConstIf`).

python/revng/internal/pipeline.yml
aleclearmind, line 1: No abbreviations in commit messages. What's `s.p.`?

lib/CliftPipes/ClifterPipe.cpp
```diff
@@ -57,3 +59,36 @@ class LLVMToCliftPipe {
...
+                         const LLVMFunctionContainer &Input,
+                         MLIRFunctionContainer &Output) :
+  Binary(*Model.get().get()),
+  Input(Input),
+  Output(Output),
+  NameBuilder(Binary) {
+  Output.getContext().loadDialect<clift::CliftDialect>();
+}
+
+void LLVMToClift::runOnFunction(const model::Function &Function) {
```
aleclearmind, line 74: @vasama-work can you review this?

lib/CliftPipes/OptimizationPipe.cpp
```diff
@@ -68,3 +69,25 @@ class OptimizationPipe {
...
+
+CliftOptimization::CliftOptimization(const Model &Model,
+                                     llvm::StringRef Config,
+                                     llvm::StringRef DynamicConfig,
+                                     MLIRFunctionContainer &ModuleContainer) :
+  MLIRFunctionMixin(ModuleContainer),
+  PM(&ModuleContainer.getContext(), clift::FunctionOp::getOperationName()) {
+  PM.addPass(mlir::createCanonicalizerPass());
+  PM.addPass(clift::createLoopDetectionPass());
+  PM.addPass(clift::createCLegalizationPass(TargetCImplementation::Default));
```
aleclearmind: I don't understand the purpose of this pipe run.
Aren't these pure MLIR passes? We should have something to run pure MLIR passes, or did you want to postpone this?

lib/CliftPipes/ModelVerifyPipe.cpp
```diff
@@ -39,3 +40,19 @@ class ModelVerifyPipe {
...
 } // namespace
+
+namespace revng::pypeline::piperuns {
+
+void ModelVerifyClift::runOnMLIRFunction(const model::Function &Function,
+                                         mlir::clift::FunctionOp MLIRFunction) {
+  if (not ModelVerifyLogger.isEnabled())
+    return;
+
+  revng::Tracking::push(Binary);
```
aleclearmind: 1. I think we have a RAII object for this.
2. Comment why we are doing this.

python/revng/internal/pipeline.yml
```diff
@@ -91,6 +91,9 @@ branches:
         arguments: [llvm-root, cfg-map]
       - savepoint: cfg-computed
         containers: [cfg-map]
+        artifacts:
+          - name: emit-cfg
+            container: cfg-map
```
aleclearmind, line 104: I'm not entirely sure what's the value of having all these micro commits on `python/revng/internal/pipeline.yml`.

Can we have a single commit at the end?

python/revng/internal/cli/revng2.py
```diff
@@ -26,9 +41,25 @@ def quick():
...
+    default=None,
+    type=click.Path(exists=True, dir_okay=False, path_type=Path),
+)
+def init(binary: Path | None):
     """Initialize a new project."""
-    # TODO
+    model_name = get_singleton(Model).model_name()  # type: ignore[type-abstract]
+    model_file = Path.cwd() / model_name
+    if model_file.exists():
+        raise click.UsageError(f"File {model_name} is present, cannot run init")
```
aleclearmind: > is already present in the current directory. Refusing to overwrite it. ~~cannot run init~~

share/revng/test/configuration/revng/pypeline-comparison.yml
```diff
@@ -10,7 +10,13 @@ commands:
...
     suffix: /
     command: |-
-      "${SOURCES_ROOT}/share/revng/test/tests/pypeline-comparison/prepare.py" "$INPUT" "$OUTPUT";
+      pushd "$OUTPUT" &> /dev/null;
+      cp "$INPUT" input;
+      revng2 project init input;
+      mv model.yml model_init.yml;
+      revng analyze revng-initial-auto-analysis --model=model_init.yml input > model.yml;
+      rm model_init.yml;
+      popd &> /dev/null;
```
aleclearmind, line 16: Comment

python/revng/internal/cli/revng2.py
```diff
@@ -26,9 +41,25 @@ def quick():
...
+
+    if binary is None:
+        model_file.touch()
+        return
+
+    with open(model_file, "w") as f:
+        yaml.safe_dump(generate_model_with_binaries([binary]), f)
```
aleclearmind: `revng2 project init` should run `initial-auto-analyses`.

python/revng/pypeline/storage/local_provider.py
```diff
@@ -598,3 +599,44 @@ def get_custom_invalidation_data(
...
+    def __init__(self):
+        self.directory = TemporaryDirectory()
+        self.dir_path = Path(self.directory.name)
+
+        model_type = get_singleton(Model)  # type: ignore [type-abstract]
+        model_name = model_type.model_name()
+        model_path = self.dir_path / model_name
+        model_path.touch()
+
+        super().__init__(self.dir_path / "db", model_path)
```
aleclearmind: Please keep this as the very first line of the function.

python/revng/pypeline/storage/local_provider.py
```diff
@@ -598,3 +599,44 @@ def get_custom_invalidation_data(
...
+                self.files[hash_] = file.contents
+            elif file.path is not None:
+                hash_ = compute_hash(file.path)
+                self.files[hash_] = file.path
+
+            result.append(hash_)
+
+        return result
+
+    def get_files_from_storage(self, requests: list[FileRequest]) -> dict[str, bytes]:
```
aleclearmind: Are you storing the whole file in memory? Please don't, this is not a debugging tool, this will run on real stuff.

include/revng/PipeboxCommon/CliftContainer.h
```diff
@@ -0,0 +1,96 @@
...
+
+class MLIRFunctionContainer {
+public:
+  static constexpr llvm::StringRef Name = "MLIRFunctionContainer";
+  static constexpr Kind Kind = Kinds::Function;
+  static constexpr llvm::StringRef MimeType = "application/x.mlir.bc";
+
+private:
+  mlir::MLIRContext Context;
+  std::map<ObjectID, mlir::OwningOpRef<mlir::ModuleOp>> Modules;
```
vasama-work, line 34: Why do we have multiple *modules*? Functions I would understand.

include/revng/PipeboxCommon/MLIRContainer.h
```diff
@@ -0,0 +1,96 @@
...
+
+inline const mlir::DialectRegistry MLIRDialectRegistry = ([]() {
+  mlir::DialectRegistry Registry;
+  Registry.insert<mlir::clift::CliftDialect>();
+  return Registry;
+})();
+
+}
+
+class MLIRFunctionContainer {
```
vasama-work: We renamed what was originally `MLIRContainer` to `CliftContainer` because it was so Clift-specific. Is this container intended to be more general than that?

include/revng/PipeboxCommon/MLIRContainer.h
```diff
@@ -0,0 +1,96 @@
...
+    return Result;
+  }
+
+  bool verify() const {
+    bool Result = true;
+    for (auto const &[_, Module] : Modules) {
+      mlir::LogicalResult ModuleResult = Module.get().verify();
+      Result = Result & ModuleResult.succeeded();
+    }
+    return true;
```
vasama-work: `Result` is unused. I suspect you meant to return it.

Also consider if `Result &=` instead of `Result = Result &` is an improvement.

include/revng/PipeboxCommon/MLIRContainer.h
```diff
@@ -0,0 +1,96 @@
...
+
+namespace revng::pypeline {
+
+namespace detail {
+
+inline const mlir::DialectRegistry MLIRDialectRegistry = ([]() {
+  mlir::DialectRegistry Registry;
+  Registry.insert<mlir::clift::CliftDialect>();
+  return Registry;
+})();
```
vasama-work: If this is only used by `MLIRFunctionContainer`, why not make it a static member variable?

include/revng/CliftPipes/ClifterPipe.h
```diff
@@ -0,0 +1,43 @@
...
+#include "revng/PipeboxCommon/Model.h"
+
+namespace revng::pypeline::piperuns {
+
+class LLVMToClift {
+private:
+  const model::Binary &Binary;
+  const LLVMFunctionContainer &Input;
+  MLIRFunctionContainer &Output;
+  model::CNameBuilder NameBuilder;
```
vasama-work: As far as I can tell, this is never used.

lib/CliftPipes/ClifterPipe.cpp
```diff
@@ -57,3 +59,36 @@ class LLVMToCliftPipe {
...
+  ObjectID Object(Function.Entry());
+  const llvm::Module &Module = Input.getModule(Object);
+  const llvm::Function
+    &LLVMFunction = getUniqueIsolatedFunction(Module, Function.Entry());
+
+  using namespace mlir;
```
vasama-work: Not a big issue, but for consistency with the rest of the Clift code, I would prefer not applying a using directive to namespace `mlir`. I've commonly done this though:
```cpp
namespace clift = mlir::clift;
using namespace clift;
```

include/revng/Model/Importer/Binary/BinaryImporter.h
mrjackv: That wold require having a header with `using BinaryReference = TupleTreeReference<model::BinaryIdentifier, model::Binary>`. I was under the impression that `using`s in header were prohibited, did that change?

include/revng/Canonicalize/SimplifySwitch.h
mrjackv: It's already dropped, will close this in the next push

include/revng/EarlyFunctionAnalysis/CollectFunctionsFromCalleesPass.h
mrjackv: The problem is as follows:
1. There's a function or a chunk of it that needs to be exposed between TUs, this is because (as in this case, there are variations) the code was inside an LLVM pass or some function that accepted pipeline (the old one)-specific inputs
2. I outline the chunk I need in a function that's called by both the old and new pipeline stuff
3. I put it in the `revng::internal` namespace because it's a free-floating chunk of code stitched to other parts of pipes that doesn't do meaningful things on it own and I can't use `detail` since you said that it cannot be used across TUs

python/revng/internal/pipeline.yml
mrjackv: **S**ave**p**oint
If I write it out it exceeds the 52 line limit imposed by check-conventions

lib/CliftPipes/ModelVerifyPipe.cpp
mrjackv: 1. `ObjectDependenciesHelper::Committer` exists but it's a private class and I don't think it's warranted to expose it since it's tightly couple with `ObjectDependenciesHelper`
2. Ok

python/revng/internal/cli/revng2.py
mrjackv: That's not what's specified in [REP21](https://pad.rev.ng/o37XT3QfScKKV8Wh7h6ZCw#revng-project-init)
Making it do the initial auto analysis is not a problem, but I'd prefer if there was a `--no-initial-auto-analysis` or similar to suppress it

python/revng/pypeline/storage/local_provider.py
mrjackv: > Are you storing the whole file in memory?

No

The `FileRequest` is composed of two fields `content` and `path`, only one of them can be non-None. In the case of CLI stuff the `path` field will be non-none. The only case where the `content` field is non-none is if the storage provider is used in conjunction with the daemon and a request to add a file arrives via HTTP (which, at that point, does not have a file associated). The content is stored as-is because:
1. It's already in RAM, we pay 0 cost in keeping it there
2. It's unlikely that this storage provider will be used with the daemon

lib/Model/Importer/Binary/ELFImporter.cpp
mrjackv: `ELFImporter` previously relied on the `TheBinary` field to retrieve the filename of the file. This is no longer possible because the API has been changed to work with `llvm::MemoryBuffer` as it's friendlier with `BinariesContainer`. The importer still needs the filename because it does a bunch of heuristics (e.g. looking for `${FILENAME}.debug`) to retrieve the debug info. So first the name is retrieved from `/Binaries/0/Name` if it's valid or the `TheBinary` as a fallback (which is compatible to what the old pipeline does)

lib/Model/Importer/WellKnownModels.cpp
mrjackv: It's the body of `WellKnownModel::run` from the legacy pipe converted to a static function so that it can be used by both the old and new pipe. I've followed the pattern of converting the `run` into a static function with a `camelCase` name consistently in the PR

include/revng/PipeboxCommon/CliftContainer.h
mrjackv: This mirrors the behavior of `LLVMFunctionContainer`. Each isolated function is in its own module, so that there isn't the need to run `cloneFiltered` or `mergeBackImpl` when serializing/deserializing.

include/revng/Model/Importer/Binary/BinaryImporter.h
aleclearmind: `using`s are not prohibited in headers. `using namespace` are prohibited.

lib/Model/Importer/WellKnownModels.cpp
aleclearmind: What does it do? Functions should be a verb or a noun representing what they return.

python/revng/internal/pipeline.yml
aleclearmind: > Add `segregate-stack-accesses` to pypeline
>
> This commit adds...

However, I doubt the usefulness of all these commits, as mentioned elsewhere.

lib/CliftPipes/ModelVerifyPipe.cpp
aleclearmind: 1. `struct DisableTracking`. Adopt it everywhere it makes sense.

python/revng/internal/cli/revng2.py
aleclearmind: > That's not what's specified in [REP21](https://pad.rev.ng/o37XT3QfScKKV8Wh7h6ZCw#revng-project-init)

Right, but I think we already said that it should do it.

> I'd prefer if there was a --no-initial-auto-analysis or similar to suppress it

Sure.

lib/Model/Importer/WellKnownModels.cpp
mrjackv: It does the `ImportWellKnownModels` analysis but without the `revng::pipeline`-isms

lib/Model/Importer/Binary/ELFImporter.h
```diff
@@ -64,6 +65,7 @@ class ELFImporter : public BinaryImporterHelper, public ELFImporterBase {
   RawBinaryView File;
   TupleTree<model::Binary> &Model;
   const llvm::object::ELFObjectFileBase &TheBinary;
+  BinaryReference &Reference;
```
aleclearmind: `Reference` is an empty term, do not use it.

lib/Model/Importer/Binary/ImportBinaryAnalysis.cpp
```diff
@@ -16,6 +16,17 @@
 
 using namespace revng::pipes;
 
+using BinaryReference = TupleTreeReference<model::BinaryIdentifier,
+                                           model::Binary>;
+
+static BinaryReference makeReference(model::Binary *Binary, size_t Index) {
```
aleclearmind: Use `model::Binary &` unless you want to store it (e.g., in a `map`) or you actually think having it `nullptr` is useful.

lib/Model/Importer/Binary/ELFImporter.cpp
aleclearmind: Add a comment clearly explaining what needs to happen once we dismiss the old pipeline.

include/revng/EarlyFunctionAnalysis/CollectFunctionsFromCalleesPass.h
aleclearmind: 3. Please do not try to circumvent rules. Ask out. `revng::internal` is not a convention we adopt and it's just a way to circumvent a rule. These are tricks to avoid rules that risk going unnoticed during code review and then become a precedent that someone else can adopt. Refrain from doing this.

A free function is fine. If you don't like it, you can make a static method in the class.

lib/Model/Importer/WellKnownModels.cpp
aleclearmind: `importWellKnownModels`?

python/revng/pypeline/storage/local_provider.py
```diff
@@ -598,3 +599,44 @@ def get_custom_invalidation_data(
...
 
         return result
+
+
+class EphemeralLocalStorageProvider(LocalStorageProvider):
+    """Specialization of LocalStorageProvider that's ephemeral, meaning that
+    all the data will be lost once this provider is deleted"""
+
+    def __init__(self):
+        self.directory = TemporaryDirectory()
```
aleclearmind: Describe in a comment the layout of this directory

python/revng/pypeline/storage/local_provider.py
```diff
@@ -598,3 +599,44 @@ def get_custom_invalidation_data(
             result[argument_index].append((obj_id_type.from_bytes(object_id), data))
 
         return result
+
+
+class EphemeralLocalStorageProvider(LocalStorageProvider):
```
aleclearmind: Do we have an URL to instantiate this?

python/revng/pypeline/storage/local_provider.py
aleclearmind: Why do we have to override `put_files_in_storage` at all?
I'd really like this to just be a `LocalStorageProvider` putting things in a temporary directory as much as we can.

I think we should be able to do this without creating a new class. We should just feed the right `__init__` arguments.
Maybe we should have a `temporary://` factory that does this (and uses the vanilla `LocalStorageProvider`).

lib/Canonicalize/SimplifySwitchPass.cpp
mrjackv: Dropped

lib/Canonicalize/SimplifySwitchPass.cpp
mrjackv: Instantiated the `DominatorTree` manually

include/revng/Model/Importer/Binary/BinaryImporter.h
mrjackv: Done

include/revng/EarlyFunctionAnalysis/CollectFunctionsFromCalleesPass.h
mrjackv: Moved the two functions outside the `revng::internal` namespace

lib/DataLayoutAnalysis/Frontend/DLACreateInterProceduralTypes.cpp
mrjackv: Done

include/revng/PipeboxCommon/CliftContainer.h
mrjackv: Closing this since there are comments to address

lib/CliftPipes/ClifterPipe.cpp
mrjackv: Closing this since there are comments to address

python/revng/internal/cli/revng2.py
mrjackv: Done

share/revng/test/configuration/revng/pypeline-comparison.yml
mrjackv: This got cut down drastically, please check if comments are still needed

python/revng/internal/cli/revng2.py
mrjackv: Implemented

include/revng/PipeboxCommon/MLIRContainer.h
mrjackv: Renamed to `CliftFunctionContainer`

include/revng/PipeboxCommon/MLIRContainer.h
mrjackv: Good catch! thanks!

include/revng/PipeboxCommon/MLIRContainer.h
mrjackv: Converted to static member

include/revng/CliftPipes/ClifterPipe.h
mrjackv: Sorry I just copy-pasted from the old pipeline, it's indeed unused. I've dropped it.

lib/Model/Importer/Binary/ELFImporter.h
mrjackv: Renamed to `BinaryReference`

lib/Model/Importer/Binary/ImportBinaryAnalysis.cpp
mrjackv: Done

include/revng/DataLayoutAnalysis/DLA.h
mrjackv: Renamed

include/revng/PipeboxCommon/CliftContainer.h
mrjackv: Closing this as it has been discussed on mattermost

include/revng/ADT/CompilationTime.h
mrjackv: Renamed to `invokeCombination`

include/revng/Model/Importer/Binary/ImportBinaryAnalysis.h
mrjackv: Done

lib/Model/Importer/Binary/ELFImporter.cpp
mrjackv: Done

lib/Model/Importer/WellKnownModels.cpp
mrjackv: renamed

include/revng/Clift/Helpers.h
mrjackv: Done

lib/CliftPipes/ModelVerifyPipe.cpp
mrjackv: Done

lib/CliftPipes/ClifterPipe.cpp
mrjackv: Thanks for the heads up, in the end I've dropped the use of `using namespace mlir;` in the PR

python/revng/internal/pipeline.yml
mrjackv: Compacted changes to `pipeline.yml` into a single commit for each branch of the pipeline

lib/CliftPipes/OptimizationPipe.cpp
mrjackv: Added `pure-mlir-passes-pipe`

python/revng/internal/pipeline.yml
mrjackv: Fixed, see [here](https://github.com/revng/revng/pull/550#discussion_r2607086212)

python/revng/pypeline/storage/local_provider.py
mrjackv: Closing, this class has been dropped

python/revng/pypeline/storage/local_provider.py
mrjackv: Closing, this class has been dropped

python/revng/pypeline/storage/local_provider.py
mrjackv: Yes, `temporary://`

python/revng/pypeline/storage/local_provider.py
mrjackv: Closing, this class has been dropped

lib/PipeboxCommon/Python/_pipebox.cpp
```diff
@@ -33,6 +33,37 @@ static void handleSignal(int SigNo) {
...
+  PythonInitRevng(llvm::ArrayRef<std::string> ArgvRef,
+                  llvm::StringRef Overview) :
+    Overview(Overview.str()) {
+    ArgvRaw.push_back("");
+    append(ArgvRef, ArgvRaw);
+
+    Argc = ArgvRaw.size();
+    Argv.resize(ArgvRaw.size());
+    for (size_t I = 0; I < ArgvRaw.size(); I++)
+      Argv[I] = ArgvRaw[I].c_str();
```
aleclearmind: Any reason not to do it in `InitRevng` itself?
I'd prefer not to have this.

mrjackv: > Any reason not to do it in InitRevng itself?

Because all other uses of `InitRevng` pass the argv directly from `main`, so they don't need to do this
Also, `InitRevng` is a child of `InitLLVM` so having _it_ store the argv would require breaking the child-parent relationship

aleclearmind: > would require breaking the child-parent relationship

I don't see that.

```patch
diff --git a/include/revng/Support/InitRevng.h b/include/revng/Support/InitRevng.h
index 695f55e1a..271de6c05 100644
--- a/include/revng/Support/InitRevng.h
+++ b/include/revng/Support/InitRevng.h
@@ -22,6 +22,8 @@ namespace revng {
 class InitRevng : public llvm::InitLLVM {
 private:
   static inline bool Initialized = false;
+  std::vector<std::string> Args;
+  std::vector<const char *> Argv;
 
 public:
   InitRevng(int &Argc, auto **&Argv, const char *Overview) :
@@ -31,7 +33,9 @@ public:
             auto **&Argv,
             const char *Overview,
             llvm::ArrayRef<const llvm::cl::OptionCategory *> CategoriesToHide) :
-    InitLLVM(Argc, Argv, true) {
+    Args(copy(Argv, Argc)),
+    Argv(something(Args)),
+    InitLLVM(Argc, this->Argv, true) {
 
     revng_assert(not Initialized);
     Initialized = true;
```

Alternatively, you can just make `const char *Argv[Argc]` a global variable.
I know you don't like them, but I think this is fine. Initialization is supposed to happen only once.

I don't think we need a "PythonInitRevng", there's nothing python about it. Either the existing one should not assume argv have infinite life time or we should just ensure lifetime is infinite in all users.

mrjackv: > I don't see that.

> ```diff
> -    InitLLVM(Argc, Argv, true) {
>+    Args(copy(Argv, Argc)),
>+    Argv(something(Args)),
>+    InitLLVM(Argc, this->Argv, true) {
>```

That cannot be done, base class constructors need to be initialized before member variables. The only way out is to make `InitRevng` look more like `PythonInitRevng` with a `std::unique_ptr<llvm::initLLVM>` or similar

> Alternatively, you can just make const char *Argv[Argc] a global variable.

This is also impractical, it would require changing all (valid) invocations of `InitRevng`. Also, the length is not known, so you'd need to have something like `static std::vector<std::string> Args`, set it and then re-wrangle it to make it work with `InitLLVM` because its constructor needs to be the first thing that's called. Again, might as well use a `std::unique_ptr<llvm::InitLLVM>`

aleclearmind: > That cannot be done

Right, dunno, we can use composition instead of inheritance. There's no real reason to inherit.

> This is also impractical

Uh? I just meant

```patch
diff --git a/lib/PipeboxCommon/Python/_pipebox.cpp b/lib/PipeboxCommon/Python/_pipebox.cpp
index 37b2296af..851da4bb8 100644
--- a/lib/PipeboxCommon/Python/_pipebox.cpp
+++ b/lib/PipeboxCommon/Python/_pipebox.cpp
@@ -52,13 +52,15 @@ NB_MODULE(_pipebox, m) {
       }
     }
 
-    int Argc = ArgVector.size() + 1;
-    const char *Argv[Argc];
+    static std::vector<std::string> ArgVector2 = ArgVector;
+    int Argc = ArgVector2.size() + 1;
+    static std::vector<const char *> Argv;
+    Argv.resize(Argc);
     Argv[0] = "";
-    for (size_t I = 0; I < ArgVector.size(); I++)
-      Argv[I + 1] = ArgVector[I].c_str();
+    for (size_t I = 0; I < ArgVector2.size(); I++)
+      Argv[I + 1] = ArgVector2[I].c_str();
 
-    const char **ArgvPtr = Argv;
+    const char **ArgvPtr = &Argv[0];
     // use a capsule to call the destructor when the Python module is unloaded
     m.attr("__init_revng__") = makeCapsule<revng::InitRevng>(Argc, ArgvPtr, "");
 ```

If you really wanna keep the `PythonInitRevng` class, you should call it `ArgvOwningInitRevng`, which is quite sad.

mrjackv: Ah ok, I was a bit confused. I've dropped the class and made the variables in `Initialize` static

PR #551, opened 2025-12-17
Pypeline crash fixes

mrjackv:


include/revng/PipeboxCommon/BinariesContainer.h
```diff
@@ -27,14 +28,20 @@ class BinariesContainer {
   static constexpr llvm::StringRef MimeType = "application/x-tar";
 
 private:
+  static constexpr llvm::StringRef Prefix = "binaries-container-file";
   struct File {
     std::string Filename;
-    Buffer Contents;
+    TemporaryFile Contents;
+    std::unique_ptr<llvm::MemoryBuffer> Buffer;
```
aleclearmind, line 35: Why `unique_ptr`?

include/revng/PipeboxCommon/BinariesContainer.h
```diff
@@ -59,7 +66,18 @@ class BinariesContainer {
       std::string Hash = hash(Entry.Data);
       revng_assert(Entry.Filename.starts_with("binaries/"));
       revng_assert(Entry.Filename == "binaries/" + Hash);
-      Files.push_back({ std::move(Hash), std::move(Entry.Data) });
+
+      auto TempFile = llvm::cantFail(TemporaryFile::make(Prefix));
+      writeToFile({ Entry.Data.data(), Entry.Data.size() }, TempFile.path());
+      // Set the file as read-only so that it cannot be (easily) changed
+      llvm::sys::fs::setPermissions(TempFile.path(),
+                                    llvm::sys::fs::perms::owner_read);
```
aleclearmind, line 77: Why? Are you scared someone is going to abuse it?

include/revng/Model/Importer/DebugInfo/DwarfImporter.h
```diff
@@ -36,9 +36,7 @@ class DwarfImporter {
   TupleTree<model::Binary> &getModel() { return Model; }
 
 public:
-  void import(llvm::StringRef FileName, const ImporterOptions &Options);
-
-  void import(llvm::MemoryBuffer &Buffer,
+  void import(llvm::StringRef Filepath,
               const ImporterOptions &Options,
               llvm::StringRef Filename);
```
aleclearmind, line 44: I'm not happy with these arguments. The first and last seem redundant.

NOTETOSELF: review meaning of `Filename`.

include/revng/Model/Importer/DebugInfo/DwarfImporter.h
```diff
@@ -36,9 +36,7 @@ class DwarfImporter {
   TupleTree<model::Binary> &getModel() { return Model; }
 
 public:
-  void import(llvm::StringRef FileName, const ImporterOptions &Options);
-
-  void import(llvm::MemoryBuffer &Buffer,
+  void import(llvm::StringRef Filepath,
```
aleclearmind: AFAIU we already have a `MemoryBuffer`.
Maybe we should have a small object containing both the file name and the `MemoryBuffer`. What do you think?

lib/FunctionIsolation/IsolateFunctions.cpp
aleclearmind, line 772: I really think this stuff shouldn't happen in a the destructor of a class called `Isolate`.

Can you make a method with a sensible name that's invoked explicitly? Maybe in a distinct commit.

lib/FunctionIsolation/IsolateFunctions.cpp
```diff
@@ -772,10 +774,14 @@ Isolate::~Isolate() {
...
     T.advance(Address.toString(), true);
     std::set<const llvm::Function *> ToClone;
-    ToClone.insert(ClonedModule->getFunction(Function->getName()));
+    ToClone.insert(Function);
     ToClone.insert(ExternalFunctions.begin(), ExternalFunctions.end());
 
-    Output.assign(ObjectID(Address), ::cloneFiltered(*ClonedModule, ToClone));
+    Output.assign(ObjectID(Address), ::cloneFiltered(*ClonedRoot, ToClone));
+
+    // Since we saved the function to the output, delete its body in ClonedRoot
```
aleclearmind: I think `ClonedRoot` is not involved here.

include/revng/Pipebox/TupleTreeContainer.h
```diff
@@ -17,6 +17,7 @@ class TupleTreeContainer {
   static constexpr llvm::StringRef MimeType = "text/x.yaml";
 
 private:
+  bool Expiring = false;
```
aleclearmind: Let's figure out some good terminology.
"Expiring" is something that has to do with time.

Here the semantics is more like "moved", i.e., "you've full ownership on this object, no one is going to use it after you". What about `bool Disposable`?

As for "pruneIfPossible" term I'd go for "disposeIfPossible".

include/revng/PipeboxCommon/CliftContainer.h
```diff
@@ -73,9 +77,21 @@ class CliftFunctionContainer {
     return Result;
   }
 
+  void setIsExpiring() { Expiring = true; }
+
+  void pruneIfPossible() {
+    if (not Expiring)
+      return;
```
aleclearmind: Try to be consistent `if (Expering) { ... }` vs `if (not Expiring) return`.

include/revng/PipeboxCommon/BinariesContainer.h
```diff
@@ -105,6 +108,15 @@ class BinariesContainer {
 
   bool verify() const { return true; }
 
+  void setIsExpiring() { Expiring = true; }
+
+  void pruneIfPossible() {
```
aleclearmind: Return a `true` if disposed?

include/revng/PipeboxCommon/CliftContainer.h
```diff
@@ -23,19 +23,23 @@ class CliftFunctionContainer {
   static constexpr llvm::StringRef MimeType = "application/x.mlir.bc";
 
 private:
+  static constexpr auto Threading = mlir::MLIRContext::Threading::DISABLED;
```
aleclearmind, line 26: 1. Do not use capital stuff.
2. Mmmh.

include/revng/PipeboxCommon/CliftContainer.h
```diff
@@ -23,19 +23,23 @@ class CliftFunctionContainer {
...
   static inline const mlir::DialectRegistry MLIRDialectRegistry = ([]() {
     mlir::DialectRegistry Registry;
     Registry.insert<mlir::clift::CliftDialect>();
     return Registry;
   })();
 
 private:
-  mlir::MLIRContext Context;
+  bool Expiring = false;
+  std::unique_ptr<mlir::MLIRContext> Context;
```
aleclearmind: Stop using `unique_ptr` where a `std::optional` would be enough.

include/revng/PipeboxCommon/CliftContainer.h
```diff
@@ -23,19 +23,23 @@ class CliftFunctionContainer {
...
   std::map<ObjectID, mlir::OwningOpRef<mlir::ModuleOp>> Modules;
 
 public:
-  CliftFunctionContainer() :
-    Context(MLIRDialectRegistry, mlir::MLIRContext::Threading::DISABLED) {}
+  CliftFunctionContainer() {
+    Context = std::make_unique<mlir::MLIRContext>(MLIRDialectRegistry,
+                                                  Threading);
```
aleclearmind: Use `: Context(...)` syntax.

include/revng/PipeboxCommon/CliftContainer.h
```diff
@@ -73,9 +77,21 @@ class CliftFunctionContainer {
...
+  void setIsExpiring() { Expiring = true; }
+
+  void pruneIfPossible() {
+    if (not Expiring)
+      return;
+
+    Modules.clear();
+    Context = std::make_unique<mlir::MLIRContext>(MLIRDialectRegistry,
+                                                  Threading);
+    Expiring = false;
```
aleclearmind: This seems to hint to the fact that the module is back in a "reusable" state.
Not sure this is what we want.

include/revng/PipeboxCommon/LLVMContainer.h
```diff
@@ -22,12 +22,14 @@ class LLVMRootContainer {
   static constexpr llvm::StringRef MimeType = "application/x.llvm.bc";
 
 private:
-  llvm::LLVMContext Context;
+  bool Expiring = false;
+  std::unique_ptr<llvm::LLVMContext> Context;
   std::unique_ptr<llvm::Module> Module;
```
aleclearmind:
```c++
struct ModuleAndContext {
  llvm::LLVMContext Context;
  llvm::Module Module;
};
std::optional<ModuleAndContext> TheModuleAndContext;

File: python/revng/pypeline/pipeline.py [hunk 14]
```diff
@@ -386,20 +389,38 @@ def schedule(
...
-                        written_out[container_decl] = scheduled_task.outgoing[container_decl]
+                container_decl = argument.to_container_decl()
+                access = argument.access
+                if access != TaskArgumentAccess.READ and container_decl in scheduled_task.outgoing:
+                    written_out[container_decl] = scheduled_task.outgoing[container_decl]
+
+                if access == TaskArgumentAccess.WRITE:
+                    written_containers.add(container_decl)
+                elif access & TaskArgumentAccess.READ:
+                    read_containers.add(container_decl)
```
aleclearmind: Suspicious piece of code.

1. What about `READ_WRITE`?
2. Why do you use `&` in the second check?

If correct, add comments.

python/revng/pypeline/pipeline.py
```diff
@@ -386,20 +389,38 @@ def schedule(
...
             if written_out.empty() and parent_scheduled_task is not None:
                 # If here the task does not actually need to produce anything
                 # and can be skipped, by "glueing" its dependencies onto the parent
                 assert [scheduled_task] == parent_scheduled_task.depends_on
                 parent_scheduled_task.depends_on = scheduled_task.depends_on
             else:
+                # Actually set the expiring containers if needed
+                for read_container in read_containers:
+                    if read_container.name not in readers_encountered:
+                        # Skip expiring containers if this task is the first one
```
aleclearmind: Why?

python/revng/pypeline/schedule/scheduled_task.py
```diff
@@ -87,6 +92,9 @@ def run(self, containers: ContainerSet) -> ScheduledTaskDependencies | None:
         bindings = self.node.bindings
         result: ScheduledTaskDependencies | None = None
 
+        for container_decl in self.expiring_containers:
+            containers[container_decl].set_is_expiring()
```
aleclearmind: Do not use abbreviations

python/revng/pypeline/pipeline.py
```diff
@@ -386,20 +389,38 @@ def schedule(
...
+                elif access & TaskArgumentAccess.READ:
+                    read_containers.add(container_decl)
 
             if written_out.empty() and parent_scheduled_task is not None:
                 # If here the task does not actually need to produce anything
                 # and can be skipped, by "glueing" its dependencies onto the parent
                 assert [scheduled_task] == parent_scheduled_task.depends_on
                 parent_scheduled_task.depends_on = scheduled_task.depends_on
             else:
+                # Actually set the expiring containers if needed
```
aleclearmind: 1. Why "actually"?
2. I'm not a fan of doing this in `schedule`. I'd prefer doing this in a separate method `identify_expiring_containers`, as a post-processing. This method is already complex enough.
   Also, one day we might parallelize schedules, merge them and so on. It'd be beneficial to identify expiring containers *after* we do all that, right before `run`.

python/revng/pypeline/pipeline.py
```diff
@@ -386,20 +389,38 @@ def schedule(
...
+                for read_container in read_containers:
+                    if read_container.name not in readers_encountered:
+                        # Skip expiring containers if this task is the first one
+                        if parent_scheduled_task is not None:
+                            scheduled_task.expiring_containers.add(read_container)
+                        readers_encountered.add(read_container.name)
```
aleclearmind: Why `.name` instead of the container itself?

include/revng/Support/IRHelpers.h
```diff
@@ -1575,3 +1576,19 @@ std::unique_ptr<llvm::Module> cloneIntoContext(const llvm::Module &Module,
 /// modules being merged.
 void linkFunctionModules(std::unique_ptr<llvm::Module> &&Source,
                          std::unique_ptr<llvm::Module> &Destination);
+
+/// Helper class that allows getting the exhaustive list of functions called by
+/// a function
+class FunctionCallEnumerator {
+private:
+  const llvm::Module &Module;
+  std::map<const llvm::Function *, std::set<const llvm::Function *>>
```
aleclearmind: Use a `DenseSet` instead of `std::set<const llvm::Function *>`

include/revng/Support/IRHelpers.h
```diff
@@ -1575,3 +1576,19 @@ std::unique_ptr<llvm::Module> cloneIntoContext(const llvm::Module &Module,
...
+    CalledFunctions;
+
+public:
+  FunctionCallEnumerator(const llvm::Module &Module) : Module(Module) {}
+
+  RecursiveCoroutine<std::set<const llvm::Function *>>
```
aleclearmind: 1. Return a `const &`, not a copy.
2. Do not expose `RecursiveCoroutine` in `IRHelpers.h`.

include/revng/Support/IRHelpers.h
```diff
@@ -1575,3 +1576,19 @@ std::unique_ptr<llvm::Module> cloneIntoContext(const llvm::Module &Module,
...
+  getCalledFunctions(const llvm::Function &Function,
+                     const std::set<const llvm::Function *> &ToSkip);
```
aleclearmind: Not sure what `ToSkip` is for, seems something the caller can do.

include/revng/Support/IRHelpers.h
```diff
@@ -1575,3 +1576,19 @@ std::unique_ptr<llvm::Module> cloneIntoContext(const llvm::Module &Module,
 /// modules being merged.
 void linkFunctionModules(std::unique_ptr<llvm::Module> &&Source,
                          std::unique_ptr<llvm::Module> &Destination);
+
+/// Helper class that allows getting the exhaustive list of functions called by
+/// a function
+class FunctionCallEnumerator {
```
aleclearmind: This class purpose seems to enumerate the calls, nowhere you state that you enumerate all the *transitively* reachable calls.

What about `ReachableFunctionsEnumerator`?

python/revng/pypeline/schedule/schedule.py
```diff
@@ -109,7 +110,12 @@ def run(
     ) -> ContainerSet:
         for task in self.tasks:
             if isinstance(task.node.task, Pipe):
-                task.node.task.check_precondition(self.model)
+                try:
+                    task.node.task.check_precondition(self.model)
+                except RuntimeError as e:
```
aleclearmind, line 114: Maybe we should review the choice of using `RuntimeError`.

aleclearmind:
NOTETOSELF: Review
* `pypeline: expand `ScheduledTask`'s constructor`
* `isolate: selectively clone helpers`
* `pypeline: move `PipelineNode.run`

include/revng/PipeboxCommon/BinariesContainer.h
mrjackv: That's what's returned by `cloneIntoContext` and friends

include/revng/PipeboxCommon/BinariesContainer.h
mrjackv: The `TemporaryFile` is created as `u=rw`, after writing it, as a precaution, it's set to `u=r` so that nobody can intentionally/accidentally write to it

include/revng/Model/Importer/DebugInfo/DwarfImporter.h
mrjackv: > The first and last seem redundant.

They express two different things:
* First one: the actual file path to the file (e.g. in the pypeline `/tmp/binaries-container-file-73ad2c`)
* Second one: the name of the file, e.g. `bash` for `/bin/bash` or `qa_calc`, which is needed to look around for `.debug` file etc.

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: Sure

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: What do you mean? we're removing the function from the `ClonedRoot` module

include/revng/Pipebox/TupleTreeContainer.h
mrjackv: Sounds good

include/revng/PipeboxCommon/BinariesContainer.h
mrjackv: Hmmm, dunno, giving visibility into this kind of thing sounds like a footgun, I'd keep the function opaque. Do you have some specific use case where a user of the function would want to know if pruning actually happened?

include/revng/PipeboxCommon/CliftContainer.h
mrjackv: 1. What do you mean?

include/revng/PipeboxCommon/CliftContainer.h
mrjackv: Ah ok, I'll correct it

include/revng/PipeboxCommon/CliftContainer.h
mrjackv: Tecnically it is, they way things are laid out right now if you have something like these 4 pipes (with R/W if they are going to read or write the container)
1. P(W)
2. P(R)
3. P(W)
4. P(R)

Then pipes (2) and (4) will be "expiring" the container

include/revng/PipeboxCommon/LLVMContainer.h
mrjackv: "naked" modules (e.g. without `std::unique_ptr`) are really hard to handle because they are not copyable or movable, and all the LLVM infrastructure (e.g. `parseIR`, `cloneModule`) always returns a `std::unique_ptr<llvm::Module>`

python/revng/pypeline/pipeline.py
mrjackv: Basically, you cannot mark a container as expiring if it is on the node where you ask for the artifact, otherwise once you try to extract it via `.deserialize` you get an error/nothing

python/revng/pypeline/pipeline.py
mrjackv: Both are there because `READ_WRITE` is handled
The first `if` basically determines if the container is "truncated", which allows to add another pipe as expiring
The second `if` basically checks "is the access READ or READ_WRITE", if so and it's the "last to read" then it marks that container as expiring

python/revng/pypeline/pipeline.py
mrjackv: 1. The part above, where the `access` is checked is just probing, we can't actually call `set_expiring` until we're check that the task hasn't been skipped because it actually wrote nothing
2. Implementing the general correct algorithm requires spinning up a proper liveness analysis, here's the poor man's version which assumes a straight line of pipes. When parallelizing, if a container was expiring in the single-threaded version then it is expiring also in the parallel version. We never talked about merging schedules.

include/revng/Support/IRHelpers.h
mrjackv: 1. Ok
2. Do you mean I should forward-declare `class RecursiveCoroutine` in the IRHelpers.h file?

include/revng/Support/IRHelpers.h
mrjackv: In `isolate`, it's to specify that other isolated functions should not be explored and skipped, since we do not want their subtree to end up in the function set.
Now that I think about it, `ToSkip` breaks the caching present in the class, maybe this should be downgraded to a simple function.

python/revng/pypeline/schedule/schedule.py
mrjackv: The long-term plan is to have proper `llvm::Error` subclasses and convert them to the proper `PypelineException` equivalent (there's a TODO in `Casters.h`) but I didn't want to front-load this PR with that

include/revng/Model/Importer/DebugInfo/DwarfImporter.h
aleclearmind: Yes, but passing `/tmp/binaries-container-file-73ad2c` for the latter is not the right thing to do.

lib/FunctionIsolation/IsolateFunctions.cpp
aleclearmind: `Root` is a function, maybe the name `ClonedModule` was more appropriate.

include/revng/PipeboxCommon/BinariesContainer.h
aleclearmind: Maybe asserting under certain conditions, but no biggie.

include/revng/Model/Importer/DebugInfo/DwarfImporter.h
mrjackv: That never happens, the argument is passed twice only in the old pipeline and CLI tools

include/revng/PipeboxCommon/LLVMContainer.h
aleclearmind: Mh, OK, maybe leave a comment.

python/revng/pypeline/pipeline.py
aleclearmind: Why don't we add to container to `written_containers` in case of `TaskArgumentAccess.READ_WRITE`?
Maybe you want a different name for `written_containers`?

python/revng/pypeline/pipeline.py
aleclearmind: This is not just a problem with the first node. It's a general problem, we can have artifacts anywhere.

include/revng/Support/IRHelpers.h
aleclearmind: 2. No, try to push its usage to a `static` function in the `.cpp` file, so it doesn't pop up here.

include/revng/Support/IRHelpers.h
aleclearmind: > In isolate, it's to specify that other isolated functions should not be explored and skipped, since we do not want their subtree to end up in the function set

Can you be more specific?

python/revng/pypeline/pipeline.py
mrjackv: The logic is like this:
* If you're the last reader of a container, the container is disposable
* If you `WRITE` (not `READ_WRITE`) a container, then if a task before you reads it that one also can treat the container as disposable
* When you `READ_WRITE` you cannot say that a task before you that reads it will be disposable. But if you're the last one to read it (effectively making you READ for the purposes of the schedule) you can mark it as disposable

python/revng/pypeline/pipeline.py
aleclearmind: > Implementing the general correct algorithm requires spinning up a proper liveness analysis

This is an optimization, so you can apply it only if it's a straight line.
My concern is on pushing it in this already non trivial code.
I really want this into an isolated function, also easier to review.

> We never talked about merging schedules.

We did at a certain point, I don't remember the context right now. We were talking with Tommy.

python/revng/pypeline/pipeline.py
mrjackv: Remeber, this is in the `def schedule` function, there is **one** node with artifacts (the last one)

include/revng/Support/IRHelpers.h
mrjackv: say I have the following call graph (with X -> Y meaning "X calls Y") when isolating `local_0x1000`
```
local_0x1000 -> local_0x2000 -> helper_foo
```

When `cloneFiltered`-ing want both `local_0x2000` **and** `helper_foo` to be converted to declarations

include/revng/Model/Importer/DebugInfo/DwarfImporter.h
aleclearmind: So, in the new pipeline if I do `revng quick artifact /bin/bash` will I get `/bin/bash` as `Filename`? I don't mean `bash`, I mean `/bin/bash`.

mrjackv: No, just `bash`

python/revng/pypeline/pipeline.py
aleclearmind: > there is one node with artifacts

`Pipeline.schedule` is not related to artifacts in any way, it just gets to some `PipelineNode`.

I really think this is not the right place to do this.
We need a separate function computing this stuff just looking at the schedule. If you only want to handle straight lines of pipes, that's fine, but conflating this into `schedule` is not nice.

python/revng/pypeline/schedule/schedule.py
aleclearmind: OK, maybe leave a comment.

python/revng/pypeline/pipeline.py
mrjackv: Ok, let me rephrase it differently
Given **just** a schedule and omitting edge cases like `W -> R -> W -> R`, you cannot make any assertion that a container is expiring because, for what the `Schedule` alone is concerned, all the containers will be read by the caller once the `run` function is concerned. So the possible solutions are:
* We give a list of `containers` that **will** be read from the output of run in the constructor of `Schedule`
* `Schedule.identify_expiring_containers` gets a list of output containers
* stays as a function of `Pipeline`

python/revng/pypeline/schedule/scheduled_task.py
```diff
@@ -35,28 +33,38 @@ class ScheduledTask:
...
-    """
+    def __init__(
+        self,
+        node: PipelineNode,
+        model: ReadOnlyModel,
+        storage_provider: StorageProvider,
+        pipeline_configuration: PipelineConfiguration,
+    ):
+        self.node = node
+        """The node of the pipeline we scheduled."""
```
aleclearmind, line 48: Docstrings go *after* the definition uh?

python/revng/pypeline/pipeline.py
```diff
@@ -637,7 +640,9 @@ def deserialize_schedule(self, schedule: str) -> Schedule:
                 raise ValueError(f"Unknown task type: \"{task['type']}\"")
 
             depends_on = [scheduled_tasks[i] for i in task["depends_on"]]
-            scheduled_task = ScheduledTask(pipeline_node, False, outgoing, incoming, depends_on)
+            scheduled_task = ScheduledTask(pipeline_node, model, storage_provider, configuration)
+            scheduled_task.request(incoming, outgoing)
+            scheduled_task.depends_on = depends_on
```
aleclearmind: 1. Is `.request` used elsewehere? If not, add to `__init__`.
2. Why not add `depends_on` to `__init__`?

tests/pypeline/test_all.py
```diff
@@ -618,14 +604,9 @@ def test_schedule_serdes(model):
...
-    begin_node.run(
-        model=ReadOnlyModel(model),
-        containers={root1: container},
-        incoming=requests,
-        outgoing=requests,
-        pipeline_configuration=pipeline_configuration,
-        storage_provider=storage_provider,
-    )
+    task = ScheduledTask(begin_node, ReadOnlyModel(model), storage_provider, pipeline_configuration)
+    task.request(requests, requests)
```
aleclearmind: I really think we should drop the `request` method in favor of a new `__init__` argument.

python/revng/pypeline/schedule/scheduled_task.py
mrjackv: Yes, just like the docstring for a function is after the `def` line

python/revng/pypeline/pipeline.py
mrjackv: 1. Yes, in `Pipeline.schedule`
2. Because it's done iteratively in `Pipeline.schedule`

tests/pypeline/test_all.py
mrjackv: See answer to the previous comment

python/revng/pypeline/pipeline.py
aleclearmind: 2. We can have `.add_requests` (as opposed to `.request`) but I think having it in `__init__` helps as well.

include/revng/Pipebox/TupleTreeContainer.h
mrjackv: Renamed to `Disposable`

include/revng/PipeboxCommon/CliftContainer.h
mrjackv: Stuck with `if (not Expiring) return;`

include/revng/PipeboxCommon/CliftContainer.h
mrjackv: Fixed everywhere

include/revng/PipeboxCommon/CliftContainer.h
mrjackv: Done

include/revng/PipeboxCommon/LLVMContainer.h
mrjackv: Done

python/revng/pypeline/schedule/scheduled_task.py
mrjackv: Fixed

python/revng/pypeline/pipeline.py
mrjackv: Used the `ContainerDeclaration`

include/revng/Model/Importer/DebugInfo/DwarfImporter.h
mrjackv: Fixed now, there's `/bin/bash` in the `BinaryIdentifier.Name` now

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: Done

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: Renamed to `ClonedModule`

include/revng/Support/IRHelpers.h
mrjackv: Done

include/revng/Support/IRHelpers.h
mrjackv: Pushed to a `static` function

include/revng/Support/IRHelpers.h
mrjackv: Fixed

include/revng/Support/IRHelpers.h
mrjackv: Renamed to `ReachableFunctionsEnumerator`

python/revng/pypeline/schedule/schedule.py
mrjackv: Added a TODO

include/revng/Model/Importer/DebugInfo/DwarfImporter.h
mrjackv: Pushed everything into an object called `{ELF|COFF|MachO}Binary`

python/revng/pypeline/pipeline.py
mrjackv: Done

include/revng/PipeboxCommon/CliftContainer.h
aleclearmind: OK, maybe take a quick look at all the containers and make sure there are not "leftovers" upon discard.

python/revng/pypeline/pipeline.py
```diff
@@ -406,11 +410,15 @@ def schedule(
...
 
-        return Schedule(
+        result = Schedule(
             {v for v in self.declarations if v.name in used_declatations},
             tasks[target_node],
             pipeline_configuration,
+            model,
+            storage_provider,
         )
+        result._identify_discardable_containers()
```
aleclearmind: 1. Calling a private method is not nice. Doesn't the linter complain about this?
2. What if we run this in the beginning of the `run` method?

python/revng/pypeline/schedule/schedule.py
```diff
@@ -244,3 +250,48 @@ def serialize(self) -> str:
...
+        """
+        Given a schedule, set the ScheduleTasks with the containers that will
+        be discarded when the schedule is executed.
+        """
+
+        scheduled_task: ScheduledTask = self.root
+        # Assume that all the outgoing request of the root task are going to
+        # be read, so the caller of the `run` method is implicitly the last reader
+        readers_encountered: set[ContainerDeclaration] = {
+            cd for cd, objects in self.root.outgoing.items() if len(objects) != 0
```
aleclearmind: Maybe we should rename `root` to `target_node`. What do you think?

I understand it's the "root" from the perspective of who builds the schedule, but it's rather confusing. Consider that the pipeline also as a "root" but it's the opposite concept: the root of the schedule is where you end up, the root of the pipeline is where you start from. Quite confusing.

python/revng/pypeline/pipeline.py
mrjackv: 1. The linter doesn't complain. The thing is having it as a public method is very wrong, the user should not be calling it.
2. Sure

python/revng/pypeline/schedule/schedule.py
mrjackv: I'd go with `target_task`, since it's a `ScheduledTask`

python/revng/pypeline/schedule/schedule.py
```diff
@@ -244,3 +250,48 @@ def serialize(self) -> str:
...
+        }
+
+        # Inspect the tasks in backward order, from the last one to the first.
+        # The logic used is the following:
+        # * If a task reads a container last (either with READ or READ_WRITE)
+        #   then it should be marked as disposable
```
aleclearmind: * If a task is the last one to read (either with READ or READ_WRITE) a container, the container should be marked as disposable.
* If a task clobbers a container (with WRITE) then the container can be marked as disposable in the preceding task

python/revng/pypeline/schedule/schedule.py
```diff
@@ -244,3 +250,48 @@ def serialize(self) -> str:
...
+        # * If a task clobbers a container (with WRITE) then a previous task
+        #   that reads it can mark the container as disposable
+        while True:
+            # Assume that this schedule is a straight line of tasks, as it
+            # simplifies the logic needed
+            assert len(scheduled_task.depends_on) in (0, 1)
+
+            for argument in scheduled_task.node.arguments_with_access:
```
aleclearmind: `s/arguments_with_access/arguments/`

python/revng/pypeline/schedule/schedule.py
```diff
@@ -244,3 +250,48 @@ def serialize(self) -> str:
...
+                container_declaration = argument.to_container_decl()
```
aleclearmind: `s/to_container_decl/declaration/`

python/revng/pypeline/schedule/schedule.py
```diff
@@ -244,3 +250,48 @@ def serialize(self) -> str:
...
+            cd for cd, objects in self.root.outgoing.items() if len(objects) != 0
+        }
+
+        # Inspect the tasks in backward order, from the last one to the first.
+        # The logic used is the following:
+        # * If a task reads a container last (either with READ or READ_WRITE)
+        #   then it should be marked as disposable
+        # * If a task clobbers a container (with WRITE) then a previous task
+        #   that reads it can mark the container as disposable
+        while True:
```
aleclearmind: Maybe

```python
while scheduled_task is not None:
  # ...
  scheudled_task = scheduled_task.depends_on[0] if ... else None
```

python/revng/pypeline/schedule/schedule.py
```diff
@@ -244,3 +250,48 @@ def serialize(self) -> str:
...
+            # Assume that this schedule is a straight line of tasks, as it
+            # simplifies the logic needed
+            assert len(scheduled_task.depends_on) in (0, 1)
```
aleclearmind: `s/depends_on/dependencies/`

`depends_on` sounds like a predicate, something returning a `bool`.

python/revng/pypeline/schedule/schedule.py
mrjackv: `arguments` is a different property which returns `list[ContainerDeclaration]`

aleclearmind: Sounds a bit backwards to me. `arguments` has a simple name but it's computed from `arguments_with_access`.

I'd consider dropping `arguments` and pushing its code to callers or, if very handy, call it `argument_declarations`.

include/revng/PipeboxCommon/CliftContainer.h
mrjackv: Checked

python/revng/pypeline/pipeline.py
mrjackv: Moved to `Schedule.run`

python/revng/pypeline/schedule/schedule.py
mrjackv: Done

python/revng/pypeline/schedule/schedule.py
mrjackv: Fixed the comment

python/revng/pypeline/schedule/schedule.py
mrjackv: Renamed to `arguments` and `argument_declarations`, respectively

python/revng/pypeline/schedule/schedule.py
mrjackv: Done

python/revng/pypeline/schedule/schedule.py
mrjackv: Done

python/revng/pypeline/schedule/schedule.py
mrjackv: Fixed

PR #554, opened 2026-01-13
Implement preliminaries for pypeline tracing

mrjackv:


lib/PipeboxCommon/Python/_pipebox.cpp
```diff
@@ -19,39 +19,63 @@
...
 static void handleSignal(int SigNo) {
-  const SignalHandler &Handler = SavedSignals[SigNo];
-  if (Handler.IsTerminating)
-    llvm::sys::RunInterruptHandlers();
-  Handler.Handler(SigNo);
+  const sighandler_t &Handler = SavedSignals[SigNo];
+
+  {
+    // re-acquire the GIL since the signal might have been received while a pipe
+    // is running
```
aleclearmind: If the signal is triggered in Python code, won't this trigger a deadlock?

lib/PipeboxCommon/Python/_pipebox.cpp
```diff
@@ -19,39 +19,63 @@
...
+    nanobind::gil_scoped_acquire X;
+
+    if (SigNo == SIGINT) {
+      // Reset the signal handler to avoid calling this function twice
+      signal(SigNo, SIG_IGN);
```
aleclearmind: I think the signal is blocked while the signal handler is running.
I think this is not necessary.

lib/PipeboxCommon/Python/_pipebox.cpp
```diff
@@ -19,39 +19,63 @@
...
+      // allocated are still alive and reports them as leaks. Since we're
+      // calling `Py_Exit` the garbage collector is not run and so all
+      // variables are still alive (but their destructors are still called).
+      nanobind::set_leak_warnings(false);
+      Py_Exit(1);
+    } else {
+      // Dispatch to the original python signal handler. If an exception is
+      // raised its throwing might be delayed if there is a piece of native
+      // code currently running.
+      Handler(SigNo);
```
aleclearmind, line 46: Are you sure Python's signal handlers are supposed to be invoked with the GIL taken? 🤔

lib/PipeboxCommon/Python/_pipebox.cpp
```diff
@@ -19,39 +19,63 @@
...
+
+    // Reset signal handlers on the python side, this avoids, e.g. `asyncio`
+    // from mistakenly thinking the default signal handler is installed
+    nanobind::object SignalFunction = importObject("signal.signal");
+    nanobind::object
+      SignalHandler = nanobind::cpp_function([](int, nanobind::object) {
+        return;
+      });
+    for (int SigNumber : std::ranges::views::keys(SavedSignals))
+      SignalFunction(nanobind::int_(SigNumber), SignalHandler);
```
aleclearmind: I'm confused. What are we doing here exactly?

Below we set the signal handlers again.
Why do we  need to blank them first?
Are we trying to fake some internal state of the `signal` module?

lib/Support/Progress.cpp
```diff
@@ -83,18 +106,44 @@ class TraceProgressListener : public llvm::ProgressListener {
...
+    std::string Buffer;
+    unsigned long long RSSKbytes = 0ULL;
+
+    while (not IS.eof()) {
+      std::getline(IS, Buffer);
+      if (Buffer.starts_with("Rss:")) {
+        llvm::StringRef BufferRef(Buffer);
+        BufferRef.consume_front("Rss:");
+        BufferRef.consume_back("kB");
+        BufferRef.trim().getAsInteger(10, RSSKbytes);
```
aleclearmind: You need to check the returned bool and bail out in case of failure.

lib/Support/Progress.cpp
```diff
@@ -83,18 +106,44 @@ class TraceProgressListener : public llvm::ProgressListener {
...
     Output << "\n";
   }
+
+  void checkMemoryPolling() {
+    auto Now = std::chrono::high_resolution_clock::now();
+    if (not MemoryPollEnabled or Now - LastMemoryPoll < MemoryPollInterval)
+      return;
+
+    // TODO: linux-specific and requires `smaps_rollup` which is linux 4.14+
+    std::ifstream IS("/proc/self/smaps_rollup");
```
aleclearmind, line 121: Avoid `std::stream`s in favor of LLVM's ones.

lib/Support/Progress.cpp
```diff
@@ -83,18 +106,44 @@ class TraceProgressListener : public llvm::ProgressListener {
...
+    Output << "\"ts\": " << getTimestamp() << ", ";
     Output << "\"pid\": " << getpid() << ", ";
     Output << "\"tid\": " << getpid();
     Output << "}";
     if (EmitTrailingComma)
       Output << ",";
     Output << "\n";
   }
+
+  void checkMemoryPolling() {
```
aleclearmind: This function does not "check" something. Just go with `emitMemoryUsage`.

lib/Support/Progress.cpp
```diff
@@ -65,6 +78,7 @@ class TraceProgressListener : public llvm::ProgressListener {
       emitEvent(T->stepName(), "task", "E");
     }
     emitEvent(T->name(), "task", "E");
+    checkMemoryPolling();
```
aleclearmind: Just call it from `emitEvent`.

lib/Support/Progress.cpp
```diff
@@ -53,10 +63,13 @@ class TraceProgressListener : public llvm::ProgressListener {
     }
   }
 
+  void enableMemoryPoller() { MemoryPollEnabled = true; }
```
aleclearmind: `MemoryPoller` is *how* you do it.
Names should reflect *what* you do. `MemoryProfiler`.

lib/Support/Progress.cpp
```diff
@@ -22,6 +23,9 @@ extern "C" {
 
 static void destroyTraceProgressListener(void *OpaqueListener);
 
+class TraceProgressListener;
+static TraceProgressListener *TPL = nullptr;
```
aleclearmind: Very very bad.

Please don't do these hacks and don't submit them for review, in particular not without any comment.

I assume this is because of command line parsing taking place after instantiation of the `class`, correct?

In any case, I think this feature can be enabled by default. We can handle turn it off if it turns out to be problematic.

python/revng/internal/pipebox.py
```diff
@@ -49,6 +49,19 @@ def signature(cls) -> tuple[TaskArgument, ...]:
             ),
         )
 
+    def _get_file_requests(self, model: ReadOnlyModel):
+        # TODO: add facilities to `model` that avoid having to serialize the
+        # whole thing
+        model_obj = Binary.deserialize(model.serialize().decode())
```
aleclearmind, line 63: 1. Never use `_obj` suffixes.
2. I don't understand, we have `.clone`.
3. Why are we even doing this? Please better comment your code.

python/revng/pypeline/task/pipe.py
```diff
@@ -166,6 +166,14 @@ def run(
         length as `SIGNATURE`.
         """
 
+    def needed_files(self, model: ReadOnlyModel) -> list[FileRequest]:
+        """
+        Request the list of file hashes that would be requested as part of the
+        `run` method via the `FileProvider`. This is required otherwise system
+        mode cannot work properly.
```
aleclearmind: > This is required otherwise system mode cannot work properly.

This does not explain much. What do you mean?
Also, it's no longer called "system mode".

python/revng/pypeline/utils/__init__.py
```diff
@@ -102,3 +106,39 @@ async def __call__(self) -> AsyncIterator[T]:
         """
         async with self._lock:
             yield self._value
+
+
+class DebugContext:
+    def __init__(
+        self,
+        use_system: bool = False,
+        argv_hook: Callable[[list[str]], list[str]] | None = None,
```
aleclearmind: Outline and document `Callable[[list[str]], list[str]]`

python/revng/pypeline/utils/__init__.py
```diff
@@ -102,3 +106,39 @@ async def __call__(self) -> AsyncIterator[T]:
...
+        base_dir: Path | None = None,
```
aleclearmind: `s/_dir/_directory/g`

python/revng/pypeline/cli/pipeline/run_pipe.py
```diff
@@ -147,6 +150,10 @@ def build_pipe_command(
         type=click.Path(exists=True, file_okay=False, path_type=Path),
         default=Path.cwd(),
     )
+    @click.option("--dependencies", type=click.Path(dir_okay=False, path_type=Path), default=None)
+    @click.option(
+        "--custom-invalidation", type=click.Path(dir_okay=False, path_type=Path), default=None
+    )
```
aleclearmind: 1. Docs the flags and the format of their outputs.
2. Given that this is a new format we're making up, let's merge them in a single file. No reason to have two options, one emitting a YAML and one a tar.
   Either let's put the YAML in the tar or let's put in base64 (or is it textual) the contents of the tar in the YAML.

python/revng/pypeline/schedule/scheduled_task.py
```diff
@@ -91,16 +100,22 @@ def add_requests(self, incoming: Requests, outgoing: Requests) -> None:
...
         """
         Run the task with the requests computed during the scheduling phase.
         """
         if self.completed:
             raise RuntimeError(f"ScheduledTask {self.node} has already been run.")
+
+        if getattr(self.node, "_dummy", False):
+            self.completed = True
+            return ScheduledTaskDependencies([], [])
+
```
aleclearmind: Bad hack.

I don't even think we should have a `DummyPipelineNode` class.

Can't you just have a regular dummy `Pipe`?
These hacks are calling for more hacks. Let's fix this.

python/revng/pypeline/schedule/scheduled_task.py
```diff
@@ -120,17 +135,101 @@ def run(self, containers: ContainerSet) -> ScheduledTaskDependencies | None:
...
+                model_type = get_singleton(Model)  # type: ignore[type-abstract]
+
+                cmd = ["pype", "pipeline", "run-pipe", task.name, "--tar"]
+                cmd += ["--static-configuration", task.static_configuration]
+                cmd += ["--configuration", configuration]
+                cmd += ["--dependencies", str(temp_dir_path / "dependencies")]
+                cmd += ["--custom-invalidation", str(temp_dir_path / "custom-invalidation")]
+
+                # Populate the temporary directory with the files needed by the
+                # pipe
```
aleclearmind: 1. Add a comment clearly describing the layout of the debug directory.
2. Split off this code into:
   ```python
   prepare_debug_directory()
   argv = prepare_pype_invocation()
   run(argv)
   load_pipe_results()
   ```

python/revng/pypeline/schedule/scheduled_task.py
```diff
@@ -120,17 +135,101 @@ def run(self, containers: ContainerSet) -> ScheduledTaskDependencies | None:
...
-                model=self.model,
-                containers=pipe_containers,
-                incoming=pipe_incoming,
-                outgoing=pipe_outgoing,
-                configuration=self.pipeline_configuration.get(task, ""),
-            )
+            configuration = self.pipeline_configuration.get(task, "")
+            file_provider = StorageProviderFileProvider(self.storage_provider)
+
+            if debug_context.use_system:
```
aleclearmind: We're inlining way too much logic here.

I think we should rename `DebugContext` into `PipeRunner` or `PipeExecutor`:

```python
class PipeRunner:
    def __init__(self, ...):
        self.use_system  = ...

    def run(pipe: Pipe, file_provider, model, ...):
        if not self.use_system:
            pipe_output = task.run(
                file_provider=file_provider,
                model=self.model,
                containers=pipe_containers,
                incoming=pipe_incoming,
                outgoing=pipe_outgoing,
                configuration=configuration,
            )
        else:
            ...
```

And here we just to `runner.run(pipe, file_provider, model, ...)`.

Same for the analyses.

python/revng/pypeline/utils/__init__.py
```diff
@@ -102,3 +106,39 @@ async def __call__(self) -> AsyncIterator[T]:
...
+    def run_command(self, cmd: list[str]):
+        assert self.use_system
+        if self._argv_hook is not None:
+            cmd = self._argv_hook(cmd)
+
+        pypeline_logger.debug_log(f"Running debug command: {cmd}")
+        result = subprocess.run(cmd)
+        if result.returncode != 0:
+            pypeline_logger.log("Debug command failed! Re-run it with:")
+            pypeline_logger.log(shlex.join(cmd))
```
aleclearmind: And then? We can't just proceed.

python/revng/pypeline/cli/utils.py
```diff
@@ -322,6 +330,30 @@ def compute_objects(
 )
 
 
+def _parse_debug_option(path: str, ctx):
+    pipebox = ctx.obj["pipebox"]
+    return DebugContext(True, getattr(pipebox, "argv_hook"), Path(path))
```
aleclearmind: You really need to avoid or *heavily* justify your usages of `getattr`.
They are a red flag.

python/revng/pypeline/cli/utils.py
```diff
@@ -322,6 +330,30 @@ def compute_objects(
...
+    type=EagerParsedPath(
+        name="debug_context",
+        parser=_parse_debug_option,
+        default=DebugContext(),
+        dir_okay=True,
+        file_okay=False,
+    ),
+    help=(
+        "Run the command in debug mode. Where possible the pipes and analyses"
+        " will be run as subcommands via `run-pipe` and `run-analysis`."
```
aleclearmind: State that you need to pass a directory and how it will be populated.

Note: the directory might not exists.

python/revng/pypeline/cli/utils.py
```diff
@@ -113,10 +117,14 @@ def __init__(
         super().__init__(*args, **kwargs)
         self.parser = parser
         self.name = name
+        self.default = default
 
     def convert(
         self, value: Any, param: Optional[click.Parameter], ctx: Optional[click.Context]
     ) -> Any:
+        if value is self.__class__.DEFAULT:
+            return self.default
```
aleclearmind, line 129: Uh?

python/revng/pypeline/schedule/scheduled_task.py
```diff
@@ -120,17 +135,101 @@ def run(self, containers: ContainerSet) -> ScheduledTaskDependencies | None:
...
+                    if argument.access & TaskArgumentAccess.WRITE:
+                        cmd_epilogue.append(str(temp_dir_path / f"{arg_name}_output.tar"))
+                        argument_outgoing = self.outgoing.get(bindings[index])
+                        if len(argument_outgoing) > 0:
+                            cmd += [
+                                f"--{arg_name}-objects",
+                                ",".join(x.serialize() for x in argument_outgoing),
+                            ]
+
+                debug_context.run_command([*cmd, *cmd_epilogue])
```
aleclearmind: I don't see the reproducer script being serialized anywhere.

lib/PipeboxCommon/Python/_pipebox.cpp
mrjackv: `gil_scoped_acquire` calls `PyGILState_Ensure` which is re-entrant, so this is not a problem.

lib/PipeboxCommon/Python/_pipebox.cpp
mrjackv: I get a crash otherwise within python, so I'm sure this is the case.

lib/PipeboxCommon/Python/_pipebox.cpp
mrjackv: There is some python library code that does things differently if `signal.getsignal(XXX) == python.default_handler_XXX`, since we're overriding the signals on the "C side" we must also modify them in python so that the above doesn't happen.

lib/Support/Progress.cpp
mrjackv: `llvm::MemoryBuffer` is really annoying to use, since it doesn't have an equivalent `std::getline` and loads the entire file in memory.

lib/Support/Progress.cpp
mrjackv: > I assume this is because of command line parsing taking place after instantiation of the class, correct?

* The `--trace-memory` callback is called after the class is constructed
* The class is owned internally by `llvm::ProgressReport` so there's no way to fish it out once you've constructed it.

> In any case, I think this feature can be enabled by default. We can handle turn it off if it turns out to be problematic.

Ok

python/revng/internal/pipebox.py
mrjackv: 1. Ok
2. This is not to have a copy, this is to have the python-equivalent object where I can do `model_obj["Binaries"]` and get a list. The `ReadOnlyModel` is an opaque C++ object.
3. See (2)

python/revng/pypeline/schedule/scheduled_task.py
mrjackv: > Can't you just have a regular dummy Pipe?

Sure, but then in the graph of the pipeline (and eventually, the docs) you would get a root pipe named `dummy-pipe` which is confusing.

lib/PipeboxCommon/Python/_pipebox.cpp
aleclearmind: OK, add a comment.

lib/PipeboxCommon/Python/_pipebox.cpp
aleclearmind: OK, add a comment.

lib/PipeboxCommon/Python/_pipebox.cpp
aleclearmind: So `getsignal` does not return the actual signal handler (that we're setting from C++ via `signal(...)`) but some cached version?

Am I getting this right?

lib/Support/Progress.cpp
aleclearmind: You could use `StringRef.split`, but OK.

I was thinking about output streams, there's no `llvm::raw_input_stream`.

lib/PipeboxCommon/Python/_pipebox.cpp
mrjackv: From python `getsignal` returns a python function which is stored in a list in the `_signal` module.
This differs from the return value of `getsignal` which must be a C function pointer.

python/revng/internal/pipebox.py
aleclearmind: Ah `Binary` is not the model from pypeline. It's our generated wrapper.
Add a comment.

Anyway, the model should be small the first time we run this and it shouldn't invalidate too much.

python/revng/pypeline/schedule/scheduled_task.py
mrjackv: The issue is that the debug context needs to be carried across between invocations (e.g. analyses list calls multiple `Schedule.run` and `Analysis.run` and they need to end up in the same dir).

I'm thinking of having 3, a `DebugContext` which does what's currently doing, and then `PipeRunner` and `AnalysisRunner` which have a `DebugContext` in the constructor and behave as you described

python/revng/pypeline/schedule/scheduled_task.py
aleclearmind: I meant a dummy pipe that's not revng-specific.

> you would get a root pipe named dummy-pipe which is confusing.

Not too bad to be honest. We're doing too many hoops to avoid something like this.

lib/PipeboxCommon/Python/_pipebox.cpp
aleclearmind: OK, then there's some internal state in `signal` that we're trying to tamper with.

Add a comment explaining this.

python/revng/pypeline/cli/utils.py
mrjackv: The `argv_hook` is an optional feature of a pipebox, hence the `getattr`.
If there should be a comment explaining that I'm all for it.

python/revng/pypeline/schedule/scheduled_task.py
aleclearmind: Or maybe they could all be a single `RunnerContext` class that has `run_pipe` and `run_analysis`.

python/revng/pypeline/cli/utils.py
mrjackv: This is a workaround for `click`, which basically does not work well if you have a custom `type` and a default value. Basically the `DEFAULT` object is a marker object that you use in conjunction with `EagerParsedPath` to allow **IT** to return the correct default.

python/revng/pypeline/schedule/scheduled_task.py
mrjackv: Sorry, I didn't write it down and it got lost in all the discussions we had.

python/revng/pypeline/cli/utils.py
aleclearmind: My concern is using `getattr` vs `is not None`.
Let's use `getattr` if we incur in a signicant, recurrent pain if we try to avoid it.

In this case, the pipebox can just expose it and set it to `None`, can't it?

python/revng/pypeline/cli/utils.py
aleclearmind: OK, add a comment.

python/revng/pypeline/cli/utils.py
mrjackv: Currently not having the attribute or doing `argv_hook = None` are equivalent. But if the pipebox ever public interface it's easier to say "a pipepox has an `initialize` function and that's it" rather than "a pipebox has an initialize function, an `argv_hook` that you can set to None" and whatever other tendrils will come up with time

aleclearmind: It's easier but also obscure, it looks like there are "secret features".
Anyway, your call. Let's try to contain the usage of `getattr`, in general.

lib/PipeboxCommon/Python/_pipebox.cpp

lib/PipeboxCommon/Python/_pipebox.cpp
mrjackv: Dropped

lib/PipeboxCommon/Python/_pipebox.cpp
mrjackv: Done

lib/PipeboxCommon/Python/_pipebox.cpp
mrjackv: Expanded the comment to better explain what's going on

lib/Support/Progress.cpp
mrjackv: Done

lib/Support/Progress.cpp
mrjackv: Done

lib/Support/Progress.cpp
mrjackv: Done

lib/Support/Progress.cpp
mrjackv: Changed the appropriate variables

lib/Support/Progress.cpp
mrjackv: Dropped the pointer to the instance and the `--trace-memory` option

python/revng/internal/pipebox.py
mrjackv: Changed the comment to better explain what's going on

python/revng/pypeline/task/pipe.py
mrjackv: Changed the comment to better explain the situation

python/revng/pypeline/utils/__init__.py
mrjackv: Done

python/revng/pypeline/utils/__init__.py
mrjackv: Done

python/revng/pypeline/schedule/scheduled_task.py
mrjackv: Given that the code has been moved to `RunnerContext.run_pipe` and the body is much simpler, should this still be done?

python/revng/pypeline/schedule/scheduled_task.py
mrjackv: Renamed `DebugContext` to `RunnerContext` and moved everything into `run_pipe` and `run_analysis`

python/revng/pypeline/cli/utils.py
mrjackv: Changed to plain attribute access

python/revng/pypeline/cli/utils.py
mrjackv: Done

python/revng/pypeline/schedule/scheduled_task.py
mrjackv: Dropped all the hacks

python/revng/pypeline/utils/__init__.py
mrjackv: added `raise PypelineError`

python/revng/pypeline/cli/utils.py
mrjackv: Expanded on the help text to better explain what the option does

python/revng/pypeline/schedule/scheduled_task.py
mrjackv: Implemented

python/revng/pypeline/schedule/scheduled_task.py
aleclearmind: Yes, the logic is still verbose and non-trivial.

mrjackv: Done

python/revng/pypeline/cli/pipeline/run_pipe.py
mrjackv: 1. Done
2. Went with tar with the normal dependencies inside as a file

python/revng/pypeline/runner_context.py
```diff
@@ -0,0 +1,276 @@
...
+            new_model = model.clone()
+            try:
+                analysis.run(
+                    model=new_model,
+                    containers=containers,
+                    incoming=incoming,
+                    configuration=configuration,
+                )
+            # TODO: eventually the pipe will raise `PypelineException` directly
+            except RuntimeError as e:
```
aleclearmind, line 295: Is this the strictest thing we can catch?

python/revng/pypeline/runner_context.py
```diff
@@ -0,0 +1,276 @@
...
+        if use_system:
+            assert base_directory is not None
+            assert base_directory.is_dir() or not base_directory.exists()
+
+            self.base_directory = base_directory
+            self.argv_hook = argv_hook
+
+        self.counter = 0
+
+    def _get_temp_dir(self, name: str) -> Path:
```
aleclearmind: 1. The fact that we never delete this directory makes me thing it's not a "temp" directory.
2. No abbreviations.

python/revng/pypeline/runner_context.py
```diff
@@ -0,0 +1,276 @@
...
+from revng.pypeline.utils.logger import pypeline_logger
+from revng.pypeline.utils.registry import get_singleton
+
+# Function defined by the pipebox that transforms the command-line before
+# `run-pipe` and `run-analysis` are invoked. This receives the `argv` of the
+# soon-to-be executed command and returns the modified argv.
+ArgvHook = Callable[[list[str]], list[str]]
+
+
+class RunnerContext:
```
aleclearmind, line 40: Comment on the role of this class and in particular on the effect of `use_system`.

python/revng/pypeline/runner_context.py
```diff
@@ -0,0 +1,276 @@
...
+
+        # Apply argv_hook if present
+        if self.argv_hook is not None:
+            cmd = self.argv_hook(cmd)
+
+        # Store the reproducer script
+        script_contents = f"""#!/usr/bin/env bash
+set -euo pipefail
+cd "$(realpath "$(dirname "${{BASH_SOURCE[0]}}")")"
+{shlex.join(cmd)}
```
aleclearmind: 1. `exec` for easier `gdb`ability?
2. Did you test this?
3. Can you provide an example usage? I'd like to give it a shot before merging.

python/revng/pypeline/runner_context.py
```diff
@@ -0,0 +1,276 @@
...
+"""
+        script_path = root_directory / "go.sh"
```
aleclearmind: `s/go.sh/run/`

python/revng/pypeline/runner_context.py
mrjackv: Currently yes, eventually it will throw the right exception (as explained in the TODO)

python/revng/pypeline/runner_context.py
mrjackv: 1. Leftover from the old design, I'll rename it to `_get_directory`
2. see (1)

Edit: renamed to `_get_directory`

python/revng/pypeline/runner_context.py
mrjackv: Done

python/revng/pypeline/runner_context.py
mrjackv: 1. Done
2. Yes
3. I've attached a log below that shows what happens when a pipe invocation fails

[debug.log](https://github.com/user-attachments/files/24905930/debug.log)

python/revng/pypeline/runner_context.py
mrjackv: Done

PR #555, opened 2026-01-21
Improve pypeline memory usage

mrjackv:
Commit `TraceProgressListener: also record memory usage` is shared with #554 
Sister PRs:
* [llvm-project](https://github.com/revng/llvm-project/pull/44)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -219,36 +219,69 @@ foreach(
     list(APPEND HELPER_MODULE_LIST "${CMAKE_BINARY_DIR}/share/revng/${OUTPUT}")
   endforeach()
 
-  # Annotate libtcg-helpers-*.bc
-
+  # Annotate `libtcg-helpers-*.bc`. This loads the `libtcg-helpers` bitcode
+  # files and runs analyses that determine which CSVs are accessed by each
+  # helper function.
   set(LIBTCG_HELPERS
       "${CMAKE_INSTALL_PREFIX}/share/libtcg/libtcg-helpers-${ARCH}.bc")
-  set(LIBTCG_HELPERS_ANNOTATED
-      "${CMAKE_BINARY_DIR}/share/revng/libtcg-helpers-annotated-${ARCH}.bc")
+  set(LIBTCG_HELPERS_FULL
+      "${CMAKE_BINARY_DIR}/share/revng/libtcg-helpers-full-${ARCH}.bc")
   add_custom_command(
-    OUTPUT "${LIBTCG_HELPERS_ANNOTATED}"
+    OUTPUT "${LIBTCG_HELPERS_FULL}"
     DEPENDS "${LIBTCG_HELPERS}" revng-all-binaries
     COMMAND
       ./bin/revng opt -sroa -instsimplify -cpu-loop-exit
       -analyze-helper-arguments -fix-helpers
       -fix-helpers-architecture="${ARCH}" "${LIBTCG_HELPERS}" -o
-      "${LIBTCG_HELPERS_ANNOTATED}")
+      "${LIBTCG_HELPERS_FULL}")
   add_custom_target("annotated-libtcg-helpers-${ARCH}" ALL
-                    DEPENDS "${LIBTCG_HELPERS_ANNOTATED}")
+                    DEPENDS "${LIBTCG_HELPERS_FULL}")
 
-  # Drop all the functions *not* tagged with revng_inline
-  set(LIBTCG_HELPERS_ANNOTATED_SLIM
-      "${CMAKE_BINARY_DIR}/share/revng/libtcg-helpers-annotated-slim-${ARCH}.bc"
-  )
-  add_custom_command(
-    OUTPUT "${LIBTCG_HELPERS_ANNOTATED_SLIM}"
-    DEPENDS "${LIBTCG_HELPERS_ANNOTATED}" revng-all-binaries
-    COMMAND ./bin/revng opt -slim-down-helpers-module -globaldce
-            "${LIBTCG_HELPERS_ANNOTATED}" -o "${LIBTCG_HELPERS_ANNOTATED_SLIM}")
-  add_custom_target("annotated-libtcg-helpers-slim-${ARCH}" ALL
-                    DEPENDS "${LIBTCG_HELPERS_ANNOTATED_SLIM}")
-  list(APPEND HELPER_MODULE_LIST "${LIBTCG_HELPERS_ANNOTATED_SLIM}")
+  function(process_helper_module)
+    set(options)
+    set(oneValueArgs NAME)
+    set(multiValueArgs PASSES)
+    cmake_parse_arguments(OPT "${options}" "${oneValueArgs}"
+                          "${multiValueArgs}" ${ARGN})
 
+    set(PROCESS_HELPER_MODULE_OUTPUT
+        "${CMAKE_BINARY_DIR}/share/revng/libtcg-helpers-${OPT_NAME}-${ARCH}.bc")
+    list(TRANSFORM OPT_PASSES PREPEND -)
+    add_custom_command(
+      OUTPUT "${PROCESS_HELPER_MODULE_OUTPUT}"
+      DEPENDS "${LIBTCG_HELPERS_FULL}" revng-all-binaries
+      COMMAND ./bin/revng opt ${OPT_PASSES} "${LIBTCG_HELPERS_FULL}" -o
+              "${PROCESS_HELPER_MODULE_OUTPUT}")
+    add_custom_target("annotated-libtcg-helpers-${OPT_NAME}-${ARCH}" ALL
+                      DEPENDS "${PROCESS_HELPER_MODULE_OUTPUT}")
+    list(APPEND HELPER_MODULE_LIST "${PROCESS_HELPER_MODULE_OUTPUT}")
+    set(HELPER_MODULE_LIST
+        "${HELPER_MODULE_LIST}"
+        PARENT_SCOPE)
+  endfunction()
+
+  # Produce to-inline version of helpers. In this version the bitcode file only
+  # contains helper functions that are in the `revng_inline` section and the
+  # CSVs that are used by them. This is used by `inline-helpers` to allow
+  # inlining the body of the `revng_inline` function into the isolated
+  # functions' bodies.
+  process_helper_module(
+    NAME
+    to-inline
+    PASSES
+    slim-down-helpers-module
+    sroa
+    instsimplify
+    simplifycfg
+    strip-dead-debug-info)
+
+  # Produce declarations-only version of helpers. In this version of the bitcode
+  # file all the helper functions have been turned into declarations and all
+  # CSVs have been dropped. This will be used in `lift` to import the helper
+  # declarations to be used in the lifting process.
+  process_helper_module(
+    NAME declarations-only PASSES helpers-module-to-declarations globaldce
+    strip-dead-debug-info)
 endforeach()
 
 # Produce well-known-models
--- a/include/revng/FunctionIsolation/InlineHelpers.h
+++ b/include/revng/FunctionIsolation/InlineHelpers.h
@@ -12,8 +12,16 @@ class InlineHelpersPass : public llvm::ModulePass {
 public:
   static char ID;
 
+  // This map stores the loaded helper module(s) to be inlined. Generally state
+  // in a pass is frowned upon, but since this is only ever read it's simpler to
+  // store it at the pass level.
+  llvm::StringMap<std::unique_ptr<llvm::Module>> HelpersModules;
+
 public:
   InlineHelpersPass() : ModulePass(ID) {}
 
   bool runOnModule(llvm::Module &M) override;
+
+private:
+  void linkRequiredHelpers(llvm::Module &M);
 };
--- a/include/revng/FunctionIsolation/IsolateFunctions.h
+++ b/include/revng/FunctionIsolation/IsolateFunctions.h
@@ -29,6 +29,7 @@ namespace revng::pypeline::piperuns {
 
 class Isolate {
 private:
+  const model::Binary &Binary;
   std::unique_ptr<llvm::Module> ClonedModule;
   LLVMFunctionContainer &Output;
   std::optional<GeneratedCodeBasicInfo> GCBI;
--- a/include/revng/FunctionIsolation/IsolateFunctions.h
+++ b/include/revng/FunctionIsolation/IsolateFunctions.h
@@ -31,7 +31,7 @@ class Isolate {
 private:
   std::unique_ptr<llvm::Module> ClonedModule;
   LLVMFunctionContainer &Output;
-  GeneratedCodeBasicInfo GCBI;
+  std::optional<GeneratedCodeBasicInfo> GCBI;
   // unique_ptr to the implementation. This is a temporary measure until the
   // old pipeline is dropped and the body of the `Impl` class can be inlined in
   // this one.
--- a/include/revng/Support/IRHelpers.h
+++ b/include/revng/Support/IRHelpers.h
@@ -25,6 +25,7 @@
 #include "llvm/IR/ValueMap.h"
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/raw_ostream.h"
+#include "llvm/Transforms/Utils/Cloning.h"
 
 #include "revng/ADT/Concepts.h"
 #include "revng/Support/BasicBlockID.h"
@@ -1079,10 +1080,13 @@ llvm::PointerType *getStringPtrType(llvm::LLVMContext &C);
 llvm::GlobalVariable *
 buildString(llvm::Module *M, llvm::StringRef String, const llvm::Twine &Name);
 
-llvm::Constant *getUniqueString(llvm::Module *M,
-                                llvm::StringRef String,
-                                bool AddNull = true,
-                                llvm::StringRef Namespace = "revng.const.");
+constexpr llvm::StringRef DefaultStringNamespace = "revng.const.";
+
+llvm::Constant *
+getUniqueString(llvm::Module *M,
+                llvm::StringRef String,
+                bool AddNull = true,
+                llvm::StringRef Namespace = DefaultStringNamespace);
 
 llvm::StringRef extractFromConstantStringPtr(llvm::Value *V);
 
@@ -1554,9 +1558,16 @@ inline void linkModules(std::unique_ptr<llvm::Module> &&Source,
   linkModules(std::move(Source), Destination, std::nullopt);
 }
 
-/// Clone the module, keeping only the functions specified in \p ToClone .
+/// Clone the module with the `Action` function dictating what should be done
+/// for each GlobalValue in the module.
 /// This requires the source module to be writable in order to guarantee that
 /// the metadata attached to the cloned functions is also preserved.
+std::unique_ptr<llvm::Module>
+cloneFiltered(llvm::Module &Module,
+              llvm::function_ref<llvm::CloneAction(const llvm::GlobalValue *)>
+                Action);
+
+// Shortcut for the above with a set of functions to clone
 std::unique_ptr<llvm::Module>
 cloneFiltered(llvm::Module &Module, std::set<const llvm::Function *> &ToClone);
 
@@ -1588,21 +1599,3 @@ llvm::GlobalVariable &getOrCreateGlobal(llvm::Module &M,
 /// modules being merged.
 void linkFunctionModules(std::unique_ptr<llvm::Module> &&Source,
                          std::unique_ptr<llvm::Module> &Destination);
-
-/// Helper class that allows getting the exhaustive list of functions called by
-/// a function
-class ReachableFunctionsEnumerator {
-private:
-  using DenseFunctionSet = llvm::DenseSet<const llvm::Function *>;
-  using FunctionSet = std::set<const llvm::Function *>;
-
-private:
-  const FunctionSet &ToIgnore;
-  std::map<const llvm::Function *, DenseFunctionSet> CalledFunctions;
-
-public:
-  ReachableFunctionsEnumerator(const FunctionSet &ToIgnore) :
-    ToIgnore(ToIgnore) {}
-
-  const DenseFunctionSet &getCalledFunctions(const llvm::Function &Function);
-};
--- a/include/revng/Support/IRHelpers.h
+++ b/include/revng/Support/IRHelpers.h
@@ -83,6 +83,9 @@ constexpr const char *FunctionEntryMDName = "revng.function.entry";
 constexpr const char *JTReasonMDName = "revng.jt.reasons";
 constexpr const char *ControlFlowGraphMDName = "revng.function.metadata";
 constexpr const char *ExplicitParenthesesMDName = "revng.explicit_parentheses";
+constexpr llvm::StringRef InlineHelpersSection = "revng_inline";
+constexpr llvm::StringRef HelpersListName = "helpers_list";
+constexpr llvm::StringRef QemuArchitectureMD = "revng.qemu_architecture";
 
 template<typename T>
 inline bool contains(T Range, typename T::value_type V) {
--- a/include/revng/Support/SimplePassManager.h
+++ b/include/revng/Support/SimplePassManager.h
@@ -1,0 +1,56 @@
+#pragma once
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "llvm/Passes/PassBuilder.h"
+
+namespace detail {
+
+class SimplePassManagerBase {
+protected:
+  llvm::LoopAnalysisManager LAM;
+  llvm::FunctionAnalysisManager FAM;
+  llvm::CGSCCAnalysisManager CGAM;
+  llvm::ModuleAnalysisManager MAM;
+  llvm::PassBuilder PB;
+
+  SimplePassManagerBase() {
+    PB.registerModuleAnalyses(MAM);
+    PB.registerCGSCCAnalyses(CGAM);
+    PB.registerFunctionAnalyses(FAM);
+    PB.registerLoopAnalyses(LAM);
+    PB.crossRegisterProxies(LAM, FAM, CGAM, MAM);
+  }
+};
+
+} // namespace detail
+
+/// Wrapper for the new llvm pass manager. Does all the required initialization
+/// and exposes all the methods to have an easy to use pass manager.
+class SimplePassManager : public detail::SimplePassManagerBase {
+private:
+  llvm::ModulePassManager PM;
+
+public:
+  template<typename T>
+  void addPass(T &&Pass) {
+    PM.addPass(std::forward<T>(Pass));
+  }
+
+  void run(llvm::Module &Module) { PM.run(Module, MAM); }
+};
+
+class SimpleFunctionPassManager : public detail::SimplePassManagerBase {
+private:
+  llvm::FunctionPassManager PM;
+
+public:
+  template<typename T>
+  void addPass(T &&Pass) {
+    PM.addPass(std::forward<T>(Pass));
+  }
+
+  void run(llvm::Function &Function) { PM.run(Function, FAM); }
+};
--- a/lib/BasicAnalyses/GeneratedCodeBasicInfo.cpp
+++ b/lib/BasicAnalyses/GeneratedCodeBasicInfo.cpp
@@ -27,14 +27,17 @@ char GeneratedCodeBasicInfoWrapperPass::ID = 0;
 using RegisterGCBI = RegisterPass<GeneratedCodeBasicInfoWrapperPass>;
 static RegisterGCBI X("gcbi", "Generated Code Basic Info", true, true);
 
-void GeneratedCodeBasicInfo::run(Module &M) {
+GeneratedCodeBasicInfo::GeneratedCodeBasicInfo(const model::Binary &Binary,
+                                               llvm::Module &M) :
+  Binary(Binary) {
+
   RootFunction = M.getFunction("root");
   NewPC = getIRHelper("newpc", M);
 
   revng_log(PassesLog, "Starting GeneratedCodeBasicInfo");
 
   using namespace model::Architecture;
-  auto Architecture = Binary->Architecture();
+  auto Architecture = Binary.Architecture();
   PC = M.getGlobalVariable(getPCCSVName(Architecture), true);
   SP = M.getGlobalVariable(getCSVName(getStackPointer(Architecture)), true);
   auto ReturnAddressRegister = getReturnAddressRegister(Architecture);
@@ -158,23 +161,20 @@ GeneratedCodeBasicInfo::blocksByPCRange(MetaAddress Start, MetaAddress End) {
 GeneratedCodeBasicInfo
 GeneratedCodeBasicInfoAnalysis::run(Module &M, ModuleAnalysisManager &MAM) {
   auto &LMA = MAM.getResult<LoadModelAnalysis>(M);
-  GeneratedCodeBasicInfo GCBI(*LMA.getReadOnlyModel());
-  GCBI.run(M);
+  GeneratedCodeBasicInfo GCBI(*LMA.getReadOnlyModel(), M);
   return GCBI;
 }
 
 GeneratedCodeBasicInfo
 GeneratedCodeBasicInfoAnalysis::run(Function &F, FunctionAnalysisManager &FAM) {
   auto &LMA = FAM.getResult<LoadModelAnalysis>(F);
-  GeneratedCodeBasicInfo GCBI(*LMA.getReadOnlyModel());
-  GCBI.run(*F.getParent());
+  GeneratedCodeBasicInfo GCBI(*LMA.getReadOnlyModel(), *F.getParent());
   return GCBI;
 }
 
 bool GeneratedCodeBasicInfoWrapperPass::runOnModule(Module &M) {
   auto &LMA = getAnalysis<LoadModelWrapperPass>().get();
-  GCBI.reset(new GeneratedCodeBasicInfo(*LMA.getReadOnlyModel()));
-  GCBI->run(M);
+  GCBI.reset(new GeneratedCodeBasicInfo(*LMA.getReadOnlyModel(), M));
   return false;
 }
 
--- a/lib/BasicAnalyses/GeneratedCodeBasicInfo.cpp
+++ b/lib/BasicAnalyses/GeneratedCodeBasicInfo.cpp
@@ -29,7 +29,7 @@ static RegisterGCBI X("gcbi", "Generated Code Basic Info", true, true);
 
 GeneratedCodeBasicInfo::GeneratedCodeBasicInfo(const model::Binary &Binary,
                                                llvm::Module &M) :
-  Binary(Binary) {
+  Binary(Binary), Module(M) {
 
   RootFunction = M.getFunction("root");
   NewPC = getIRHelper("newpc", M);
--- a/lib/FunctionIsolation/InlineHelpers.cpp
+++ b/lib/FunctionIsolation/InlineHelpers.cpp
@@ -6,6 +6,7 @@
 
 #include "llvm/ADT/SCCIterator.h"
 #include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/StringSet.h"
 #include "llvm/Analysis/CallGraph.h"
 #include "llvm/Transforms/Utils/Cloning.h"
 
@@ -14,6 +15,7 @@
 #include "revng/FunctionIsolation/InlineHelpers.h"
 #include "revng/Support/IRHelpers.h"
 #include "revng/Support/OpaqueFunctionsPool.h"
+#include "revng/Support/ResourceFinder.h"
 
 using namespace llvm;
 
@@ -54,7 +56,7 @@ bool InlineHelpers::shouldInline(Function *F) const {
   if (Recursive.count(F) != 0)
     return false;
 
-  return F->getSection() == "revng_inline";
+  return F->getSection() == InlineHelpersSection;
 }
 
 CallInst *InlineHelpers::getCallToInline(Instruction *I) const {
@@ -86,6 +88,7 @@ bool InlineHelpers::doInline(Function *F) const {
 
   return ToInline.size() > 0;
 }
+
 static void dropDebugOrPseudoInst(Function *F) {
   SmallVector<Instruction *, 16> ToErase;
   for (BasicBlock &BB : *F) {
@@ -102,14 +105,62 @@ static void dropDebugOrPseudoInst(Function *F) {
 }
 
 void InlineHelpers::run(Function *F) {
-  // Fixed-point inlining
-  while (doInline(F))
-    ;
-
+  // Since all the helpers have been already inlined the `doInline` function can
+  // be called only once to inline all the helper functions.
+  doInline(F);
   dropDebugOrPseudoInst(F);
 }
 
+void InlineHelpersPass::linkRequiredHelpers(llvm::Module &M) {
+  NamedMDNode *Node = M.getNamedMetadata(QemuArchitectureMD);
+  revng_assert(Node != nullptr);
+  revng_assert(Node->getNumOperands() > 0);
+
+  // Multiple linkings could have lead to this node having more than one
+  // operand, if so check that they are all the same.
+  llvm::StringSet Strings;
+  for (size_t I = 0; I < Node->getNumOperands(); I++) {
+    revng_assert(Node->getOperand(0)->getNumOperands() == 1);
+    const MDOperand &StringNode = Node->getOperand(0)->getOperand(0);
+    Strings.insert(cast<MDString>(StringNode)->getString());
+  }
+
+  revng_assert(Strings.size() == 1,
+               "QEMU helpers of multiple architectures were linked inside the "
+               "module");
+
+  // Given the architecture MD, retrieve the correct libtcg module (if not
+  // already in the `HelpersModules` map) and link it into the main module
+  StringRef ArchName = Strings.begin()->first();
+  if (HelpersModules.count(ArchName) == 0) {
+    const std::string LibHelpersName = ("/share/revng/"
+                                        "libtcg-helpers-to-inline-"
+                                        + ArchName + ".bc")
+                                         .str();
+    auto OptionalHelpers = revng::ResourceFinder.findFile(LibHelpersName);
+    revng_assert(OptionalHelpers.has_value(), "Cannot find libtcg helpers");
+    HelpersModules[ArchName] = parseIR(M.getContext(), OptionalHelpers.value());
+  }
+
+  llvm::Module &HelpersModule = *HelpersModules[ArchName];
+
+  std::set<const llvm::Function *> ToClone;
+  for (llvm::Function &F : M) {
+    if (not F.isDeclaration() or F.getSection() != InlineHelpersSection)
+      continue;
+
+    llvm::Function *HelperFunction = HelpersModule.getFunction(F.getName());
+    revng_assert(HelperFunction != nullptr);
+    ToClone.insert(HelperFunction);
+  }
+
+  auto ClonedHelpersModule = ::cloneFiltered(HelpersModule, ToClone);
+  linkModules(std::move(ClonedHelpersModule), M);
+}
+
 bool InlineHelpersPass::runOnModule(llvm::Module &M) {
+  linkRequiredHelpers(M);
+
   SmallVector<Function *, 32> Isolated;
   for (Function &F : M)
     if (FunctionTags::Isolated.isTagOf(&F))
@@ -120,6 +171,13 @@ bool InlineHelpersPass::runOnModule(llvm::Module &M) {
     T.advance(F->getName());
     InlineHelpers IH(*F->getParent());
     IH.run(F);
+  }
+
+  // Since we're done inlining helpers they are no longer needed, delete their
+  // body
+  for (Function &F : M) {
+    if (F.getSection() == InlineHelpersSection)
+      deleteOnlyBody(F);
   }
 
   return true;
--- a/lib/FunctionIsolation/IsolateFunctions.cpp
+++ b/lib/FunctionIsolation/IsolateFunctions.cpp
@@ -19,6 +19,7 @@
 #include "llvm/IR/Verifier.h"
 #include "llvm/Support/raw_os_ostream.h"
 #include "llvm/Transforms/IPO.h"
+#include "llvm/Transforms/IPO/GlobalDCE.h"
 #include "llvm/Transforms/Utils/BasicBlockUtils.h"
 #include "llvm/Transforms/Utils/Cloning.h"
 #include "llvm/Transforms/Utils/CodeExtractor.h"
@@ -52,6 +53,7 @@
 #include "revng/Support/Debug.h"
 #include "revng/Support/IRHelpers.h"
 #include "revng/Support/MetaAddress.h"
+#include "revng/Support/SimplePassManager.h"
 
 // This name is not present after `enforce-abi`.
 RegisterIRHelper FDispatcher("function_dispatcher");
@@ -832,6 +834,12 @@ void Isolate::runOnFunction(const model::Function &TheFunction) {
 }
 
 void Isolate::splitIsolatedFunctionsToOutput() {
+  // Run globaldce on the isolated root module to remove any cruft present. This
+  // should speed up cloning.
+  SimplePassManager PM;
+  PM.addPass(llvm::GlobalDCEPass());
+  PM.run(*ClonedModule);
+
   llvm::Task T(IsolatedFunctions.size(),
                "Splitting functions into individual modules");
 
@@ -841,6 +849,25 @@ void Isolate::splitIsolatedFunctionsToOutput() {
     T.advance(Address.toString(), true);
 
     auto IsolatedModule = Cloner.cloneModule(*Function);
+    // Check that running globaldce on the isolated module does nothing
+    if (VerifyLog.isEnabled()) {
+      llvm::StringSet GlobalNames;
+      for (GlobalValue &GV : IsolatedModule->global_values()) {
+        if (GV.hasName())
+          GlobalNames.insert(GV.getName());
+      }
+
+      SimplePassManager PM;
+      PM.addPass(llvm::GlobalDCEPass());
+      PM.run(*IsolatedModule);
+
+      for (GlobalValue &GV : IsolatedModule->global_values()) {
+        if (GV.hasName())
+          GlobalNames.erase(GV.getName());
+      }
+      revng_assert(GlobalNames.size() == 0,
+                   "GlobalDCE removed names from an isolated module");
+    }
     Output.assign(ObjectID(Address), std::move(IsolatedModule));
 
     // Since we saved the function to the output, delete its body in
--- a/lib/FunctionIsolation/IsolateFunctions.cpp
+++ b/lib/FunctionIsolation/IsolateFunctions.cpp
@@ -676,6 +676,125 @@ bool IF::runOnModule(Module &TheModule) {
   return false;
 }
 
+/// Helper class that wraps `llvm::cloneModule` and specializes in the case
+/// where a single function needs to be cloned. It does aggressive changes to
+/// the Module, such as detaching globals and reattaching them only when needed.
+/// While an instance of this class is active the module will *not* verify and
+/// should not be passed to other pieces of LLVM infrastructure (e.g.
+/// PassManager).
+class MinimalModuleCloner {
+private:
+  struct FunctionUses {
+    std::set<const llvm::Function *> CalledFunctions;
+    std::set<const llvm::GlobalVariable *> UsedGVs;
+  };
+
+private:
+  llvm::Module &Module;
+
+  std::set<const llvm::GlobalVariable *> IgnorableGVs;
+
+public:
+  MinimalModuleCloner(const model::Binary &Binary, llvm::Module &Module) :
+    Module(Module) {
+    revng::verify(&Module);
+
+    // Always copy SP because it's needed by another pipe down the pipeline
+    // (InjectStackSizeProbesAtCallSitesPass). The proper fix would be to have
+    // that pipe detect the absence of `SP` and assert that the stack pointer
+    // could not have moved.
+    std::string SPName = getCSVName(getStackPointer(Binary.Architecture()));
+
+    // Construct the set of ignorable globals from string globals and CSVs
+    // (except for SP)
+    for (llvm::GlobalVariable &GV : Module.globals()) {
+      StringRef Name = GV.getName();
+      if (Name.starts_with(DefaultStringNamespace)
+          or (FunctionTags::CSV.isTagOf(&GV) and Name != SPName))
+        IgnorableGVs.insert(&GV);
+    }
+  }
+
+  std::unique_ptr<llvm::Module> cloneModule(llvm::Function &Function) {
+    FunctionUses AnalysisResult = analyzeFunction(Function);
+
+    auto ActionFunction =
+      [this, &Function, &AnalysisResult](const GlobalValue *V) {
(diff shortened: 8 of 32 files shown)


lib/BasicAnalyses/GeneratedCodeBasicInfo.cpp
```diff
@@ -28,6 +28,7 @@ using RegisterGCBI = RegisterPass<GeneratedCodeBasicInfoWrapperPass>;
 static RegisterGCBI X("gcbi", "Generated Code Basic Info", true, true);
 
 void GeneratedCodeBasicInfo::run(Module &M) {
+  TheModule = &M;
```
aleclearmind: After a quick check I verified all users of `GCBI` invoke `run` right away.

I think it makes sense to move `run` into the constructor, there's no reason to have two step initialization.

include/revng/Support/IRHelpers.h
```diff
@@ -1586,20 +1589,66 @@ llvm::GlobalVariable &getOrCreateGlobal(llvm::Module &M,
...
+/// global variables called/used by a function
+class FunctionReachabilityEnumerator {
 private:
-  using DenseFunctionSet = llvm::DenseSet<const llvm::Function *>;
-  using FunctionSet = std::set<const llvm::Function *>;
+  // The implementation uses recursive coroutines, which should not be exposed
+  // in public headers. To avoid this the actual implementation is hidden in
+  // this `Impl` class that's defined in `IRHelpers.cpp`.
+  class Impl;
+  std::unique_ptr<Impl> ImplInstance;
```
aleclearmind: Never use PImpl. The fact that it was allowed once with a `TODO: drop` does not allow you to use it elsewhere.

include/revng/Support/IRHelpers.h
```diff
@@ -1586,20 +1589,66 @@ llvm::GlobalVariable &getOrCreateGlobal(llvm::Module &M,
...
+    /// Set of `GlobalVariable`s used in all the CalledFunctions and
+    /// the entry function
+    const llvm::DenseSet<llvm::GlobalVariable *> &UsedGlobals;
+  };
+
+public:
+  /// Constructor, the `ToIgnore` set marks which functions should not be looked
+  /// into when running `analyze`
+  FunctionReachabilityEnumerator(const std::set<const llvm::Function *>
+                                   &ToIgnore);
```
aleclearmind: Why do we go for a blacklist approach as opposed to a whitelist?

include/revng/Support/IRHelpers.h
```diff
@@ -1586,20 +1589,66 @@ llvm::GlobalVariable &getOrCreateGlobal(llvm::Module &M,
...
+  Result analyze(llvm::Function &Function);
+};
+
+/// Helper class that wraps `llvm::cloneModule` and specializes in the case
+/// where a single function needs to be cloned. It does aggressive changes to
+/// the Module, such as detaching globals and reattaching them only when needed.
+/// While an instance of this class is active the module will *not* verify and
+/// should not be passed to other pieces of LLVM infrastructure (e.g.
+/// PassManager).
+class ConservativeModuleCloner {
```
aleclearmind: I'm not sure what's "conservative" about this.
Minimal?

include/revng/Support/IRHelpers.h
```diff
@@ -1586,20 +1589,66 @@ llvm::GlobalVariable &getOrCreateGlobal(llvm::Module &M,
...
 private:
-  const FunctionSet &ToIgnore;
-  std::map<const llvm::Function *, DenseFunctionSet> CalledFunctions;
+  llvm::Module &Module;
+  std::set<llvm::GlobalVariable *> DetachedGVs;
+  std::vector<llvm::GlobalVariable *> OriginalGVsOrder;
```
aleclearmind: Do these two contain the same globals?
Consider using `SetVector`.

include/revng/Support/IRHelpers.h
```diff
@@ -1586,20 +1589,66 @@ llvm::GlobalVariable &getOrCreateGlobal(llvm::Module &M,
...
+
+  std::set<const llvm::Function *> FREToIgnore;
```
aleclearmind: `FRE`? No abbreviations.
Just drop the prefix.

include/revng/Support/IRHelpers.h
```diff
@@ -1586,20 +1589,66 @@ llvm::GlobalVariable &getOrCreateGlobal(llvm::Module &M,
...
+  std::optional<FunctionReachabilityEnumerator> FRE;
```
aleclearmind: Mmmh, do not use `optional` like this unless you *really* need to.

lib/FunctionIsolation/IsolateFunctions.cpp
```diff
@@ -743,14 +743,40 @@ void Isolate::splitIsolatedFunctionsToOutput() {
...
-  ReachableFunctionsEnumerator Enumerator(InternalFunctions);
+
+  // Construct a set of special CSV that need to be preserved. These are SP, PC
+  // and RA which are needed later in the pipeline.
+  const model::Architecture::Values &Architecture = Binary.Architecture();
+  llvm::StringSet SpecialCSVs{ getPCCSVName(Architecture),
+                               getCSVName(getStackPointer(Architecture)) };
+  auto ReturnAddressRegister = getReturnAddressRegister(Architecture);
+  if (ReturnAddressRegister != model::Register::Invalid)
+    SpecialCSVs.insert(getCSVName(ReturnAddressRegister));
```
aleclearmind: I'm not sure we need to do any special casing.
Have you tried not doing any special casing?

I'd expect, at most, to special case SP, but maybe not even that.

lib/FunctionIsolation/IsolateFunctions.cpp
```diff
@@ -743,14 +743,40 @@ void Isolate::splitIsolatedFunctionsToOutput() {
...
+    }
+
+    const auto *Function = llvm::dyn_cast<llvm::Function>(GV);
+    if (Function != nullptr)
+      return FunctionTags::Isolated.isTagOf(Function);
+
+    return false;
+  };
+
+  ConservativeModuleCloner Cloner(*ClonedModule, ToDetach);
```
aleclearmind: `ConservativeModuleCloner` is doing some pretty nasty stuff to `ClonedModule`. Also, `ClonedModule` is approaching death anyway.
I feel like `ConservativeModuleCloner` should take ownership of `ClonedModule` (i.e., `std::move`).

I propose we:

1. Drop `root`
2. Destroy `IsolateFunctionsImpl` (which holds references to `ClonedModule`, correct?).
3. `std::move` `ClonedModule` into `ConservativeModuleCloner`

Is this viable?

lib/Support/IRHelpers.cpp
```diff
@@ -798,7 +801,9 @@ cloneFiltered(llvm::Module &Module, std::set<const llvm::Function *> &ToClone) {
 
   FunctionsMetadata::backup(Module);
 
-  revng::verify(&Module);
+  if (Verify)
+    revng::verify(&Module);
```
aleclearmind: Uh? No.

lib/Support/IRHelpers.cpp
```diff
@@ -1058,52 +1068,96 @@ void linkFunctionModules(std::unique_ptr<llvm::Module> &&Source,
...
+  using DenseGVSet = DenseSet<GlobalVariable *>;
+  friend FunctionReachabilityEnumerator;
 
-  if (Function.isDeclaration()) {
-    CalledFunctions[&Function] = {};
-    co_return &CalledFunctions[&Function];
-  }
+private:
+  const FunctionSet &ToIgnore;
+  std::map<llvm::Function *, DenseFunctionSet> CalledFunctions;
```
aleclearmind: `s/Called/Reachable/`

lib/Support/IRHelpers.cpp
```diff
@@ -1139,3 +1193,74 @@ llvm::GlobalVariable &getOrCreateGlobal(llvm::Module &M,
...
+  for (llvm::GlobalVariable *GV : DetachedGVs)
+    GV->removeFromParent();
+}
+
+ConservativeModuleCloner::~ConservativeModuleCloner() {
+  reattachGVs(DetachedGVs);
+}
+
+std::unique_ptr<llvm::Module>
+ConservativeModuleCloner::cloneModule(llvm::Function &Function) {
```
aleclearmind: Comment this code

lib/Support/IRHelpers.cpp
```diff
@@ -1139,3 +1193,74 @@ llvm::GlobalVariable &getOrCreateGlobal(llvm::Module &M,
...
+    GlobalVariable *Pointer = OriginalGVsOrder[Index];
+    if (DetachedGVs.contains(Pointer)) {
+      if (ToReattach.contains(Pointer)) {
+        auto NewIterator = List.insert(Iterator, Pointer);
+        Iterator = ++NewIterator;
+      }
+    } else {
+      Iterator++;
+    }
+  }
```
aleclearmind: This code is very confusing.
Needs further review after other questions have been answered.

include/revng/FunctionIsolation/IsolateFunctions.h
aleclearmind, line 1: I'm struggling to get a global pictures of this commit and I smell something broken. This kind of things have been source of subtle bugs in the past and I don't want to repeat that situation.

A summary of what's going on would help.

`ReachableFunctionsEnumerator` pretends to be something generic and reusable but it does a very small subset of what it should be doing to make sure we don't have dangling references. This might be fine under some assumptions, e.g., we know we don't do certain things in isolated functions, but it's not OK for helpers which might be doing anything.

Let's discuss this in person.

lib/Support/IRHelpers.cpp
```diff
@@ -1139,3 +1193,74 @@ llvm::GlobalVariable &getOrCreateGlobal(llvm::Module &M,
...
+void ConservativeModuleCloner::reattachGVs(const std::set<GlobalVariable *>
+                                             &ToReattach) {
+  revng_assert(llvm::set_is_subset(ToReattach, DetachedGVs));
+
+  // This loop re-inserts the elements of ToReattach to the `GlobalVariable`s
+  // list, in the same order they were present before they got detached.
+  Module::GlobalListType &List = Module.getGlobalList();
+  auto Iterator = List.begin();
+  for (size_t Index = 0; Index < OriginalGVsOrder.size(); Index++) {
+    GlobalVariable *Pointer = OriginalGVsOrder[Index];
```
aleclearmind: You can use a range-based `for` loop here.

include/revng/Support/IRHelpers.h
```diff
@@ -1652,3 +1655,25 @@ class ConservativeModuleCloner {
 private:
   void reattachGVs(const std::set<llvm::GlobalVariable *> &ToReattach);
 };
+
+/// Wrapper for the new llvm pass manager. Does all the required initialization
+/// and exposes all the methods to have an easy to use pass manager.
+class PassManagerWrapper {
```
aleclearmind: 1. Stop adding stuff to `IRHelpers.h`. Make a separate header.
2. Use a dedicated commit.
3. See if there are trivial places where this can be adopted.

lib/FunctionIsolation/IsolateFunctions.cpp
```diff
@@ -741,6 +743,11 @@ void Isolate::splitIsolatedFunctionsToOutput() {
       InternalFunctions.insert(&F);
   }
 
+  PassManagerWrapper PMW;
+  PMW.addPass(llvm::GlobalDCEPass());
+  PMW.addPass(llvm::StripDeadDebugInfoPass());
+  PMW.run(*ClonedModule);
```
aleclearmind:
```
  dumpModule("1.ll");
  PassManagerWrapper PMW;
  PMW.addPass(llvm::GlobalDCEPass());
  PMW.run(*ClonedModule);

  dumpModule("2.ll");

  PassManagerWrapper PMW;
  PMW.addPass(llvm::StripDeadDebugInfoPass());
  PMW.run(*ClonedModule);

  dumpModule("3.ll");
```

Please report on a significant example `diff -u 1.ll 2.ll` and `diff -u 2.ll 3.ll`.

Since we're here, I'd like to have a clear idea about what we're cloning, just to be discarded right after.

CMakeLists.txt
```diff
@@ -236,19 +237,39 @@ foreach(
...
+  add_custom_command(
+    OUTPUT "${LIBTCG_HELPERS_ANNOTATED_DECLARATIONS}"
+    DEPENDS "${LIBTCG_HELPERS_ANNOTATED}" revng-all-binaries
+    COMMAND
+      ./bin/revng opt -helpers-module-to-declarations -globaldce
+      -strip-dead-debug-info "${LIBTCG_HELPERS_ANNOTATED}" -o
+      "${LIBTCG_HELPERS_ANNOTATED_DECLARATIONS}")
+  add_custom_target("annotated-libtcg-helpers-declarations-${ARCH}" ALL
+                    DEPENDS "${LIBTCG_HELPERS_ANNOTATED_DECLARATIONS}")
+  list(APPEND HELPER_MODULE_LIST "${LIBTCG_HELPERS_ANNOTATED_DECLARATIONS}")
```
aleclearmind: 1. Make a summary about who's going to use what (e.g., who is going to use the slim version)?
2. Please see if we can make a macro for all of this CMake code.
   I want to take my eyes out of my orbits.
   ```
   process_helper_module("-transformation1", "-suffix1")
   process_helper_module("-transformation2", "-suffix2")
   process_helper_module("-transformation3", "-suffix3")
   ```

include/revng/Support/IRHelpers.h
```diff
@@ -1677,3 +1677,8 @@ class PassManagerWrapper {
 
   void run(llvm::Module &Module) { MPM.run(Module, MAM); }
 };
+
+constexpr llvm::StringRef InlineHelpersSection = "revng_inline";
+constexpr llvm::StringRef HelpersListName = "helpers_list";
+constexpr llvm::StringRef QemuArchitectureMD = "revng.qemu_architecture";
+constexpr llvm::StringRef ReachableFunctionsTag = "revng.reachable_functions";
```
aleclearmind: These should be at the top of the file.

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
```diff
@@ -25,26 +52,102 @@ class SlimDownHelpersModule : public llvm::ModulePass {
...
+        Data.push_back(ValueAsMetadata::get(CF));
+
+      F.addMetadata(ReachableFunctionsTag, *MDTuple::get(M.getContext(), Data));
+    }
+
+    return true;
+  }
+};
+
+class HelpersModuleToDeclarations : public llvm::ModulePass {
```
aleclearmind: Please make an effort and stop just adding stuff to existing files.
This clearly belongs to a separate file.

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
```diff
@@ -25,26 +52,102 @@ class SlimDownHelpersModule : public llvm::ModulePass {
...
+    for (Function &F : M) {
+      // Remove the body of all the functions except `helper_initialize_env`,
+      // which needs to survive.
+      if (F.getName() != "helper_initialize_env") {
+        F.eraseMetadata(DbgKindID);
+        SmallVector<std::pair<unsigned, MDNode *>, 8> MDs;
+        F.getAllMetadata(MDs);
+        F.deleteBody();
+        for (auto [ID, Node] : MDs)
+          F.setMetadata(ID, Node);
```
aleclearmind: This looks like `deleteOnlyBody`.

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
```diff
@@ -25,26 +52,102 @@ class SlimDownHelpersModule : public llvm::ModulePass {
...
+
+  void getAnalysisUsage(llvm::AnalysisUsage &AU) const override {}
+
+  bool runOnModule(llvm::Module &M) override {
+    auto DbgKindID = M.getContext().getMDKindID("dbg");
+    for (Function &F : M) {
+      // Remove the body of all the functions except `helper_initialize_env`,
+      // which needs to survive.
+      if (F.getName() != "helper_initialize_env") {
+        F.eraseMetadata(DbgKindID);
```
aleclearmind: Mention that we need to run strip dead debug info afterwards.

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
```diff
@@ -25,26 +52,102 @@ class SlimDownHelpersModule : public llvm::ModulePass {
...
         F.eraseMetadata(DbgKindID);
         SmallVector<std::pair<unsigned, MDNode *>, 8> MDs;
         F.getAllMetadata(MDs);
         F.deleteBody();
         for (auto [ID, Node] : MDs)
           F.setMetadata(ID, Node);
+      } else {
+        // For functions to be inlined, strip their debug info, as it will not
+        // be needed when the module will be used.
+        llvm::stripDebugInfo(F);
```
aleclearmind: I think you can just strip all the debug info in `opt`, can you?

lib/Lift/CodeGenerator.cpp
```diff
@@ -154,32 +154,10 @@ CodeGenerator::CodeGenerator(const RawBinaryView &RawBinary,
...
 
-  // Tag all global objects in HelpersModule as QEMU
-  for (GlobalVariable &G : HelpersModule->globals())
-    FunctionTags::QEMU.addTo(&G);
-
-  for (Function &F : HelpersModule->functions()) {
-    if (F.isIntrinsic())
-      continue;
-
-    F.setDSOLocal(false);
```
aleclearmind, line 173: Lost?

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
```diff
@@ -25,26 +52,102 @@ class SlimDownHelpersModule : public llvm::ModulePass {
...
+      if (GVLinkageSave.count(GV.getName()))
+        GV.setLinkage(GVLinkageSave[GV.getName()]);
+    }
+
+    // Tag all globals
+    tagGlobals(M);
+
+    // Compute the list of transitive calls to be later used in cloning
+    std::set<const llvm::Function *> ToIgnore;
+    FunctionReachabilityEnumerator FRE(ToIgnore);
```
aleclearmind: I think we're using `FunctionReachabilityEnumerator` too liberally.
We need to discuss this.

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
```diff
@@ -25,26 +52,102 @@ class SlimDownHelpersModule : public llvm::ModulePass {
...
     // Mark all global variables as internal, so we can purge those that are now
     // unused due to the removal of most of helpers' bodies.
-    // Also preserve the `arch_cpu_type_beacon` and all the variables with
-    // revng.tags metadata, since they are CSVs produced by VariableManager.
+    // Also preserve the `arch_cpu_type_beacon` and `cpu_loop_exiting`.
+    // CSVs produced by the VariableManager need to be handled carefully, we
+    // want only the ones with uses to survive, so we save their current linkage
+    // now, run `globaldce` within the pass and restore the previous linkage to
+    // the surviving ones.
+    StringMap<GlobalValue::LinkageTypes> GVLinkageSave;
```
aleclearmind: Do not use abbreviations. Just `GlobalLinkageBackup`.

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
```diff
@@ -25,26 +52,102 @@ class SlimDownHelpersModule : public llvm::ModulePass {
...
+      }
+    }
+
+    // Run globaldce
+    PassManagerWrapper PMW;
+    PMW.addPass(llvm::GlobalDCEPass());
+    PMW.run(M);
+
+    // Restore the linkage to CSVs after running globaldce
+    for (GlobalVariable &GV : M.getGlobalList()) {
```
aleclearmind: Avoid `getGlobalList` if you can. It's going away in future LLVM versions.

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
```diff
@@ -25,26 +52,102 @@ class SlimDownHelpersModule : public llvm::ModulePass {
...
+      if (GVLinkageSave.count(GV.getName()))
```
aleclearmind: Why iterate on all globals? Just do `M.getGlobalVariable()`

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
```diff
@@ -25,26 +52,102 @@ class SlimDownHelpersModule : public llvm::ModulePass {
...
 
     // Mark all global variables as internal, so we can purge those that are now
     // unused due to the removal of most of helpers' bodies.
-    // Also preserve the `arch_cpu_type_beacon` and all the variables with
-    // revng.tags metadata, since they are CSVs produced by VariableManager.
+    // Also preserve the `arch_cpu_type_beacon` and `cpu_loop_exiting`.
+    // CSVs produced by the VariableManager need to be handled carefully, we
+    // want only the ones with uses to survive, so we save their current linkage
+    // now, run `globaldce` within the pass and restore the previous linkage to
+    // the surviving ones.
```
aleclearmind: I was trying to rewrite this comment and I came up with the following observation.

We have the following type of globals: special globals, CSVs and QEMU globals.
Special globals must not be touched.
CSVs need to be external and retain their initializers.
QEMU globals do not need to have their initializers.
So we can:

1. Leave QEMU globals as external *but* turn them into declarations (i.e., drop their initializer). GlobalDCE will collect them if unused, even if external.
2. Mark CSVs as internal and restore them.

New version of the comment above:

> Now that all of the non-revng_inline functions have gone, we want to reduce the number of global variables with some help from GlobalDCE.
> We have three type of variables: 
> [rephrase the above]

Also rewrite the code below to reflect this simplified approach.

include/revng/Support/IRHelpers.h
mrjackv: I'm not seeing a clear way to fix this, it's either this, including `RecursiveCoroutine.h` or writing a very ugly function that takes all the attributes of the `Impl` class by reference

include/revng/Support/IRHelpers.h
mrjackv: Because we want to exclude the isolated functions to be looked into when analyzing

include/revng/Support/IRHelpers.h
mrjackv: No, the first has a list of the GVs we detached, the second has a list of all the GVs to allow re-inserting them in the same order, since one GV could have an initializer that refers to another one.

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: >  Have you tried not doing any special casing?

Yes, and it lead to `InjectStackSizeProbesAtCallSitesPass` to complain about the missing `SP`, I've also added PC and RA for good measure since they are used by GCBI

lib/Support/IRHelpers.cpp
mrjackv: Check carefully, this is a `static` function, the public `cloneFiltered` has `Verify` set to true. This is needed with `ConservativeModuleCloner` to allow it to re-use the `cloneFiltered` while using a known-broken module

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
mrjackv: I considered it, but:
1. There is code sharing with `SlimDownHelpersModule`
2. These two passes operate on the same context (the analyzed helpers module), IMHO it makes sense for them to live in the same TU

include/revng/Support/IRHelpers.h
aleclearmind: Just move this class to its own header file and use `RecursiveCoroutine.h`.

include/revng/Support/IRHelpers.h
aleclearmind: Asymptotically, the set of isolated functions is going to be bigger than its complement set.

lib/Lift/CodeGenerator.cpp
mrjackv: No, setting it lead to verify failing, I read the IR reference and I'm not sure why it's needed, removing it leads to the module verifying successfully and no crashes.

lib/FunctionIsolation/IsolateFunctions.cpp
aleclearmind: Just keep `SP` and add a note on why.

lib/Support/IRHelpers.cpp
aleclearmind: How does this fail verification? Detaching a global variable should not lead to verification failures.

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
aleclearmind: It's not OK to have `X.cpp` contain class `X` and class `Y` which has the same dignity as `Y`.

If you need to share code, make a header in `lib/HelperArgumentsAnalysis`.

lib/Lift/CodeGenerator.cpp
aleclearmind: Can you report the failure?

lib/Support/IRHelpers.cpp
mrjackv: It does because if a non-detached functions has a reference to a detached GV it fails verification

aleclearmind: Well, that sounds like a bug on our part.

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: > which holds references to ClonedModule, correct?

Yes

> Is this viable?

yes

lib/Support/IRHelpers.cpp
mrjackv: I think this is best left to be discussed tomorrow, there are multiple moving parts to this.

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
mrjackv: There's `-strip` but that drop *all* debug information, here (at least, in my understanding) we want to preserve debug info for globals

lib/Lift/CodeGenerator.cpp
mrjackv:
```
GlobalValue with local linkage or non-default visibility must be dso_local!
ptr @do_calc_cc
in function do_calc_cc
LLVM ERROR: Broken function found, compilation aborted!
```

This is when processing `libtcg-helpers-annotated-s390x.bc`

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
mrjackv: We expect globaldce to cut down 99% of global variables, so `M.getGlobalList().size() << GVLinkageSave.size()`

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
aleclearmind: Uhm, no. I don't think we want to preserve any debug info in helpers on this path of the pipeline.

lib/Lift/CodeGenerator.cpp
aleclearmind: Try to keep `setDSOLocal(false)` and add `setVisibility(Default)`.

Let's have less "weird" functions.

lib/FunctionIsolation/IsolateFunctions.cpp
aleclearmind: Add TODO about the fact that this is necessary due to `InjectStackSizeProbesAtCallSitesPass`, but that, in truth, in case the function does not change the stack pointer value, we can safely assume that all the call site have the stack pointer at the value it assumed at the function entry.

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
aleclearmind: Something along the lines of:
```c++
    StringMap<GlobalValue::LinkageTypes> GVLinkageSave;
    for (GlobalVariable &GV : M.globals()) {
      if (not(GV.isDeclaration() or isSpecialGV(GV))) {
        // Save linkage to be restored after we run globaldce
        if (FunctionTags::CSV.isTagOf(&GV)) {
          GVLinkageSave[GV.getName()] = GV.getLinkage();
          GV.setLinkage(GlobalValue::InternalLinkage);
        } else {
          GV.setInitializer(nullptr);
        }
      }
    }
```

aleclearmind:
As per our discussion:

## Point #1

* The helpers declarations module should not contain CSVs, we don't need them
* We don't need the helper_initialize_env body on the decompilation path
* ConservativeModuleCloner main goal is to clone only the correct revng.const.* (O(n)) and CSVs (O(qemu))
* What do:
  * Detach all CSVs and revng.const global variables
  * Detach all the basic blocks from isolated functions
  * Detach all the isolated functions and all the helper declarations
  * Verify
  * For each isolated function   
    * Reattach their body
    * Visit their body to identify all the globals and isolated functions that are used
    * Reattach the globals they use
    * Reattach the *declarations* of the isolated functions they use
    * Reattach the helpers they use
    * Verify
    * cloneFiltered
    * Verify all teh things
    * Re-detach everything

## Point #2

* Inliner helpers should no longer be fixed-point
* When preparing the slim module, we should mark as internal and always_inline all the non-helpers (i.e., those whose name does not start with helper_).
  Then we run the --always-inline pass and we get a module with only helpers that do not call any other function with a body.
* At this point, in -inline-helpers no longer need to do recursive visits.

lib/BasicAnalyses/GeneratedCodeBasicInfo.cpp
mrjackv: Done

include/revng/Support/IRHelpers.h
mrjackv: Done

include/revng/Support/IRHelpers.h
mrjackv: Moved to top

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
mrjackv: Split in two separate files

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
mrjackv: Done

lib/Lift/CodeGenerator.cpp
mrjackv: Done

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
mrjackv: Done

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
mrjackv: Switched to `globals()`

include/revng/Support/IRHelpers.h
mrjackv: Dropped, closing

include/revng/Support/IRHelpers.h
mrjackv: Dropped, closing

include/revng/Support/IRHelpers.h
mrjackv: Renamed to `MinimalModuleCloner`

include/revng/Support/IRHelpers.h
mrjackv: These have been dropped, closing

include/revng/Support/IRHelpers.h
mrjackv: Dropped

include/revng/Support/IRHelpers.h
mrjackv: Dropped

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: Added

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: Closing as it no longer applies

lib/Support/IRHelpers.cpp
mrjackv: Closing, no longer applies

lib/Support/IRHelpers.cpp
mrjackv: Dropped

lib/Support/IRHelpers.cpp
mrjackv: This code has been dropped, closing

lib/Support/IRHelpers.cpp
mrjackv: Code dropped, closing

include/revng/FunctionIsolation/IsolateFunctions.h
mrjackv: Closing as this was discussed

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
mrjackv: As discussed, the `-strip` pass removes too much (e.g. it removes the names of internal functions), sticking with `llvm::stripDebugInfo`

lib/Support/IRHelpers.cpp
mrjackv: Leaving this open for the current implementation of `cloneModule` which does manual cloning

CMakeLists.txt
mrjackv: Done

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
mrjackv: Done

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
mrjackv: Done

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: Closing as `GlobalDCE` is no longer used

aleclearmind: Enterprise.

include/revng/Support/PassManagerWrapper.h
```diff
@@ -0,0 +1,56 @@
...
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "llvm/Passes/PassBuilder.h"
+
+namespace detail {
+
+class PassManagerWrapperBase {
```
aleclearmind: Think 100 times before calling something wrapper.

"Wrapping" is "how", names should reflect "what".

This is a `SimplePassManager`.

lib/Lift/CodeGenerator.cpp
```diff
@@ -661,13 +664,11 @@ void CodeGenerator::translate(LibTcg &LibTcg,
 
   T.advance("Optimize lifted IR");
 
-  legacy::FunctionPassManager InstCombinePM(&*TheModule);
-  InstCombinePM.add(createSROAPass());
-  InstCombinePM.add(createInstructionCombiningPass());
-  InstCombinePM.add(createDeadCodeEliminationPass());
-  InstCombinePM.doInitialization();
+  FunctionPassManagerWrapper InstCombinePM;
+  InstCombinePM.addPass(llvm::SROAPass(SROAOptions::PreserveCFG));
```
aleclearmind, line 646: Are you sure about the option?

I really don't want to accidentally change the semantics.

include/revng/Support/SimplePassManager.h
aleclearmind, line 1: The commit message claims "all the applicable uses".

When we discussed this you mentioned we still have old pass managers around.

CMakeLists.txt
```diff
@@ -236,19 +237,51 @@ foreach(
...
+  endfunction()
+
+  # Produce slim version of helpers. In this version the bitcode file only
+  # contains helper functions that are in the `revng_inline` section and the
+  # CSVs that are used by them. This is used by `inline-helpers` to allow
+  # inlining the body of the `revng_inline` function into the isolated
+  # functions' bodies.
+  process_helper_module(
+    NAME
+    slim
```
aleclearmind: I propose the new nomenclature:

* `full`
* `to-inline`
* `declarations-only`

I'd also drop "annotated". Let's try to make things more straightforward.

lib/FunctionIsolation/InlineHelpers.cpp
```diff
@@ -102,14 +105,60 @@ static void dropDebugOrPseudoInst(Function *F) {
 }
 
 void InlineHelpers::run(Function *F) {
-  // Fixed-point inlining
-  while (doInline(F))
-    ;
-
+  doInline(F);
```
aleclearmind, line 110: Add a comment why we no longer need to this fixed point.

include/revng/FunctionIsolation/InlineHelpers.h
```diff
@@ -11,9 +11,13 @@
 class InlineHelpersPass : public llvm::ModulePass {
 public:
   static char ID;
+  llvm::StringMap<std::unique_ptr<llvm::Module>> HelpersModules;
```
aleclearmind, line 18: Add a comment explaining why this is halal.

lib/FunctionIsolation/InlineHelpers.cpp
```diff
@@ -102,14 +105,60 @@ static void dropDebugOrPseudoInst(Function *F) {
...
   dropDebugOrPseudoInst(F);
 }
 
+Module &InlineHelpersPass::loadHelpersModule(llvm::Module &M) {
+  NamedMDNode *Node = M.getNamedMetadata(QemuArchitectureMD);
+  revng_assert(Node != nullptr);
+  revng_assert(Node->getNumOperands() > 0);
+
+  // Multiple linkings could have lead to this node having more than one
+  // operand, if so check that they are all the same.
```
aleclearmind, line 120: Do they grow indefinitely? Do we end up having one per function?

If so, we need to make the merger aware of this and avoid it. We had this problem in the past.

lib/FunctionIsolation/InlineHelpers.cpp
```diff
@@ -102,14 +105,60 @@ static void dropDebugOrPseudoInst(Function *F) {
...
+  for (size_t I = 0; I < Node->getNumOperands(); I++) {
+    revng_assert(Node->getOperand(0)->getNumOperands() == 1);
+    const MDOperand &StringNode = Node->getOperand(0)->getOperand(0);
+    Strings.insert(cast<MDString>(StringNode)->getString());
+  }
+
+  revng_assert(Strings.size() == 1,
+               "QEMU helpers of multiple architectures were linked inside the "
+               "module");
+  StringRef ArchName = Strings.begin()->first();
```
aleclearmind, line 134: Add a comment to state you're doing something different than before.

lib/FunctionIsolation/InlineHelpers.cpp
```diff
@@ -102,14 +105,60 @@ static void dropDebugOrPseudoInst(Function *F) {
...
+  if (not HelpersModules.count(ArchName)) {
```
aleclearmind: `.contains`

lib/FunctionIsolation/InlineHelpers.cpp
```diff
@@ -102,14 +105,60 @@ static void dropDebugOrPseudoInst(Function *F) {
...
+    const std::string LibHelpersName = ("/share/revng/"
+                                        "libtcg-helpers-annotated-slim-"
+                                        + ArchName + ".bc")
+                                         .str();
+    auto OptionalHelpers = revng::ResourceFinder.findFile(LibHelpersName);
+    revng_assert(OptionalHelpers.has_value(), "Cannot find tinycode helpers");
```
aleclearmind: `s/tinycode/libtcg/`

lib/FunctionIsolation/InlineHelpers.cpp
```diff
@@ -102,14 +105,60 @@ static void dropDebugOrPseudoInst(Function *F) {
...
+    ToClone.insert(HelperFunction);
+  }
+
+  auto ClonedHelpersModule = ::cloneFiltered(HelpersModule, ToClone);
+  linkModules(std::move(ClonedHelpersModule), MainModule);
+}
+
 bool InlineHelpersPass::runOnModule(llvm::Module &M) {
+  Module &HelpersModule = loadHelpersModule(M);
+  cloneAndLinkHelpers(M, HelpersModule);
```
aleclearmind: Move the invocation to `loadHelpersModule` into `cloneAndLinkHelpers` and rename it to just `linkRequiredHelpers`.

lib/FunctionIsolation/InlineHelpers.cpp
```diff
@@ -122,5 +171,10 @@ bool InlineHelpersPass::runOnModule(llvm::Module &M) {
     IH.run(F);
   }
 
+  for (Function &F : M) {
+    if (F.getSection() == InlineHelpersSection)
+      deleteOnlyBody(F);
+  }
```
aleclearmind, line 181: Add comment on why we do this

lib/HelperArgumentsAnalysis/HelpersModuleToDeclarations.cpp
```diff
@@ -0,0 +1,79 @@
...
+using namespace llvm;
+
+class HelpersModuleToDeclarations : public llvm::ModulePass {
+public:
+  static char ID;
+
+public:
+  HelpersModuleToDeclarations() : llvm::ModulePass(ID) {}
+
+  void getAnalysisUsage(llvm::AnalysisUsage &AU) const override {}
```
aleclearmind: Drop?

lib/HelperArgumentsAnalysis/HelpersModuleToDeclarations.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+  bool runOnModule(llvm::Module &M) override {
+    for (Function &F : M) {
+      // This module does not require debugging information on the function
```
aleclearmind: This module does not require -> Only the full module requires

lib/HelperArgumentsAnalysis/HelpersModuleToDeclarations.cpp
```diff
@@ -0,0 +1,79 @@
...
+      // declarations. Remove the !dbg tag. Later the `-strip-dead-debug-info`
+      // will take care of dropping all the unused metadata.
+      F.eraseMetadata(LLVMContext::MD_dbg);
+      deleteOnlyBody(F);
+    }
+
+    // Mark all global variables with internal linkage, a couple
```
aleclearmind: Missing "except"?

lib/HelperArgumentsAnalysis/PostProcessing.h
```diff
@@ -0,0 +1,38 @@
...
+inline bool isSpecialGV(llvm::GlobalVariable &GV) {
+  llvm::StringRef Name = GV.getName();
+  return Name == "cpu_loop_exiting" or Name == "arch_cpu_type_beacon";
+}
+
+inline bool isHelper(llvm::Function &F) {
+  return F.getName().starts_with("helper_");
+}
+
+inline void tagGlobals(llvm::Module &Module) {
```
aleclearmind: Name too generic. `tagQEMUGlobalsAndFunctions`?

lib/HelperArgumentsAnalysis/HelpersModuleToDeclarations.cpp
```diff
@@ -0,0 +1,79 @@
...
+    ArrayType *FunctionArray = ArrayType::get(PointerType,
+                                              FunctionPointers.size());
+    Constant *Initializer = ConstantArray::get(FunctionArray, FunctionPointers);
+
+    new GlobalVariable(M,
+                       FunctionArray,
+                       true,
+                       GlobalValue::ExternalLinkage,
+                       Initializer,
+                       HelpersListName);
```
aleclearmind, line 67: 1. We need to do this since they have internal linkage, correct?
2. Who sets the linkage to internal? Add a comment.

lib/HelperArgumentsAnalysis/PostProcessingHelpers.h
aleclearmind, line 1: There isn't much "post processing" going on here. `PostProcessingHelpers`?

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
```diff
@@ -20,59 +29,74 @@ class SlimDownHelpersModule : public llvm::ModulePass {
...
       }
     }
 
-    // Assume the module M already contains the functions.
-    // Collect all functions whose name starts with "helper_" to prevent DCE
-    llvm::DenseSet<Function *> Helpers;
-    for (Function &F : M)
-      if (F.getName().startswith("helper_"))
-        Helpers.insert(&F);
+    // Run always-inline and globaldce
```
aleclearmind: 1. Elaborate on why `always-inline` is important.
2. How big does the module get after `always-inline`? If it's too big we might be slowing down loading, but I don't think so.

CMakeLists.txt
```diff
@@ -236,19 +237,51 @@ foreach(
...
-  add_custom_command(
-    OUTPUT "${LIBTCG_HELPERS_ANNOTATED_SLIM}"
-    DEPENDS "${LIBTCG_HELPERS_ANNOTATED}" revng-all-binaries
-    COMMAND ./bin/revng opt -slim-down-helpers-module -globaldce
-            "${LIBTCG_HELPERS_ANNOTATED}" -o "${LIBTCG_HELPERS_ANNOTATED_SLIM}")
-  add_custom_target("annotated-libtcg-helpers-slim-${ARCH}" ALL
-                    DEPENDS "${LIBTCG_HELPERS_ANNOTATED_SLIM}")
-  list(APPEND HELPER_MODULE_LIST "${LIBTCG_HELPERS_ANNOTATED_SLIM}")
-
+  function(process_helper_module)
```
aleclearmind, line 240: Nice

include/revng/Support/IRHelpers.h
```diff
@@ -1558,6 +1562,12 @@ inline void linkModules(std::unique_ptr<llvm::Module> &&Source,
 /// This requires the source module to be writable in order to guarantee that
 /// the metadata attached to the cloned functions is also preserved.
 std::unique_ptr<llvm::Module>
+cloneFiltered(llvm::Module &Module,
```
aleclearmind, line 1566: Update doxygen

lib/FunctionIsolation/IsolateFunctions.cpp
```diff
@@ -841,6 +847,21 @@ void Isolate::splitIsolatedFunctionsToOutput() {
     T.advance(Address.toString(), true);
 
     auto IsolatedModule = Cloner.cloneModule(*Function);
+    // Check that running globaldce on the isolated module does nothing
+    if (VerifyLog.isEnabled()) {
```
aleclearmind, line 853: My idea was to push `GlobalDCEPass` within this `if`, so we run it only to check that it did not do anything.

lib/Lift/CodeGenerator.cpp
mrjackv: > Are you sure about the option?

100%, see [here](https://github.com/revng/llvm-project/blob/develop/llvm/include/llvm/Transforms/Scalar.h#L108)

include/revng/Support/SimplePassManager.h
mrjackv: Yes, there are places, e.g. [here](https://github.com/revng/revng/blob/develop/lib/Lift/JumpTargetManager.cpp#L1307) where we schedule passes that we wrote ourselves. These need to be converted to the new pass manager. Other instances are related to the old pipeline and those will go away when the old pipeline is dropped.
The meaning of the commit is "I've replaced the old pass manager with the new one in all the places where all the passes were already supporting it (e.g. where we use only llvm passes)"

lib/FunctionIsolation/InlineHelpers.cpp
mrjackv: > Do they grow indefinitely? Do we end up having one per function?

No, I've observed at most two in the old pipeline, but no more than that.
In the new pipeline it's always 1.

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: It is that way, the `GlobalDCE` on `ClonedModule` is beneficial because it removes a bunch of globals before cloning

aleclearmind: Mh, I got confused. But what is the invocation to `GlobalDCEPass` at the beginning of `splitIsolatedFunctionsToOutput` for?

mrjackv: The isolation process leaves some cruft in the LLVM module (e.g. unused MetaAddresses, unused helpers), removing them before cloning means they don't need to be iterated upon by `CloneModule` so it's beneficial to do so.

lib/FunctionIsolation/IsolateFunctions.cpp
```diff
@@ -735,22 +834,35 @@ void Isolate::runOnFunction(const model::Function &TheFunction) {
...
-    Output.assign(ObjectID(Address), ::cloneFiltered(*ClonedModule, ToClone));
+    auto IsolatedModule = Cloner.cloneModule(*Function);
+    // Check that running globaldce on the isolated module does nothing
+    if (VerifyLog.isEnabled()) {
+      llvm::StringSet GlobalNames;
+      for (GlobalValue &GV : IsolatedModule->global_values()) {
+        if (GV.hasName())
+          GlobalNames.insert(GV.getName());
+      }
+      PM.run(*IsolatedModule);
```
aleclearmind, line 862: Recreate the pipeline here, I didn't even noticed it.
I don't want to have side-effects if we change the code above.

mrjackv: Done

lib/HelperArgumentsAnalysis/PostProcessingHelpers.h
mrjackv: Renamed

lib/HelperArgumentsAnalysis/PostProcessing.h
mrjackv: Renamed

lib/HelperArgumentsAnalysis/HelpersModuleToDeclarations.cpp
mrjackv: I had another look at the comment and re-wrote it from scratch to better explain the situation.

lib/HelperArgumentsAnalysis/HelpersModuleToDeclarations.cpp
mrjackv: Done

lib/HelperArgumentsAnalysis/HelpersModuleToDeclarations.cpp
mrjackv: Dropped on both passes

include/revng/Support/PassManagerWrapper.h
mrjackv: Renamed to `SimplePassManager`

lib/FunctionIsolation/InlineHelpers.cpp
mrjackv: This has been added in [llvm 17](https://github.com/llvm/llvm-project/commit/21700677cb44f58a8f4c9d9ed6c05dd049108854), we're on 16

lib/FunctionIsolation/InlineHelpers.cpp
mrjackv: I'm a bit confused, what is the before relative to? The previous implementation where the helpers had a body in the main module?

lib/HelperArgumentsAnalysis/HelpersModuleToDeclarations.cpp
mrjackv: 1. We need to do this because the helper functions are all declarations and they might get globaldce'd away otherwise
2. See (1)

In any case I've reworded the comment just above to clarify the situation

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
mrjackv: 1. ACK
2. They are actually *WAY* smaller than their pre-PR counterpart (e.g. x86_64 is 1.3M (old) vs 34k (new))

lib/FunctionIsolation/InlineHelpers.cpp
aleclearmind:
```patch
   revng_assert(Strings.size() == 1,
                "QEMU helpers of multiple architectures were linked inside the "
                "module");

+  // Now I'm doing something else
   StringRef ArchName = Strings.begin()->first();
```

lib/FunctionIsolation/InlineHelpers.cpp
aleclearmind: Ah! Use `.count != 0`

lib/HelperArgumentsAnalysis/HelpersModuleToDeclarations.cpp
aleclearmind: Ah, right.

CMakeLists.txt
mrjackv: Done

lib/FunctionIsolation/InlineHelpers.cpp
mrjackv: Done

include/revng/FunctionIsolation/InlineHelpers.h
mrjackv: Added

lib/FunctionIsolation/InlineHelpers.cpp
mrjackv: Ah ok, done

lib/FunctionIsolation/InlineHelpers.cpp
mrjackv: Done

lib/FunctionIsolation/InlineHelpers.cpp
mrjackv: Fixed

lib/FunctionIsolation/InlineHelpers.cpp
mrjackv: Done

lib/FunctionIsolation/InlineHelpers.cpp
mrjackv: Added

lib/HelperArgumentsAnalysis/SlimDownHelpersModule.cpp
mrjackv: 1. Done

include/revng/Support/IRHelpers.h
mrjackv: Fixed

PR #558, opened 2026-02-10
Fully implement pypeline tracing

mrjackv:
Sister PRs:
* [orchestra](https://github.com/revng/orchestra/pull/194)
* [llvm-project](https://github.com/revng/llvm-project/pull/45)


python/revng/internal/cli/revng2.py
```diff
@@ -130,6 +140,8 @@ def artifact(
     container_format: ContainerFormat,
     runner_context: RunnerContext,
 ):
+    exec_wrapper_if_needed()
```
aleclearmind: I'm confused, why is `artifact` special? 🤔

python/revng/internal/cli/revng2.py
```diff
@@ -264,11 +312,18 @@ def patch_pype():
         elif param.name == "cache_dir":
             param.default = cache_directory()
 
+    for wrapper in WRAPPERS:
+        pype = wrapper.add_to_command(pype)
+
+    pype_main.pype = pype
+    pipeline.callback = add_wrapper_exec(pipeline.callback)
+    project.callback = add_wrapper_exec(project.callback)
```
aleclearmind: This is not a "callback". A callback is some sort of hook.

AFAIU `project.callback` is not a hook, it the method called when you want to run things, and we override/wrap it to do something different.

python/revng/internal/cli/support.py
```diff
@@ -223,3 +224,79 @@ def file_wrapper(path: str | None, mode: Literal["r", "rb", "w", "wb"]):
...
+
+
+def exec_with_wrapper(args: list[str], env: Mapping[str, str] | None = None):
+    if env is None:
+        env = os.environ
+    ctx = click.get_current_context()
+    wrapper: Wrapper | None = ctx.obj.get("wrapper")
+    if wrapper is not None:
+        args = [*wrapper.prefix, *args]
+    os.execvpe(args[0], args, env)
```
aleclearmind: The fact that we have this non-returning functions in random places is quite confusing.

I'd like to see in the places where it's used something like:

```python
if should_rerun:
  exec_...()
else:
  regular_path()
```

include/revng/PipeboxCommon/Helpers/Native/Analysis.h
```diff
@@ -19,6 +19,8 @@ class Analysis {
                           std::vector<Container *> Containers,
                           const pypeline::Request &Incoming,
                           llvm::StringRef Configuration) = 0;
+
+  virtual std::vector<ContainerArgument> signature() const = 0;
```
aleclearmind, line 23: `arguments`?

include/revng/PipeboxCommon/Runners/ContainerHandler.h
```diff
@@ -0,0 +1,109 @@
...
+
+class ContainerHandler {
+private:
+  using ContainerPath = std::pair<Container *, llvm::StringRef>;
+
+private:
+  llvm::ArrayRef<ContainerArgument> Signature;
+  std::vector<std::unique_ptr<Container>> Containers;
+  std::vector<ContainerPath> Loads;
+  std::vector<ContainerPath> Stores;
```
aleclearmind: `ToLoad` and `ToStore`?

include/revng/PipeboxCommon/Runners/ContainerHandler.h
```diff
@@ -0,0 +1,109 @@
...
+
+public:
+  ContainerHandler(llvm::ArrayRef<ContainerArgument> Signature) :
+    Signature(Signature) {}
+
+  size_t getExpectedNumberOfArguments() {
```
aleclearmind: `getExpectedNumberOfCLIArguments`

include/revng/PipeboxCommon/Helpers/Native/CLI/ContainerHandler.h
```diff
@@ -0,0 +1,109 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "revng/PipeboxCommon/Helpers/Native/Pipe.h"
+#include "revng/PipeboxCommon/Helpers/Native/Registry.h"
+#include "revng/Support/Tar.h"
+
+namespace revng::pypeline::helpers::native {
+
+class ContainerHandler {
```
aleclearmind, line 18: 1. Very generic name. This seems to be specific to CLI and the name should reflect it. `revng::pypeline::helpers::native::cli::ContainerHandler`?
2. Add some docs. Mention `.tar`.

tools/pypeline/run-analysis/CMakeLists.txt
aleclearmind, line 1: Follow the current pattern: `tools/pypeline/run-analysis`

include/revng/PipeboxCommon/Helpers/Native/CLI/ContainerHandler.h
aleclearmind, line 1: This file and `Runners/Helpers.h` are `Native`-specific and should be under `Helpers/Native/CLI`.

include/revng/PipeboxCommon/Runners/Helpers.h
```diff
@@ -0,0 +1,76 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/StringMap.h"
+#include "llvm/ADT/StringRef.h"
+
+#include "revng/PipeboxCommon/Helpers/Native/Helpers.h"
+
+namespace revng::pypeline::helpers::native {
```
aleclearmind: `::cli`

tools/pypeline-run-analysis/Main.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include "revng/PipeboxCommon/Helpers/Native/Registry.h"
+#include "revng/PipeboxCommon/Runners/ContainerHandler.h"
+#include "revng/PipeboxCommon/Runners/Helpers.h"
+#include "revng/Support/InitRevng.h"
+
+static llvm::ExitOnError AbortOnError;
+
+namespace Options {
+using namespace llvm::cl;
+using optstring = opt<std::string>;
```
aleclearmind: Fix case

tools/pypeline/run-analysis/Main.cpp
```diff
@@ -0,0 +1,103 @@
...
+                               cat(RunAnalysisCategory));
+
+static list<std::string> Objects("objects",
+                                 ZeroOrMore,
+                                 multi_val(2),
+                                 desc("pair-wise container outgoing requests"),
+                                 cat(RunAnalysisCategory));
+
+static optstring
+  Output("o", desc("output model path"), cat(RunAnalysisCategory));
```
aleclearmind: Add `--output`. Let's always have an explicit version.

tools/pypeline/run-analysis/Main.cpp
```diff
@@ -0,0 +1,103 @@
...
+  // Get the vector of container pointers
+  auto ContainersPointers = ContainerHandler.getContainers();
+
+  // Run the actual pipe
+  AbortOnError(TheAnalysis->run(TheModel,
+                                ContainersPointers,
+                                Outgoing,
+                                Options::Configuration));
+
+  if (not Options::Output.empty()) {
```
aleclearmind: Don't we write to stdout if `-o` is missing?

The old pipeline used to do this. We can change that, but let's be coherent with `revng2 analyze`.

tools/pypeline-run-pipe/Main.cpp
```diff
@@ -0,0 +1,186 @@
...
+  Model(Positional, Required, desc("MODEL"), cat(RunPipeCategory));
+
+static list<std::string> Arguments(Positional,
+                                   ZeroOrMore,
+                                   desc("CONTAINER_FILES..."),
+                                   cat(RunPipeCategory));
+
+} // namespace Options
+
+struct SerializedInvalidation {
```
aleclearmind: 1. If you use `private:` or `public:`, switch to `class`.
2. I think we want this in a more public header.

tools/pypeline-run-pipe/Main.cpp
```diff
@@ -0,0 +1,186 @@
...
+private:
+  friend llvm::yaml::ScalarTraits<SerializedInvalidation>;
+
+private:
+  size_t Index;
```
aleclearmind: `= 0`

tools/pypeline-run-pipe/Main.cpp
```diff
@@ -0,0 +1,186 @@
...
+  std::string Object;
+  std::string Path;
+
+public:
+  // Must be default-constructible due to its use in YAMLTraits
+  SerializedInvalidation(){};
```
aleclearmind: 1. `= default`
2. `clang-format`?

tools/pypeline-run-pipe/Main.cpp
```diff
@@ -0,0 +1,186 @@
...
+    Out << '[' << Value.Index << ", " << Value.Object << ", " << Value.Path
+        << ']';
+  }
+
+  static StringRef input(StringRef, void *, SerializedInvalidation &) {
+    revng_abort();
+  }
+
+  static QuotingType mustQuote(StringRef) { return QuotingType::None; }
+};
```
aleclearmind: 1. What you're emitting is not a scalar. This is a bad hack. At this point, I'd just bypass YAML traits entirely. Just write by hand.
2. Are you sure we're never going to need quotes?

tools/pypeline-run-pipe/Main.cpp
```diff
@@ -0,0 +1,186 @@
...
+  revng::InitRevng X(Argc, Argv, "", { &Options::RunPipeCategory });
+
+  // Create the pipe instance
+  revng_assert(Registry.Pipes.count(Options::Pipe));
+  std::unique_ptr<Pipe>
+    ThePipe = Registry.Pipes[Options::Pipe](Options::StaticConfiguration);
+  auto Signature = ThePipe->signature();
+
+  // Deserialize model
+  Model TheModel = AbortOnError(modelFromPath(Options::Model));
```
aleclearmind: `ExitOnError`? 🤔

include/revng/PipeboxCommon/Runners/Helpers.h
```diff
@@ -0,0 +1,76 @@
...
+    else
+      Result += std::tolower(C);
+  }
+
+  return Result;
+}
+
+inline llvm::Error
+parseRequest(llvm::ArrayRef<ContainerArgument> Signature,
+             llvm::ArrayRef<std::string> CmdObjects,
```
aleclearmind: No abbreviations

include/revng/PipeboxCommon/Runners/Helpers.h
```diff
@@ -0,0 +1,76 @@
...
+  for (auto &Argument : Signature) {
+    std::string Name = normalizeName(Argument.Name);
+    ObjectMapping[Name] = Index;
+    Index++;
+  }
+
+  size_t ObjectsIndex = 0;
+  while (ObjectsIndex < CmdObjects.size()) {
+    llvm::StringRef Name = CmdObjects[ObjectsIndex];
+    llvm::StringRef Objects = CmdObjects[ObjectsIndex + 1];
```
aleclearmind: Possible out of bound access?

python/revng/internal/pipebox.py
```diff
@@ -37,9 +56,21 @@ def initialize(argv: list[str] = []):
...
     assert input_[0] == "pype"
-    output = input_[:]
-    output[0] = "revng2"
-    return output
+    assert input_[1] == "pipeline"
+    assert input_[2] in ("run-pipe", "run-analysis")
+
+    if input_[2] == "run-pipe":
+        pipe_name = input_[3]
+        if pipe_name not in _native_pipes:
```
aleclearmind, line 69: :enterprise:

python/revng/internal/pipebox.py
```diff
@@ -37,9 +56,21 @@ def initialize(argv: list[str] = []):
 
 def argv_hook(input_: list[str]):
```
aleclearmind, line 59: Docs!

python/revng/internal/pipebox.py
```diff
@@ -16,7 +18,24 @@
...
+        continue
+
+    _entry = getattr(_module, _entry_name)
+    if not inspect.isclass(_entry):
+        continue
+
+    if issubclass(_entry, Pipe):
+        _native_pipes[_entry.name] = _entry
+    elif issubclass(_entry, Analysis):
+        _native_analyses[_entry.name] = _entry
```
aleclearmind: 1. Do we really need to do this at module level?
2. Wrap into a function.

python/revng/internal/cli/revng2.py
```diff
@@ -280,6 +284,129 @@ def generate_prefix(self, value: Any) -> list[str]:
 ]
 
 
+def _generate_load_arguments(ctx):
+    native_libraries: list[Path] = getattr(ctx.obj["pipebox"], "_native_libraries", [])
+    return [f"-load={p.resolve()!s}" for p in native_libraries]
```
aleclearmind, line 292: How many of these do we have in practice? I'd be really happy to get rid of a thousand `-load`s.

python/revng/internal/cli/revng2.py
```diff
@@ -280,6 +284,129 @@ def generate_prefix(self, value: Any) -> list[str]:
...
+    help="Run a pipe without python",
+)
+def run_pipe_native() -> None:
+    pass
+
+
+class RunAnalysisNativeGroup(PypeGroup):
+    @staticmethod
+    def _get_analyses(ctx) -> dict[str, type[Analysis]]:
+        return getattr(ctx.obj["pipebox"], "_native_analyses", {})
```
aleclearmind: Do we really need `getattr`?

python/revng/internal/cli/revng2.py
```diff
@@ -280,6 +284,129 @@ def generate_prefix(self, value: Any) -> list[str]:
...
+
+@click.group(
+    cls=RunPipeNativeGroup,
+    help="Run a pipe without python",
+)
+def run_pipe_native() -> None:
+    pass
+
+
+class RunAnalysisNativeGroup(PypeGroup):
```
aleclearmind, line 360: Docs, what does this do in practice?

python/revng/internal/cli/revng2.py
mrjackv: `callback` is the way `click` calls the original function that has been wrapped in a `Command`, nothing we can control

python/revng/internal/cli/revng2.py
mrjackv: There's 2 categories of commands:
* Commands that run code you want to gdb into (e.g. `revng2 project artifact`)
* Commands that will run commands that you want to gdb into (e.g. `revng2 pipeline run-pipe-native`, `revng opt`)

include/revng/PipeboxCommon/Helpers/Native/Analysis.h
mrjackv: It's the same name as in python, IMHO it makes more sense

tools/pypeline-run-pipe/Main.cpp
mrjackv: `AbortOnError` and `ExitOnError` are both used under `tools`, I can switch to the `Exit` version if you prefer

include/revng/PipeboxCommon/Runners/Helpers.h
mrjackv: No, because we checked it beforehand via `getExpectedNumberOfArguments`

python/revng/internal/cli/revng2.py
mrjackv: Just one, it could be maybe 2 if we have some closed-source pipes in the future.
This is to also support unit-tests where the library is different from `librevngPipebox.so`

python/revng/internal/cli/revng2.py
aleclearmind: What about `revng project analyze`?

python/revng/internal/cli/revng2.py
aleclearmind: Great.

tools/pypeline-run-pipe/Main.cpp
mrjackv: 1. Ok
2. Index is an int, it doesn't need quotes
    Object is an ObjectID, so the only thing that can potentially have quoting issues is MetaAddress which is safe
    Path *could* have quotes, but I think it's highly unlikely that the model will have quotable keys anytime soon

python/revng/internal/cli/revng2.py
mrjackv: This is a `revng`-specific thing, from the POV of pypeline these attributes don't exist, so I think it's warranted they're optional

python/revng/internal/cli/revng2.py
mrjackv: Call `libexec/pypeline-run-analysis` with the correct arguments and a wrapper if needed

tools/pypeline-run-pipe/Main.cpp
aleclearmind: 2. Let's err on the safe side and escape strings. This doesn't need to be nice to look at.

python/revng/internal/cli/revng2.py
aleclearmind: OK, docs.

include/revng/PipeboxCommon/Runners/ContainerHandler.h
mrjackv: Renamed

include/revng/PipeboxCommon/Runners/ContainerHandler.h
mrjackv: This function has been dropped, now `parseCommandLine` performs the check and returns an `llvm::Error` if there is a mismatch

include/revng/PipeboxCommon/Helpers/Native/CLI/ContainerHandler.h
mrjackv: 1. Done
2. Done

tools/pypeline/run-analysis/CMakeLists.txt
mrjackv: Done

include/revng/PipeboxCommon/Helpers/Native/CLI/ContainerHandler.h
mrjackv: Moved there

include/revng/PipeboxCommon/Runners/Helpers.h
mrjackv: Done

tools/pypeline-run-analysis/Main.cpp
mrjackv: Fixed

tools/pypeline/run-analysis/Main.cpp
mrjackv: Added the alias

tools/pypeline-run-pipe/Main.cpp
mrjackv: This `struct` has been dropped

tools/pypeline-run-pipe/Main.cpp
mrjackv: This has been dropped

tools/pypeline-run-pipe/Main.cpp
mrjackv: This has been dropped

tools/pypeline-run-pipe/Main.cpp
mrjackv: 1. Done
2. Used `llvm::yaml::escape` to escape the strings

include/revng/PipeboxCommon/Runners/Helpers.h
mrjackv: Fixed

python/revng/internal/cli/revng2.py
mrjackv: Dropped `getattr`

include/revng/PipeboxCommon/Runners/ContainerHandler.h
mrjackv: Renamed

python/revng/internal/cli/revng2.py
mrjackv: Switched to `pypeline` having wrappers facilities and `revng2` defining them

python/revng/internal/cli/support.py
mrjackv: Converted to decorator

tools/pypeline/run-analysis/Main.cpp
mrjackv: Fixed

python/revng/internal/pipebox.py
mrjackv: Done

python/revng/internal/pipebox.py
mrjackv: 1. I thought about it but this is the simplest thing, I'm not sure having something more complicated is warranted
2. Done

python/revng/internal/cli/revng2.py
mrjackv: Added docs

PR #560, opened 2026-02-19
Port tests to pypeline

mrjackv:


include/revng/TupleTree/VersionField.h
aleclearmind, line 1: This commit seems rather evil:

1. Pattern matches on name
2. Uses mutable
3. Has some weird `operator uint64_t`

Please explain:

1. What are you trying to achieve?
2. What were we doing before?

include/revng/PipeboxCommon/Model.h
```diff
@@ -84,6 +84,12 @@ class Model {
     return Result;
   }
 
+  bool operator==(const Model &Other) const {
+    if (this == &Other)
+      return true;
+    return this->get().get()->localCompare(*Other.get().get());
```
aleclearmind: This is not a deep comparison.

You're not likely to want this.

include/revng/PipeboxCommon/Model.h
aleclearmind, line 1: > Change the `get_model` and `set_model` interface of `StorageProvider` so that it is responsibility of the `StorageProvider` to serialize/deserialize the model before returning to the caller.

Why do we do this?

lib/PipeboxCommon/Python/_pipebox.cpp
```diff
@@ -48,6 +49,42 @@ static void handleSignal(int SigNo) {
   }
 }
 
+static llvm::Expected<std::pair<Model, bool>>
+deserializeModel(llvm::ArrayRef<uint8_t> Input) {
```
aleclearmind, line 59: 1. Docs. What's the returned `bool`?
2. This function would benefit from some logging.

lib/PipeboxCommon/Python/_pipebox.cpp
```diff
@@ -48,6 +49,42 @@ static void handleSignal(int SigNo) {
...
+  auto MaybeModel2 = Model::deserialize({ BytesPointer, OutputBytes.size() });
+  if (MaybeModel2) {
+    // Here consumeError is valid because the error came from the fact that the
+    // model was of a previous version and it got updated
+    llvm::consumeError(MaybeModel.takeError());
+    return std::pair{ *MaybeModel2, true };
+  }
+
+  // If here, we didn't manage to deserialize a valid model, return the errors
+  return revng::joinErrors(MaybeModel.takeError(), MaybeModel2.takeError());
```
aleclearmind: This will results in a quite obscure double error.

Create a new error stating that the model was invalid, we tried to upgrade it and it failed again, reporting the respective errors.

python/revng/pypeline/model.py
```diff
@@ -122,7 +122,7 @@ def serialize(self) -> bytes:
 
     @classmethod
     @abstractmethod
-    def deserialize(cls, data: bytes) -> Model:
+    def deserialize(cls, data: bytes) -> tuple[Model, bool]:
```
aleclearmind, line 128: Add docs

python/revng/pypeline/storage/local_provider.py
```diff
@@ -473,7 +496,10 @@ def invalidate(
                         ),
                     )
 
-            cursor.execute(INVALIDATE_QUERY.format(objectid_mask=_OBJECTID_MASK, uuid=table_uuid))
+            cursor.executescript(
```
aleclearmind, line 524: Add a big docstring on `invalidate` explaining what's going on and stressing that we manage to let do the whole work to the DB engine.

Maybe merge the docs of `INVALIDATE_QUERY` and `INVALIDATE_QUERY2` here.

share/revng/test/configuration/revng/end-to-end-translation.yml
```diff
@@ -11,12 +11,14 @@ commands:
       - type: revng-qa.compiled-with-debug-info
         filter: for-runtime and for-comparison
     command: |-
-      revng artifact
-        --analyses=import-binary,detect-abi
```
aleclearmind, line 15: I use this quite a lot.

Can we have it?

share/revng/test/configuration/revng/for-collect-cfep.yml
```diff
@@ -8,7 +8,6 @@ commands:
       - type: revng-qa.compiled-with-debug-info
         filter: for-collect-cfep
     command: |-
-      revng analyze
-        revng-initial-auto-analysis
-        "${INPUT}" |
-        FileCheck --ignore-case "${SOURCE}".yml.filecheck
+      cd $$(temp -d);
+      revng2 project init "$INPUT";
```
aleclearmind: Can't we do this with `revng2 quick`?

share/revng/test/configuration/revng/for-collect-cfep.yml
aleclearmind, line 1: Please make a commit with a `TODO` for all the tests that still need to be ported but you didn't port due to crashes.

share/revng/test/configuration/revng/for-import-types.yml
```diff
@@ -8,5 +8,7 @@ commands:
       - type: revng-qa.compiled-with-debug-info
         filter: for-import-types
     command: |-
-      revng analyze import-binary "$INPUT"
+      cd "$$(temp -d)";
+      revng2 project init --no-initial-auto-analysis "$INPUT";
+      revng2 project analyze parse-binary
```
aleclearmind: Can we have `revng quick analyze --analyses=`?

If you deem this inappropriate, let's discuss it.

lib/FunctionIsolation/IsolateFunctions.cpp
```diff
@@ -705,14 +705,26 @@ class MinimalModuleCloner {
...
+    auto Architecture = Binary.Architecture();
+    std::string SPName = getCSVName(getStackPointer(Architecture));
+    llvm::StringRef PCName = getPCCSVName(Architecture);
+    auto IsSpecialCSV = [&SPName,
+                         &PCName,
+                         &Architecture](llvm::GlobalVariable &GV) {
+      StringRef Name = GV.getName();
+      return Name == SPName or Name == PCName
+             or (Architecture == model::Architecture::arm and Name == "_thumb");
+    };
```
aleclearmind, line 715: NOTETOSELF: review this, suspicious

aleclearmind (commented):
NOTETOSELF: Review again `Port tests to the new pypeline`

include/revng/PipeboxCommon/Model.h
mrjackv: So that in the next commit we can have the model upgrade logic happen in the storage provider and be transparent to downstream users

include/revng/PipeboxCommon/Model.h
mrjackv: > This is not a deep comparison.

Looking at the code *it is* a deep comparison and does work correctly as opposed to `operator==`

include/revng/TupleTree/VersionField.h
mrjackv: > What are you trying to achieve?

```
# echo '' | revng model opt
---
Version:         8
...
# echo '{}' | revng model opt 
---
Version:         8
...
# echo '{Version: 0}' | revng model opt 
---
Version:         8
...
```

> What were we doing before?

```
# echo '' | revng model opt
---
Version:         8
...
# echo '{}' | revng model opt 
---
...
# echo '{Version: 0}' | revng model opt 
---
...
```

The problem stems from `YAMLTraits` having `0` as the default value and nowhere nothing does `if (Version == 0) { Version = SchemaVersion }`

> Pattern matches on name

This is done in other places in `struct.h.tpl`

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: Here's the places where the new GVs are used:
* https://github.com/revng/revng/blob/develop/lib/BasicAnalyses/GeneratedCodeBasicInfo.cpp#L53 (for PC)
* https://github.com/revng/revng/blob/develop/lib/Model/ProgramCounterHandler.cpp#L130 (for `_thumb`)

python/revng/pypeline/storage/local_provider.py
mrjackv: NOTETOSELF: add extra comments to `def invalidate` explaining what's going on

include/revng/PipeboxCommon/Model.h
aleclearmind: `localCompare` does not compare `TupleTreeReference` fields.

```c++
bool model::generated::DefinedType::localCompare(
    const model::DefinedType &Other) const {
  if (this->Kind() != Other.Kind())
    return false;

  if (this->IsConst() != Other.IsConst())
    return false;

  // The following field will not be compared here as it's
  // a reference field: `Definition`.

  return true;
}
```

include/revng/PipeboxCommon/Model.h
aleclearmind: State that in the commit message.

include/revng/PipeboxCommon/Model.h
mrjackv: Eh, `operator==` is more broken because of:
```
bool operator==(const Function &Other) const { return key() == Other.key(); }
```

If you want to introduce a `deepCompare` I'm all for it

aleclearmind: Why do you need this again?

In any case, if you need it, the correct thing would be to change all the `KeyedObjectContainers` to use the key as sorting key and then promote `deepCompare` to be `operator==`. I'd need to take a deeper look at why the `operator==` method is implemented this way.

mrjackv:
NOTE: `Model.enableCaching` and `Model.disableCaching` still need to be implemented

include/revng/TupleTree/VersionField.h
mrjackv: Fixed in `TupleTree.h`

include/revng/PipeboxCommon/Model.h
mrjackv: Done

lib/PipeboxCommon/Python/_pipebox.cpp
mrjackv: 1. Done
2. Done

lib/PipeboxCommon/Python/_pipebox.cpp

python/revng/pypeline/model.py
mrjackv: Done

python/revng/pypeline/storage/local_provider.py
mrjackv: Improved the comment above `INVALIDATE_QUERY` and added a bunch of comments to the `invalidate` function explaining what's going on

share/revng/test/configuration/revng/end-to-end-translation.yml
mrjackv: Added

share/revng/test/configuration/revng/for-collect-cfep.yml
mrjackv: It needed `revng2 quick analyze`, now there is

share/revng/test/configuration/revng/for-collect-cfep.yml
mrjackv: Done

share/revng/test/configuration/revng/for-import-types.yml
mrjackv: Done

include/revng/PipeboxCommon/Model.h
mrjackv: Fixed `operator==`

include/revng/TupleTree/TupleTreeReference.h
```diff
@@ -22,19 +22,25 @@ class TupleTreeReference {
 public:
   using RootT = RootType;
   using RootVariant = std::variant<RootT *, const RootT *>;
-  using TargetVariant = std::variant<T *, const T *>;
+
+private:
+  static inline T *const NoCachedTarget = static_cast<T *>(nullptr);
+  static inline T
+    *const CachingDisabled = reinterpret_cast<T *>(static_cast<uintptr_t>(-1));
```
aleclearmind, line 29: `inline`? o_O

include/revng/TupleTree/TupleTreeReference.h
```diff
@@ -125,82 +104,83 @@ class TupleTreeReference {
...
-      return Pointer == nullptr;
+      return Pointer != nullptr;
     };
     return std::visit(IsNullVisitor, Root);
   }
 
-  bool canGet() const {
-    return Root.index() != std::variant_npos and not hasNullRoot();
+  template<typename U>
+    requires std::is_same_v<U, T> or std::is_same_v<U, const T>
```
aleclearmind: `ConstOrNot`?

include/revng/TupleTree/TupleTreeReference.h
```diff
@@ -125,82 +104,83 @@ class TupleTreeReference {
...
+  U *setCache(U *Ptr) const {
+    // The caller might have legitimately returned one of the two special
+    // pointer we use as placeholders, in this case return the pointer directly
+    // and don't cache it to avoid issues.
+    if (Ptr == CachedTarget or Ptr == CachingDisabled)
+      return Ptr;
```
aleclearmind: Uh? Maybe `assert`.

include/revng/TupleTree/TupleTreeReference.h
```diff
@@ -125,82 +104,83 @@ class TupleTreeReference {
...
+
+    // Here we assume that the caller has called `isCached` beforehand
+    if (CachedTarget != CachingDisabled) {
```
aleclearmind: `isCachingEnabled()`

include/revng/TupleTree/TupleTreeReference.h
```diff
@@ -125,82 +104,83 @@ class TupleTreeReference {
...
+      if constexpr (std::is_same_v<U, const T>)
+        CachedTarget = const_cast<T *>(Ptr);
+      else
+        CachedTarget = Ptr;
```
aleclearmind: You can unconditionally use `const_cast`

include/revng/TupleTree/TupleTreeReference.h
```diff
@@ -125,82 +104,83 @@ class TupleTreeReference {
...
+      return Pointer != nullptr;
     };
     return std::visit(IsNullVisitor, Root);
   }
 
-  bool canGet() const {
-    return Root.index() != std::variant_npos and not hasNullRoot();
+  template<typename U>
+    requires std::is_same_v<U, T> or std::is_same_v<U, const T>
+  U *setCache(U *Ptr) const {
```
aleclearmind: I wouldn't make it return anything.

include/revng/TupleTree/TupleTreeReference.h
```diff
@@ -22,19 +22,25 @@ class TupleTreeReference {
...
 
 private:
   RootVariant Root = static_cast<RootT *>(nullptr);
   TupleTreePath Path;
-  TargetVariant CachedTarget = static_cast<T *>(nullptr);
+  // This field needs to work even if the object is const. Moreover the
+  // const-ness of the returned pointer is dictated by the constness of `Root`.
+  // All public interfaces (`get*()`) respect the const-ness by checking the
+  // const-ness of `Root`.
+  mutable T *CachedTarget = CachingDisabled;
```
aleclearmind: Ah this is why we didn't do this, we don't want `mutable`.
I think your patch is worth it though.

> This field needs to work even if the object is const.

Explain why (lazy caching).

lib/FunctionIsolation/IsolateFunctions.cpp
aleclearmind: This is not OK, you should use `ProgramCounterHandler::affectsPC`.
Can you get a reference to it here?

include/revng/TupleTree/TupleTreeReference.h
mrjackv: I tried, but apparently there are cases where `nullptr` is returned

include/revng/TupleTree/TupleTreeReference.h
mrjackv: You need it to say that you're initializing the value inline, as opposed to:
```
struct Foo {
  static int Bar;
}

Foo::Bar = 42;

// vs

struct Foo {
  static inline int Bar = 42;
}
```

include/revng/TupleTree/TupleTreeReference.h
aleclearmind: For sure it should never give you `CachingDisabled`.

include/revng/TupleTree/TupleTreeReference.h
mrjackv: It composes well with the callers, e.g.
```
return setCache(getByPath<T>(Path, *std::get<RootT *>(Root)));
```

Instead of:
```
T *Ptr = getByPath<T>(Path, *std::get<RootT *>(Root))
setCache(Ptr);
return Ptr;
```

include/revng/TupleTree/TupleTreeReference.h
mrjackv: Sure, but really, the policy of "we don't cache magic values" is much more robust, tomorrow we could have that `CachingDisabled = 42` and it would still work

include/revng/TupleTree/TupleTreeReference.h
mrjackv: Depends on [this](https://github.com/revng/revng/pull/560#discussion_r2841741205), if the function is `void` then the signature is much simpler and `ConstOrNot` can be used, otherwise you have to write a trailing type annotation and do dances with `decltype()`

include/revng/TupleTree/TupleTreeReference.h
aleclearmind: Rename to `cacheAndGet` or something like that.

include/revng/TupleTree/TupleTreeReference.h
aleclearmind: I don't get what you mean. We don't want to allow this value, let's not try to be flexible "just in case".
Just assert it's different.

lib/FunctionIsolation/IsolateFunctions.cpp
mrjackv: Done

include/revng/TupleTree/TupleTreeReference.h
mrjackv: Added and dropped the `requires`

include/revng/TupleTree/TupleTreeReference.h
mrjackv: Added the assert

include/revng/TupleTree/TupleTreeReference.h
mrjackv: Done

include/revng/TupleTree/TupleTreeReference.h
mrjackv: Done

include/revng/TupleTree/TupleTreeReference.h
mrjackv: Renamed

include/revng/TupleTree/TupleTreeReference.h
mrjackv: Reworked the explaination, adding that it's done for lazy caching.

PR #562, opened 2026-03-02
Clift based headers (part 1)

Cvelth:


pfez:
This is being merged.

lib/CliftEmitC/CBackend.cpp
```diff
@@ -94,8 +88,17 @@ class CliftToCEmitter : CEmitter {
       CInteger = CIntegerKind::Int;
     }
 
-    Tokens.emitIntegerLiteral(makeIntegerValue(Primitive, Value),
-                              *CInteger,
+    bool IsSigned = Primitive.getKind() == PrimitiveKind::SignedKind;
+    if (IsSigned and static_cast<int64_t>(Value) < 0) {
+      Tokens.emitOperator(ptml::CTokenEmitter::Operator::Minus);
+      Value = static_cast<uint64_t>(-static_cast<int64_t>(Value));
```
vasama-work: This negation has undefined behaviour for the signed minimum value. To solve, you can pick any one of these options:
```cpp
Value = ~Value + 1;
Value = ~(Value - 1);
Value = static_cast<uint64_t>(-(static_cast<int64_t>(Value) + 1)) + 1;
```

lib/PTML/CTokenEmitter.cpp
```diff
@@ -571,34 +572,37 @@ void CTokenEmitter::emitLiteralIdentifier(llvm::StringRef Identifier) {
...
-  revng_assert(IsValidRadix(Radix),
-               "The specified integer radix does not correspond to any "
-               "C integer literal token.");
-
   llvm::SmallString<64> String;
   Value.toString(String,
-                 Radix,
-                 /*Signed=*/Value.isSigned(),
-                 /*FormatAsCLiteral=*/true);
+                 verifyRadix(Radix),
```
vasama-work: Passing the radix through the verification function seems needlessly obfuscating. I would prefer to keep the assert, although converting the local lambda to a namespace scope function is fine by me.

vasama-work (changes requested):

lib/CliftEmitC/CBackend.cpp
Cvelth: Wait! Right! Good catch!

PR #564, opened 2026-03-05
Clift based headers (part 3: emitters and headers)

Cvelth:

PR #566, opened 2026-03-05
Clift based headers (part 5: extras)

Cvelth:


include/revng/Clift/CliftTypeInterfacesClass.td
```diff
@@ -17,6 +17,7 @@ def Clift_ClassType : Clift_TypeInterface<"ClassType", [Clift_DefinedType]> {
         return $_type.isComplete();
       }]
     >,
+
```
Cvelth, line 20: (attaching to an arbitrary place because otherwise github does not allow closing threads)

Let this thread serve as a reminder to not forget about this: https://github.com/revng/revng/pull/564#discussion_r3101094307

Cvelth: (done)

tests/filecheck/clift/backend/types.mlir
```diff
@@ -7,85 +7,85 @@
 
 !void = !clift.void
 
-!f = !clift.func<"f_t" as "f" : !void()>
-!s = !clift.struct<"s" as "s" : size(1) {}>
+!f = !clift.func<"/type-definition/0-CABIFunctionDefinition" as "f" : !void()>
+!s = !clift.struct<"/type-definition/1-StructDefinition" as "s" : size(1) {}>
```
pfez: why?

Cvelth: Because this IR is emitted as C.

And eventually emitting "actions" for invalid handles will become a hard error.

For some context, this is a left-over from the time I was asserting that. Since there is currently exactly one test triggering this, might as well fix it.

lib/CliftEmitC/CBackend.cpp
```diff
@@ -877,9 +877,9 @@ class CliftToCEmitter : CEmitter {
     if (mlir::isa<GotoOp>(S))
       Tokens.emitKeyword(CTE::Keyword::Goto);
     else if (mlir::isa<BreakToOp>(S))
-      Tokens.emitLiteralIdentifier("break_to");
+      Tokens.emitMacro("break_to");
     else if (mlir::isa<ContinueToOp>(S))
-      Tokens.emitLiteralIdentifier("continue_to");
+      Tokens.emitMacro("continue_to");
```
pfez: For these I want the same highlighting as regular `break` and `continue`.

include/revng/PTML/CTokenEmitter.h
```diff
@@ -187,6 +187,11 @@ class CTokenEmitter {
   /// \pre \param Identifier matches `[_a-zA-Z][_a-zA-Z0-9]*`.
   void emitMacro(llvm::StringRef Identifier);
 
+  // IMPORTANT: unwrapped PTML sections do not get *any* syntax highlighting.
+  //            They should only be used when no other options are available.
+  //            If nothing else, abusing `macro` is preferable to not wrapping
+  //            at all.
+
```
pfez: It's not clear to my why this comment is here. What code does it refer to? `emitMacro` above?

lib/ImportFromC/ImportFromCAnalysis.cpp
```diff
@@ -293,6 +293,9 @@ struct ImportFromCAnalysis {
 
     // Check if an error was reported by clang or revng during parsing of C
     // code.
+    //
+    // TODO: adjusting line numbers to account for the lines we append would
+    //       make UX considerably better.
```
pfez, line 288: Can you describe what's the current UX and how it is bad? I think this should be an easy fix.

tests/filecheck/clift/backend/types.mlir
pfez: > And eventually emitting "actions" for invalid handles will become a hard error.

We'll deal with this when that becomes a hard error, if ever.

It is a design goal to be able to emit C code from Clift that wasn't generated by revng, and whose handles are not generated by revng. So leave this as it was.

lib/ImportFromC/ImportFromC.cpp
```diff
@@ -275,14 +275,14 @@ DeclVisitor::makePrimitive(const BuiltinType *UnderlyingBuiltin,
...
+  case BuiltinType::Char_U:
+  case BuiltinType::UChar:
   case BuiltinType::Char8:
   case BuiltinType::Bool:
     return model::PrimitiveType::makeUnsigned(1);
 
-  case BuiltinType::Char_U:
-  case BuiltinType::UChar:
+  case BuiltinType::Char_S:
+  case BuiltinType::SChar:
```
pfez, line 340: Oh wow! Nice catch!

lib/CliftEmitC/CBackend.cpp
Cvelth: That means adding them explicitly as `keywords`.

I don't really mind it, but something tells me @vasama-work might. So let's give him a chance to convince you out of this!

P. S. This reminds me, are we banning these names in the `NameBuilder`. I need to check.

lib/ImportFromC/ImportFromC.cpp
```diff
@@ -2,7 +2,11 @@
 // This file is distributed under the MIT License. See LICENSE.md for details.
 //
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-anon-enum-enum-conversion"
 #include "clang/AST/Comment.h"
+#pragma clang diagnostic pop
+
```
pfez: @aleclearmind what's your take on this?

I see 2 options
1. accept `pragma clang diagnostic` `push`/`pop`/`ignore` around `include`s from external headers. AFAIK, we've never accepted this.
2. go and fix `clang/ADT/Comment.h` so that warning doesn't trigger anymore. This is the approach we've usually adopted with `llvm` for example. It should be easy to check if the problem is perhaps already fixed upstream and backport the commit. Otherwise even a local fix we maintain is fine. The only thing I'm not sure is if the commit fixing this should be included in `clang-release` or regular `llvm`.

Possibly, I'd go with 2. @Cvelth can you investigate how hard it is to change the code in `Comment.h` to ensure the warning doesn't trigger? Should be easy, no?

tests/filecheck/clift/backend/types.mlir
Cvelth: This is valid clift. But this IR explicitly goes through the `--emit-c=ptml`. And we cannot emit valid PTML without valid locations, can we?

include/revng/PTML/CTokenEmitter.h
Cvelth: Mostly as a preventative measure against the future instances of `emitLiteralIdentifier` I removed. It can probably be worded better though.

lib/ImportFromC/ImportFromCAnalysis.cpp
Cvelth: The line numbers in error messages we pass back to the user do not match the input.

lib/ImportFromC/ImportFromC.cpp
Cvelth: Sometimes adding random extra tests helps catch the most bizarre things!

lib/ImportFromC/ImportFromC.cpp
Cvelth: Let's hear @aleclearmind's take first, but I'd imagine the patch itself should be reasonable. I'm worried about the maintenance burden of rebasing it each time though. Not sure that is worth it to avoid a pragma.

lib/Model/NameBuilder.cpp
```diff
@@ -596,8 +597,35 @@ llvm::Error model::CNameBuilder::isNameReserved(llvm::StringRef Name) const {
...
+using CNB = model::CNameBuilder;
+bool CNB::shouldAllowThisTypedefNameAsAnException(const model::TypedefDefinition
+                                                    &Typedef,
+                                                  llvm::StringRef Name) const {
+  revng_assert(!Name.empty());
+
+  static constexpr std::array<llvm::StringRef, 2> AllowTypedefList = {
+    "size_t",
+    "ptrdiff_t",
+  };
```
pfez: @aleclearmind in our list of quickies we mentioned to dynamically unban `size_t`. Here @Cvelth has set up a way to have a list of names that can be dynamically unbanned.

The list is now composed of `size_t` and `ptrdiff_t`, even if our quick task only included `size_t`. What's your take on this?

I think that we either a) just keep `size_t` for now (and only add others on a per-need basis), or b) we add at least `ssize_t` (`size_t` and `ssize_t` are also POSIX, so they might show up quite often) or c) we also add `intptr_t` and `ptrdiff_t`.

I definitely don't  want to go beyond these 4 though. Otherwise we'll end up with all those `intmax` `intleast` `atomic` which I definitely want to keep out for now.

tests/filecheck/clift/comments/3-bonus/StandardTypedefs.mlir
```diff
@@ -0,0 +1,23 @@
...
+//
+
+// RUN: %root/bin/revng clift-opt --emit-type-and-global-header %s -o /dev/null | FileCheck %s
+// RUN: %root/bin/revng clift-opt --emit-type-and-global-header=ptml %s -o /dev/null | %root/bin/revng ptml | FileCheck %s
+
+!uint64_t = !clift.int<unsigned 8>
+!int64_t = !clift.int<signed 8>
+
+// CHECK: typedef uint64_t size_t;
+// CHECK: typedef int64_t ptrdiff_t;
```
pfez: Note to ensure we only check those we've decided to support.

lib/Model/NameBuilder.cpp
pfez: To clarify, I haven't looked into the logic of how these are dynamically unbanned yet, and I'm not asking you to do it @aleclearmind. I'm only asking we converge on a list of names we want to dynamically unban.

tests/filecheck/clift/backend/types.mlir
```diff
@@ -14,78 +14,78 @@ module attributes {clift.module} {
   // CHECK: void f(void) {
   clift.func @f<!f>() attributes {
     clift.handle = "/function/0x1000:Code_aarch64",
-    clift.name = "f"
+    name = "f"
```
pfez, line 17: Note to self to investigate this.

lib/Model/NameBuilder.cpp
Cvelth: note-to-self: rename this into `AllowedTypedefList` because grammar is hard apparently xD

lib/CliftEmitC/CBackend.cpp
pfez: @vasama-work will not convince me. This is just for highlighting. And we want to highlight them as keywords, even if they are not, because they need to stick out in the same way when you look at the code.

lib/ImportFromC/ImportFromCAnalysis.cpp
pfez: Should be easy to fix though, no?

lib/ImportFromC/ImportFromC.cpp
pfez: > I'm worried about the maintenance burden of rebasing it each time though.

Don't worry about that, I'm the one who's going to be doing the rebase anyway :) . This doesn't look like something painful to rebase. And it's the kind of change that tends to be well accepted upstream anyway. So this should not have any weight in the decision.

lib/ImportFromC/ImportFromCAnalysis.cpp
Cvelth: I'd assume so. I mostly didn't do it outright because I have no idea how the new daemon looks.

But, in hindsight, it probably doesn't matter. We just subtract the number of lines we inject and call it a day 🤔.

lib/ImportFromC/ImportFromC.cpp
Cvelth: Well, excuse me for not wanting to push extra work on future @pfez! /j

Jokes aside, I don't have a strong stance one way or the other.

lib/CliftEmitC/CBackend.cpp
vasama-work: It's fine by me.

lib/ImportFromC/ImportFromCAnalysis.cpp
mrjackv: If I may butt in, the best thing for the UI would be having clang emit SARIF (I tried to make it work but stopped because I got too sidetracked), subtract the lines, filter out the negative ones and then feed it back to the UI, because, at that point, I can put squiggly red lines on the text and that 1000% better.
However all the above is a bit of work, and requires some changes on the UI as well, but we should pursue it eventually.

Cvelth: Hmm, sounds to me like we want both. UI can definitely benefit from using SARIF, but I'm not sure CLI would 🤔.

Either way, we need line number fixes. So I guess tweaking that and leaving a better todo mentioning the SARIF idea is the best course of action for now.

lib/ImportFromC/ImportFromC.cpp
vasama-work: The usual solution is to use something like `-isystem` for external headers (I think there might have been another similar, but more appropriate option also.)

lib/Model/NameBuilder.cpp
aleclearmind: We can dynamically unban things, but the question is not the list of types, but more whether this in scope and *how* we do it.

I don't think this was in scope, but more importantly...

```c++
  if (Underlying.isSignedPrimitive() or Underlying.isUnsignedPrimitive())
    return true;
```

This is not what we meant by "dynamically unban". We need to query the target ABI and have it tell us the *exact* type `size_t` has. Here we're just checking whether is a signed primitive or an unsigned primitive, which looks quite arbitrary.

Cvelth: Ah, so you want to be even more strict about it? If anything I would prefer that too.

It would, I guess, mean a dependency between the name builder and the data-model. Except the data-model is not even aware of `size_t`'s existence currently! So I guess this is not as trivial as I hoped :)

pfez: I think that we need to think this through and desing it properly, and not rush it in so quickly. @Cvelth drop the commit from this PR and consider this suspended until I don't come back to you with a more thorough design. We're already tracking in our roadmap so there's no need to keep this into your backlog either.

Cvelth: Sounds good!

pfez: Then close this thread when you're done.

lib/ImportFromC/ImportFromCAnalysis.cpp
pfez: Yes. Leave the TODO in place for now and mention SARIF as well. Let's not fix it for now.

tests/filecheck/clift/backend/types.mlir
Cvelth: (ping)

lib/CliftEmitC/CBackend.cpp
Cvelth: Done.

lib/ImportFromC/ImportFromCAnalysis.cpp
Cvelth: Please take a look. It's in `[new] import-from-c: correct error line numbers`.

include/revng/PTML/CTokenEmitter.h
Cvelth: (ping)

lib/ImportFromC/ImportFromC.cpp
Cvelth: (this can be postponed until the next PR, let's keep this thread open until I open one there)

tests/filecheck/clift/comments/3-bonus/StandardTypedefs.mlir
Cvelth: (this is postponed, and I have an entry in my notes, so I'm closing this)

tests/filecheck/clift/backend/types.mlir
Cvelth: I adjusted the commit to also strip `clift.` from the handle. As well as fixed the commit message.

lib/ImportFromC/ImportFromC.cpp
pfez: I've got a one-line fix in clang for this. Drop the `pragma clang diagnostic` from here.

Cvelth: Ah, great, no need for this commit then :tada:

pfez: Take it from here if you want https://github.com/revng/llvm-project/pull/51

I turned it into a PR so that if this ends up being merged by ale next week he has a clear reference.

lib/CliftEmitC/CBackend.cpp
```diff
@@ -797,15 +797,31 @@ class CliftToCEmitter : CEmitter {
...
+      .Identifier = S.getName(),
+      .Location = S.getHandle(),
+      .CAttributes = getDeclarationOpCAttributes(S),
+      .Kind = CTE::EntityKind::LocalVariable,
+    };
+
+    auto Type = S.getResult().getType();
+    auto Class = mlir::dyn_cast<ClassType>(Type);
+    if (Configuration.InlineStackFrame and isStackFrameVariable(S) and Class
+        and Class.getHandle() != Configuration.TypeToOmit) {
```
vasama-work: Is this condition ever going to be hit? Isn't `TypeToOmit` only relevant when emitting types for editing or something?

tests/filecheck/clift/backend/stack.mlir
```diff
@@ -0,0 +1,53 @@
...
+    %s = clift.local : !stack attributes {
+      handle = "/stack-frame-variable/0x1000:Code_aarch64",
+      name = "var_0"
+    }
+
+    clift.expr {
+      clift.yield %s : !stack
+    }
+  }
+}
```
pfez: What happens in case the stack frame type has a field whose type is:
1. a pointer type
2. another struct type
3. an array of non-pointer types
4. an array of pointers.

lib/CliftEmitC/TypeDefinitionEmitter.cpp
```diff
@@ -148,7 +148,9 @@ void TypeDefinitionEmitter::emitPaddingField(clift::ClassType Class,
   Tokens.emitNewline();
 }
 
-void TypeDefinitionEmitter::emitClassDefinition(clift::ClassType Class) {
+using TDE = TypeDefinitionEmitter;
+void TDE::emitClassDefinition(clift::ClassType Class,
+                              CEmitter::DeclaratorInfo const *Declarator) {
```
vasama-work: I don't like taking a declarator here. This function is doing too much. It would be better to move the emission of the semicolon out and let the function backend compose the stack variable declarator. I don't imagine we're going to need to emit declarators of this form elsewhere, and if we are, then we'll probably need it for more than just classes.

lib/CliftEmitC/CBackend.cpp
```diff
@@ -797,15 +797,31 @@ class CliftToCEmitter : CEmitter {
 
   //===---------------------------- Statements ----------------------------===//
 
+  static bool isStackFrameVariable(LocalVariableOp L) {
+    return pipeline::locationFromString(revng::ranks::StackFrameVariable,
+                                        L.getHandle())
+      .has_value();
+  }
+
```
vasama-work: I don't really find this very appropriate... So far we've wanted to avoid looking at the handles for information. I think I would prefer to set some guidance attribute on the stack frame variable in `ImportDescriptiveInfo` instead. Like `UnitAttr` `clift.inline_type_definition` on the local variable op that tells the backend to emit a full definition.

lib/CliftEmitC/TypeDefinitionEmitter.cpp
pfez: > It would be better to move the emission of the semicolon out and let the function backend compose the stack variable declarator. I don't imagine we're going to need to emit declarators of this form elsewhere.

I agree.

lib/CliftEmitC/CBackend.cpp
pfez: My understanding is that it should never be hit. Also if `isStackFrameVariable(S)` is true, my understanding is that `Class` should always be guaranteed to be true.

If these things go wrong, the result is that we're emitting something broken anyway. I presume the extra checks are just a way to be kind with the UX and degrade gracefully is something very weird ends up here, which isn't necessarily wrong.

In this way you get a user report like "I don't see the inline struct I expected" instead of a crash report. But for our tests it would be much better to have these as assertions. I think it would be good guard this with an `if (VerifyLog.isEnabled())` so that if we hit it in the tests it never goes silent, but users still get nice gracefully degrading UX.

include/revng/CliftEmitC/CEmitter.h
```diff
@@ -40,10 +40,12 @@ class CEmitter {
   void emitCAttribute(clift::CAttributeAttr Attribute);
   void emitCAttributes(llvm::ArrayRef<clift::CAttributeAttr> Attributes,
                        bool SpaceBefore,
-                       bool SpaceAfter);
+                       bool SpaceAfter,
+                       bool NewlineAfter = false);
```
pfez, line 44: Commenting here on the commit message. I understand you have second thoughts, this is a clearly subjective preference. But let's not introduce another parameter for that for now. 99% of the functions will only have the `ABI` attribute, and for the others I consider having one attribute per line a saner default.

If external or internal need arises for this, we can still add a parameter later. Not now.

tests/filecheck/clift/backend/stack.mlir
Cvelth: Currently nothing special, but as I already mentioned on mattermost, it's a matter of adding an `if` for each of those. Once we have the complete list compiled.

lib/CliftEmitC/CBackend.cpp
pfez: This is the thing that worries me the most about this commit as well. On its own it's probably the main reason to pull out this commit on stack frame inlining from this PR.

The proper way to do this would be to have an attribute on the local variable op yes. And then we need a model-aware pipe that runs late before the backend that uses a location-based logic to add these attributes, so that the backend can stay agnostic.

Probably something we can do later.

lib/CliftEmitC/CBackend.cpp
Cvelth: > Isn't `TypeToOmit` only relevant when emitting types for editing or something?

Exactly. That's why I'm just doing it as a basic sanity check here. In case we re-use `TypeToOmit` for something else in the future or whatever. Obviously it should never happen. Maybe it should even be an assert. It would have been an error if we already had a system for reporting those :P

> Also if `isStackFrameVariable(S)` is true, my understanding is that Class should always be guaranteed to be true.

Sure, but that's an external condition. The idea behind the `and Class` condition here is very basic check to make sure `Class.getHandle()` never asserts! :)

> `if (VerifyLog.isEnabled())`

Yeah, but that's for expensive checks. Comparing two strings is hardly a big performance degradation.

lib/CliftEmitC/TypeDefinitionEmitter.cpp
Cvelth: Sure!

lib/CliftEmitC/CBackend.cpp
Cvelth: Yes, I 100% agree that abusing locations like this is ugly.

On the other hand, I also find attributes dedicated to something like this to be pretty awkward. Suddenly we have multiple places stating the same thing. What if they are conflicting? I'd argue if we introduce the attribute like that, we should drop the `StackFrameVariable` location. One way or the other, this deserves more attention. Good thing we're postponing this.

include/revng/CliftEmitC/CEmitter.h
Cvelth: This function is used for *all* the attributes. We need to differentiate the ones on functions (with newlines) against those on other things (without).

lib/CliftEmitC/CBackend.cpp
pfez: We've decided to pull the commits related to stack frame inlining out of this PR.

I'm not closing this thread so it's easier to find again later when we revive this effort. But I consider the discussion closed here for now.

lib/CliftEmitC/TypeDefinitionEmitter.cpp
pfez: We've decided to pull the commits related to stack frame inlining out of this PR.

I'm not closing this thread so it's easier to find again later when we revive this effort. But I consider the discussion closed here for now.

lib/CliftEmitC/CBackend.cpp
pfez: We've decided to pull the commits related to stack frame inlining out of this PR.

I'm not closing this thread so it's easier to find again later when we revive this effort. But I consider the discussion closed here for now.

tests/filecheck/clift/backend/stack.mlir
pfez: We've decided to pull the commits related to stack frame inlining out of this PR.

I'm not closing this thread so it's easier to find again later when we revive this effort. But I consider the discussion closed here for now.

tests/filecheck/clift/backend/types.mlir
pfez: Depends what you mean by "valid". The backend should be agnostic w.r.t. locations. They are supposed to just be opaque strings for what the backend is concerned.

Cvelth: Clift-side - sure. PTML-side - I'd argue that it's impossible to be location-agnostic.

> Depends what you mean by "valid".

"Valid" as in "valid PTML, UI would be able to make sense of".

include/revng/PTML/CTokenEmitter.h
pfez: I see, but in its current form and location this is informative only to you. It's not clear what it means nor what code it refers to. It's just confusing and causes FUD without explaining what's it's FUDding about. No other part of the codebase seems to refer to "unwrapped PTML sections", so I guess this term refers to some concept you're implicitly relying upon from your old implementation effort, and you've just put this comment down for your piece of mind. Just drop it.

Cvelth: That's why review is useful. It's not always easy to write code/notes/docs/etc that can be easily understood by other people :D

---

I'd still prefer we kept _something_, what about
```cpp
// IMPORTANT: If you are about to add a new method in here, please make sure
//            what it emits is wrapped into a ptml `<span>` tag with:
//            - `data-token` set (determines syntax highlighting),
//            - either `data-location-definition` or `data-location-references` set
//              (ensures correct ctrl+click behavior),
//            - `data-allowed-actions` and `data-action-context-location` set
//              iff it is supposed to be actionable (ensures user interactivity).
```
?

lib/ImportFromC/ImportFromCAnalysis.cpp
pfez: I decided to pull this out of the PR as well.

I think that the right way to fix this is simply to use the virtual filesystem machinery available to clang tooling. But this should be investigated more thoroughly, since I'm not super familiar with that.

I think that if we use that properly we can actually have a header file that is separate from the C file, and the line numbers are just correct from the get go, without having to mess around to adjust them, and without using SARIF.

pfez: Please change the TODO to reflect this. And pull the commit out of the PR.

Cvelth: I wonder if that would work 🤔.

Note that _both_ c source and the header are _ours!_

And the source file will have to include the header to use names declared in it. And the user does not need to be aware of that header's existence.

Cvelth: Also, SARIF is not because it makes line number adjusting easier. It's preferable because UI can actually easily parse it and act on the errors in more predictable ways (think squigglies).

pfez: About SARIF, I understand, but it's a sign that we're dealing with two problems at the same time here, which doesn't make me confident to rush this in.

pfez: About using the virtual filesystem. I understand your doubts, but I'm pretty sure it would work.

The fact that both the `.c` and `.h` are ours is a non issue. If anything, it makes things even simpler: we emit the header, and put it in the VFS, then we take the user-provided `.c`, prepend a single line that includes the header using the name we've picked in the VFS, and then we're done (modulo some magic to adjust for the single line we've added for the `include` directive, but in the UI I'm pretty sure we can handle that without the user knowing and without parsing any error message). Then all error messages have the correct line number. Just like that, independently of SARIF vs textual output. 

> the user does not need to be aware of that header's existence.

True, until there are no errors. If there's an error, the user, or us for debug, might be interested to see the header. This is another things that needs to be thought through thoroughly though. And it's another reason for not rushing this in.

include/revng/PTML/CTokenEmitter.h
pfez: Much better. I'd put this on top of the data structure though, not floating through in the middle.

Also, there's no mention to abusing `macro` as the old comment suggested. I believe that's on purpose, but I want to double check.

tests/filecheck/clift/backend/types.mlir
pfez: What UI? The backend is not supposed to assume any UI.

pfez: In a nutshell, if locations are broken, the backend should emit them as broken. It's the responsiblity of whoever emits those locations to shape them into a form that the UI they want to use recognizes. The backend just funnels them through.

include/revng/CliftEmitC/CEmitter.h
pfez: Still, the only place passing `true` at the moment is when we function attributes. Let's leave it like this. The very moment when some different part of the code starts asking to emit newlines AND they want a single newline at the end of all attributes, then we'll worry about it.

tests/filecheck/clift/backend/types.mlir
pfez: I'm puzzled. There's a commit where you switch from `name` to `clift.name` and then here back to `name`. If you have to switch back, just don't change it do begin with.

lib/ImportFromC/ImportFromCAnalysis.cpp
Cvelth: So, we still generate the very same c file. And the very same header. We still arrange them in the exact same way. 🤔 And we still manually fix the line numbers (for some context, the current offset is also 1).

I don't see any difference between using the virtual file system vs the real one. Delta maybe performance. But we are compiling *one* tiny file. Doubt it's going to matter much.

---

I guess we can add an artifact for that header too. But, if anything, _that_ feels excessive. We _for sure_ have bigger fish to fry!

tests/filecheck/clift/backend/types.mlir
Cvelth: From what I understand, this has been due to rebases. We had two branches in flight at the same time by different, one of which changed behavior around these.

One way or the other, this commit instills the correct format, unless I'm misunderstanding something.

tests/filecheck/clift/backend/types.mlir
Cvelth: How can the backend emit valid PTML if it doesn't want to _know_ what counts as a valid PTML or not?

Note that the backend is the last stage of generation. Unless we want an additional "verifier" (which I doubt), the backend should be the one ensuring stuff we emit actually makes sense.

The same way we ensure whatever we emit in the `plain` mode is valid C, it's equally important that `ptml` mode produces valid ptml.

include/revng/CliftEmitC/CEmitter.h
Cvelth: Ah, wait, I'm dumb!!!! There's a misunderstanding

What I meant by the commit message is actually whether we really want
```cpp
_ABI(whatever)
_ALWAYS_INLINE
_NORETURN
void whatever(void);
```
as we decided during our last call, as opposed to

```cpp
_ABI(whatever) _ALWAYS_INLINE _NORETURN
void whatever(void);
```
which is closer to what the old backend would have emitted.

include/revng/PTML/CTokenEmitter.h
Cvelth: Yeah, I got a bit of FUD about opening the flood-gates to everything becoming a macro :P

Also, it didn't fit as well with this higher-level overview.

Maybe it's worth adding two more paragraphs, about the more practical side of this, something like:
```cpp
// Practically, you should use:
// - `PTML.initializeOpenTag(ptml::tags::Span)` for the tag,
// - `CTokenEmitter::getEntityKindAttribute` to select the `data-token` attribute,
// - `CTokenEmitter::getAllowedActions` to select the `data-allowed-actions`.
//
// Also, take a look at `emitMacro`, odds are it'll be able to emit what you
// need. At the very least, it's sufficient temporarily until you figure out
// the exact way you want your token formatted.
```

what do you think?

pfez: Ok for adding the first paragraph.

No for the second. Let's keep the flood-gates closed.

tests/filecheck/clift/backend/types.mlir
Cvelth: TL;DR: one was correct, then another one became. All while I had a too big of a branch in flight to keep track of everything :D So this file ended up being locked in while broken.

tests/filecheck/clift/backend/types.mlir
pfez: Valid PTML is a thing, and the emitters already ensure that in terms of matching open and closed tags. Valid PTML for revng-specific UI and locations is another thing, that the backend is not supposed to assume, or check. Otherwise we lose the capability of emitting C code from clift that doesn't have revng-conforming locations. Here you're assuming that handles are revng-specific ptml locations. I want the backend to be capable of emitting C code if locations are random string the user made up for their own purposes, or even placeholders, like they originally were in this test. Of course is someone uses our backend to emit C code starting form non-revng-conforming location, it will not work in the UI, but that's what you opt out for when you decide to hand-code your locations and go out of our PTML reference for locations.

lib/ImportFromC/ImportFromCAnalysis.cpp
pfez: > So, we still generate the very same c file. And the very same header. We still arrange them in the exact same way. 🤔 And we still manually fix the line numbers

More or less, except that the clang VFS machinery takes care of the platform specific part of the newlines I guess, so we have less chances for subtle errors.

And it clearly decouples line numbering from choosing SARIF output I think, which is not evident to me it was such a clear cut in the other way.

> I guess we can add an artifact for that header too.

Let's not put the cart ahead of the cows (if that's even a thing you can say in english). This nuances are part of why I want to pull this out of the PR. Let's pull this commit out.

As other threads in this PR, I'll leave this open even if I consider the discussion closed, so it's easier to find it when we revive the discussion later for actually implementing this.

tests/filecheck/clift/backend/types.mlir
Cvelth: Wait, I think there's a misunderstanding. The backend will always be able to emit *C* no matter what the handles are.

But *wrapping* that C into tags correctly (more than just putting "tokens" into spans without any regard of what those tokens are) means being aware of PTML specifics. There's no way around that.

Sure we can do "if this is an unknown location, do the generic inoffensive thing", but that's probably already covering bugs. And will cover more bugs in the future!

---

As for this test, if we just remove the `--emit-c=ptml` variant of it (which, let's be honest, doesn't really check anything), it can stay as it is!

include/revng/PTML/CTokenEmitter.h
Cvelth: Done.

include/revng/CliftEmitC/CEmitter.h
pfez: Yes, this was clear to me since the beginning.

I'm ok with the first form you posted, and I believe that's clear in hindsight if you re-read this thread.

My idea is just that `_ALWAYS_INLINE` and `_NORETURN` are rare, and when they are present the default it that they should stand out.

What the old backend did around this was just the first thing that came to my mind when I wrote that, nothing that deserves any great credit.

Are we on the same page?

Cvelth: Sure, closing the thread then! I was just trying to make _sure_-sure we're on the same page. Because, frankly, I find the first one ugly. But, as you said, it's rare enough to not really matter.

tests/filecheck/clift/backend/types.mlir
pfez: What I'm saying is that *in this PR* you have a commit changing `name` to `clift.name` ("lit-tests: only use real handles") and another commit then switching it back to `name` ("lit-tests: drop `clift.` in function attributes"). don't do this. if `name` is correct, just leave it alone to begin with.

Cvelth: Fuck, the rebase result I was thinking about was not merged yet 🤦.

Thanks for catching that!

tests/filecheck/clift/backend/types.mlir
Cvelth: (gone)

tests/filecheck/clift/backend/types.mlir
Cvelth: Fixed, my bad!

PR #567, opened 2026-03-05
Clift based headers (part 6: extras)

Cvelth:


Cvelth:
(everything we decided to keep from this PR is already in `develop`, so I'm closing this one)

PR #568, opened 2026-03-05
Integrate UI with pypeline

mrjackv:


typescript/mass-testing-report/src/main.ts
```diff
@@ -230,7 +230,8 @@ function populateNav(element: HTMLElement) {
 // showHours=false => MM:ss.mmm
 // showHours=true  => hh:MM:ss.mmm
 function formatTime(elapsedSeconds: number, showHours: boolean): string {
```
aleclearmind, line 232: What is this used for?

python/CMakeLists.txt
```diff
@@ -468,6 +468,11 @@ python_module(
   MODULE_FILES
   ${REVNG_PYPELINE_MODULE_FILES})
 
+configure_file(
+  revng/pypeline/pipeline-description-schema.yml
+  "${CMAKE_BINARY_DIR}/share/revng/pipeline/pipeline-description-schema.yml"
+  COPYONLY)
```
aleclearmind, line 477: Can't this be in `share/revng/pipeline/pipeline-description-schema.yml` in the source directory as well?

python/revng/pypeline/daemon/daemon.py
```diff
@@ -53,18 +53,26 @@ def get_pipeline_description(pipeline: Pipeline) -> dict[str, Any]:
     Build and validate the web representation of the pipeline.
     """
 
+    model_type: type[Model] = get_singleton(Model)  # type: ignore [type-abstract]
     # Build the pipeline description
     pipeline_description = {
         "version": revng.pypeline.__version__,
+        "model": {
+            "type": model_type.__module__ + "." + model_type.__qualname__,
```
aleclearmind: Mmmh, why are we exposing the name of a Python type? Doesn't sound right.

tests/pypeline/python/daemon/json_daemon.py
```diff
@@ -86,7 +89,12 @@ def get_model(self) -> Response:
 
     def run_analysis(self, analysis_request) -> Response:
         analysis_request.setdefault("project_id", self.project_id)
-        response = asyncio.run(self.daemon.analyze(analysis_request))
+        try:
+            response = asyncio.run(self.daemon.analyze(analysis_request))
+        except DaemonException as e:
+            response = DaemonResponse(e.code, e.body)
+        except PypelineException as e:
+            response = DaemonResponse(500, {"message": str(e)})
```
aleclearmind: What's the ratio to distinguish these two?

python/revng/pypeline/daemon/app.py
```diff
@@ -127,6 +139,7 @@ async def analysis_endpoint(request: Request, data: dict) -> Response:
         Route("/api/epoch", epoch_endpoint, methods=["GET"]),
         Route("/api/pipeline", pipeline_endpoint, methods=["GET"]),
         Route("/api/model", model_endpoint, methods=["GET"]),
+        Route("/api/put_file", put_file_endpoint, methods=["POST"]),
```
aleclearmind: Kebab case

python/revng/pypeline/cli/common_options.py
```diff
@@ -124,3 +124,19 @@ def _parse_debug_option(path: str, ctx):
     default=EagerParsedPath.DEFAULT,
     show_default=False,
 )
+
+
+class _ShowHiddenArtifactsParameter(click.ParamType):
+    def convert(self, value, param, ctx):
+        if value:
+            ctx.obj["show_hidden_artifacts"] = True
```
aleclearmind: `obj` seems to be a dictionary, what about a dataclass?

python/revng/pypeline/cli/utils.py
```diff
@@ -305,3 +307,18 @@ def compute_objects(
                 objects={obj_id_type.deserialize(obj) for obj in objects.split(",") if obj.strip()},
             )
     return model.all_objects(kind)
+
+
+def get_current_root_name(ctx: click.Context) -> str:
```
aleclearmind: What is a `root` in this context? Docs.

python/revng/internal/pipeline.yml
```diff
@@ -2,6 +2,18 @@
...
+  # User-facing artifacts that the end-user is expected to see
+  - name: user
+    show_by_default: true
+  # Artifacts that the user is not expected to see but machine users
+  # (e.g. scripting) are expected to use
+  - name: internal
+    show_by_default: false
+  # Debugging artifacts, these are for debugging purposes and/or testing
+  - name: dev
+    show_by_default: false
```
aleclearmind: 1. Is it possible to have an artifact belonging to no category?
2. Did we agree on the naming? Claude suggests

   primary / auxiliary / diagnostic
   output / intermediate / debug

   What do you think?

python/revng/internal/pipeline.yml
```diff
@@ -171,6 +173,7 @@ branches:
           - name: enforce-abi
             container: llvm-functions
             category: dev
+            single_object_filename: abi_enforced.bc
```
aleclearmind: 1. Just `filename`.
2. Kebab-cases everywhere.

python/revng/pypeline/pipeline-description-schema.yml
```diff
@@ -178,6 +180,18 @@ properties:
...
           description: The filename to use when representing this artifact
+        defined_locations:
+          type: array
+          description: PTML locations that are present in this artifact
+          items:
+            type: string
+        preferred_artifacts:
+          type: array
+          description: >
+            other artifacts where it's preferred to look for definitions of
```
aleclearmind: Capital letter?

python/revng/pypeline/pipeline-description-schema.yml
```diff
@@ -178,6 +180,18 @@ properties:
...
+            locations defined in this one
```
aleclearmind: `s/defined/referenced/`

python/revng/pypeline/schema.yml
```diff
@@ -199,3 +199,15 @@ $defs:
...
+        description: PTML locations that are present in this artifact
+        items:
+          type: string
+      preferred_artifacts:
+        type: array
+        description: >
+          other artifacts where it's preferred to look for definitions of
+          locations defined in this one
+        items:
+          type: string
```
aleclearmind, line 213: Are we OK with this duplication?

python/revng/pypeline/daemon/daemon.py
```diff
@@ -158,14 +161,38 @@ async def put_file(self, request) -> Response:
...
 
         # Validate data
-        for artifact_name, _ in artifacts.items():
-            if artifact_name not in self.pipeline.artifacts:
-                raise MalformedRequestError(f"Artifact {artifact_name} not found in the pipeline")
+        if artifact_name not in self.pipeline.artifacts:
+            raise MalformedRequestError(f"Artifact {artifact_name} not found in the pipeline")
+
+        if format_ not in ("json", "tar"):
+            raise MalformedRequestError(f"Format {format_} is not valid, valid values: json, tar")
```
aleclearmind, line 129: I think it'd be cool to use the [`Accept` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept).

python/revng/pypeline/pipeline-description-schema.yml
```diff
@@ -165,6 +166,11 @@ properties:
         cacheable:
           type: boolean
           description: Whether this artifact can be cached
+        pipe_dependencies:
+          type: array
+          description: List of pipes that this artifacts depends on
```
aleclearmind: 1. Add "since the start of the pipeline".
2. Explain when this is needed in the description

python/revng/pypeline/pipeline.py
```diff
@@ -66,12 +66,18 @@ def is_cacheable(self) -> bool:
         """An artifact is cacheable if it's backed by a savepoint."""
         return isinstance(self.node.task, SavePoint)
 
+    def pipe_dependencies(self) -> list[str]:
+        if not self.is_cacheable():
```
aleclearmind: Comment

python/revng/pypeline/pipeline.py
```diff
@@ -189,12 +189,14 @@ def __init__(
...
         for name, artifact in self.artifacts.items():
-            self.savepoint_id_to_artifact[artifact.node.id] = artifact
             if name != artifact.name:
                 raise ValueError(
                     f"Artifact name {artifact.name} does not match the key "
                     f"{name} in the artifacts map."
                 )
+            if isinstance(artifact.node.task, SavePoint):
+                assert artifact.node.savepoint_range is not None
+                self.savepoint_id_to_artifact[artifact.node.savepoint_range.start] = artifact
```
aleclearmind, line 196: Can you elaborate on the reason for this change?

python/revng/pypeline/daemon/app.py
```diff
@@ -50,19 +56,34 @@ def get_project_id(headers: Mapping[str, str]) -> str | None:
...
-        await subscriber.listen_for_messages()
+        _, pending = await asyncio.wait(
+            (
+                asyncio.create_task(shutdown_begun.wait()),
+                asyncio.create_task(subscriber.listen_for_messages()),
+            ),
+            return_when=asyncio.FIRST_COMPLETED,
+        )
+    except WebSocketDisconnect:
+        pass
```
aleclearmind, line 75: Log?

python/revng/pypeline/daemon/app.py
```diff
@@ -167,6 +188,12 @@ async def analysis_endpoint(request: Request, data: dict) -> Response:
     if "REVNG_EXPOSE_HEADERS" in os.environ:
         expose_headers.extend(os.environ["REVNG_EXPOSE_HEADERS"].split(","))
 
+    @asynccontextmanager
+    async def lifespan(app):
+        global shutdown_begun
+        shutdown_begun = asyncio.Event()
+        yield
```
aleclearmind, line 195: So this is a fancy way of doing initialization/destruction code?

typescript/mass-testing-report/src/main.ts
mrjackv: The tables in mass testing, e.g. https://mass.rev.ng/crashes.html

python/CMakeLists.txt
mrjackv: It can, but it's impractical for 3 reasons:
* The pypeline does not have a logic related to relative directories (e.g. `get_root` in `revng.support`)
* The `share/revng` subdirectory is not correct w.r.t. pypeline which should be standalone
* The pythonic version of doing things is to have the file within the python package. As we want pypeline to be as standalone as possible making it depend on a file installed under `share` is wrong

python/revng/pypeline/daemon/daemon.py
mrjackv: Eh it's something, at least clients can assert that the model is `revng.internal._pipebox.Model` instead of `foo.custom.Model`

tests/pypeline/python/daemon/json_daemon.py
mrjackv: `DaemonException`: we are aware we're in a HTTP-ish environment, so we return json-like output and a HTTP return code
`PypelineException`: this is the generic exception that's thrown from under `revng.pypeline`. It knows nothing about HTTP, json or status codes

python/revng/pypeline/cli/common_options.py
mrjackv: It can be done, however it's kind of distributed, in the sense that each `ParamType` puts its own keys into `object`. You can do it centrally but I think it's detrimental.

python/revng/pypeline/cli/utils.py
mrjackv: it's the click-compliant way of retrieving the equivalent of `sys.argv[0]`, so for instance a `revng2 ...` command this function will return `"revng2"`

python/revng/internal/pipeline.yml
mrjackv: 1. Yes
2. We agreed on `user/dev` but I wanted something in-between for things like `cross-relations.yml` where it's not user but e.g. revng-ui depends on it

python/revng/pypeline/schema.yml
mrjackv: Do you mean about the `items: { type: string }` part?

python/revng/pypeline/daemon/daemon.py
mrjackv: Let's please try to take a clear stance on the matter, we do all the jump through hoops to have _less_ HTTP semantics in the daemon, as far as not having it paired with the HTTP server, so that we can run some tests. Now we're pushing to re-introduce HTTP semantics. I don't think this is a good way to handle things.

python/revng/pypeline/pipeline.py
mrjackv: The previous way was wrong period. `artifact.node.id` is a auto-incremented number that allows referring to a node in an unique way in the pipeline. `artifact.node.savepoint_range.start` refers to another number that refers to the visit of savepoints and only increments when a savepoint is encountered.

python/revng/pypeline/daemon/app.py
mrjackv: I don't think that a websocket disconnecting is grounds for logging, it's pretty normal for it to happen

python/revng/pypeline/daemon/app.py
mrjackv: Yes, related to ASGI, in the sense that:
* When the function is called the server has finished initializing, but hasn't started accepted requests
* You run your code and when when you `yield` the server starts accepting requests
* The `yield` returns once all connections from the server are closed and the server is shutting down

python/revng/pypeline/daemon/daemon.py
aleclearmind: Add some identifier to the pipebox or the model.
We're not exposing a Python implementation detail.

tests/pypeline/python/daemon/json_daemon.py
aleclearmind: We need to pay attention `PypelineException` does not contain sensitive data.

Maybe we should have some daemon part catching `PypelineException`s and turning them into a `DaemonException`s?

python/revng/pypeline/cli/common_options.py
aleclearmind: How many of these do we have?
I'd argue that if they start to grow and centralization look bad we can have some registry.

The current thing is subject to typos.

python/revng/pypeline/cli/utils.py
aleclearmind: `get_root_command_name`?

python/revng/internal/pipeline.yml
aleclearmind: 1. I think we should force all artifacts to belong to a category or drop one category and make it default one.
2. `user` means "UI", but we other "final" artifacts such as "cleanup-ir".
   `cross-relations.yml` is more like `metadata` or `support` or `summary`.
   `dev` is clear but I'd call it `debug`.

   I propose:

   ```
   user: decompiled code, call graph
   metadata: cross-relations.yml 
   automation: cleanup-ir
   debug: others
   ```

   What do you think?

python/revng/pypeline/schema.yml
aleclearmind: `git grep 'PTML locations that are present in this artifact'`

It's in two files.

tests/pypeline/python/daemon/json_daemon.py
mrjackv: Eh, but the entire point of `DaemonException`s is to then return an http error response, which is done exactly in the snippet above

python/revng/pypeline/schema.yml
mrjackv: Ah, yes, it's present in `pipeline-description-schema.yml` and `schema.yml`. Unfortunately there is some overlap but they also share differences, so it's very impractical to merge them.

python/revng/pypeline/daemon/daemon.py
aleclearmind: Let me reformulate: `Daemon` is agnostic to the protocol, but we can make it work with MIME types (not HTTP-specific). Then, the starlette daemone fetches this from the `Accept:` header.

Sounds reasonable?

python/revng/pypeline/cli/common_options.py
mrjackv: * `base_directory`
* `cache_dir`
* `pipebox`
* `pipebox_args`
* `pipebox_path`
* `pipeline`
* `show_hidden_artifacts`
* `storage_provider`
* `wrapper`

python/revng/internal/pipeline.yml
mrjackv: 1. Just to rephrase to make things clear:
   * The pipeline must define at least one category
   * All artifacts must belong to a category
   Right?
2. Looks good

python/revng/pypeline/daemon/daemon.py
mrjackv: Hmmm, maybe we should talk about this IRL

tests/pypeline/python/daemon/json_daemon.py
aleclearmind: I'm thinking we could wrap `self.pipeline.run_analysis` into a `try/except` block turning `PypelineException` into `DaemonException`s.

But no biggie.

python/revng/pypeline/cli/common_options.py
aleclearmind: I doesn't seem too bad to have all of this in a single place.

python/revng/internal/pipeline.yml
aleclearmind: 1. Sounds good.
2. OK.

python/revng/pypeline/daemon/daemon.py
mrjackv: Added `.identifier`, the model reports this with `"revng"`

tests/pypeline/python/daemon/json_daemon.py
mrjackv: Reworked by using a decorator, this simplifies things

python/revng/pypeline/daemon/app.py
mrjackv: Done

python/revng/pypeline/cli/common_options.py
mrjackv: Implemented in "pypeline/cli: use dataclass for `ctx.obj`"

python/revng/pypeline/cli/utils.py
mrjackv: Renamed

python/revng/internal/pipeline.yml
mrjackv: Implemented

python/revng/pypeline/pipeline-description-schema.yml
mrjackv: Fixed both

python/revng/pypeline/pipeline.py
mrjackv: Re-looking at the code the `if` was incorrect, not returning pipe dependencies means that the client cannot correctly cache the artifacts locally. I've dropped it and unconditionally return the list.

python/revng/pypeline/pipeline-description-schema.yml
mrjackv: Fixed

python/revng/internal/pipeline.yml
mrjackv: Done

python/revng/pypeline/pipeline-description-schema.yml
mrjackv: Fixed

python/revng/pypeline/daemon/daemon.py
mrjackv: As discussed, we're deferring this at a later time

python/revng/pypeline/cli/context.py
```diff
@@ -0,0 +1,87 @@
...
+    from revng.pypeline.cli.wrappers import Wrapper
+    from revng.pypeline.pipeline import Pipeline
+
+
+# Placeholder value for an uninitialized field of `ContextObject`
+_MISSING = object()
+
+
+@dataclass(kw_only=True)
+class ContextObject:
```
aleclearmind, line 24: Docs for each field

mrjackv: Done

aleclearmind:
Merged in f75a122. Thanks.

PR #571, opened 2026-03-11
Integrate scripting with pypeline

mrjackv:
Sister PRs:
* [orchestra](https://github.com/revng/orchestra/pull/198)
* [revng-ui](https://rev.ng/gitlab/revng-private/revng-ui/-/merge_requests/51)


python/revng/pypeline/storage/local_provider.py
```diff
@@ -651,8 +651,9 @@ def _find_file(self, request: FileRequest) -> Path:
                 if not path_path.is_absolute():
                     path_path = (self._model_directory / path_path).resolve()
 
-                if path_path.is_file():
-                    found_paths.append(path_path)
+                with suppress(OSError):
```
aleclearmind: Emit a debug message

python/revng/pypeline/cli/project/analyze.py
```diff
@@ -40,13 +43,28 @@
     help="Write invalidation data to the specified file",
 )
 
+pipe_configuration_option = click.option(
+    "--pipe-configuration",
```
aleclearmind: Let's be coherent with the terminology.

I see "configuration", "pipeline configuration" and "pipe configuration (at most it should be "pipes configuration").

Can we make all of these coherent?

We could call these "dynamic configuration" as opposed to the "static configuration".

Can you propose a unified terminology?

python/revng/pypeline/utils/pipeline.py
```diff
@@ -57,3 +59,21 @@ def get_pipeline_description(pipeline: Pipeline) -> dict[str, Any]:
     validator.validate(pipeline_description)
 
     return pipeline_description
+
+
+def deserialize_pipe_configuration(pipeline: Pipeline, input_: dict[str, str]) -> dict[Pipe, str]:
```
aleclearmind: 1. Docs
2. More like "apply_pipe_configuration"

python/revng/pypeline/daemon/app.py
```diff
@@ -169,6 +169,9 @@ async def analysis_endpoint(request: Request, data: dict) -> Response:
         """Process analysis requests"""
         return await daemon.analyze({**await request.json(), **data})
 
+    async def status(request):
+        return PlainTextResponse("OK")
```
aleclearmind, line 173: Lol

python/CMakeLists.txt
```diff
@@ -231,12 +231,48 @@ python_module(TARGET_NAME revng-merge-dynamic WHEEL revng_internal MODULE_FILES
 #
 # Install revng.project
 #
+add_custom_command(
```
aleclearmind, line 240: Every single time I've seen a `add_custom_command` without a corresponding `add_custom_target`, it was a bug.

If it's not a bug, add a comment.

python/CMakeLists.txt
```diff
@@ -231,12 +231,48 @@ python_module(TARGET_NAME revng-merge-dynamic WHEEL revng_internal MODULE_FILES
...
+  OUTPUT
+    "${CMAKE_BINARY_DIR}/${PYTHON_INSTALL_PATH}/revng/project/model/_generated.py"
+  DEPENDS
+    "${CMAKE_BINARY_DIR}/${PYTHON_INSTALL_PATH}/${PYTHON_GENERATED_MODEL_PATH}"
```
aleclearmind: Where does this file come from?
Do we need two copies or can we directly copy it to the destination here?

scripts/generate-project-mixins.py
aleclearmind, line 1: 1. Docs
2. You sure this is the right place?

scripts/generate_project_mixins.py
```diff
@@ -0,0 +1,82 @@
...
+    return re.sub(r"[^a-zA-Z]", "_", string)
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-o", "--output", default=sys.stdout)
+    parser.add_argument("input")
+    args = parser.parse_args()
+
+    process = run(["./bin/revng2", "project", "dump-pipeline"], stdout=PIPE)
```
aleclearmind: Let's avoid running subprocesses when we can read their output from stdin.

python/revng/project/model/mixins.py.tpl
aleclearmind, line 1: These are currently called "mixins" or "python mixins".

Maybe we should call them "pipeline mixins". AFAIU they are generated starting from the pipeline schema.

python/revng/project/daemon_project.py
```diff
@@ -2,180 +2,184 @@
...
+from revng.project.model import Binary  # type: ignore[attr-defined]
+from revng.project.project import Project
+from revng.support import TarMapper
 
-from .project import Project
+
+class BufferedReader(IOBase):
+    """
+    Adapter class that buffers an urllib3 HTTPResponse content in a spooled file
+    """
```
aleclearmind, line 24: 1. Why do we need this?
2. What's `urllib3` and why do we need it when we use `requests`?

python/revng/project/daemon_project.py
```diff
@@ -2,180 +2,184 @@
...
-        )
+        self._base_url = urlparse(url)
+        self._session = requests.Session()
+        self._session.hooks["response"].append(self._response_hook)
+
         super().__init__()
         self._set_model(self._get_model())
 
-    def set_binary_path(self, binary_path: str):
-        query = gql(
```
aleclearmind, line 36: Oh God

python/revng/project/project.py
```diff
@@ -201,97 +151,30 @@ def _get_pipeline_description(self) -> PipelineDescription:
...
-        the binary and runs the `revng` initial auto analysis.
+        the binary and runs the initial auto analysis.
         """
-        self.set_binary_path(binary_path)
-        self._analyses_list("revng-initial-auto-analysis")
+        hash_ = self.upload_binary(binary_path)
+
+        size = Path(binary_path).stat().st_size
+        new_model = {
+            "Binaries": [{"Index": 0, "Hash": hash_, "Size": size, "Name": Path(binary_path).name}]
```
aleclearmind: Unrelated: in your demo today I saw we `revng2 project init` saves absolute paths in the model. Did we agree on this? What was the ratio?

Ideally I'd like a relative path so that if you have a directory with `model.yml` and `executable` they are "portable".

python/revng/project/project.py
```diff
@@ -201,97 +151,30 @@ def _get_pipeline_description(self) -> PipelineDescription:
...
-
-    def __init__(self, revng_executable_path: Optional[str]):
+    def __deepcopy__(self, memo):
         """
-        Check if the path of the user supplied or default revng executable
-        is present on the system
+        In _set_model the model is deeep-copied. Inside the model we have
+        references to this class. To avoid having a copy of the Project object
+        we implement the `__deepcopy__` method which returns the same instance,
+        effectively disabling any object of this class from being deep-copied.
```
aleclearmind: 1. deeep
2. This sounds extremely suspicious to me. Seems like we're fixing a problem in the wrong place. Can you elaborate?

python/revng/support/__init__.py
```diff
@@ -129,3 +132,31 @@ def get_example_binary_path() -> str:
         if re.match(r"^calc-x86-64-static-revng-qa\.compiled-[0-9a-f]{8}$", entry.name):
             return str(entry.resolve())
     raise ValueError("Could not find calc binary")
+
+
+class TarMapper(Mapping):
```
aleclearmind: 1. Docs
2. `TarDictionary`?

python/CMakeLists.txt
```diff
@@ -255,9 +255,22 @@ add_custom_command(
...
+add_custom_command(
+  OUTPUT
+    "${CMAKE_BINARY_DIR}/${PYTHON_INSTALL_PATH}/revng/project/pipeline_description.py"
+  DEPENDS
+    "${CMAKE_SOURCE_DIR}/scripts/jsonschema_to_py.sh"
+    "${CMAKE_SOURCE_DIR}/python/revng/pypeline/pipeline-description-schema.yml"
+  COMMAND
+    "${CMAKE_SOURCE_DIR}/scripts/jsonschema_to_py.sh"
+    "${CMAKE_SOURCE_DIR}/python/revng/pypeline/pipeline-description-schema.yml"
+    "${CMAKE_BINARY_DIR}/${PYTHON_INSTALL_PATH}/revng/project/pipeline_description.py"
```
aleclearmind, line 268: Please always put `COMMAND` first. CMake is already quite difficult to read.

scripts/jsonschema_to_py.sh
aleclearmind, line 1: Docs

share/doc/revng/references/cli/revng2-project-analyze.md
aleclearmind, line 1: I wonder if we could generate this from `--help`?

share/doc/revng/user-manual/tutorial/model-from-scratch.md
```diff
@@ -229,11 +240,10 @@ Almost everything in the model can have a name. Let's add a name to the function
...
 
 ```bash
-$ revng artifact decompile sum --model model.yml | revng ptml
-0x400000:Code_x86_64: |-
-  _ABI(SystemV_x86_64)
+$ revng2 project artifact emit-c | revng ptml
+/function/0x400000:Code_x86_64: |-
   uint64_t Sum(uint64_t FirstAddend, uint64_t SecondAddend) {
-    return FirstAddend + SecondAddend;
+    return (uint64_t) ((generic64_t) FirstAddend + (generic64_t) SecondAddend);
```
aleclearmind, line 247: Lauri seems to think some pipes are missing.

Can you compare the old and new pipeline? I think thinks are going out of sync (and will keep going out of sync as long as the two pipelines exist).

aleclearmind (commented):
NOTETOSELF: review with split view from `revng.project: port to pypeline` on.

python/revng/pypeline/cli/project/analyze.py
mrjackv: `pipeline_configuration` and `pipe_configuration` refer to the same thing (the dynamic configuration of a set of pipes)
`configuration` varies on the context, e.g. for analyses it refers to the configuration of the analysis.

python/revng/pypeline/daemon/app.py
mrjackv: Do you have any objections to this endpoint?

python/CMakeLists.txt
mrjackv: Ok, it's not a bug because the output file is then passed to `python_module`.
`add_custom_target` is not really needed, it just adds an alias target to ninja for the file(s) you're producing. The only special case is when you do `add_custom_target(my_name ALL ...)` because then you're adding the file to the default target.

python/CMakeLists.txt
mrjackv: > Where does this file come from?

`lib/Model/CMakeList.txt` from the invocation of `target_tuple_tree_generator`

>  Do we need two copies or can we directly copy it to the destination here?

We need to copies because of the `from .mixins import ...`. The first copy (the one generated by `target_tuple_tree_generator`) lives in `revng.model` and has a dummy mixin. The second copy lives under `revng.project.model` and has the magic artifact mixin.

scripts/generate-project-mixins.py
mrjackv: 1. Ok
2. It's a compile-time script, so I think the `scripts/` directory is correct

scripts/generate_project_mixins.py
mrjackv: I'm not understanding the comment, what's the problematic part?

python/revng/project/model/mixins.py.tpl
mrjackv: `mixin` is a general term for when you have an incomplete class that augments an existing one. From the POV of tuple-tree-generator this is true.
The fact is that once that hits the implementation, they are related to the pypeline but this is opaque to the generating infrastructure.

python/revng/project/daemon_project.py
mrjackv: 1. This allows us to read the tar response gradually and without loading it entirely in ram if it gets too big
2. requests is the nice porcelain interface, but under the hood it uses `urllib3` for the plumbing, in some spots (e.g. `request.raw`) this leaks onto the user interface.

python/revng/project/project.py
mrjackv: > in your demo today I saw we revng2 project init saves absolute paths in the model. Did we agree on this? What was the ratio?

We did it to enable the debug info logic to find `foo.debug` if it was in the same directory as the `foo` binary

> Ideally I'd like a relative path so that if you have a directory with model.yml and executable they are "portable".

We had this discussion as well, the problem is two-fold:
1. Relative paths only work if you move things in tandem, e.g. if the file is not in the same directory of the model the relative path breaks the moment you only move the `model.yml`
2. The pypeline so far does not have the concept of a project directory, that's an aspect internal to the `LocalStorageProvider`. Other providers, e.g. HSS, the project directory is not something that makes sense. It would be possible this to be trickled down to the pipes/analyses, but we decided that it wasn't so important and threw this in the lowprio pile

python/revng/project/project.py
mrjackv: 2. Due to the mixins, e.g. `BinaryMixin` we need a reference to the `Project` class in order to call the `get_artifact` function. Due to the mixin re-organization passing the `project` parameter has become a compulsory part of `Binary.deserialize` (which is good). The problem lies in with `revert`, since we want it to work we, in `_set_model` do `self._last_saved_model = deepcopy(self.model)` and then the revert becomes trivial. However since now you have a model with `_project` fields scattered around they are affected by the `deepcopy` so either we use this hook and basically prevent the deep copy or use some other mechanism such as WeakRef, but that also has drawbacks.

share/doc/revng/user-manual/tutorial/model-from-scratch.md
mrjackv: I've re-check just now, all the passes that are present in `OptimizationPipe.cpp` are also present in `pipeline.yml`. I'm not sure how to dig deeper in the matter.

python/revng/pypeline/cli/project/analyze.py
aleclearmind: > pipeline_configuration and pipe_configuration refer to the same thing (the dynamic configuration of a set of pipes)

Can we pick one or the other?

> configuration varies on the context, e.g. for analyses it refers to the configuration of the analysis.

Maybe we should be strict and switch to `static_configuration` and `dynamic_configuration` for pipes + `analysis_configuration` for analyses.

Let's try to get to the bottom of this issues before it spirals out of control.

python/revng/pypeline/daemon/app.py
aleclearmind: No.

python/CMakeLists.txt
aleclearmind: Add a comment about all of this.

scripts/generate-project-mixins.py
aleclearmind: 2. OK, use kebab-case if you don't need to import this.

scripts/generate_project_mixins.py
aleclearmind: This script should not spawn `revng2`.

We should do

```
./bin/revng2 project dump-pipeline | ./generate_project_mixins.py
```

mrjackv: Ah like that, but then it would require an extra bash script that does just the piping just to workaround CMake being really awful (this script is invoked inside a `add_custom_command`)

python/revng/project/daemon_project.py
aleclearmind: 1. If we always read the whole file (no early returns), can't we just do the following?
   ```
   buf = SpooledTemporaryFile(max_size=2 * 1024 * 1024)                                                                                                                                        
   shutil.copyfileobj(response, buf)
   buf.seek(0)                                                                                                                                                                                 
   ```
2. Add a comment about this

python/revng/project/project.py
aleclearmind: Right.

Let me rephrase. Imagine the following scenario:

1. We specify `/bin/ls`
2. `/bin/ls` does not exist.
3. In the same directory of the model there's `ls` (correct hash).

Do we catch the `ls` in the model directory?

python/revng/project/daemon_project.py
mrjackv: 1. Yes, but then we have to wait for the file to download first and then read it, in the current way it's downloaded gradually, so we process each tar entry while the OS is downloading the following chunks
2. Ok

python/revng/project/project.py
mrjackv: Yes, but the the analysis that finds the debug information will not load `ls.debug` if it's also in the model directory

scripts/generate_project_mixins.py
aleclearmind: I think you can do this without an additional script.

Anyway, let's not let the CMake disease spread. Build tools should be sensible independently of how much CMake sucks.

python/revng/project/daemon_project.py
aleclearmind: 1. What is the practical benefit of this? Are we actually enabling some form of pipelining?

python/revng/pypeline/cli/project/analyze.py
aleclearmind: 1. Let's collapse configuration of analyses and pipes into a single dictionary.
2. Let's add a `--$PIPE-configuration` CLI option
3. Let's make sure that in the codebase the static configuration of the pipes is called `static_configuration` and let's purge all the `analysis_configuration` and `pipeline_configuration` in favor of just `configuration`.

python/revng/project/daemon_project.py
```diff
@@ -2,180 +2,184 @@
...
-            for k, v in data.items()
-        }
-
-    def _mapped_artifact_mime(self, artifact_name: str) -> str:
-        container_mime = self._get_artifact_container(artifact_name).MIMEType
-        if container_mime.endswith("+tar+gz"):
-            return container_mime.removesuffix("+tar+gz")
-        raise ValueError
+        response = self._session.post(self._join_url("/api/artifact"), json=body, stream=True)
+        return dict(TarMapper(BufferedReader(response.raw)))
```
aleclearmind: Leave a TODO that one day we could use a generator to pipeline further processing of the file (decompression? parsing?).

python/revng/project/daemon_project.py
```diff
@@ -2,180 +2,184 @@
...
+            while size > 0:
+                size_to_read = chunk_size if size > chunk_size else size
+                buffer = self.response.read(size_to_read)
+                if buffer == b"":
+                    self.end = True
+                    break
+
+                self.file.write(buffer)
+                self.max_offset += len(buffer)
+                size -= chunk_size
```
aleclearmind: The read might return less data than requested, subtract `len(buffer)`.

python/revng/project/daemon_project.py
```diff
@@ -2,180 +2,184 @@
...
+
+    def seek(self, offset: int, whence=os.SEEK_SET):
```
aleclearmind, line 88: 1. `seek` should return the new position.
2. Shall we override `seekable` and `readable` as well?
3. Why don't you forward to `self.file.seek`?

python/revng/project/common.py
```diff
@@ -0,0 +1,74 @@
...
+        # Handle the project_directory parameter, using a temporary directory
+        # if not supplied
+        if project_directory is None:
+            self._tmp_dir = TemporaryDirectory()
+            project_directory = self._tmp_dir.name
+        if not os.path.exists(project_directory):
+            os.makedirs(project_directory)
+        assert os.path.isdir(project_directory)
+        self._project_directory = project_directory
+        self._project_directory_path = Path(project_directory)
```
aleclearmind: Mmmh, do we really need both?

python/revng/project/common.py
```diff
@@ -0,0 +1,74 @@
...
+
+        # Since the project directory might not contain a model, touch one if
+        # missing
+        pipeline_description = Revng2ProjectMixin._get_pipeline_description(self)
+        model_name: str = pipeline_description["model"]["name"]
+        self._model_path = Path(project_directory) / model_name
+        if not self._model_path.exists():
+            self._model_path.touch()
```
aleclearmind, line 74: Isn't `revng2` supposed to do this for us?

python/revng/project/common.py
```diff
@@ -0,0 +1,74 @@
...
+    def _run_revng2_cli(self, cmd_args: Iterable[str], *args, **kwargs):
+        try:
+            return run(
+                [self._executable_path, "-C", self._project_directory, "project", *cmd_args],
+                *args,
+                **kwargs,
+                check=True,
+            )
+        except CalledProcessError as e:
+            raise CLIError() from e
```
aleclearmind: 1. Where is stderr going? Should we wrap it into the error message?
2. If you want to leave stderr to stderr, OK, but at least put in the exit code.
3. I'm a bit confused about exception chaining here.
   What the end result of chaining? Do people typically look into the underyling exception? Does the message of the underlying exception appears in `str(e)` of the new exception?

python/revng/project/daemon_project.py
```diff
@@ -2,180 +2,184 @@
...
+        self.max_offset = 0
+        self.position = 0
+        self.end = False
+
+    def close(self):
+        self.response.close()
+        self.file.close()
+
+    def fileno(self):
+        raise OSError
```
aleclearmind, line 42: `return self.file.fileno()`?

python/revng/project/daemon_project.py
```diff
@@ -2,180 +2,184 @@
...
-                    __typename
-                }
-            }
-        """
-        )
-
-        response = self.client.execute(query, variable_values=variables)
-        assert response["runAnalysis"]["__typename"] == "Diff"
-        self._set_model(self._get_model())
+        if analysis_name in self._analyses_list:
```
aleclearmind: Given the new design we discussed this should become simpler, hopefully.

python/revng/project/daemon_project.py
```diff
@@ -2,180 +2,184 @@
...
+        return response.json()
 
     def _get_model(self) -> Binary:
-        response = self.client.execute(gql("""{ getGlobal(name: "model.yml") }"""))
-        return Binary.deserialize(response["getGlobal"])
+        response = self._session.get(self._join_url("/api/model"))
+        data = response.json()
+        return Binary.deserialize(data["model"], self)
+
+    def _get_epoch(self) -> int:
```
aleclearmind, line 184: This method seems to be called a lot.

Do we really want the client to be stateless? No copy of the model locally?

python/revng/project/local_daemon_project.py
```diff
@@ -2,60 +2,68 @@
...
 from urllib.request import urlopen
 
-from .daemon_project import DaemonProject
-from .project import CLIProjectMixin, ResumeProjectMixin
+from revng.project.common import Revng2ProjectMixin
+from revng.project.daemon_project import DaemonProject
 
 
-class LocalDaemonProject(DaemonProject, CLIProjectMixin, ResumeProjectMixin):
+class LocalDaemonProject(DaemonProject, Revng2ProjectMixin):
```
aleclearmind: I would use inheritance only if you want to expose method of `Revng2ProjectMixin` to users of `LocalDaemonProject` which doesn't seem to be the case.

Here I think you can just demote to *composition*, i.e., turn `Revng2ProjectMixin` into `Revng2CLIHelper` and set it as a private member.

Also, let's not put `Revng` in names, let's go for `CLI2xxx`.

python/revng/project/local_daemon_project.py
```diff
@@ -2,60 +2,68 @@
...
-        ResumeProjectMixin.__init__(self, resume_path)
+        Revng2ProjectMixin.__init__(self, project_directory, executable_path)
 
+        self._host = "127.0.0.1"
         self._port: int = self._get_port()
+        self._daemon_url = f"http://{self._host}:{self._port}/"
         self._daemon_process: Optional[Popen] = None
         self.start_daemon(connection_retries)
-        super().__init__(f"http://127.0.0.1:{self._port}/graphql/")
+        super().__init__(self._daemon_url)
```
aleclearmind, line 40: As I always ask you, call the parent's `__init__` first.

Not doing this is a code smell and needs to be justified with a comment. This PR does it also in other place, review them.

python/revng/project/local_daemon_project.py
```diff
@@ -2,60 +2,68 @@
...
+            "-C",
+            self._project_directory,
+            "project",
+            "daemon",
+            "--host",
+            self._host,
+            "--port",
+            str(self._port),
+        ]
+        self._daemon_process = Popen(cli_args, stdout=DEVNULL, stderr=STDOUT)
```
aleclearmind: Make the method private and `return Popen(...)`?

python/revng/project/local_daemon_project.py
```diff
@@ -70,7 +78,7 @@ def stop_daemon(self) -> int:
 
     def _is_server_running(self) -> bool:
         try:
-            urlopen(f"http://127.0.0.1:{self._port}/status", timeout=5)
+            urlopen(f"{self._daemon_url}status", timeout=5)
```
aleclearmind: 1. Any reason to use `urlopen` as opposed to requests here?
2. Print the error somewhere. Never ever silently suppress errors.

python/revng/project/project.py
```diff
@@ -36,156 +31,111 @@ def __init__(self):
...
-        self._analyses_list_names = [x.Name for x in self._pipeline_description.AnalysesLists]
+        self._analysis_names: Set[str] = {
+            entry["name"] for entry in self._pipeline_description["analyses"]
+        }
+        self._analyses_list: Dict[str, List[str]] = {
+            entry["name"]: entry["analyses"]
+            for entry in self._pipeline_description["analyses_lists"]
+        }
+
+        # Compute the container type of an artifact
```
aleclearmind: Add a comment about why we need this

python/revng/project/project.py
```diff
@@ -36,156 +31,111 @@ def __init__(self):
...
+            objects is None
+            or len(objects) == 0
+            or (
+                len(objects) == 1
+                and isinstance(objects[0], AllMixin)
+                and objects[0]._location == "/binary"
+            )
         ):
-            actual_targets = all_targets
+            actual_objects = set()
```
aleclearmind: What's going on here? A set with a singleton element is not the same as an empty set.

python/revng/project/project.py
```diff
@@ -36,156 +31,111 @@ def __init__(self):
...
 
-        if isinstance(results, dict):
-            if rank.Depth == 0:
-                keys = list(results.keys())
-                assert len(keys) == 1
-                return Artifact.make(results[keys[0]], container_mime)
-            else:
-                container_mime = self._mapped_artifact_mime(artifact_name)
-                return {k: Artifact.make(v, container_mime) for k, v in results.items()}
+        if kind == "binary":
```
aleclearmind: Is this special-casing actually needed? Feel very wrong.

python/revng/project/project.py
```diff
@@ -201,97 +151,30 @@ def _get_pipeline_description(self) -> PipelineDescription:
...
-    def import_and_analyze(self, binary_path: str):
+    def import_and_analyze(self, binary_path: Union[str, Path]):
         """
         A helper method for setting up the project. This method imports
-        the binary and runs the `revng` initial auto analysis.
+        the binary and runs the initial auto analysis.
         """
-        self.set_binary_path(binary_path)
-        self._analyses_list("revng-initial-auto-analysis")
+        hash_ = self.upload_binary(binary_path)
```
aleclearmind, line 195: Maybe we should have an endpoint to ask the daemon if a file with a certain hash is already available.
TODO?

python/revng/project/project.py
```diff
@@ -201,97 +151,30 @@ def _get_pipeline_description(self) -> PipelineDescription:
...
+
+        size = Path(binary_path).stat().st_size
+        new_model = {
+            "Binaries": [{"Index": 0, "Hash": hash_, "Size": size, "Name": Path(binary_path).name}]
+        }
+        self._analyze("set-model", yaml.safe_dump(new_model))
+
+        self._analyze("initial-auto-analysis")
```
aleclearmind, line 202: 1. Maybe we should try to be a bit more idempotent: let's add the binary to `Binaries` and let's just `.commit`.
   We can consider `initial-auto-analysis` "idempotent".
2. Why are you using a "raw" model instead of our nice classes?

scripts/jsonschema_to_py.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
aleclearmind, line 1: Docs

python/revng/internal/cli/_commands/ptml.py
```diff
@@ -68,24 +74,41 @@ def handler(args) -> int:
...
+    content: bytes | IO[bytes], output: IO[str], color: ColorMode, filters: str | list[str]
 ):
     with to_bytes(content) as wrapped:
-        artifact = ptml_artifact_autodetect(wrapped)
-        if isinstance(artifact, PTMLArtifact):
-            artifact.print(output, color)
+        if len(wrapped) == 0:
+            raise ValueError("Input is empty!")
+
+        with suppress(tarfile.ReadError):
```
aleclearmind, line 86: Why?

Can you pretty please never ever suppress errors? This something that keeps happening again and again.
If you really really want to suppress an error you *have to* leave a detailed comment explaining why, and then think again three times.

python/revng/internal/cli/_commands/ptml.py
```diff
@@ -3,16 +3,22 @@
...
 from revng.internal.cli.commands_registry import Command, CommandsRegistry, Options
-from revng.internal.cli.support import file_wrapper
-from revng.ptml.printer import ColorMode
+from revng.internal.cli.support import TarMapper, file_wrapper
+from revng.ptml.printer import ColorMode, ptml_print, ptml_print_mapping
 from revng.support import to_bytes
-from revng.support.artifacts import PTMLArtifact, ptml_artifact_autodetect
+
+# Get yaml.CSafeLoader if present, fall back to yaml.SafeLoader otherwise
+YAMLLoader = getattr(yaml, "CSafeLoader", yaml.SafeLoader)
```
aleclearmind: I don't think we need this. If it was no pypi I'd understand, but this is our own code.

We should always use the C version and assert.

python/revng/internal/cli/_commands/ptml.py
```diff
@@ -68,24 +74,41 @@ def handler(args) -> int:
 
 
 def handler_inner(
-    content: bytes | IO[bytes], output: IO[str], color: ColorMode, filters: Union[str, List[str]]
+    content: bytes | IO[bytes], output: IO[str], color: ColorMode, filters: str | list[str]
 ):
     with to_bytes(content) as wrapped:
-        artifact = ptml_artifact_autodetect(wrapped)
-        if isinstance(artifact, PTMLArtifact):
```
aleclearmind, line 75: Do we no longer need to check this?

python/revng/internal/cli/_commands/ptml.py
```diff
@@ -68,24 +74,41 @@ def handler(args) -> int:
...
-            artifact.print(output, color)
-        else:
-            artifact.print(output, color, lambda x: x in cast(List[str], filters))
+        if re.match(rb"\s*<", wrapped) is None:
+            data = None
+            with suppress(yaml.YAMLError):
+                data = yaml.load(wrapped, Loader=YAMLLoader)
+            if data is not None:
+                handle_filters(data, filters, output, color)
+                return 0
```
aleclearmind, line 100: What's going on here? Docs.

python/revng/internal/cli/_commands/ptml.py
aleclearmind, line 1: Do we still support printing tars containing PTML?

share/doc/revng/user-manual/tutorial/model-from-scratch.md
aleclearmind: Give him a reproducer and ask for help.

python/revng/project/project.py
aleclearmind: 2. OK, let's have a `IgnoreDeepCopy` wrapper or something like that.

python/revng/project/daemon_project.py
aleclearmind: Add a comment

python/revng/project/daemon_project.py
aleclearmind: 1. Save the epoch in the model, update it after each analysis and ensure we throw an error in case of mismatch.
2. Add a comment like...

```
In the future we need to have more features for interactivity/multi-user

1. Cache artifacts
2. Handle cache invalidations
3. Have some form of locking of the model to enable running "client-side analysis".
   Without this, complex analysis in Python would be problematic since, without a lock, we might spend a long time doing our computations and then just fail because someone else has renamed a function.
   The idea would be to have a Python context manager holding the lock.
4. We might also have to offer an `async` flavor of the class project.
   The non-`async` version would have an opt-in thread to monitor notifications coming from the websocket.
```

python/revng/internal/cli/_commands/ptml.py
mrjackv: Yes, the burden has been shifted from `revng.support.artfact` to the `revng ptml` command

python/revng/internal/cli/_commands/ptml.py
mrjackv: > Why?

In this piece of code we have some `bytes` as input and we need to figure out what kind of PTML they are (tar, yaml or plain). There's no good way to autodetect either format so the way this is done is by trying each one and moving onto the next if an exception is caught.

> Can you pretty please never ever suppress errors? This something that keeps happening again and again.
If you really really want to suppress an error you have to leave a detailed comment explaining why, and then think again three times.

I'm aware of this, but this is already existing code that was part of [this function](https://github.com/revng/revng/blob/develop/python/revng/support/artifacts.py#L241) that has been re-stiched back into the `revng ptml` command

python/revng/internal/cli/_commands/ptml.py
mrjackv: `ptml_artifact_autodetect` has been inlined into the function

python/revng/project/common.py
mrjackv: No, must `revng2 project` stuff assumes that `model.yml` exists somewhere, otherwise the git-like search is kind of moot

python/revng/project/local_daemon_project.py
mrjackv: This was done with the previous `LocalDaemonProject`, [link](https://github.com/revng/revng/blob/develop/python/revng/project/local_daemon_project.py#L37)

I'll add a comment

share/doc/revng/user-manual/tutorial/model-from-scratch.md
mrjackv: Given the discussion on mattermost, should I close this?

share/doc/revng/references/cli/revng2-project-analyze.md
mrjackv: We can have _another_ page that's basically the `--help` output converted to markdown (e.g. using [mkdocs-click](https://github.com/mkdocs/mkdocs-click)). However I think this has its own dignity, this is a bird's eye view of the command, rather than going through each command-line option

python/revng/project/local_daemon_project.py
mrjackv: 1. this was leftover before we started using requests
2. Here we're checking if the server got up, we expect the server to not respond. If things fail we throw an exception. Printing that we got an error, every 5 seconds, to a user in the REPL doesn't seem nice.

python/revng/project/project.py
mrjackv: Basically, if we get a request that looks like the user wants to produce `*`, we set the object set to the empty set since that implicitly produces all objects when making the request to the server

python/revng/project/project.py
mrjackv: It's convenient, if you do `_get_artifact("lift")` you get the `Artifact` out instead of a `dict[str, Artifact]` with just one entry. If you prefer I check for the name of the 0-ranked kind in the pipeline description I can do that

python/revng/pypeline/storage/local_provider.py
mrjackv: Done

python/CMakeLists.txt
mrjackv: Done

scripts/generate-project-mixins.py
mrjackv: 1. Done
2. Done

scripts/generate_project_mixins.py
mrjackv: Done

python/revng/project/daemon_project.py
mrjackv: Closing this since we discussed this IRL

python/revng/support/__init__.py
mrjackv: 1. Done
2. Renamed

python/CMakeLists.txt
mrjackv: Fixed in all instances

scripts/jsonschema_to_py.sh
mrjackv: Done

python/revng/project/daemon_project.py
mrjackv: Done

python/revng/project/daemon_project.py
mrjackv: Thanks for spotting this! fixed

python/revng/project/common.py
mrjackv: Dropped `_project_directory_path`

python/revng/project/daemon_project.py
mrjackv: Done

python/revng/project/local_daemon_project.py
mrjackv: Done, went with `CLIHelper`

python/revng/project/local_daemon_project.py
mrjackv: Added comments in all the instances where the constructor of the parent class is delayed

python/revng/project/local_daemon_project.py
mrjackv: Done

python/revng/project/local_daemon_project.py
mrjackv: 1. Switched to `requests`

python/revng/project/project.py
mrjackv: Done

python/revng/project/project.py
mrjackv: 1. I don't think this is wise, right now the pypeline will throw an error if `len(Binaries) > 1`. With the current setup the only two options are:
   1. Overwrite the first entry of `Binaries` with the specified file
   2. No-op/throw if there is already an entry

scripts/jsonschema_to_py.sh
mrjackv: Done

python/revng/internal/cli/_commands/ptml.py
mrjackv: Fixed

python/revng/internal/cli/_commands/ptml.py
mrjackv: Added

python/revng/project/daemon_project.py
mrjackv: 3. The class' `seek` does not actually seek, only when `read`-ing it reads as many bytes as `(self.position + size) - self.max_offset`

python/revng/project/common.py
aleclearmind: OK, please add a comment

python/revng/project/project.py
aleclearmind: 1. This is a temporary problem. I'm fine with throwing + `TODO`.

python/revng/project/project.py
aleclearmind: I think that the fact that `ArtifactResult` is `Union[Artifact, Mapping[str, Artifact]]`, forces the caller to be aware of which one of the two it's going to get.

This would mean that they are actually two distinct methods.

Is this public API? I think it needs some more love.

python/revng/project/project.py
aleclearmind: Please entirely review the logic of this method.

Right now sending in `None`, `[]` or `["/binary"]` does the some thing.
And this same thing is requesting an empty set of things which actually means "all the things".
This doesn't make any sense to me.

Clearly define the semantics of the `objects` argument of `_get_artifact` and `_get_artifact_impl`.

Do not abuse `None` or `[]` to mean "everything". You can introduce an explicit singleton object to mean "everything" and let it handle explicitly in the implementor of `Project`.

Something like:

```python
class AllObjects:
    pass

# ...

class Project:
    def _get_artifact(
        self,
        artifact_name: str,
        objects: Union[AllObjects, Sequence[Union[str, StructBaseT]]]],
        configuration: Dict[str, str] = {},
    ) -> ArtifactResult:
```

And make sure an empty list means an empty list.

Then, the part of the code that knows that an empty list "implicitly produces all objects when making the request to the server" can handle that, but for sure this should not be in `Project`.

python/revng/project/project.py
mrjackv: Yes and no, if you use the magical accessors (e.g. `function.emit_c`) you'll get just the artifact, if you call `model.get_artifact` which is the generic version you get `ArtifactResult`

aleclearmind: What's preventing us from splitting this into `get_artifact` and `get_single_object_artifact` (or something like that)?

python/revng/project/project.py
mrjackv: Fixed both (1) and (2)

python/revng/project/project.py
mrjackv: Done

python/revng/project/project.py
mrjackv: Made the new `_get_artifact` much more uniform

python/revng/project/project.py
mrjackv: See the new `_get_artifact` and `BinaryMixin.get_artifact`

python/revng/project/local_daemon_project.py
mrjackv: 1. Done

python/revng/project/common.py
mrjackv: Done

python/revng/project/daemon_project.py
mrjackv: Fixed (1) and (2)

python/revng/project/project.py
mrjackv: Done

python/revng/project/common.py
mrjackv: I think we discussed this at some point IRL, but I forgot what was the solution we agreed upon

python/revng/pypeline/cli/project/analyze.py
mrjackv: Implemented everything

python/revng/pypeline/utils/pipeline.py
mrjackv: 1. Done
2. I've renamed the function to `deserialize_configuration` as it now takes care of both pipes and analyses. I'm still convinced that `deserialize` applies to the function

python/revng/project/common.py
mrjackv: As discussed in person, changed this so that `CLIError` has rich error message

python/revng/project/daemon_project.py
mrjackv: Closing this as this code is gone

python/revng/project/daemon_project.py
mrjackv: 1. Done
2. Done

share/doc/revng/references/cli/revng2-project-analyze.md
aleclearmind: I'm really concerned that this will go out of sync. We strive hard to test the docs and this would be an exception.

Maybe there's a way to add examples and a longer description to click's doc?

I'd prefer to have documentation for the whole CLI generated and available online, rather than a couple of key commands that risk to go out of sync.

We can postpone this, in case.

mrjackv: Let me preface this by saying that I've only made these two markdown documents because they were also present for `revng analyze` and `revng artifact` and those were referenced in the tutorial.

That said, my main concern is that, in my mind, there are two distinct things:
* The cli `--help` needs to be concise, it's needed most to refresh things while you're using the tool live and should not be a book
* The `man` (which is essentially what the docs are) should be more verbose and with examples

> We strive hard to test the docs and this would be an exception.

I'm not against the idea to test the examples, but obviously the doc testing system needs some love, right now it works well for tutorials (where you write all the steps) but not so much for examples

> Maybe there's a way to add examples and a longer description to click's doc?

You can add free-form text to a command description, but again, I think it's counter-productive:
* CLI users don't want 20 examples when running `--help`
* It's again free-form text, which can happily bitrot just as much as the docs, but at the very least at some point we might test the examples in the doc

> I'd prefer to have documentation for the whole CLI generated and available online, rather than a couple of key commands that risk to go out of sync.

Again, not against this per-se (e.g. using mkdocs-click), but the docs really need to have that man-like lens where you first explain things in plain english with examples, then show all the options.

> We can postpone this, in case.

I can add a TODO to both mds reminding us of this discussion, but I really think this is the best middle-ground for now. We can come up with a solution to make docs at a later time.

python/revng/pypeline/cli/common_options.py
```diff
@@ -140,3 +149,123 @@ def convert(self, value, param, ctx):
...
+                type=_ConfigurationOptionType(type_),
+                expose_value=False,
+                cls=_HidableOption,
+                hide_flag=hide_flag,
+                metavar="CONFIGURATION",
+                help=f"Configuration for the {type_.name} {type_string}",
+            )(func)
+
+        func = click.option(
+            "--show-configuration-options",
```
aleclearmind: Maybe we can have `--help` and `--help-full`? LLVM has `--help-list-hidden`.

It's not obvious to me that `--show-configuration-options` affects `--help`.

python/revng/pypeline/storage/local_provider.py
```diff
@@ -383,23 +389,74 @@ def _init_tables(self):
...
+                cursor.execute(
+                    "UPDATE project SET model_hash = ?, model_mtime = ? WHERE id is 0",
+                    (model_hash, model_mtime),
+                )
+            return
+
+        # Check if we actually got a non-null result, if so check that the
+        # mtime and hash of the model match (in this order for speed), if
+        # not prune
+        if model_metadata[1] == model_mtime:
```
aleclearmind, line 429: So if the file changes but the date is somehow preserved, we won't detect it, correct? Add an explicit warning about this.

Maybe we could throw it the file size in as well as an extra measure to counter this.

Sometimes files installed by the package manager have their date reset'd to epoch 0.

Not urgent though, we the warning is enough.

python/revng/pypeline/cli/common_options.py
mrjackv: Sounds good, I guess this should also remove the need for `--show-hidden-artifacts`.
Just to make sure, you want this to substitutive, e.g. a user types `--help-full` and it shows the help, not `--help-full --help`, right?

aleclearmind: Yes, I think that's what I'd like as a user.

aleclearmind: Yes, I think that's what I'd like as a user.

share/doc/revng/references/cli/revng2-project-analyze.md
aleclearmind: I'm not saying to add examples in `--help`.

I'm saying, can we specify the examples to click and then only render them with `mkdocs-click`? Like `extra_help="...", examples="..."`.

If we specify the twice the list of switches they are doomed to go out of sync.

However, looking at how `mkdocs-click` works, we can probably just replace the list of switches with something like:

```
::: mkdocs-click
    :module: myapp.cli
    :command: cli
    :prog_name: myapp
    :depth: 1
    :style: table
```

So the other way around. Let's leave a markdown comment about this.

> Let me preface this by saying that I've only made these two markdown documents because they were also present for revng analyze and revng artifact and those were referenced in the tutorial.

Fair enough.

mrjackv: > I'm saying, can we specify the examples to click and then only render them with mkdocs-click?

I don't think it can be done with the stock `mkdocs-click`, but if we fork it I'm sure it's a pretty trivial change

> If we specify the twice the list of switches they are doomed to go out of sync.

I see what you mean, yeah using something like `mkdocs-click` is the best way to go, however we have some switches (such as `-C`) that IMHO should be shown but are in previous layers. Again probably if we fork `mkdocs-click` we can make this work easily

For examples I still think they should be in plain markdown and have a better testing system that allows us to run them and check they are correct.

In any case I've left a comment in the markdowns for future reference

python/revng/pypeline/cli/common_options.py
mrjackv: Implemented

python/revng/pypeline/storage/local_provider.py
mrjackv: Added

aleclearmind:
Merged in d440e20. Thanks.

PR #572, opened 2026-03-12
Clift dialect changes

vasama-work:


lib/Support/CTarget.cpp
```diff
@@ -10,7 +10,8 @@ const TargetCImplementation TargetCImplementation::Default = {
     { 1, CIntegerKind::Char },
     { 2, CIntegerKind::Short },
     { 4, CIntegerKind::Int },
-    { 8, CIntegerKind::Long },
+    { 4, CIntegerKind::Long },
+    { 8, CIntegerKind::LongLong },
```
pfez, line 14: Why? Leave it to 8

lib/CliftEmitC/CBackend.cpp
```diff
@@ -81,7 +81,8 @@ class CliftToCEmitter : CEmitter {
     auto Primitive = getUnderlyingIntegerType(Type);
 
     auto CInteger = Target.getIntegerKind(Primitive.getSize());
-    if (not CInteger or not mlir::isa<PrimitiveType>(Type)) {
+    if (not CInteger or *CInteger >= CIntegerKind::Extended
+        or not mlir::isa<PrimitiveType>(Type)) {
```
pfez: Have we got a unit test for this?

include/revng/Clift/CliftTypeInterfacesObject.td
```diff
@@ -12,6 +12,13 @@ def Clift_ObjectType : Clift_TypeInterface<"ObjectType", [Clift_ValueType]> {
     The object types is the set of types usable as types of variables and
     fields.
   }];
+
+  let methods = [
+    InterfaceMethod<
+      "Returns the size of the object type.",
+      "uint64_t", "getObjectSize", (ins)
+    >,
+  ];
```
pfez, line 22: Ok to have this on `ObjectType` now, but I want the `getByteSize` name.

include/revng/Clift/Clift.td
```diff
@@ -157,43 +157,20 @@ def Clift_FunctionOp : Clift_Op<"func",
...
+                        mlir::DictionaryAttr::get(getContext());
+      return AttrDictView(DictAttr);
+    }
+
+    mlir::clift::AttrDictView getResAttrs(unsigned Index) {
+      std::optional<mlir::ArrayAttr> Attrs = getArgAttrs();
+      mlir::DictionaryAttr DictAttr = Attrs ?
+                        mlir::cast<mlir::DictionaryAttr>((*Attrs)[Index]) :
+                        mlir::DictionaryAttr::get(getContext());
+      return AttrDictView(DictAttr);
```
pfez, line 195: As per one of my comments on Ivan's PR on headers, we're going to move this on `FunctionType`, so it's probably not worth the effort.

include/revng/Clift/CliftTypes.h
```diff
@@ -38,19 +38,47 @@ TypedefDecomposition decomposeTypedef(ValueType Type);
...
+
+//===-------------------------- Object type size --------------------------===//
+
+/// Returns the size of the given type, assuming it is an ObjectType (after
+/// unwrapping typedefs).
+[[nodiscard]] uint64_t getObjectSize(mlir::Type Type);
+
+/// Returns the size of the given type, if it is an ObjectType (after unwrapping
+/// typedefs). Otherwise returns zero.
+[[nodiscard]] uint64_t getObjectSizeOrZero(mlir::Type Type);
```
pfez, line 152: Like I noted somewhere else, use `ByteSize` instead of `ObjectSize`.

lib/Clift/Clift.cpp
```diff
@@ -1425,6 +1416,13 @@ mlir::LogicalResult IndirectionOp::verify() {
 //===------------------------------ AssignOp ------------------------------===//
 
 mlir::LogicalResult AssignOp::verify() {
+  clift::ValueType Type = dealias(getLhs().getType(),
+                                  /*IgnoreQualifiers=*/true);
+
+  if (not isObjectType(Type) or mlir::isa<ArrayType>(Type))
+    return emitOpError() << getOperationName()
+                         << " left operand must be a non-array object type.";
```
pfez: We should verify this for RHS too.

include/revng/Clift/Clift.td
```diff
@@ -737,7 +737,7 @@ def Clift_ImmediateOp : Clift_ExpressionOp<"imm", [NoMemoryEffect]> {
 
 def Clift_StringOp : Clift_ExpressionOp<"str", [NoMemoryEffect]> {
   let arguments = (ins StrAttr:$value);
-  let results = (outs Clift_AnyValueType:$result);
```
pfez, line 725: Adding this comment here but it applies to the whole commit.

The commit message says: "Remove unnecessary Clift operand type constraints". What do you mean exactly by "unnecessary" here. I don't think we're really happy with having `AnyType` in the places you're relaxing the constraints.

I presume you mean that in all these places we can relax the constraints since `verify` catches them anyway. If that's the case, fine but clarify in the commit message.

pfez: On the other hand, one could imagine, for at least some of these constraints that you're relaxing, to actually make them stricter instead.
This would remove custom logic in C++ from `verify` and make the declaration here in tablegen more explicit and declarative about what types of operands are valid.
This would come at the cost of having to specify more constraints in tablegen, but it would easier to understand by whoever is reading the dialect specification in tablegen, without having to go and ready the `verify` in C++.

Like here, for `StringOp` you could constrain the result on a const array type.

include/revng/Clift/Clift.td
```diff
@@ -916,8 +916,8 @@ def Clift_PtrDiffOp
                        [NoMemoryEffect,
                         Clift_AllTypesMatch<["lhs", "rhs"]>]> {
 
-  let arguments = (ins Clift_AnyValueType:$lhs, Clift_AnyValueType:$rhs);
```
pfez, line 904: `PtrDiff` could be constrainted to have pointer arguments.

include/revng/Clift/Clift.td
```diff
@@ -991,7 +991,7 @@ def Clift_AddressofOp
                                        "object",
                                        "mlir::cast<PointerType>($_self).getPointeeType()">]> {
 
-  let arguments = (ins Clift_AnyValueType:$object);
```
pfez, line 979: This could be constrained to have and addressable type argument.

include/revng/Clift/Clift.td
```diff
@@ -1050,11 +1050,11 @@ def Clift_AssignOp
 }
 
 def Clift_AccessOp : Clift_ExpressionOp<"access", [NoMemoryEffect]> {
-  let arguments = (ins Clift_AnyValueType:$value,
```
pfez, line 1038: This could be constrained to have an argument of class type or pointer type.

This is definitely less compelling than the other examples I made.

I don't want to say that this always makes sense. But it's not clear what are the trade offs here. You seem to just assume that it always make sense to defer all checks to `verify`. If that was the case what are type constraints for in MLIR? Can you clarify your reasoning here?

include/revng/Clift/CliftTypeInterfacesObject.td
```diff
@@ -0,0 +1,18 @@
...
+#define MLIR_CLIFT_TYPE_INTERFACES_OBJECT
+
+include "revng/Clift/CliftTypeInterfacesBasic.td"
+
+def Clift_ObjectType : Clift_TypeInterface<"ObjectType", [Clift_ValueType]> {
+  let description = [{
+    The object types is the set of types usable as types of variables and
+    fields.
+  }];
+}
```
pfez: Shouldn't this have the `getByteSize` (or `getObjectSize`) method?

pfez: Ah it comes in a following commit. This is one of the limits of looking at the PR per-commit. Anyway, I would put the the method directly in this commit.

include/revng/Clift/CliftAttributes.td
```diff
@@ -64,7 +64,7 @@ def Clift_FieldAttr
   let parameters = (ins StringRefParameter<>:$handle,
                         "mlir::clift::MutableStringAttr":$mutable_name,
                         "uint64_t":$offset,
-                        "mlir::clift::ValueType":$type);
```
pfez, line 72: Can you explain the rationale here?
Leaving the more restrictive `ValueType` makes it clearer directly in tablegen what are some of the contraints of `FieldAttr`. Like one of the other commits I have already commented on, one could argue for making this stricter and require an `clift::ObjectType` instead of the more general `mlir::Type`.

include/revng/Clift/CliftTypes.td
```diff
@@ -278,19 +278,6 @@ def Clift_FunctionType
...
-                     "mlir::clift::ValueType":$ReturnType,
-                     "llvm::ArrayRef<mlir::clift::ValueType>":$ArgumentTypes), [{
-      return $_get($_ctxt,
-                   Handle,
-                   Name,
-                   ReturnType,
-                   llvm::SmallVector<mlir::Type, 16>(ArgumentTypes));
-    }]>,
-  ];
-
```
pfez, line 284: Were these builders just unused?

include/revng/CliftImportModel/ImportModel.h
```diff
@@ -27,8 +27,8 @@ importModelType(llvm::function_ref<mlir::InFlightDiagnostic()> EmitError,
 /// Convert the specified qualified model type to a Clift type in the specified
 /// context.
 ///
-/// \return The corresponding Clift ValueType, or null on failure.
-ValueType
```
pfez: In general I'm in favor of making argument types more permissive (e.g. `mlir::Type` vs `clift::ValueType`), but I think it backfires for return types. If we know this always returns a `clift::ValueType`, why relax it?

pfez: The same applies to a bunch of other cases in this commit.

pfez: Like, a ton of other places. Please dicuss the reasoning.

lib/Support/CTarget.cpp
vasama-work: Ah yes, I should have explained in the commit message.

The issue is that we currently only have a single default target, but we actually test the emitted code using multiple implementations, some of which seem to have a 32 bit long. This never caused issues before, because we incorrectly emitted some C-style casts when bitcasting floats to ints or vice versa, but now we correctly emit `bit_cast`, and this caused failures in cases such as `bit_cast(float64_t, 0UL)`, where `0UL` is 4 bytes in size.

Changing long to 4 for now doesn't really matter for the implementations where it's actually 8. It just means we will emit 64-bit integer literals as `0LL` instead of `0L`.

include/revng/Clift/Clift.td
vasama-work: The effort is done, so we're not saving anything. And it's a much better interface for accessing argument attributes. I suspect we'll have more use for them in the future.

include/revng/Clift/CliftTypes.h
vasama-work: So `getByteSize` seems fine when it's a member function, but a lot worse as a free function. I would really like to signal the fact that we're getting the size of an `ObjectType` in the naming.

Going forward, we're going to be using these free functions and there's going to be no direct users of the `ObjectType` member function. To the point that I was considering giving it a worse name or some tag parameter to discourage its use.

include/revng/Clift/Clift.td
vasama-work: Yes, the reason is to avoid duplicate verification. There's no point checking for `AddressableType` when we next check for `IntegerType` in the verify function. But yes, if you like, we could type verification to tablegen in some cases.

include/revng/Clift/Clift.td
vasama-work: It has been a matter of picking the most convenient way to verify. If we just verify category, I've mostly done that in tablegen. If we verify some more complicated properties, I've usually done it in C++.

include/revng/Clift/CliftAttributes.td
vasama-work: So more generally about the use of `ValueType` (now `AddressableType`): I also talked about this in my ruminations pad. Until now, we've kind of used `ValueType` as a default C++ type for any MLIR types, and at its surface that makes a lot of sense, right? Why would we want to accept any other types? But in practice, this has not panned out. Trust me when I say that the net effect has not been positive. Because any MLIR interfaces deal in `mlir::Type` (e.g. `mlir::Value`, `mlir::TypeAttr`), we're constantly having to convert to `ValueType` or just inconsistently using `mlir::Type` instead. Additionally, this has never actually prevented a bug. Some non-Clift type or even a label type sneaking in where it isn't supposed to be used just doesn't happen in practice. It was very satisfying seeing all the code that this change just simplified because we no longer had to deal with `ValueType`. So going forward, we should just use `mlir::Type` as our default for types, and that's what this commit does for existing code.

We cannot use `ObjectType` here, because that would not permit typedefs. We could use `AddressableType`, but that wouldn't be helpful. It's too wide and we still have to verify. It's more valuable to be consistent and just use `mlir::Type`.

include/revng/Clift/CliftTypes.td
vasama-work: No, but without the widespread use of `ValueType` they became unnecessary.

include/revng/Clift/Clift.td
vasama-work: > Like here, for `StringOp` you could constrain the result on a const array type.

Sure, we could, but in verify we also have to check that its size matches that of the content string.

lib/CliftEmitC/CBackend.cpp
vasama-work: No. I could add one, but I actually think this might be better done as part of legalization. Let me know what you think.

lib/Clift/Clift.cpp
vasama-work: In this case we actually verify in tablegen that the two types are the same.

lib/CliftEmitC/CBackend.cpp
```diff
@@ -504,7 +513,14 @@ class CliftToCEmitter : CEmitter {
...
 
-    if (mlir::isa<ImmediateOp>(E)) {
+    if (auto Immediate = mlir::dyn_cast<ImmediateOp>(E.getOperation())) {
+      if (isNullPointerConstant(Immediate)) {
+        return {
+          .Precedence = OperatorPrecedence::Primary,
+          .Emit = &CliftToCEmitter::emitNullPointerConstant,
+        };
+      }
+
```
pfez, line 543: Eh ok. Why not. The commit message says WIP. Is something missing?

vasama-work: There's no unit test, but more important is the question of whether we want to keep this commit at all.

include/revng/Clift/CliftEnums.td
```diff
@@ -21,13 +21,3 @@ def Clift_PrimitiveKind: I32EnumAttr<"PrimitiveKind", "primitive type kinds", [
   // attribute wrapping it.
   let genSpecializedAttr = 0;
 }
-
-def Clift_CastKind : I32EnumAttr<"CastKind", "Cast operation semantic kinds", [
-    I32EnumAttrCase<"Extend",           1,      "extend">,
-    I32EnumAttrCase<"Truncate",         2,      "truncate">,
-    I32EnumAttrCase<"Bitcast",          3,      "bitcast">,
-    I32EnumAttrCase<"Decay",            4,      "decay">,
-    I32EnumAttrCase<"Convert",          5,      "convert">,
```
pfez, line 30: Wait, where did conversion casts go? Did you drop them entirely?

vasama-work: Yes, I did. We discussed removing it for now for lack of any use, but didn't make a final decision one way or the other. I can add a `FloatCastOp`, but it wouldn't be used at all at this time. The Clifter doesn't generate any such casts and never did.

include/revng/Clift/Clift.td
pfez, line 1: 1. Why do we have `PtrResizeOp` and `PtrCastOp` as separate operations? It's not consistent with the integer casts where `IntCastOp` works both as resize and as integer cast.
2. I'm thinking it could be beneficial to have an common interface for all but `DecayOp` and `BitCastOp`. Something like `Clift_CStyleCastOpInterface`. It would make it easier to catch all clift casts that are emitted as c-style casts in C, assuming some transformation will only want to reason about those. Any thoughts?

include/revng/CliftTransforms/Expressions.td
```diff
@@ -72,69 +71,61 @@ def AddressofIndirectionPattern
...
-def CollapsibleCastPattern
-  : Pat<(Clift_CastOp:$res (Clift_CastOp $arg, $inner_kind), $outer_kind),
-        (Clift_CastOp $arg, $inner_kind),
-        [(ValuesAreEqual $inner_kind, $outer_kind),
-         (Constraint<CPred<"isCollapsibleCastKind($0.getValue())">> $inner_kind)]>;
+// Constraint: the cast pair is collapsible (implemented in .cpp file).
+def CollapsibleIntCastPattern
+  : Pat<(Clift_IntCastOp:$outer (Clift_IntCastOp:$inner $value)),
+        (Clift_IntCastOp $value),
+        [(Constraint<CPred<"isCollapsibleIntCast($0, $1, $2)">> $outer, $inner, $value)]>;
```
pfez: These should be part of canonicalization.

include/revng/Clift/Clift.td
vasama-work: 1. The reasoning is that we do a lot of rewrites on reinterpret casts between pointers, but not on pointer resizing casts, which only appear due to legalization. Rather than constantly check that the pointer sizes match, it's much simpler to keep them as separate operations.
2. We could do that... I don't realistically see any rewrite doing that. The only user I can foresee for this would be the backend, but it would do very little to simplify the backend. I would defer adding this until we actually have a non-trivial use case.

include/revng/CliftTransforms/Expressions.td
vasama-work: I know we've discussed this before, and I absolutely understand why you want this, but I've come to disagree more and more. Canonicalisation is applied very aggressively, but we do sometimes want to retain non-canonical forms, especially when going into the backend. So I think the better solution might be to explicitly opt into sets of canonicalisation rewrites as needed. Like how we already do with the boolean rewrites.

lib/CliftTransforms/Expressions.cpp
```diff
@@ -37,8 +37,19 @@ static uint64_t truncateIntegerValue(mlir::IntegerAttr ValueAttr,
   return Value & (static_cast<uint64_t>(-1) >> (64 - 8 * T.getSize()));
 }
 
-static bool isCollapsibleCastKind(CastKind Kind) {
-  return Kind != CastKind::Convert;
+static bool
+isCollapsibleIntCast(mlir::Value Outer, mlir::Value Inner, mlir::Value Value) {
```
pfez: Shouldn't we have have the concept of collapsible pointer casts too? This including pointer resize and pointer bitcats.

include/revng/Clift/Clift.td
pfez: Basically you're saying that when Ivan's PR moves the argument attributes to `FunctionType`s he'll just reuse this interface there.

include/revng/Clift/CliftTypes.h
pfez: > So `getByteSize` seems fine when it's a member function, but a lot worse as a free function

I disagree. It's more important to be non-ambiguous about the fact that the size is in bytes, than being non-ambiguous about it being an `ObjectType`.

include/revng/Clift/CliftEnums.td
pfez: Let's leave it out. I just wanted to double check if it was on purpose or it went under the radar.

include/revng/CliftTransforms/Expressions.td
pfez: > we do sometimes want to retain non-canonical forms, especially when going into the backend

Are you talking about casts specifically, or as a general argument?
If it's a general argument ok, but it boils down to not including the rewrites we don't want into canonicalization.
If it's about casts I have pretty strong opinions about ensuring integer bitcasts and collapsible casts always being canonicalized away. If we don't, every transform (including implicit cast elision) will have to deal with the full range of casts that can happen, unless we explicitly opt in to optimizing them away.

lib/CliftEmitC/CBackend.cpp
pfez: It's not perfect, but it has a minimal impact for now and it directly improves the quality of the generated C code in a lot of simple but common situations. I'm in favor of keeping it, but with at least a unit test.

include/revng/Clift/Clift.td
pfez: 1. As I suggested in another comment, we should be able to collapse pointer-resize casts. So there's a case for matching pointer casts alltogether (resize+bitcasts). Perhaps we need a `Clift_PointerCastOpInterface` that both the pointer-resize and the pointer-bitcast implement. If we do that, integer casts should follow a similar pattern for consistency.
2. I was thinking something that had to do with collapsible casts. As of now collapsible casts only consider integer casts IIUC. I think we should be more ambitious in the casts we can collapse. A few handwavy examples of things we should support and we currently don't to the best of my knowledge.
    * pointer-resize casts that don't lose bits overall
    * back-and-forth casts from integer to pointers and back
  
    These have in common the fact that they are c-style casts. And if we have to write all the possible patterns that include all of them they are quite many. If we have the right interface things can get much easier.
    But perhaps the right interface is not `CStyleCastOpInterface`. The `PointerCastOpInterface` I proposed above can help with the first example I made. And `IntPtrCastOp` solves the other.

include/revng/Clift/Clift.td
pfez: I think we have to come up with some sensible general principle here. Ideally I think that if it's possible to do all the verification in tablegen we should.
The most debatable cases are those where you can express some constraints on tablegen, but you still have to write some extra code in `verify`.
My question here is: have you laid out a clear general principle you're trying to follow everywhere? If so explain it and we can discuss it. If you have been doing this change based on your personal preference, we have to try and distill that into a clear general principle that we can follow everywhere.

include/revng/CliftImportModel/ImportModel.h
pfez: This is basically another side of the argument of using `mlir::Type` or some `clift::` type in APIs. I think I understand why you want `mlir::Type` in function arguments, but on return types I think the strictest types is always better.

include/revng/Clift/CliftTypeInterfacesObject.td
pfez: We've discussed this, and decided to never do `ByteSize`

include/revng/Clift/Clift.td
pfez: Let's keep this.

include/revng/Clift/Clift.td
pfez: We ironed out this discussion in person.

include/revng/Clift/Clift.td
pfez: This discussion was also ironed out.

lib/CliftEmitC/CBackend.cpp
pfez: What exactly do you want to do as part of legalization?

share/revng/include/primitive-types.h
```diff
@@ -9,13 +9,18 @@
...
 
-#define static_assert_size(TYPE, EXPECTED_SIZE) \
-  typedef char static_assertion[sizeof(TYPE) == (EXPECTED_SIZE) ? 1 : -1]
+#if __STDC_VERSION__ >= 201112L
+#define revng_static_assert(condition, ...) \
+  _Static_assert(condition, __VA_ARGS__ "")
+#else
+#define revng_static_assert(condition, ...) \
+  typedef char static_assert_typedef[(condition) ? 1 : 0]
+#endif
```
pfez, line 24: I'd just detect C23 as well, and
* in case of C23, `static_assert` is available
* in C11 `#define static_assert(condition, ...) _Static_assert(condition, __VA_ARGS__ "")
* in the other case `#define static_assert` to the negative-sized array

Also, at the end of the file remember to `undef` it.

And for extra paranoia, use a negative size for the array size, not 0. I don't want some compiler not to be pedantic enough on that.

lib/Model/NameBuilder.cpp
```diff
@@ -321,6 +321,7 @@ const std::set<llvm::StringRef> ReservedKeywords = {
   "break_to",
   "continue_to",
   "bit_cast",
+  "static_assert_typedef",
```
pfez, line 324: reserve `static_assert` here as well.

include/revng/Clift/CliftAttributes.td
pfez: Also this discussion was ironed out in person.

lib/Clift/Clift.cpp
pfez: This has been moved to tablegen in the last WIP commit

include/revng/Clift/Clift.td
pfez: Did you forget about this?

pfez:
@vasama-work I finished the next round of review. There is still some open thread. Can you answer to those?

Also, the last commit still seems WIP from the commit message. Is that a leftover or there's is something you're still working on? If you are still working on something that should go in there, what's that?

lib/CliftEmitC/CBackend.cpp
vasama-work: Convert `clift.imm 123 : !T`, where a literal of type `!T` is not representable in the target implementation to a literal followed by cast:
```cpp
%0 = clift.imm 123 : !U
%1 = clift.some-cast %0 : !U -> !T
```

vasama-work:
The WIP is mainly because it's new and unreviewed.

share/revng/include/primitive-types.h
vasama-work: You're right, we should use `static_assert` instead of `_Static_assert` in C23. However, I don't want to define over a keyword. I considered undefining `revng_static_assert` at the end of file, but decided not to do it for now. But there's no important reason for that, and I can do it. I don't think the `revng` prefix here is harmful since the user never sees it in generated code. I think that's better than UB.

lib/CliftTransforms/Expressions.cpp
vasama-work: `PtrCastOp` is gone now. We collapse `BitCastOp`, but not `PtrResizeOp`.

include/revng/CliftImportModel/ImportModel.h
vasama-work: We discussed this in person. Is there still something more to discuss?

include/revng/CliftTransforms/Expressions.td
vasama-work: However we do this, I need it in the implicit cast elision PR. I would like to defer changes in this area until then if that's okay with you?

include/revng/Clift/Clift.td
vasama-work: Yep. I've added an amendment.

include/revng/Clift/Clift.td
vasama-work: Is there something more you'd like to discuss on this? The constraints on `AccessOp` are much stricter than `AddressableType` and are verified in C++.

lib/Model/NameBuilder.cpp
vasama-work: It's already reserved up above along with other standard C names.

include/revng/CliftImportModel/ImportModel.h
pfez: Nope. Just forgot to close this thread as well.

include/revng/Clift/Clift.td
pfez: No. Until we don't split `AccessOp` in two separate operations for direct vs indirect access it's fine like this.

lib/CliftTransforms/Expressions.cpp
pfez: :+1: this was just another obsolete thread :)

include/revng/CliftTransforms/Expressions.td
pfez: Fine by me.

lib/CliftEmitC/CBackend.cpp
vasama-work: If we do this, we can also mark the cast as implicit. Emitting casts directly in the backend wouldn't allow that. So I'm now very convinced that we should do this in legalisation instead.

share/revng/include/primitive-types.h
pfez: > I don't want to define over a keyword

Why not? You would only define over a keyboard whenever you detect that the language version is such that they "keyboard" hasn't been reserved yet. So you wouldn't strictly be defining over it. In this way we don't have to reserve the `revng_static_assert` name. I agree that it's less problematic here to have a `revng_` prefix, but just going with `static_assert` would is strictly better.

lib/CliftEmitC/CBackend.cpp
pfez: Do you have plans to implement this in legalization?
Or you just want to leave this check here until we do that in legalization (if we ever do it)?

pfez:
Ok, drop the WIP from the commit message then.

share/revng/include/primitive-types.h
vasama-work: Actually defining over a keyword is not the problem. That's my mistake. In C23 mode, if we otherwise defined `static_assert`, we would not have to define anything. But in pre-C23 mode `static_assert` is a standard macro and I don't defining over that. So I really prefer using a non-standard name here. Especially since this is undefined at the end of the header. And we don't need to reserve `revng_static_assert` since it is undefined.

lib/CliftEmitC/CBackend.cpp
vasama-work: I can implement it for this PR.

share/revng/include/primitive-types.h
pfez: I think you're referring to the fact that in from C11 (included) to C23 (excluded) the libc defines `#define static_assert _Static_assert`.
If this is the case, I think that's even better, we can just do:
```c
#if __STDC_VERSION__ >= 202311L
// don't do anything since static_assert is present in C23
#elif __STDC_VERSION__ >= 201112L
// do the same as the libc does in C11
#define static_assert _Static_assert
#else
#define static_assert(condition, ...) \
  typedef char static_assert_typedef[(condition) ? 1 : -1]
#endif
```

vasama-work: I really don't want to define `static_assert`, being that it is a standard library definition (in C11) and could already be defined. We would then end up defining over it (causing a warning) and later undefining it would not be desirable.

It's also not desirable to just use any existing C11 `static_assert` definition if available, because we want the C23 behaviour. That's why I define
```c
#define revng_static_assert(condition, ...) \
  _Static_assert(condition, __VA_ARGS__ "")
```
instead of just
```c
#define revng_static_assert _Static_assert
```

Using a separate wrapper really is the best solution here. And it is entirely local to this header, so it doesn't affect anything else.

pfez: > being that it is a standard library definition (in C11)

In C11 `static_assert` it's mandated by the standard to be a macro that expands to `_Static_assert`. See the C11 standard, section 7.2. "Diagnostics `<assert.h>`.

This is exactly what I did in my last proposed snippet. Redefining the macro to the same thing the standard mandates ensures it does what we need, and that if users include `assert.h` we're still fine because redefining a macro with the exact same definition is not an error. I also tried on godbolt with clang 16 (the same version we use), clang 22 and gcc 15 and it works fine. https://godbolt.org/z/azK6PeG66

vasama-work: Do we really want an inconsistent definition depending on the target standard version?

If we just use `#define static_assert _Static_assert` in C11, then `static_assert(1)` is invalid.

pfez: If someone is compiling that header in C11, `static_assert(1)` is invalid according to the standard, since the message-less version did not exist. So what's wrong with that?

vasama-work: Resolved offline.

include/revng/CliftImportModel/CAttributeListBuilder.h
```diff
@@ -138,7 +138,7 @@ class CAttributeListBuilder {
   }
 
   template<ConstexprString Macro>
-  void setOrUpdate(mlir::clift::ValueType Type) {
```
pfez, line 141: Ok, but make a sensible commit message.

lib/CliftEmitC/TypeDependencyGraph.cpp
```diff
@@ -248,8 +248,8 @@ void Builder<Mode>::addDependenciesFrom(const AssociatedNodes Dependent,
   // also depends on the full definition of the depended-on, across typedefs.
   if (ForwardDeclaration
       and mlir::isa<clift::TypedefType>(DefinitionDependedOn)) {
-    auto Underlying = unwrapTypedefs(DefinitionDependedOn);
-    if (auto Defined = mlir::dyn_cast<clift::DefinedType>(Underlying)) {
```
pfez: Ok, but either use a sensible commit message, or squash this in other relevant commits.

include/revng/Clift/Clift.td
```diff
@@ -896,10 +896,10 @@ def Clift_RemOp : Clift_BinaryArithmeticOp<"rem">;
 def Clift_PtrAddOp : Clift_PointerArithmeticOp<"ptr_add", [Commutative]>;
 def Clift_PtrSubOp : Clift_PointerArithmeticOp<"ptr_sub">;
 
-def Clift_PtrDiffOp
-  : Clift_ExpressionOp<"ptr_diff", [Clift_AllTypesMatch<["lhs", "rhs"]>]> {
+def Clift_PtrDiffOp : Clift_ExpressionOp<"ptr_diff"> {
+  let arguments = (ins Clift_AnyPointerType:$lhs,
+                       Clift_AnyPointerType:$rhs);
```
pfez, line 901: Much better now. Thanks!

include/revng/Clift/Clift.td
```diff
@@ -26,62 +26,84 @@ class Clift_Op<string mnemonic, list<Trait> traits = []>
...
 
 def Clift_AnyObjectType
   : Type<CPred<"mlir::clift::unwrapped_isa<mlir::clift::ObjectType>($_self)">,
-         "Clift object type",
-         "mlir::Type">;
+         "object type">;
+
+def Clift_AnyValueType
+  : Type<CPred<"mlir::clift::unwrapped_isa<mlir::clift::ValueType>($_self)">,
+         "value type">;
```
pfez, line 37: The commit message still says WIP. I believe it's not WIP anymore.

include/revng/CliftImportModel/CAttributeListBuilder.h
vasama-work: Yes, the amendments are just going to be squashed into the commit above.

lib/CliftEmitC/TypeDependencyGraph.cpp
vasama-work: Yes, the amendments are just going to be squashed into the commit above.

include/revng/Clift/Clift.td
vasama-work: Yes, don't worry about the WIPs and the amend commits. Just check if you're happy with the commit contents.

lib/CliftEmitC/CBackend.cpp
pfez: Closing this since this just goes away in one of the following commits where you deal with this in legalization.

lib/CliftTransforms/Legalization.cpp
```diff
@@ -335,15 +323,97 @@ struct IntegerPromotionPattern : mlir::OpRewritePattern<OpT> {
...
+                                  IntegerType Type,
+                                  mlir::PatternRewriter &Rewriter) const {
+    if (isRepresentableLiteralSize(Type.getSize()))
+      return mlir::failure();
+
+    // Sizes in the range [sizeof(int), 8] must be representable in the target.
+    uint64_t NewSize = std::clamp<uint64_t>(Type.getSize(), IntSize, 8);
+
+    revng_assert(NewSize != Type.getSize());
+    revng_assert(isRepresentableLiteralSize(NewSize));
```
pfez, line 391: So this crashes if the literal size requires more than 8 bytes.
I'm wondering if we should just verify that in `ImmediateOp`.
What do you think?

vasama-work: I'm not sure in which case it would crash. We allow e.g. 128-bit immediates, but their values are still capped to 64 bits. That's why we can always clamp the final immediate width to 64. I don't think there's a need for any changes on `ImmediateOp`. If in the future we replace the value with an arbitrary precision integer, then we will have to do something more complicated in legalisation. E.g. combining multiple immediates using shifts.

pfez: nevermind.

PR #575, opened 2026-03-27
EmitFieldAccesses: add support for `pointer` as `array`

andrealinux1:
This enable us to wrap `BasePointerType` into an implicit array. This enables pointer arithmetic rewriting of `*(p + i)` as `p[i]`.

### DIFF
--- a/lib/CliftTransforms/BestTraversal.cpp
+++ b/lib/CliftTransforms/BestTraversal.cpp
@@ -3,6 +3,7 @@
 //
 #include <compare>
 #include <cstdint>
+#include <limits>
 #include <optional>
 
 #include "revng/ADT/RecursiveCoroutine.h"
@@ -105,6 +106,34 @@ void Traversal::dump() const {
   Log.flush();
 }
 
+mlir::Type deriveBaseType(mlir::Value BasePointer) {
+
+  auto BasePtrType = getPointerType(BasePointer.getType());
+  revng_assert(BasePtrType);
+  auto PointeeType = BasePtrType.getPointeeType();
+  auto UnderlyingType = dealias(PointeeType, /*IgnoreQualifiers=*/true);
+
+  // If the pointee is a struct, union, or array, use it directly — the
+  // traversal analyzer can walk its fields and arrays.
+  // Wrapping also a `struct`, would mean rewriting a constant offset access
+  // into it as `p[0].field` instead of `p->field`.
+  if (mlir::isa<StructType, UnionType, ArrayType>(UnderlyingType)) {
+    return PointeeType;
+  }
+
+  // We don't want to wrap `void` or `function` `Type`s into `array`s.
+  if (not isObjectType(PointeeType)) {
+    return PointeeType;
+  }
+
+  // In all the other situations, wrap into an implicit array. We use
+  // `ImplicitArrayNumElements` a _very large_ array in order to cover any
+  // reasonable constant offset.
+  static constexpr uint64_t
+    ImplicitArrayNumElements = std::numeric_limits<uint64_t>::max();
+  return ArrayType::get(PointeeType, ImplicitArrayNumElements);
+}
+
 namespace {
 
 // =============================================================================
@@ -128,8 +157,8 @@ static bool isCompatible(const ArrayPath &Path, llvm::APInt BaseOffset) {
     // `BaseOffset`, consuming it
     if (BaseOffset.uge(Shape.Stride)) {
 
-      // If we're jumping over the whole array, past it, we just bail out
-      if (BaseOffset.uge(Shape.Stride * Shape.NumElements)) {
+      // If we're jumping over the whole array, past it, we just bail out.
+      if (BaseOffset.udiv(Shape.Stride).uge(Shape.NumElements)) {
         return false;
       }
 
@@ -658,9 +687,11 @@ private:
                         const ArrayPath &ArrayPath);
 
   /// Obtain all the explicit rewritings of the input `Arithmetic` following all
-  /// the `ArrayPath`s for the `BaseType`
+  /// the `ArrayPath`s for the `BaseType`. We explicitly specify the `BaseType`
+  /// to handle the _pointer as array_ situation.
   std::vector<PointerArithmetic>
-  toExplicitArrayAccesses(const PointerArithmetic &Arithmetic);
+  toExplicitArrayAccesses(const PointerArithmetic &Arithmetic,
+                          mlir::Type BaseType);
 
   /// Helper which trivially spill a `PointerArithmetic` into a `Traversal`
   Traversal toTraversal(const PointerArithmetic &PA,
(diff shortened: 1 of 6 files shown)


lib/CliftTransforms/BestTraversal.cpp
```diff
@@ -105,6 +105,33 @@ void Traversal::dump() const {
...
+
+  // We don't want to wrap `void` or `function` `Type`s into `array`s.
+  if (not isObjectType(PointeeType)) {
+    return PointeeType;
+  }
+
+  // In all the other situations, wrap into an implicit array. We use
+  // `ImplicitArrayNumElements` to virtually represent a _very large_ array in
+  // order to cover any reasonable constant offset.
+  static constexpr uint64_t ImplicitArrayNumElements = 1ULL << 32;
```
pfez: I believe you want `std::numeric_limits<uint64_t>::max()`.
Why shifting by 32?

lib/CliftTransforms/FieldAccessReplacement.cpp
```diff
@@ -264,23 +263,39 @@ void Replacement::replace(ExpressionOpInterface PointerToReplace,
...
+      // We need to explicitly handle the `pointer as array` case, where
+      // `CurrentValue` is not a `ptr<T>` of `ArrayType` (we virtually wrap it
+      // ourselves), so the `indirection` and `cast<decay>` is not needed.
+      auto CurrentValueType = dealias(CurrentValue.getType(),
+                                      /*IgnoreQualifiers=*/true);
+      bool IsPointerAsArray = false;
+      if (auto PtrType = mlir::dyn_cast<clift::PointerType>(CurrentValueType)) {
+        auto Pointee = dealias(PtrType.getPointeeType(),
+                               /*IgnoreQualifiers=*/true);
+        IsPointerAsArray = not mlir::isa<clift::ArrayType>(Pointee);
```
pfez: Extract this into a helper function.

lib/CliftTransforms/PointerArithmetic.cpp
```diff
@@ -228,6 +228,16 @@ PointerArithmeticBuilder::computePointerArithmetic(ExpressionOpInterface
     return std::nullopt;
   }
 
+  // We skip every replacement where the `PointerToReplace` is equal to the
+  // `BasePointer`. Replacing such expression would create a circular reference:
+  // the new `clift.subscript` uses the `BasePointer` as operand, and
+  // `replaceAllUsesWith` would replace the `BasePointer` uses with the new
+  // `clift.subscript` result, and this insert the circular reference between
+  // mlir `Value`s.
+  if (PointerToReplace->getResult(0) == Result->BasePointer) {
```
pfez, line 237: Interesting. I did not expect this could happen. Does this even happen with a non-empty `Result->Offset`? If so, there is something weird going on.

lib/CliftTransforms/FieldAccessReplacement.cpp
```diff
@@ -264,23 +263,39 @@ void Replacement::replace(ExpressionOpInterface PointerToReplace,
...
+      // the needed `IndirectionOp` and `Decay` cast accordingly.
+      clift::ValueType ArrayElementType;
+
+      // We need to explicitly handle the `pointer as array` case, where
+      // `CurrentValue` is not a `ptr<T>` of `ArrayType` (we virtually wrap it
+      // ourselves), so the `indirection` and `cast<decay>` is not needed.
+      auto CurrentValueType = dealias(CurrentValue.getType(),
+                                      /*IgnoreQualifiers=*/true);
+      bool IsPointerAsArray = false;
+      if (auto PtrType = mlir::dyn_cast<clift::PointerType>(CurrentValueType)) {
```
pfez: I believe you want `isPointer`, not `dyn_cast<clift::PointerType>`.

lib/CliftTransforms/FieldAccessReplacement.cpp
pfez: Actually, just inline this in the only place where you check `IsPointerAsArray`.

andrealinux1: Ok, the check is now inlined [here](https://github.com/andrealinux1/revng/blob/cecf6379d983a6118625ee8e11b30b6435d24ed3/lib/CliftTransforms/FieldAccessReplacement.cpp#L273)

lib/CliftTransforms/FieldAccessReplacement.cpp
andrealinux1: I used the `getPointerType` helper [here](https://github.com/andrealinux1/revng/blob/cecf6379d983a6118625ee8e11b30b6435d24ed3/lib/CliftTransforms/FieldAccessReplacement.cpp#L272) so this together with the _inlining of the only use_ of the comment above, is more compact

lib/CliftTransforms/BestTraversal.cpp
andrealinux1: I changed it to use the `max` limit [here](https://github.com/andrealinux1/revng/blob/cecf6379d983a6118625ee8e11b30b6435d24ed3/lib/CliftTransforms/BestTraversal.cpp#L133).

I went with the intermediate approach of shifting and not going to `max`, because I was worried about overflow when checking the compatibility of the `ArrayPath`. We performed our check as `if (BaseOffset.uge(Shape.Stride * Shape.NumElements))`, so as soon the stride was > 1, and we had a _virtual_ with `ImplicitArrayNumElements` set to `max()`  we would have add overflow I think.

The new check [here](https://github.com/andrealinux1/revng/blob/6a4ad12e5da2a8a503f42928810c54a48e2749d2/lib/CliftTransforms/BestTraversal.cpp#L161), avoids the overflow.

lib/CliftTransforms/PointerArithmetic.cpp
andrealinux1: No, I don't foresee it happening with a non empty offset.

It triggers when the expression is the `BasePointer` without arithmetic, e.g., `addressof(local_variable)`. The `BasePointer` is the `addressof` itself, with `offset = 0`. Then the `PointerToReplace` is rewritten with the `subscript`, and `replaceAllUsesWith` is called, and this insert the circular dependency as explained in the comment.

lib/CliftTransforms/FieldAccessReplacement.cpp
pfez: Do this instead. I'm worried about `PtrType` begin visible and usable down below.

```c++
if (auto PtrType = getPointerType(CurrentValue.getType());
    PtrType and not mlir::isa<ArrayType>(dealias(PtrType.getPointeeType(),
                                                   /*IgnoreQualifiers=*/true))) {
        ArrayElementType = PtrType.getPointeeType();
```

lib/CliftTransforms/PointerArithmetic.cpp
pfez: I kinda expected that. So it's worth an assertion that the offset is zero and has an empty linear combination.

lib/CliftTransforms/BestTraversal.cpp
pfez: Seems legit. Thanks!

lib/CliftTransforms/FieldAccessReplacement.cpp
pfez: See my other comment on that. Closing this thread.

lib/CliftTransforms/FieldAccessReplacement.cpp
andrealinux1: Uh ok, changed this [here](https://github.com/andrealinux1/revng/blob/28d9f919613606f477d8b336ce6edfb8b2d82ab1/lib/CliftTransforms/FieldAccessReplacement.cpp#L272)

lib/CliftTransforms/PointerArithmetic.cpp
andrealinux1: Right, better control what's slips in here, changed this [here](https://github.com/andrealinux1/revng/blob/28d9f919613606f477d8b336ce6edfb8b2d82ab1/lib/CliftTransforms/PointerArithmetic.cpp#L241)

PR #576, opened 2026-04-07
EmitFieldAccesses: propagate `Type` through `IndirectionOp`

andrealinux1:
After `emit-fa` perform a replacement, propagate the `Type` of the `access` to the `IndirectionOp` uses.

If such propagation takes place, perform another run of `emit-fa` in order to unlock further replacements.

### DIFF
--- a/include/revng/CliftTransforms/Expressions.td
+++ b/include/revng/CliftTransforms/Expressions.td
@@ -140,4 +140,12 @@ def DirectAccessPattern
         (Clift_AccessOp $arg, (NativeCodeCall<"mlir::UnitAttr(nullptr)">), $member_index),
         [(Constraint<CPred<"not $0">> $indirect)]>;
 
+// x[0] -> *x
+// Constraint: the subscript index is a constant zero.
+def SubscriptZeroPattern
+  : Pat<(Clift_SubscriptOp $ptr, (Clift_ImmediateOp $value)),
+        (Clift_IndirectionOp $ptr),
+        [(Constraint<CPred<"$0.getValue().isZero()">> $value)]>;
+
+
 #endif
--- a/lib/CliftTransforms/BestTraversal.cpp
+++ b/lib/CliftTransforms/BestTraversal.cpp
@@ -198,17 +198,20 @@ static uint64_t commonPrefixStrides(const llvm::ArrayRef<ArrayShape> &LHS,
 }
 
 /// The `typeDistance` function computes the distance between two `Type`s.
-/// Returns 0 if the types are equal (after stripping `typedef`s), or infinity
-/// otherwise. The return type is kept as `uint64_t` to allow future refinement
-/// of the scoring criterion, enabling more fine grained control over this
-/// score.
+/// Returns 0 if the types are exactly equal (after stripping `typedef`s), 1 if
+/// they only differ in CV-qualifiers, or infinity otherwise. The return type is
+/// kept as `uint64_t` to allow future refinement of the scoring criterion,
+/// enabling more fine grained control over this score.
 static uint64_t typeDistance(mlir::Type Explicit, mlir::Type Ideal) {
+  if (unwrapTypedefs(Explicit) == unwrapTypedefs(Ideal)) {
+    return 0;
+  }
 
-  // Unwrap any typedefs to compare the underlying types
-  Explicit = unwrapTypedefs(Explicit);
-  Ideal = unwrapTypedefs(Ideal);
+  if (equivalentUnwrapped(Explicit, Ideal)) {
+    return 1;
+  }
 
-  return Explicit == Ideal ? 0 : std::numeric_limits<uint64_t>::max();
+  return std::numeric_limits<uint64_t>::max();
 }
 
 // =============================================================================
--- a/lib/CliftTransforms/FieldAccessReplacement.cpp
+++ b/lib/CliftTransforms/FieldAccessReplacement.cpp
@@ -71,8 +71,9 @@ struct Replacement {
                           const PointerArithmetic &Arithmetic,
                           const Traversal &BestTraversal);
 
-  /// This method performs the actual `clift` IR rewriting,
-  void replace(ExpressionOpInterface PointerToReplace,
+  /// This method performs the actual `clift` IR rewriting. Returns `true` if
+  /// a type was propagated through an indirection.
+  bool replace(ExpressionOpInterface PointerToReplace,
                const PointerArithmetic &Arithmetic) const;
 };
 
@@ -205,8 +206,9 @@ Replacement Replacement::make(unsigned PointerBitWidth,
   return Result;
 }
 
-void Replacement::replace(ExpressionOpInterface PointerToReplace,
+bool Replacement::replace(ExpressionOpInterface PointerToReplace,
                           const PointerArithmetic &Arithmetic) const {
+  bool PropagatedThroughIndirection = false;
 
   // We need the `PointerSize` in order to generate the `ImmediateOp`s used to
   // access the `struct` fields and `array` members, and to generate the
@@ -357,6 +359,12 @@ void Replacement::replace(ExpressionOpInterface PointerToReplace,
                                              CurrentValuePointerType,
                                              CurrentValue);
 
+  // After we emit the `addressof`, we save the resulting `RichValue`, which may
+  // contain a _rich_ type information of the emitted access. We collect this
+  // here before the subsequent `cast` strips it of the type information`. We
+  // then propagate the type information into `indirection` uses.
+  mlir::Value RichValue = CurrentValue;
+
   // If there is a non-null `LeftoverOffset`, we add it as integer arithmetic
   if (not LeftoverOffset.BaseOffset.isZero()
       or not LeftoverOffset.LinearCombination.empty()) {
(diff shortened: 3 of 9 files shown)


lib/CliftTransforms/BestTraversal.cpp
```diff
@@ -216,7 +216,7 @@ static uint64_t typeDistance(mlir::Type Explicit, mlir::Type Ideal) {
                      /*IgnoreQualifiers=*/true);
   Ideal = dealias(mlir::cast<ValueType>(Ideal), /*IgnoreQualifiers=*/true);
 
-  return Explicit == Ideal ? 0 : std::numeric_limits<uint64_t>::max();
+  return equivalent(Explicit, Ideal) ? 0 : std::numeric_limits<uint64_t>::max();
```
pfez: Nice. But can we have a unit test for this case? I don't want this kind of thing to slip through again.

lib/CliftTransforms/FieldAccessReplacement.cpp
```diff
@@ -434,6 +440,76 @@ void Replacement::replace(ExpressionOpInterface PointerToReplace,
...
+        // the `indirection` safely
+        auto IndirResultType = mlir::cast<ValueType>(IndirOp.getResult()
+                                                       .getType());
+        if (RichPointeeType.getByteSize() != IndirResultType.getByteSize()) {
+          continue;
+        }
+
+        // In case the types are already equal, we do not have anything do
+        // rewrite
+        if (RichPointeeType == IndirResultType) {
```
pfez: `equivalent` perhaps?

pfez: actually no, `==` is fine here, since if it was off-by-const we'd still need recreate the indirection and att the `const` cast via bitcast below.

lib/CliftTransforms/FieldAccessReplacement.cpp
```diff
@@ -434,6 +440,76 @@ void Replacement::replace(ExpressionOpInterface PointerToReplace,
     PointerToReplace->getResult(0).replaceAllUsesWith(CurrentValue);
   }
 
+  // We want to try to propagate type information from the _rich_ `addressof`,
+  // to a potential `indirection` (originally using the untyped pointer, e.g.,
+  // `ptr<generic8>`) using it.
+  auto RichPtrType = getPointerType(RichValue.getType());
```
pfez: No abbreviations. `RichPointerType` should fit just fine.

lib/CliftTransforms/FieldAccessReplacement.cpp
```diff
@@ -434,6 +440,76 @@ void Replacement::replace(ExpressionOpInterface PointerToReplace,
...
+        // In case the types are already equal, we do not have anything do
+        // rewrite
+        if (RichPointeeType == IndirResultType) {
+          continue;
+        }
+
+        // Build a `indirection` with the typed operand, so that the `result`
+        // type is inferred from `RichValue` pointee type
+        Builder.setInsertionPoint(IndirOp);
+        auto NewIndirOp = Builder.create<IndirectionOp>(PointerToReplaceLoc,
```
pfez: No abbreviations. `NewIndirection`.

lib/CliftTransforms/FieldAccessReplacement.cpp
```diff
@@ -434,6 +440,76 @@ void Replacement::replace(ExpressionOpInterface PointerToReplace,
...
+    // We are only interested in propagating _rich_ types which are wrapped in
+    // `ptr<ptr<...>>`, so that the `indirection` load a pointer through which
+    // further field accesses can be chained
+    auto RichPointeeType = RichPtrType.getPointeeType();
+    if (isPointerType(RichPointeeType)) {
+
+      // We want to search for `indirection` users to propagate the _rich_ type
+      // information to
+      for (auto *User : CurrentValue.getUsers()) {
+        auto IndirOp = mlir::dyn_cast<IndirectionOp>(User);
```
pfez: No abbreviations. `Indirection` should work.

lib/CliftTransforms/FieldAccessReplacement.cpp
```diff
@@ -434,6 +440,76 @@ void Replacement::replace(ExpressionOpInterface PointerToReplace,
...
+        if (not IndirOp)
+          continue;
+
+        // We must check that the peeled pointee type of `RichValue` has the
+        // same byte size as the `indirection` result, so that we can _retype_
+        // the `indirection` safely
+        auto IndirResultType = mlir::cast<ValueType>(IndirOp.getResult()
```
pfez: No abbreviations. `NewResultType`.

lib/CliftTransforms/BestTraversal.cpp
andrealinux1: There is now a unit test for this here https://github.com/andrealinux1/revng/blob/f5cb6e06ebcefe8310684a4ee637b8dc5637d51e/tests/unit/mlir_lit_tests/emit-field-accesses/struct-const-field-access.mlir

lib/CliftTransforms/FieldAccessReplacement.cpp
andrealinux1: Mmm I guess this is resolved I guess?

lib/CliftTransforms/FieldAccessReplacement.cpp
andrealinux1: Removed the abbreviation

lib/CliftTransforms/FieldAccessReplacement.cpp
andrealinux1: Removed the abbreviation

lib/CliftTransforms/FieldAccessReplacement.cpp
andrealinux1: Removed the abbreviation

lib/CliftTransforms/FieldAccessReplacement.cpp
andrealinux1: Removed the abbreviation

lib/CliftTransforms/BestTraversal.cpp
```diff
@@ -197,18 +197,21 @@ static uint64_t commonPrefixStrides(const llvm::ArrayRef<ArrayShape> &LHS,
...
 /// of the scoring criterion, enabling more fine grained control over this
 /// score.
 static uint64_t typeDistance(mlir::Type Explicit, mlir::Type Ideal) {
-
-  // Unwrap any typedefs to compare the underlying types
-  Explicit = unwrapTypedefs(Explicit);
-  Ideal = unwrapTypedefs(Ideal);
-
-  return Explicit == Ideal ? 0 : std::numeric_limits<uint64_t>::max();
+  return unwrappedEquivalent(Explicit, Ideal) ?
```
pfez: This way we don't distinguish `const` from non-`const`. Let's return 0 if they are actually equal, 1 if they are off-by-`const` and `max` in all the other cases. At some point we'll come down to properly implementing type distance. But this is a low hanging fruit.

lib/CliftTransforms/BestTraversal.cpp
```diff
@@ -197,18 +197,21 @@ static uint64_t commonPrefixStrides(const llvm::ArrayRef<ArrayShape> &LHS,
   return Count;
 }
 
+/// Determine if the two types are equivalent after unwrapping typedefs and
+/// ignoring Clift CV-qualifiers.
+static bool unwrappedEquivalent(mlir::Type Lhs, mlir::Type Rhs) {
+  return equivalent(unwrapTypedefs(Lhs), unwrapTypedefs(Rhs));
+}
```
pfez: Like Lauri said in one of the comments on mattermost, let's put this in clift directly, so other users can benefit.

include/revng/CliftTransforms/Expressions.td
```diff
@@ -140,4 +140,14 @@ def DirectAccessPattern
         (Clift_AccessOp $arg, (NativeCodeCall<"mlir::UnitAttr(nullptr)">), $member_index),
         [(Constraint<CPred<"not $0">> $indirect)]>;
 
+// &(x[0]) -> x
```
pfez: Following Lauri's advice, change this to turn `p[0]` into `*p`. The rest should just click.

lib/CliftTransforms/BestTraversal.cpp
andrealinux1: Right, the function [here](https://github.com/andrealinux1/revng/blob/4e3af009fe52761e12029c5cc7b9ca17b712025a/lib/CliftTransforms/BestTraversal.cpp#L205) now reflects your intended behavior. I also modified the unit test [here](https://github.com/andrealinux1/revng/blob/4e3af009fe52761e12029c5cc7b9ca17b712025a/tests/unit/mlir_lit_tests/emit-field-accesses/struct-const-field-access.mlir) to take this into account.

lib/CliftTransforms/BestTraversal.cpp
andrealinux1: The helper is now in the `CliftTypes.h` header [here](https://github.com/andrealinux1/revng/blob/4e3af009fe52761e12029c5cc7b9ca17b712025a/include/revng/Clift/CliftTypes.h#L140). The naming has also been changed to reflect the suggestion.

include/revng/CliftTransforms/Expressions.td
andrealinux1: Ok the rewrite now performs the suggested rewrite [here](https://github.com/andrealinux1/revng/blob/4e3af009fe52761e12029c5cc7b9ca17b712025a/include/revng/CliftTransforms/Expressions.td#L143). However, in the unit tests, I also kept [here](https://github.com/andrealinux1/revng/blob/4e3af009fe52761e12029c5cc7b9ca17b712025a/tests/unit/mlir_lit_tests/expression-rewrites/subscript-zero.mlir#L29) a test checking for the full rewriting we expect (so we don't loose it if the transformation composition stops to work at a certain point and we have a regression).

PR #577, opened 2026-04-08
Fix Clift comment emission

vasama-work:
1. Currently empty lines in comments are emitted without slashes:
A comment with content `A\n\nB` comes out as:
   ```
   //A

   //B
   ```
   With this change, the missing slashes are emitted.

2. Effectively the same problem applies to Doxygen comments, where (with the previous change), empty lines come out without the relevant prefixes:
   ```
   /// A
   //
   /// B
   ```
   With this change, the missing slash (but not the trailing space) is emitted.

3. Because `IndentingEmitter` is no longer used to emit anything non-trivial (e.g. the line comment slashes), the `IndentableEmitter` concept is removed and the indentation string of `IndentingEmitter` is instead made configurable. The default of `"  "` matches what was emitted by `PTMLStreamEmitter`.

### DIFF
--- a/include/revng/ADT/LineRange.h
+++ b/include/revng/ADT/LineRange.h
@@ -1,0 +1,69 @@
+#pragma once
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include <algorithm>
+#include <string_view>
+
+/// Iterates over the lines of a string. If the string is empty, the range is
+/// empty. Each element line (except the last one) ends in a newline character.
+class LineRange {
+public:
+  class Sentinel {
+  public:
+    explicit Sentinel() = default;
+  };
+
+  class Iterator {
+    const char *LineBegin = nullptr;
+    const char *LineEnd = nullptr;
+    const char *End = nullptr;
+
+  public:
+    using value_type = std::string_view;
+    using difference_type = ptrdiff_t;
+
+    Iterator() = default;
+
+    explicit Iterator(std::string_view String) :
+      LineBegin(String.data()),
+      LineEnd(LineBegin),
+      End(LineBegin + String.size()) {
+      if (LineBegin != End)
+        ++(*this);
+    }
+
+    [[nodiscard]] std::string_view operator*() const {
+      return std::string_view(LineBegin, LineEnd);
+    }
+
+    Iterator &operator++() & {
+      const char *LF = std::find(LineEnd, End, '\n');
+      LineBegin = LineEnd;
+      LineEnd = LF == End ? End : LF + 1;
+      return *this;
+    }
+
+    [[nodiscard]] Iterator operator++(int) & {
+      auto Result = *this;
+      ++(*this);
+      return Result;
+    }
+
+    [[nodiscard]] friend bool operator==(const Iterator &Self,
+                                         const Sentinel &) {
+      return Self.LineBegin == Self.End;
+    }
+  };
+
+private:
+  Iterator Begin;
+
+public:
+  explicit LineRange(std::string_view String) : Begin(String) {}
+
+  [[nodiscard]] Iterator begin() const { return Begin; }
+  [[nodiscard]] Sentinel end() const { return Sentinel(); }
+};
--- a/include/revng/PTML/DoxygenEmitter.h
+++ b/include/revng/PTML/DoxygenEmitter.h
@@ -16,6 +16,9 @@ struct DoxygenCommentConfiguration {
   llvm::StringRef LinePrefix;
 };
 
+/// An Emitter capable of emitting Doxygen comments. While at the time of
+/// creation only C comments are emitted (using CDoxygenEmitter via
+/// CTokenEmitter::CommentEmitter), future use for assembly comments is planned.
 template<PTMLEmitter EmitterT>
 class DoxygenEmitter : EmitterT {
   DoxygenCommentConfiguration Configuration;
--- a/include/revng/PTML/DoxygenEmitter.h
+++ b/include/revng/PTML/DoxygenEmitter.h
@@ -4,6 +4,7 @@
 // This file is distributed under the MIT License. See LICENSE.md for details.
 //
 
+#include "revng/ADT/LineRange.h"
 #include "revng/PTML/Constants.h"
 
 namespace ptml {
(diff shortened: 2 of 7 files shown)


include/revng/ADT/LineIterator.h
```diff
@@ -0,0 +1,67 @@
...
+
+  LineIterator &operator++() & {
+    const char *LF = std::find(LineEnd, End, '\n');
+    LineBegin = LineEnd;
+    LineEnd = LF == End ? End : LF + 1;
+    return *this;
+  }
+
+  [[nodiscard]] LineIterator operator++(int) & {
+    auto R = *this;
```
aleclearmind: No abbreviations

include/revng/ADT/LineIterator.h
```diff
@@ -0,0 +1,67 @@
...
+
+#include <algorithm>
+#include <string_view>
+
+class LineSentinel {
+public:
+  explicit LineSentinel() = default;
+};
+
+class LineIterator {
```
aleclearmind: 1. Docs. Does this yield `asd\n` or just `asd`?
2. Have you considered implementing this as a generator? This iterator is fancy but I hate writing iterators `:P`

include/revng/ADT/LineIterator.h
```diff
@@ -0,0 +1,67 @@
+#pragma once
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include <algorithm>
+#include <string_view>
+
+class LineSentinel {
```
aleclearmind: `LineIteratorEndSentinel`?

include/revng/ADT/LineIterator.h
```diff
@@ -0,0 +1,67 @@
...
+public:
+  explicit LineSentinel() = default;
+};
+
+class LineIterator {
+  const char *LineBegin;
+  const char *LineEnd;
+  const char *End;
```
aleclearmind: Always `nullptr`/`0`/`false` initialize all primitive types (C.14), no matter what.

lib/PTML/CTokenEmitter.cpp
```diff
@@ -773,13 +774,15 @@ CTokenEmitter::CommentEmitter::~CommentEmitter() {
...
 void CTokenEmitter::CommentEmitter::emit(llvm::StringRef Content) {
   if (not Content.empty()) {
-    for (auto [I, R] : llvm::enumerate(std::views::split(Content, '\n'))) {
-      llvm::StringRef Line = std::string_view(R.begin(), R.end());
+    bool EmitLinePrefix = IsAtBeginningOfLine;
 
-      if (I != 0)
-        PTML.emit("\n");
+    for (auto Line : LineRange(Content)) {
+      if (std::exchange(EmitLinePrefix, true))
```
aleclearmind, line 780: TIL

include/revng/PTML/DoxygenEmitter.h
```diff
@@ -15,45 +16,20 @@ struct DoxygenCommentConfiguration {
...
-  }
-};
-
-} // namespace detail
-
 template<PTMLEmitter EmitterT>
-class DoxygenEmitter : IndentingEmitter<detail::DoxygenEmitterBase<EmitterT>> {
-  using Base = detail::DoxygenEmitterBase<EmitterT>;
-  using IndentingEmitter = IndentingEmitter<Base>;
+class DoxygenEmitter : EmitterT {
```
aleclearmind, line 23: Why do we need this to be templated again?
To emit doxygen in disassembly?

Add docs.

include/revng/ADT/LineIterator.h
vasama-work: 1. The docs are on `LineRange`. I can put it on both if you like though. For `"A\nB"` it yields `{ "A\n", "B" }`.
2. I did not, and I would not. Generators are very heavy and this iterator is very simple.

lib/PTML/CTokenEmitter.cpp
vasama-work: It's not sure what you learned, but okay. 😄

include/revng/ADT/LineIterator.h
vasama-work: Perhaps I should make `LineIterator` and `LineSentinel` members of `LineRange` and then there would be no confusion?

EDIT: I did that.

include/revng/PTML/DoxygenEmitter.h
vasama-work: The fundamental reason is to decouple it from the `CTokenEmitter`. Yes, perhaps in the future we may wish to emit comments in assembly. In any case, the templating doesn't introduce any complexity here.

lib/PTML/CTokenEmitter.cpp
aleclearmind: `std::exchange`

include/revng/ADT/LineRange.h
```diff
@@ -0,0 +1,69 @@
...
+//
+
+#include <algorithm>
+#include <string_view>
+
+/// Iterates over the lines of a string. If the string is empty, the range is
+/// empty. Each element line (except the last one) ends in a newline character.
+class LineRange {
+public:
+  class sentinel {
```
aleclearmind: Always use our casing convention, unless we need to be compatible with something from the standard library.

include/revng/ADT/LineIterator.h
aleclearmind: 2. > Generators are very heavy

   Are they?
   This is not performance critical and the readability gains would be quite good. We accept the overhead in far more critical code with recursive coroutine.

include/revng/PTML/DoxygenEmitter.h
aleclearmind: Add docs about the future uses for the template (assembly).

include/revng/ADT/LineRange.h
vasama-work: > unless we need to be compatible with something from the standard library

How do we define this?

The standard naming scheme for iterator and sentinel member types is `iterator` and `sentinel`.

include/revng/ADT/LineIterator.h
vasama-work: They are. We accept it with recursive coroutines because we have to. Because to avoid recursive coroutine and stack overflow both would require turning all our code inside out with stacks and loops. Anyway, if you don't want to have this iterator, I can do the same thing with internal iteration (think visitor).

include/revng/ADT/LineRange.h
aleclearmind: Do you get a compile error if you use `Iterator`? If not, let's go with it.

Let's deviate only when strictly necessary.

include/revng/ADT/LineIterator.h
aleclearmind:
```
$ ssh claudietto@macsux
$ cd /home/claudietto/test
$ claude --continue
  GCC 13 vs Clang 18, both -O2                                                                                                                                               
                                                                                                                                                                             
  ┌────────────────────────┬─────────────┬───────────┐                                                                                                                       
  │        Scenario        │ Clang ratio │ GCC ratio │                                                                                                                       
  ├────────────────────────┼─────────────┼───────────┤                                                                                                                       
  │ 1M lines × 40 chars    │ 1.04x       │ 1.28x     │
  ├────────────────────────┼─────────────┼───────────┤
  │ 5M lines × 1 char      │ 1.04x       │ 2.38x     │                                                                                                                       
  ├────────────────────────┼─────────────┼───────────┤
  │ 100K lines × 500 chars │ 0.99x       │ 1.03x     │                                                                                                                       
  └────────────────────────┴─────────────┴───────────┘                                                                                                                       
```

`clang` seems to be rather good at this.

I really want to push the adoption of generators over iterators, in particular for non-performance critical code.

This said, we can go with this iterator, but it's much more preferable to have a generator, unless there's a provably good reason not to use generators, which are much easier to read and maintain.

Keep in mind that you're not the only one working on the codebase and most of us really don't want to deal with iterators.

vasama-work: How's the comparison for 1M iterations of 1 line with 1 char? It's not the running cost I'm worried about, but the setup and teardown.

aleclearmind: I expanded the tests to measure the set up time and the data doesn't look bad at all.

`clang` manages to get rid of heap allocations and inline everything. This is one of the cases where you actually want to use coroutines because the compiler did a great job.

In any case, this is not performance critical code, so the observations above hold.

vasama-work: Mmhm. I'm still quite sceptical of coroutine optimisations. They're very fragile.

aleclearmind: I don't have much to add. `clang` does HALO, this is not performance critical code, generators look great, we want to push them, compilers are going to get better at this.

Let's not FUD and embrace coroutines! `:)`

vasama-work: But in any case, you're okay with the status quo and we can resolve the conversation, am I correct?

PR #579, opened 2026-04-20
Clift implicit cast elision

vasama-work:


include/revng/Clift/CliftTypeInterfacesDefined.td
```diff
@@ -7,8 +7,9 @@
 
 include "revng/Clift/CliftTypeInterfacesBasic.td"
 
-def Clift_DefinedType : Clift_TypeInterface<"DefinedType",
-                                            [Clift_AliasableType]> {
+def Clift_DefinedType
+  : Clift_TypeInterface<"DefinedType",
+                        [Clift_AliasableType, Clift_AddressableType]> {
```
pfez, line 12: This is now possible thanks to the commit you backported in LLVM right? Nice.

share/revng/compile-flags.cfg
```diff
@@ -152,6 +152,7 @@
 # Both cases should be handled correctly, so eventually we should be able to
 # re-enable this warning.
 -Wno-unused-but-set-variable
+-Wno-unused-but-set-parameter
```
pfez, line 164: It's fine if you disable this, but add a comment to describe what was the scenario that forces you to disable this. Look at other disabled warnings in this file for examples.

include/revng/Clift/CliftTypeInterfacesDefined.td
vasama-work: Yes.

python/revng/internal/pipeline.yml
```diff
@@ -384,6 +384,12 @@ branches:
             - canonicalize
             - optimize-expressions
             - terminal-branch-complement-hoisting
+      - pipe: import-function-data-model
+        arguments: [clift-functions]
+      - pipe: pure-mlir-passes-pipe
+        arguments: [clift-functions]
+        configuration:
+          passes:
```
pfez, line 397: You forgot `elide-implicit-casts` after `c-legalization`.

share/revng/compile-flags.cfg
vasama-work: I'm not sure what leads to this, so I'm not sure I can make a very good comment on it.

lib/ABI/Definition.cpp
```diff
@@ -418,4 +426,16 @@ Definition::hasNaturalAlignment(const model::TypeDefinition &Type,
   return Result->IsNatural;
 }
 
+CDataModel Definition::getDataModel() const {
+  CDataModel DM = CDataModel::getDefaultDataModel(getPointerSize());
```
pfez, line 479: I don't like that we always start from the same default data model and then we only have to specify some fields that override the default.

This requires someone that edits the model by hand to know what's the default data model, and that's only defined in C++ code, not in user facing code.

I think that it would be better that each ABI fully defines its own default data model in the schema.

share/revng/abi/AAPCS64.yml
```diff
@@ -98,3 +98,7 @@ FloatingPointScalarTypes:
   - Size: 4
   - Size: 8
   - Size: 16
+
+CStandardTypes:
+  - Name: long double
+    Size: 16 # IEEE Quad
```
pfez: For example here if I don't know what's the default data model, and I don't know the AAPCS64 ABI it's impossible to know how large an `int` is.

include/revng/Support/CDataModel.h
```diff
@@ -0,0 +1,143 @@
...
+  return std::nullopt;
+}
+
+struct CDataModel {
+  uint8_t PointerSize = 0;
+  uint8_t StandardTypeSize[static_cast<uint8_t>(CStandardType::Count)] = {};
+
+  [[nodiscard]] static CDataModel getDefaultDataModel(uint8_t PointerSize);
+
+  /* Convenience accessors */
```
aleclearmind: We do not use C-style comments. Use doxygen grouping if you want to.

include/revng/Support/CDataModel.h
aleclearmind, line 1: Is `Support` the best place to place this file?
This is very backend-specific.

include/revng/Support/CDataModel.h
```diff
@@ -0,0 +1,143 @@
...
+  auto E = std::end(CStandardTypeName);
+
+  auto I = std::find(B, E, Name);
+  if (I != E)
+    return static_cast<CStandardType>(I - B);
+
+  return std::nullopt;
+}
+
+struct CDataModel {
```
aleclearmind: We use `struct` for plain data objects, I think this deserves to be a class with data hiding and everything. Also group methods as usual.

include/revng/Model/model-schema.yml
```diff
@@ -33,6 +33,10 @@ definitions:
         doc: The default ABI to adopt for analysis purposes.
         type: ABI
 
+      - name: TargetABI
```
aleclearmind, line 47: What's the default in the end? It is `Invalid` here, but what's the semantics?

include/revng/ABI/abi-internal-schema.yml
```diff
@@ -128,3 +133,15 @@ definitions:
 
     key:
       - Size
+
+  - name: CStandardType
+    type: struct
+    fields:
+      - name: Name
+        type: string
```
Cvelth: I'm not sure whether a string is better here than an enum? Do we expect there to be a lot of variation between the architectures? Also, don't we want to ensure some of the values are always present?

lib/ABI/Definition.cpp
```diff
@@ -14,8 +14,8 @@
 #include "revng/Support/YAMLTraits.h"
 
 template<std::ranges::range RegisterContainer>
-bool verifyRegisters(const RegisterContainer &Registers,
-                     model::Architecture::Values Architecture) {
+static bool verifyRegisters(const RegisterContainer &Registers,
```
Cvelth: Why static? This is a local template, it's going to be instantiated on demand anyway, won't it?

share/revng/abi/AAPCS64.yml
```diff
@@ -98,3 +98,7 @@ FloatingPointScalarTypes:
   - Size: 4
   - Size: 8
   - Size: 16
+
+CStandardTypes:
+  - Name: long double
```
Cvelth: Wait, how does one tell wherever this is an integer or a float type? We can't even verify that the given size is available without that information, can we?

include/revng/Support/CDataModel.h
```diff
@@ -0,0 +1,143 @@
...
+
+  Float,
+  Double,
+  LongDouble,
+
+  Count
+};
+
+inline constexpr llvm::StringRef CStandardTypeName[] = {
+  "char", "short", "int", "long", "long long", "float", "double", "long double",
```
Cvelth, line 29: Wait, we _do_ have an exhaustive list! This strongly feels like a hand-rolled enum! Should we just bite the bullet and push this enum under the TTG?

include/revng/Support/CDataModel.h
Cvelth: This seems to be tightly tied with the `abi::Definition`, maybe it's better to just fold this in? 🤔 And even if the two data-structures are better separately, they should probably still live relatively close to each other!

lib/Support/CTarget.cpp
```diff
@@ -1,17 +0,0 @@
-//
-// This file is distributed under the MIT License. See LICENSE.md for details.
-//
-
-#include "revng/Support/CTarget.h"
-
-const TargetCImplementation TargetCImplementation::Default = {
```
Cvelth, line 7: 🎉🎉🎉

include/revng/Support/CDataModel.h
vasama-work: This is also used by the Clift dialect. That's why I a) placed it under `Support/` and b) made a separate class in the first place.

include/revng/Support/CDataModel.h
vasama-work: How do we define plain data? No member functions? Anyway, all of the member variables of this type are intentionally exposed. I can make it a class and I can even hide the member variables, but then I'll also have to add getters and setters and we'll be Java no for good reason.

include/revng/Model/model-schema.yml
vasama-work: Currently, in this order:
1. `TargetABI`
2. `DefaultABI`
3. `SystemV_x86_64`

There is a WIP about maybe replacing that third one with something smarter. It's something to discuss.

share/revng/abi/AAPCS64.yml
vasama-work: The default data model is fairly simple and we can document it. But I'm not opposed to requiring a full accounting of it in each ABI either. 🤷 

> I don't know the AAPCS64 ABI it's impossible to know how large an int is.

In practice, it's almost impossible to guess wrong. That's how brain dead simple the defaults are.

include/revng/ABI/abi-internal-schema.yml
vasama-work: An enum would probably be fine here...

lib/ABI/Definition.cpp
vasama-work: All TU-local entities should have internal linkage. You wouldn't declare an `inline` TU-local non-template function or variable either, right?

share/revng/abi/AAPCS64.yml
vasama-work: I don't follow. Can you elaborate?

include/revng/Support/CDataModel.h
vasama-work: The reason is the same as the reason for the existence of `CDataModel` as a separate entity under `Support/`: it's used by the Clift dialect, which doesnt't know about the model.

include/revng/Support/CDataModel.h
vasama-work: I changed it from struct to class, but kept the data public.

include/revng/HeadersGeneration/ConfigurationHelpers.h
```diff
@@ -13,12 +13,21 @@ inline uint64_t getExplicitPointerSize(const model::Binary &Model) {
...
+  model::ABI::Values TargetABI = Model.TargetABI();
+  if (TargetABI == model::ABI::Invalid) {
+    TargetABI = Model.DefaultABI();
 
-  // Currently we hardcode the target pointer size as 8 (64-bit), so there is
-  // no reason to emit explicit pointer sizes for binaries with matching size.
-  if (PointerSize == 8)
+    if (TargetABI == model::ABI::Invalid)
+      TargetABI = model::ABI::SystemV_x86_64;
+  }
```
pfez: This duplicates the logic in `importDataModel`. Please refactor.

include/revng/Support/CDataModel.h
aleclearmind: The clift part seems to be not coupled with revngABI so I guess we can keep it in revngSupport. If this grow we'll have another library i guess.

include/revng/CliftEmitC/CBackend.h
```diff
@@ -9,8 +9,8 @@
 
 #include "revng/Clift/Clift.h"
 #include "revng/PTML/CTokenEmitter.h"
-#include "revng/Support/CTarget.h"
+#include "revng/Support/CDataModel.h"
 
 void decompile(clift::FunctionOp Function,
                ptml::CTokenEmitter &Emitter,
-               const TargetCImplementation &Target);
```
pfez, line 16: There are many places that used to take `TargetCImplementation` (or a reference) as argument (or have it as a data member) and now take `CDataModel`. In almost all cases, in the pieces of code that use the `CDataModel` we have either a `clift::ModuleOp` and/or a `clift::FunctionOp` available.

In all these cases, taking the `CDataModel` epxlicitly is not necessary anymore, since if we have a `ModuleOp` we can just get the data model attribute directly from it, and if have a `FunctionOp` we can take the data model attribute from its parent.

Please take a wider look at this commit, and remove the additional `CDataModel` arguments/members where they are not necessary and can be inferred from the `ModuleOp`/`FunctionOp`.

include/revng/Support/CDataModel.h
```diff
@@ -0,0 +1,143 @@
...
+
+    CStandardType Max = Min;
+    while (Max < CStandardType::LongLong
+           and getStandardTypeSize(next(Max)) <= Size)
+      Max = next(Max);
+
+    return std::pair<CStandardType, CStandardType>(Min, Max);
+  }
+
+  /* Hashing and comparison */
```
aleclearmind: No C-style comments, review the whole PR.
Check the rest of the clift codebase as well and remove any C-style comments.

include/revng/CliftEmitC/Headers.h
```diff
@@ -10,10 +10,10 @@
...
 
 void emitTypeAndGlobalHeader(ptml::CTokenEmitter &Tokens,
-                             const TargetCImplementation &Target,
+                             const CDataModel &DataModel,
                              mlir::ModuleOp Module,
                              TypeEmitterConfiguration Configuration);
 
 void emitHelperHeader(ptml::CTokenEmitter &Tokens,
-                      const TargetCImplementation &Target,
+                      const CDataModel &DataModel,
```
pfez: In particular in this case there are many `ModuleOp`, so in each case we want the `CDataModel` to be taken by the actual module that is being processed.

include/revng/CliftEmitC/Headers.h
```diff
@@ -10,10 +10,10 @@
 #include "revng/PTML/CTokenEmitter.h"
 
 void emitTypeAndGlobalHeader(ptml::CTokenEmitter &Tokens,
-                             const TargetCImplementation &Target,
+                             const CDataModel &DataModel,
```
pfez: This is now redundant.

include/revng/CliftEmitC/CSemantics.h
```diff
@@ -5,9 +5,9 @@
...
 
 #include "revng/Clift/Clift.h"
-#include "revng/Support/CTarget.h"
+#include "revng/Support/CDataModel.h"
 
 // TODO: does this really belong with the emitters?
 
 mlir::LogicalResult verifyCSemantics(mlir::ModuleOp Module,
-                                     const TargetCImplementation &Target);
+                                     const CDataModel &DataModel);
```
pfez: This is now redundant.

include/revng/CliftTransforms/Legalization.h
```diff
@@ -5,11 +5,10 @@
 //
 
 #include "revng/Clift/Clift.h"
-#include "revng/Support/CTarget.h"
 
 namespace clift {
 
 mlir::LogicalResult legalizeForC(clift::FunctionOp Function,
-                                 const TargetCImplementation &Target);
+                                 const CDataModel &DataModel);
```
pfez: This is now redundant.

include/revng/Model/model-schema.yml
aleclearmind: OK, another option would be to fail in `checkPrecondition` in case neither `TargetABI` nor `DefaultABI` are set.

lib/CliftEmitC/CEmissionPasses.cpp
```diff
@@ -100,12 +100,9 @@ using HHBase = clift::impl::CliftEmitHelperHeaderBase<T>;
 
 clift::PassPtr<mlir::ModuleOp> clift::createEmitHelperHeaderPass() {
   static constexpr auto Impl = [](mlir::ModuleOp Module,
+                                  const CDataModel &DataModel,
```
pfez: This is now redundant.

lib/CliftEmitC/CSemantics.cpp
```diff
@@ -94,25 +96,21 @@ class CVerifier : public ModuleVisitor<CVerifier> {
...
-
-  const TargetCImplementation &Target;
 };
 
 } // namespace
 
 mlir::LogicalResult verifyCSemantics(mlir::ModuleOp Module,
-                                     const TargetCImplementation &Target) {
-  return CVerifier::visit(Module, Target);
+                                     const CDataModel &DataModel) {
```
pfez: This is now redundant.

lib/CliftEmitC/CSemantics.cpp
```diff
@@ -32,12 +32,14 @@ static clift::PointerType getPointerOperationType(mlir::Operation *Op) {
 }
 
 class CVerifier : public ModuleVisitor<CVerifier> {
+  CDataModel DataModel;
+
 public:
-  explicit CVerifier(const TargetCImplementation &Target) : Target(Target) {}
+  explicit CVerifier(const CDataModel &DataModel) : DataModel(DataModel) {}
```
pfez: `DataModel` should be set when calling `visit` on a `ModuleOp`, from the module's data model.

lib/CliftEmitC/Headers.cpp
```diff
@@ -206,10 +205,10 @@ class CHeaderEmitterImpl {
 };
 
 void emitTypeAndGlobalHeader(ptml::CTokenEmitter &Tokens,
-                             const TargetCImplementation &Target,
+                             const CDataModel &DataModel,
```
pfez: This is now redundant.

lib/Clift/CliftAttributes.cpp
```diff
@@ -257,7 +257,132 @@ template class ClassAttrImpl<UnionAttr>;
...
+      Name += ' ';
+      Name += std::string_view(Keyword);
+    }
+
+    uint8_t *SizeField = &DM.PointerSize;
+    if (Name != "pointer") {
+      // In order to provide good diagnostics like GCC, the special case of
+      // extremely long integer types is handled separately.
+      if (Name == "long long long")
+        return Parser.emitError(KWLoc) << Name << " is too long for Clift.";
```
pfez, line 297: What? Why do we catch specifically `long long long`?

share/revng/abi/AAPCS64.yml
pfez: It's true that the defaults are very simple and hard to get wrong. Let's keep this like it is and only fix it if we grow uncomfortable with that.

lib/Clift/CliftAttributes.cpp
vasama-work: Why, to provide better diagnostics, [just like GCC](https://godbolt.org/z/jz4WzxPeW).

include/revng/CliftEmitC/CBackend.h
vasama-work: Ah, that's a good point. However, for `clift-opt` I specifically wanted to fall back to a default data model (SystemV x86-64) for convenience. This avoids having to copy paste the data model in all our unit tests for example, and it's a nicer user experience when doing manual stuff.

lib/ABI/Definition.cpp
Cvelth: Theoretically, yes, but practically, considering the mangled name and the fact that it does not appears in any headers, the chance of linker actually pushing any instantiation of this anywhere unexpected is non-existent. So it exists, pretty much, until the linker runs.

share/revng/abi/AAPCS64.yml
Cvelth: The data structures above this one define primitive sizes available. We intentionally split them up into generic and floating point a while back to create clear distinction.

What I'm trying to say is, maybe it's worth, on top of size, also encode whether this type is generic or floating point. That way we can (1) verify it actually makes sense under the given architecture and (2) avoid having to duplicate any extra type-related information (like alignment) as it will implicitly be shared between the two.

lib/ABI/Definition.cpp
vasama-work: In this particular case perhaps the probability of an ODR violation occurring is very low, but still, all TU-local entities should have internal linkage and I'm not going to spend time evaluating the probability of failure in each individual case when the solution is so trivial.

share/revng/abi/AAPCS64.yml
vasama-work: > The data structures above this one define primitive sizes available.

Ah, that's not quite what it does. It specifies the sizes of the standard C fundamental types. That set of types is fixed and their constraints are well specified. `int` is never a floating point type and `float` is never an integer type.

We could also specify alignment, but for now we don't *need* that information.

vasama-work: Or rather, we rely on the assumption that the size and alignment of any given fundamental type are equal. This assumption will eventually fail for floating point types at least, but for now we don't have a lot of support for floating point types. When that support gets further along, we'll probably have to add both alignment and format information here.

include/revng/Support/CDataModel.h
Cvelth: Yes, but this leads to us having two identical enums (or enum-like things, does not matter) representing the same exact list of values in different parts of the code base.

Maybe it makes sense to either introduce one more TTG layer for this enum (for consistency, the only thing worse than ttg-enums is a mix of ttg and non-ttg enums :p)? Or bite the bullet and say that `revngClift` depends on the `revngABI` (so, indirectly, on `revngModel`, which you seem to try to avoid), allowing us to specify this information in one central location.

cc: @aleclearmind, what do you think?

vasama-work: I personally don't find duplication to be a big problem here. This set of types is defined in C99 and will never change.

vasama-work: Oh and while we could introduce an enum for this in the ABI schema, as you noted, for now I did not. Instead I used the names as the key and ensured that those names match types in this set. So currently there is no duplication.

lib/ABI/Definition.cpp
Cvelth: Yes, I'm 100% agree that we should use the same pattern for _all_ such templates across the codebase. I'm just trying to reason whether a practically non-existent chance of ODR is worth extra 7 characters, often introducing extra lines (and sometimes even extra typedefs).

vasama-work: 1. Correct, reduces mental overhead, slightly longer.
2. Incorrect, increases mental overhead, slightly shorter.

I know which one I'll pick. :)

share/revng/abi/AAPCS64.yml
Cvelth: No-no, that's not what I mean. Above this we have data structures defining the set of supported primitive sizes. It's kind of awkward that we could end up with a case where `CStandardTypes` mentions a primitive of a size we declare to not support.

What I'm trying to say is, maybe it's worth it to treat this as a "map of a name to one of the primitives above" as opposed to a "map of a name to size".

include/revng/Support/CDataModel.h
Cvelth: > So currently there is no duplication.

That comes with the price of not limited what values can be specified in the yaml, if I understand correctly.

IMO, a string that can contain a strict set of values is practically an enum, just with a different shape.

vasama-work: The names are checked by verification. Using an enum in the schema, maybe it would be checked during parsing? It's not clear to me. In any case, I'm just saying there is currently no duplication.

lib/ABI/Definition.cpp
Cvelth: 2. Does it increase the mental overload if it's the same pattern in _every_ similar case?

That being said, this isn't that strong of a preference. Just every time I have to use a  `using` I'm sad :P And this can alleviate the need for them in a handful of cases.

vasama-work: Ah, if you never give any templates internal linkage? Sure you don't get introduced mental overhead, but in the trade you increase the likelihood of those latent bugs manifesting.

The line length thing is a separate issue, and it either is a problem or it isn't. In either case, we shouldn't let it affect technical decisions. And if it is problem, the solution is to change the formatting rules.

share/revng/abi/AAPCS64.yml
vasama-work: Ah, I see. Yes, this is a little awkward. We could verify that each C type is listed in the set of supported primitive sizes. It's not entirely clear to me what the meaning of the set of supported primitive sizes is. Are you able to elaborate? Is it about what is supported native by the CPU architecture? Is it about integer types provided by the C ABI? Or something else?

share/revng/compile-flags.cfg
pfez: I checked and there are basically two possible causes for this, each worth mentioning in the explanatory comment.
1. An argument that is present in the model on the function prototype is effectively not used in the binary. This may be caused by misdetection of function arguments by revng, or by a user setting the model with a prototype with an erroneous additional argument.
2. Identification of function boundaries and/or control flow is wrong, leading to a situation where the lifted function's body is truncated via `llvm.unreachable`. Truncation of the body may render the arguments unused, if they were used in the part of the body that gets removed.

include/revng/CliftEmitC/CBackend.h
pfez: You can hardcode the SystemV x86-64 data model in `clift-opt` and check that in whatever clift module it parses the data model is either totally missing (and you add it) or equal to SystemV x86-64.

I wouldn't design all the APIs for `clift-opt` which is ostensibly for tests.

lib/Clift/CliftAttributes.cpp
pfez: Eh but why `long long long` and not, say, `long long long long`?
I would just drop this.

include/revng/CliftEmitC/CBackend.h
vasama-work: I would really prefer not to *add* the data model as a side-effect of `clift-opt`. Perhaps there's a middle ground here... Perhaps we could get rid of the data model in all the internal APIs, and either always fall back to a global default, or only when one is set by `clift-opt`. We could either use a global variable or store on the dialect object, which would effectively make it global for the MLIRContext.

lib/Clift/CliftAttributes.cpp
vasama-work: Oh, we can change it to `Name.starts_with("long long long")` too. Anyway, I just couldn't resist making a reference to that famous GCC diagnostic. 😄

pfez:
Some tests are failing.

`orc install --test revng` gives:

```
FAILED: share/revng/test/tests/runtime/calc-revng.test-pypeline-comparison-hexdump-1ca7f289
source "repos/orchestra-ivan/root/libexec/revng/common.sh" "revng.test-pypeline-comparison-hexdump"; cd "share/revng/test/tests/runtime/calc-revng.test-pypeline-comparison-prepare-7f640f7c
/"; "repos/orchestra-ivan/root/share/revng/test/tests/pypeline-comparison/compare.sh" model.yml input hexdump;
[483/587] revng artifact --resume "share/revng/test/tests/...b81509//decompiled.tar.gz" | sed 's|\.c\.ptml$||' | sort);
```

aleclearmind:
> FAILED: share/revng/test/tests/runtime/calc-revng.test-pypeline-comparison-hexdump-1ca7f289

These tests have non-deterministic problems. We should try to have a reproducer and send it to @mrjackv.

include/revng/CliftEmitC/Headers.h
vasama-work: Why do we have multiple modules here? Is that just an artifact of the new pipeline storing functions in separate modules? And more importantly, are you implying that we could emit a single header using multiple different data models? That does not make sense.

lib/ABI/Definition.cpp
Cvelth: What do you mean about the "separate issue". They are tightly linked! It's a question of whether an extra keyword that guarantees that "if we have 2 _identical_ instantiations of two _identical_ templates (with _identical_ names and parameters) in different compilation units", things don't go south. Let's be honest, what are odds of that happening? There are not _that_ many people editing our codebase in parallel for that to be a concern.

share/revng/abi/AAPCS64.yml
Cvelth: Practically, this is a set of primitives that are either (1) explicitly mentioned in whatever specification the ABI is based on, or (2) were emitted by a compiler (gcc or msvc) as a part of the ABI testing.

That being said, it's not a tight list. So, in case we have evidence of a named primitive occupying size we don't otherwise support, it's probably a good idea to just add the size to the list!

include/revng/Support/CDataModel.h
Cvelth: I'm talking less about the specific implementation and more about the idea of it.

We are ending up with the same exact enum-like construct (a flat value that can be set to any one value out of a small set) being scattered across two different libraries with duplicated logic. Just for the sake of avoiding a dependency that we might end up eventually making anyway.

lib/ABI/Definition.cpp
vasama-work: The use of internal linkage has nothing to do with the limit on line length. These are separate issues. Our choice of formatting style should not have any bearing on technical decisions.

> Let's be honest, what are odds of that happening?

You're up against the birthday problem here. It'll happen eventually. I've been bitten by this and I won't be again.

include/revng/Support/CDataModel.h
vasama-work: It's unclear to me what logic is duplicated.

Cvelth: Verification. As in, making sure the value is actually in the expected range.

vasama-work: Do you mean this in `abi::Definition::verify()`?
```cpp
  for (const abi::CStandardType &Type : CStandardTypes()) {
    if (not findCStandardType(Type.Name()))
      return false;
  }

  if (not getDataModel().verify())
    return false;
```

Cvelth: Again, I am not talking about this specific data structure choice. I'm talking about the general situation.

As things stand, we have two separate places where these values could appear (in two different shapes): one as part of the `revngSupport` and one - as part of `revngABI`. Which means we need to have validation separately(-ish, of course there are ways of sharing the logic by using the conversion) in both cases. As well as a way to convert from one to the other.

If we were able to get away with a single `enum` covering both needs, the situation would have been cleaner.

vasama-work: I see. Actually by this metric, there is a third: the Clift dialect, which must also parse names and sizes and assign them into a `CDataModel` in `#clift.data_model`. But that's all these latter two do. They just take names and sizes and plug them in, letting `CDataModel` handle all the logic.

Cvelth: Right. But we kind of have to have a clift-side one. That ship already sailed. Clift duplicates a lot of logic from elsewhere (think most of the model!), one more does not change anything.

On the other hand, there is current no duplication between `revngABI` and `revngSupport`. So introducing it does feel like a bit of a degradation.

That being said, it's important to note this is not a deal-breaker for me. I _would_ prefer less duplication, but hey, real life is messy.

That being said, at least `enum` in the TTG instead of a string is probably still worth it one way or the other. If for nothing else, to make it easier to discover accepted value ranges user-side. And better documentation (`doc` fields are required in TTG).

vasama-work: I'd be fine with duplicating this enum in TTG. As I said elsewhere, it's defined by C99 and won't change.

Cvelth: Yep, that's what the other thread is for: https://github.com/revng/revng/pull/579#discussion_r3138268695
TTG is specifically designed to be user-facing after all.

This thread is an exploration of whether it's viable to _only_ have the TTG enum.

share/revng/abi/AAPCS64.yml
vasama-work: So what is the set of supported primitive types used for?

Cvelth: 1. To ensure we don't try to convert obviously broken RFTs into CFTs.
2. To figure out alignment of primitives.

vasama-work: I see... So we're kind of duplicating the information here. Only `CStandardType` does not *yet* provide alignment information. It could be added though. However, that would still leave out extended types, which the `ScalarTypes` and `FloatingPointScalarTypes` fields presumably include. My plan has never been to include extended types as part of the `CStandardTypes` field, but instead to introduce a separate field for those...

Cvelth: No, what I'm suggesting is to let `ScalarTypes` and `FloatingPointScalarTypes` hold the full type list. And for the `CStandardTypes` to just be a map from the c-name to one of its entries. You can even use `TupleTreeReference`, the same way types a linked. TTG takes care of all the boilerplate to make that happen.

include/revng/CliftEmitC/Headers.h
pfez: Yes, it is a result of the functions being in separate modules.

I don't imply that we could emit a single header using different data models. But since we introduced the data model into the dialect, taking it of the module must be the become the only way we use it. Providing an out-of-band data model just opens the door to situations where they end up mismatching.

In this specific case the responsibility for setting the various data models of the various modules will be of whoever calls `emitHelperHeader`. And if you want to be extra safe you can assert inside `emitHelperHeaders` that the data model is the same across all the modules. But the data model must not go out-of-band anymore.

include/revng/CliftEmitC/CBackend.h
pfez: > I would really prefer not to add the data model as a side-effect of `clift-opt`

Makes sense, but only if the data model is missing in the input.

> we could get rid of the data model in all the internal APIs

This is something that I really want to do. Out-of-band data models considered harmful!

> either always fall back to a global default, or only when one is set by `clift-opt`

I think there are two orthogonal decisions to take here.
1. Out-of-band vs in-module data model. I have strong opinions on this. I always want it in-module. I don't want to see out-of-band data models passed into any API, internal or external. It must be clear that the data model is a property of the module.
2. What to do when we have to initialize the data model in a module? I think that we need two different mechanisms for a) when we initialize it from the Model vs b) in `clift-opt`. In both cases we need a backup default, which would need to be the same for consistency. For now this default can be the SystemV x86_64 for convenience.
  * a) the current logic you have implemented for importing from the model makes sense, and it already has a backup default.
  * b) in `clift-opt` we can preserve the data model if present, or always set the default when the clift module doesn't have it.

include/revng/CliftEmitC/Headers.h
vasama-work: Yes, it will not go "out-of-band". I'll just assert that the data models of all modules are equivalent.

lib/Clift/CliftAttributes.cpp
pfez: You want to have this easter egg? OK :shrug: 
Just put `starts_with` and we'll keep it :smile:
Feel free to close this thread.

include/revng/CliftEmitC/CBackend.h
vasama-work: Here's the solution I have:
1. No out-of-band data models - all the data model parameters go away. Instead, the module is always queried.
2. If no data model is available, the consumers do not make up a default. Instead they end up aborting (indirectly).
3. The dialect object carries an optional default data model which can be assigned. If a given module does not have its own data model, this default is used (only if available).
4. In `clift-opt` only, we assign the default data model to SystemV x86-64.

The effect is that a) we never accidentally use a default data model in the pipeline, but b) `clift-opt` chugs along as it always has with its convenient defaults.

Does this seems reasonable?

pfez: This also reminds me that we have to make sure that the data model is properly imported from the model even when generating the headers from clift. This is not necessary yet, because the pipe that actually generates the helper header from clift will be added in a future PR, but it's important we don't forget it. Which is also one of the reasons why I don't want to conjure default data models out of thin air if they are not in the module.

lib/ABI/Definition.cpp
pfez: The problem here is internal linkage vs brevity.
Correctness is more important than brevity, so we must have internal linkage.
If you also want brevity, just put it into an unnamed namespace and you'll have internal linkage without `static`.

lib/Clift/CliftAttributes.cpp
vasama-work: Hahaha, yes!!! I didn't actually expect you to approve it. 😆

include/revng/CliftEmitC/CBackend.h
vasama-work: Yes. I also don't want to conjure them out of thin air. That's why I'm careful to only set this default in `clift-opt`, where it is useful. It is not useful in the pipeline, and so it is not available.

lib/ABI/Definition.cpp
Cvelth: IMO, my priorities are usually:

1. Consistency on different instances of the same thing.
2. Correctness where it really matters (not a one-in-a-million-ODR-violation).
3. Readability (which includes brevity among other things).
4. Correctness in less practical cases.

-----------------

That being said, an anonymous namespace _does_ sound like the perfect solution here because it, as you mentioned, solves both concerns.

The bigger issue I have is related to 1. We currently have a mix of all three (`non-internal`, `static` and `namespace{}`) scattered everywhere. It would be nice to adopt one and enforce it everywhere.

pfez: > The bigger issue I have is related to 1. We currently have a mix of all three (non-internal, static and namespace{}) scattered everywhere. It would be nice to adopt one and enforce it everywhere.

I agree that this is an issue, but this is not the place to raise this issue.

In general, for issues on coding conventions, PRs are not the place to discuss them, because they turn minor things into large discussions that potentially affect the whole project. This is a priority inversion problem.

If you want to bring this up, you can propose this on mattermost. But whatever new coding convention we may want to introduce should be easy to evaluate objectively on the code during review without personal involving personal taste or - even better - machine checkable and enforcable via `revng check-conventions`. Anything that isn't machine checkable and that hasn't caused direct problems (like ODR here) is hardly going to be considered seriously, since there are already many non-checkable rules in our coding conventions and they already tend to slip through.

pfez: For this, let's do the unnamed namespace and call it a day.

include/revng/CliftEmitC/CBackend.h
pfez: > Here's the solution I have

This seems good, except that I'd like you to elaborate on 3.

> 3. The dialect object carries an optional default data model which can be assigned

What dialect object is going to carry the default data model? And who's adding that? Or you mean that there is going to be a dialect-wide hardcoded default?

> If a given module does not have its own data model, this default is used (only if available).

It's not clear to me how the fact that the default is used also guarantees the following:

> a) we never accidentally use a default data model in the pipeline

lib/ABI/Definition.cpp
Cvelth: Sure!

include/revng/HeadersGeneration/ConfigurationHelpers.h
vasama-work: I would prefer to duplicate this for now, because we'll soon get rid of this code anyway. It's only needed for configuring the old backend header generation to be compatible with Clift (the foreign pointer issue).

include/revng/CliftEmitC/CBackend.h
vasama-work: > What dialect object is going to carry the default data model? And who's adding that? Or you mean that there is going to be a dialect-wide hardcoded default?

Each dialect loaded into an `MLIRContext` (including Clift) is represented by a C++ object (`CliftDialect`). We can store dialect-specific context-global state in the dialect object. That's what I've done. The default data model is a runtime property of the dialect, and is not hardcoded, except in `clift-opt`, which is the only case where the default data model is assigned.

See [here (`clift-opt`)](https://github.com/vasama-work/revng/blob/feature/clift-implicit-cast-elision/tools/clift-opt/Main.cpp#L21-L24) and [here (`CliftDialect`)](https://github.com/vasama-work/revng/blob/feature/clift-implicit-cast-elision/lib/Clift/CliftDialect.cpp#L249-L258).

> It's not clear to me how the fact that the default is used also guarantees the following:
> > a) we never accidentally use a default data model in the pipeline

The default is optional (`std::optional`) and is only returned by `getDataModel(mlir::ModuleOp)` when set. It is only set by `clift-opt`, never in the pipeline.

include/revng/Model/model-schema.yml
vasama-work: Here's where I'm at right now:
```
  if (ABI == model::ABI::Invalid) {
    ABI = Model.DefaultABI();

    // TODO: We should do something smarter here:
    //       * Pick a better fallback (maybe other model properties), and/or
    //       * after exhausting all fallbacks, reject invalid and additionally
    //         check the availability of a valid ABI in `checkPrecondition`.
    if (ABI == model::ABI::Invalid)
      ABI = model::ABI::SystemV_x86_64;
  }
```
Let me know if you want me to do the `checkPrecondition` instead as part of this PR.

share/revng/compile-flags.cfg
vasama-work: I added this explanation in a comment.

include/revng/ABI/abi-internal-schema.yml
vasama-work: An enum is now used instead, and `CStandardTypes` was removed in favour of a `CTypes` field in `ScalarType`. Additionally, `ScalarTypes` and `FloatingPointScalarTypes` were merged and a disambiguating kind enum was introduced instead.

pfez:
Now that cast elision is more or less on par with the old backend, let's drop the old tests for cast reduction based on the old backend, that were very broken anyway.
This needs changes to both `revng` and `revng-qa`.

Just `find ./ -iname '*reduce-c-casts*'` in `revng` and `revng-qa` and remove everything you can find starting from there.

include/revng/CliftEmitC/CBackend.h
pfez: Neat.

include/revng/Model/model-schema.yml
aleclearmind: This works, I'd put it in `Binary::targetABI`. We do this elsewhere (e.g., `Function.Prototype()` vs `Function.prototype()`). It's not the best but it's a pattern we already have.

include/revng/HeadersGeneration/ConfigurationHelpers.h
pfez: Ok. But put a `TODO` comment on this function to explain it's only used by the old backend, it's deprecated, and it will have to be removed after we drop the old header generation.

@Cvelth please take notice and integrate this into your upcoming PR on clift-based headers (part 4 I believe).

pfez: @vasama-work when you're done feel free to close this thread

Cvelth: Actually, I'm ahead of you! This file is no more on my branch :P

include/revng/ABI/abi-internal-schema.yml
pfez: This looks good to me. I'm closing this thread. @Cvelth open a new one if you find something that you still don't like.

lib/ABI/FunctionType/Conversion.cpp
```diff
@@ -698,7 +698,7 @@ TCC::tryConvertingReturnValue(RFTReturnValues Registers) {
       //       into preserving it at least partially.
       uint64_t PointerSize = model::ABI::getPointerSize(ABI.ABI());
       uint64_t PrimitiveSize = Ordered.size() * PointerSize;
-      if (ABI.ScalarTypes().contains(PrimitiveSize)) {
+      if (ABI.findIntegerType(PrimitiveSize) != nullptr) {
```
Cvelth, line 701: Technically, this changes semantics. Mostly due to how invalidation tracking works.

The old version used `count` which only marked the accessed element or lack thereof, the new one puts in an "exact" as the requirement.

Technically, this is irrelevant as of now, because only model writes invalidation data. But there are plans to merge this into the model eventually.

P. S. See the next thread.

lib/ABI/Definition.cpp
```diff
@@ -74,8 +76,28 @@ static bool verifyReturnValueLocation(const abi::Definition &D) {
   return true;
 }
 
+} // namespace
+
 namespace abi {
 
+const ScalarType *Definition::findIntegerType(uint64_t Size) const {
+  auto I = ScalarTypes().find({ ScalarKind::Integer, Size });
```
Cvelth, line 96: Is there a reason to use `find` which does "exact" invalidation as opposed to `tryGet` which only writes one key (or lack  thereof)?

(same below)

lib/ABI/Definition.cpp
```diff
@@ -104,17 +126,34 @@ bool Definition::verify() const {
...
     if (Type.Size() == 0)
       return false;
 
+    for (auto CT : Type.CTypes()) {
+      if (CT == CType::Invalid)
+        return false;
+
+      if (std::exchange(CTypes[CT], true))
+        return false;
+    }
```
Cvelth, line 167: Nice! Thanks for this!

lib/ABI/Definition.cpp
Cvelth: For context, I'm referring to this `tryGet`:
https://github.com/revng/revng/blob/2e9abef4808233dce94139be07d88248782abd29/include/revng/ADT/TrackingContainer.h#L364

lib/ABI/FunctionType/Conversion.cpp
vasama-work: I added `hasIntegerType` and `hasFloatingPointType`.

lib/ABI/Definition.cpp
vasama-work: I changed this to use `tryGet`. Although I don't understand what you mean by "exact".

vasama-work: I spoke too soon. `ScalarTypes` is `SortedVector`, not `TrackingSortedVector` and does not appear to have a `tryGet` member.

include/revng/Model/model-schema.yml
vasama-work: I think until such time that we come up with some other user of the target ABI (besides this data model import pipe before the final target-specific Clift passes), I would prefer to keep this logic in there.

But if you prefer, I can introduce `Binary::targetABI` as well.

lib/ABI/Definition.cpp
vasama-work: Looking at this, I realised it deserved an explanation, so I added some comments along with additional verification I realised was missing.

lib/ABI/Definition.cpp
Cvelth: Wait, you're right. Even though these will eventually become a part of a tracking data-structure. They are currently not.

I guess `find` is the correct solution for now, but please leave a `TODO` as a reminder to switch to `tryGet` when that becomes relevant.

---------------

> Although I don't understand what you mean by "exact".

As you know, we track model accesses using so-called "invalidation" logic. But not all accesses are the same. The `find` is considered an "exact-level" access. Because it returns an iterator. And who knows what the callee does with it.

Which means that if _anything_ changes about the data-structure, related to the element passed inside `find` or not - invalidation trigger.

`tryGet` is identical to `find` functionality-wise. Except it doesn't leak the iterator (It returns a pointer). Hence it just writes presence or lack of a specific element. And invalidation only triggers on changes to that element - not to the entire data-structure.

lib/ABI/FunctionType/Conversion.cpp
Cvelth: Sorry, I wasn't clear enough. The same helper is good for enough both use-cases! It was specifically worried about the tracking `find`.

If we are ok with eventually switching the implementation of these to `tryGet` (looks like we are), `findX` is all we need.

(sorry I made you change this just to switch back, my bad on not explaining my thought process better)

lib/ABI/Definition.cpp
Cvelth: Where can I see those? Are you still working on them?

vasama-work: https://github.com/vasama-work/revng/blob/feature/clift-implicit-cast-elision/lib/ABI/Definition.cpp#L156-L168

lib/ABI/FunctionType/Conversion.cpp
vasama-work: I reverted this change.

lib/ABI/Definition.cpp
vasama-work: I added comments to use `tryGet` instead if invalidation tracking is added.

lib/ABI/Definition.cpp
Cvelth: Look good!

P. S. Please close this thread and the other two! All three look good to me, and I don't have the permission to close threads on other people's PRs :D

include/revng/Model/model-schema.yml
aleclearmind: If we don't do it, we'll end up with some unrelated user to sloppily use `TargetABI()`. Let's follow the existing patterns so it's not something backend-specific.

vasama-work: I added `targetABI` and changed the callers of `TargetABI` to use it instead.

PR #581, opened 2026-05-11
Floating Point Frontend

andrealinux1:
Introduce the support for making the calls to QEMU softfloat calls emerge in the IR used for decompilation.

This is obtained by reworking the machinery used by `InlineHelpers` to broaden the surface of the QEMU helpers which handle the floating point instruction.


lib/HeadersGeneration/HelpersToHeader.cpp
```diff
@@ -73,7 +74,14 @@ static void printLLVMTypeDeclaration(const llvm::StructType *S,
       PreviousOffsetInBits = FieldOffsetInBits + FieldSizeInBits;
     }
     revng_assert(PreviousOffsetInBits % 8 == 0);
-    revng_assert(PreviousOffsetInBits / 8 == StructByteSize);
+    uint64_t Alignment = StructLayout->getAlignment().value();
+    uint64_t ByteSize = PreviousOffsetInBits / 8;
+    uint64_t NumUnalignedBytes = ByteSize % Alignment;
```
aleclearmind: Never abbreviate.

lib/HeadersGeneration/HelpersToHeader.cpp
```diff
@@ -73,7 +74,14 @@ static void printLLVMTypeDeclaration(const llvm::StructType *S,
...
+    uint64_t AlignedByteSize = NumUnalignedBytes ?
+                                 (ByteSize - NumUnalignedBytes + Alignment) :
+                                 ByteSize;
```
aleclearmind: This doesn't check out

1. Are we handling non-packed structs?
2. We're printing padding, this still looks like packed to me, so we need `_PACKED`.

cc/ @pfez

share/doc/revng/developer-manual/qemu-helpers.md
aleclearmind, line 1: This needs to be referenced somewhere, in the index and possibly from other parts of the docs where we talk about helpers.

lib/HeadersGeneration/HelpersToHeader.cpp
andrealinux1: I think I can reply to this. The correct solution would be to handle the `_PACKED` attribute emission, the padding was a temporary solution that slipped in probably in a rebase, I’ll rebase on `develop` and correct this

lib/Lift/VariableManager.cpp
```diff
@@ -170,6 +170,7 @@ VariableManager::VariableManager(Module &M,
   IntegerType *IntPtrTy = AllocaBuilder.getIntPtrTy(*ModuleLayout);
   Env = cast<GlobalVariable>(TheModule.getOrInsertGlobal("env", IntPtrTy));
   Env->setInitializer(ConstantInt::get(IntPtrTy, LibTcgEnvOffset));
+  Env->setConstant(true);
```
aleclearmind, line 179: Comment why this is important

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+#include "revng/Support/CommandLine.h"
+#include "revng/Support/Debug.h"
+#include "revng/Support/IRHelpers.h"
+
+using namespace llvm;
+
+static Logger Log("tag-softfloat-helpers");
+
+static cl::opt<std::string> SoftfloatDir("tag-softfloat-helpers-dir",
```
aleclearmind: 1. Never abbreviate, review all the PRs for this.
2. When the the option description is long, it makes sense to declare an auxiliary global variable to avoid splitting it in 12 lines.
3. I don't see why we would want to customize this. Drop the option.

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
aleclearmind, line 1: Let's reframe this pass as `MarkInlineHelpersUpTo`.
Let's make this softfloat- and QEMU-agnostic.
Let's just have a list of paths that represent a runtime library we want to "emerge" by inlining all the callers.
This list should only contain `fpu/` as of now.

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
```diff
@@ -0,0 +1,179 @@
...
+static unsigned instructionSize(const Function &F) {
+  unsigned InstructionCount = 0;
+  for (const BasicBlock &BB : F) {
+    InstructionCount += BB.size();
+  }
+
+  return InstructionCount;
+}
+
+class TagSoftfloatHelpers : public ModulePass {
```
aleclearmind: Docs: what does this pass do?
When does it bail out from doing what it wants to do?

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+static cl::opt<unsigned> MaxInlinableSize("tag-softfloat-helpers-max-size",
+                                          cl::init(100),
+                                          cl::desc("Skip tagging functions "
+                                                   "whose body exceeds this "
+                                                   "many IR "
+                                                   "instructions."),
+                                          cl::cat(MainCategory));
+
+/// Helper function to identify a `Function` folder definition from the metadata
```
aleclearmind: 1. Never use the term "folder". Directory.
2. I'm not happy with `.contains`. I'd parse the path, strip the file name and check if the path ends with `DirName`.
   `../sources/qemu/fpu/lol.c` -> `../sources/qemu/fpu` -> `.endswith("fpu/")`. Something like this.

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
```diff
@@ -0,0 +1,179 @@
...
+}
+
+class TagSoftfloatHelpers : public ModulePass {
+public:
+  static char ID;
+
+public:
+  TagSoftfloatHelpers() : ModulePass(ID) {}
+
+  void getAnalysisUsage(AnalysisUsage &AU) const override {}
```
aleclearmind: Drop?

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+  bool runOnModule(Module &M) override {
+
+    // 1: Identify all the softfloat `Function`s using the debug info provided
```
aleclearmind: Do not number things, they go out of sync and don't add much

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
```diff
@@ -0,0 +1,179 @@
...
+    SmallSet<const Function *, 4> SoftfloatSet;
+    for (Function &F : M) {
+      if (F.isDeclaration()) {
+        continue;
+      }
+      if (isDefinedInDir(F, SoftfloatDir)) {
+        SoftfloatSet.insert(&F);
+        revng_log(Log, "softfloat: " << F.getName());
+      }
+    }
```
aleclearmind: Always use the following pattern:

```c++
{
revng_log(Log, "Collecting functions to defined in ...");
LoggerIndent Indent(Log);

    for (Function &F : M) {
      // ..
      if (...) {
        revng_log(Log, F.getName());
        // ...
      }
    }
}
```

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
```diff
@@ -0,0 +1,179 @@
...
+    // 2: Build the reverse edges in the call graph (callee -> {callers})
+    DenseMap<const Function *, SmallVector<const Function *>> Callers;
+    for (const Function &F : M) {
+      if (F.isDeclaration()) {
+        continue;
+      }
+      for (const BasicBlock &BB : F) {
+        for (const Instruction &I : BB) {
+          if (const auto *Call = dyn_cast<CallInst>(&I)) {
+            const Function *Callee = Call->getCalledFunction();
```
aleclearmind: Always use `getCalledFunction(Call)`

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+            // We are not interested in indirect function calls
+            if (Callee == nullptr || Callee->isIntrinsic()) {
```
aleclearmind: Always use `or` instead of `||`

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
```diff
@@ -0,0 +1,179 @@
...
+          }
+        }
+      }
+    }
+
+    // 3: Perform a reverse upward walk on the call graph to find all the
+    //    transitive callers of any softfloat function
+    SmallSet<const Function *, 4> Reachable = SoftfloatSet;
+    SmallVector<const Function *> Worklist(SoftfloatSet.begin(),
+                                           SoftfloatSet.end());
```
aleclearmind: Use `OnceQueue`

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
```diff
@@ -0,0 +1,179 @@
...
+      // The pass did not change the module
+      return false;
+    }
+
+    // 2: Build the reverse edges in the call graph (callee -> {callers})
+    DenseMap<const Function *, SmallVector<const Function *>> Callers;
+    for (const Function &F : M) {
+      if (F.isDeclaration()) {
+        continue;
+      }
```
aleclearmind: For super-simple cases like these, let's use

```c++
if (F.isDeclaration())
  continue;
```

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
```diff
@@ -0,0 +1,179 @@
...
+    for (Function &F : M) {
+      if (F.isDeclaration()) {
+        continue;
+      }
+      if (SoftfloatSet.contains(&F)) {
+        continue;
+      }
+      if (not Reachable.contains(&F)) {
+        continue;
+      }
```
aleclearmind: 1. Merge these in a single one.
2. I'm not sure the first check (`isDeclaration`) is necessary

CMakeLists.txt
```diff
@@ -237,9 +237,10 @@ foreach(
   add_custom_command(
     OUTPUT "${LIBTCG_HELPERS_FULL}"
     DEPENDS "${LIBTCG_HELPERS}" revng-all-binaries
+            $<TARGET_FILE:revngHelperArgumentsAnalysis>
```
aleclearmind: 1. Why this? If it's a bugfix, put it in a separate commit.
2 Do you need? `$<TARGET_FILE:...>`? I think the target name is enough.

CMakeLists.txt
```diff
@@ -237,9 +237,10 @@ foreach(
...
     COMMAND
-      ./bin/revng opt -sroa -instsimplify -cpu-loop-exit
-      -analyze-helper-arguments -fix-helpers
+      ./bin/revng opt -tag-softfloat-helpers -sroa -instsimplify -cpu-loop-exit
+      -analyze-helper-arguments -loop-unroll -fix-helpers
```
aleclearmind: Add a comment about the logic for this pipeline.
We should explain why each pass is in the position it is in.

tests/unit/llvm_lit_tests/TagSoftfloatHelpers.ll
```diff
@@ -0,0 +1,92 @@
...
+; CHECK:      define i64 @helper_unrelated{{.*}}{{$}}
+; CHECK-NOT:  section "revng_inline"
+; SIZE:       define i64 @helper_unrelated(
+; SIZE-NOT:   section "revng_inline"
+define i64 @helper_unrelated(i64 %a) !dbg !40 {
+  %r = add i64 %a, 42
+  ret i64 %r, !dbg !41
+}
+
+; -----------------------------------------------------------------------------
```
aleclearmind: Minimize the unnecessary stuff after this line

tests/unit/llvm_lit_tests/TagSoftfloatHelpers.ll
```diff
@@ -0,0 +1,92 @@
...
+; RUN:     -tag-softfloat-helpers-max-size=2 %s -S -o - \
+; RUN:     | FileCheck --check-prefix=SIZE %s
+
+; -----------------------------------------------------------------------------
+; A function defined in `fpu/`: it is the softfloat leaf. It must NOT be
+; tagged (it's the end of the call chain).
+; CHECK:      define i64 @softfloat_leaf{{.*}}{{$}}
+; CHECK-NOT:  section "revng_inline"
+define i64 @softfloat_leaf(i64 %a) !dbg !10 {
+  ret i64 %a, !dbg !11
```
aleclearmind: Drop all the superflous debug info

lib/HelperArgumentsAnalysis/FixHelpers.cpp
```diff
@@ -273,9 +273,17 @@ void AccessFixer::handle(Instruction &I,
...
+  // read naturally as offsets within the env region.
   Builder.SetInsertPoint(Before);
-  auto *Switch = Builder.CreateSwitch(Address, DefaultCase, Targets.size());
+  auto *EnvGlobal = Variables.envGlobal();
+  Value *EnvInt = Builder.CreateLoad(EnvGlobal->getValueType(), EnvGlobal);
+  Value *EnvRelativeAddress = Builder.CreateSub(Address, EnvInt);
+  auto *Switch = Builder.CreateSwitch(EnvRelativeAddress,
+                                      DefaultCase,
+                                      Targets.size());
+  unsigned EnvBaseOffset = Variables.envBaseOffset();
```
aleclearmind, line 286: This change should be a self-contained commit

include/revng/FunctionIsolation/InlineHelpersAnalysis.h
```diff
@@ -0,0 +1,30 @@
...
+//
+
+#include <optional>
+
+#include "llvm/ADT/BitVector.h"
+#include "llvm/IR/Function.h"
+
+namespace InlineHelpersAnalysis {
+
+/// True if the `Ptr` addresses a `GlobalVariable` which is marked as
```
aleclearmind: `\return true if...`

include/revng/FunctionIsolation/InlineHelpersAnalysis.h
```diff
@@ -0,0 +1,30 @@
...
+/// `constant`.
+bool isLoadFromConstantGlobal(const llvm::Value *Ptr);
+
+/// This helper performs a dataflow backward walk from every critical operand
+/// (`switch` conditions and `getelementptr` index operands), in order to
+/// classify each possible argument reached as a critical argument. The return
+/// value is:
+/// - `std::nullopt`: it means that the helper cannot be inlined at any call
+/// site (we have a `load` from runtime memory as a leaf).
```
aleclearmind: Indent
```
/// - `std::nullopt`: it means that the helper cannot be inlined at any call
///   site (we have a `load` from runtime memory as a leaf).
```

Also, as above, do not keep critical information in a parenthesis ("we have a `load` from runtime memory as a leaf"). Expand.

include/revng/FunctionIsolation/InlineHelpersAnalysis.h
```diff
@@ -0,0 +1,30 @@
...
+#include <optional>
+
+#include "llvm/ADT/BitVector.h"
+#include "llvm/IR/Function.h"
+
+namespace InlineHelpersAnalysis {
+
+/// True if the `Ptr` addresses a `GlobalVariable` which is marked as
+/// `constant`.
+bool isLoadFromConstantGlobal(const llvm::Value *Ptr);
```
aleclearmind: Never abbreviate. No `Ptr`, no` Args`.

lib/FunctionIsolation/InlineHelpersAnalysis.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include "llvm/Support/raw_ostream.h"
+
+#include "revng/FunctionIsolation/InlineHelpersAnalysis.h"
+#include "revng/Support/Debug.h"
+
+using namespace llvm;
+
+namespace InlineHelpersAnalysis {
+
+bool isLoadFromConstantGlobal(const Value *Ptr) {
```
aleclearmind: This function works on pointers, not loads. The name is misleading.

include/revng/FunctionIsolation/InlineHelpersAnalysis.h
```diff
@@ -0,0 +1,30 @@
...
+
+/// This helper performs a dataflow backward walk from every critical operand
+/// (`switch` conditions and `getelementptr` index operands), in order to
```
aleclearmind: Expand the definition of "critical" in a list, not just a parenthesis.

lib/FunctionIsolation/InlineHelpersAnalysis.cpp
```diff
@@ -0,0 +1,85 @@
...
+  const Value *Stripped = Ptr->stripPointerCasts();
+  const auto *GV = dyn_cast<GlobalVariable>(Stripped);
+  return GV != nullptr && GV->isConstant();
+}
+
+std::optional<BitVector> computeCriticalArgsFor(const Function &F) {
+  BitVector Critical(F.arg_size(), false);
+  if (F.isDeclaration()) {
+    return Critical;
```
aleclearmind: We can inline declarations? 🤔

lib/FunctionIsolation/InlineHelpersAnalysis.cpp
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  SmallPtrSet<const Value *, 32> Visited;
+  SmallVector<const Value *, 32> Worklist;
+
+  for (const BasicBlock &BB : F) {
```
aleclearmind: Iterate once over the whole function and do both things.

lib/FunctionIsolation/InlineHelpersAnalysis.cpp
```diff
@@ -0,0 +1,85 @@
...
+}
+
+std::optional<BitVector> computeCriticalArgsFor(const Function &F) {
+  BitVector Critical(F.arg_size(), false);
+  if (F.isDeclaration()) {
+    return Critical;
+  }
+
+  SmallPtrSet<const Value *, 32> Visited;
+  SmallVector<const Value *, 32> Worklist;
```
aleclearmind: `OnceQueue`?

lib/FunctionIsolation/InlineHelpersAnalysis.cpp
```diff
@@ -0,0 +1,85 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "llvm/ADT/SmallPtrSet.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/IR/Argument.h"
+#include "llvm/IR/Constants.h"
+#include "llvm/IR/GlobalVariable.h"
+#include "llvm/IR/Instructions.h"
+#include "llvm/Support/raw_ostream.h"
```
aleclearmind: Remove?

lib/FunctionIsolation/InlineHelpersAnalysis.cpp
```diff
@@ -0,0 +1,85 @@
...
+        for (const auto &Index : GEP->indices()) {
+          Worklist.push_back(Index);
+        }
+      }
+    }
+  }
+
+  while (not Worklist.empty()) {
+    const Value *V = Worklist.pop_back_val();
+    if (not Visited.insert(V).second) {
```
aleclearmind: 1. Reduce your usage of `continue`. Early `return`s are good, `continue` is evil.
   The whole body of the loop can be rewritten without using `continue` but with proper `if (...) { } else if (...) { } else { }`.
2. Comment each case with the reasoning.

lib/FunctionIsolation/PromoteCSVs.cpp
```diff
@@ -236,6 +238,21 @@ Function *PromoteCSVs::createWrapper(const WrapperKey &Key) {
...
+  if (HelperWrapper->getSection() == InlineHelpersSection) {
+    auto
+      Critical = InlineHelpersAnalysis::computeCriticalArgsFor(*HelperWrapper);
+    revng_assert(Critical.has_value(),
+                 "wrapper body unexpectedly produced an un-foldable "
+                 "result; the wrapper has neither switches nor "
+                 "structured-pointer GEPs and should always be "
+                 "foldable");
+    InlineHelpersPolicy::attach(*HelperWrapper, *Critical);
+  }
```
aleclearmind: I don't understand. A helper wrapper is always inlinable, we know by construction it will be, don't we?

lib/FunctionIsolation/InlineHelpersStaticPolicy.cpp
```diff
@@ -0,0 +1,117 @@
...
+    for (const Instruction &I : BB) {
+      if (isa<InsertValueInst>(&I)) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
+class InlineHelpersStaticPolicy : public ModulePass {
```
aleclearmind: The pass is not a "policy".
The name should reflect what the pass does.

aleclearmind:
I'm confused by how various parts are organized:

1. Two namespaces for rather simple things (`InlineHelpersPolicy`, `InlineHelpersAnalysis`). Can we merge them?
   If you want to have multiple small related files, create a new subdirectory, but beware not to use `../` in includes then.
2. I'm not sure why you put `InlineHelpersAnalysis`, `InlineHelpersPolicy` and `InlineHelpersStaticPolicy` under `librevngFunctionIsolation`, these are all things that happen at build-time, so they belong more under `HelperArgumentsAnalysis`, or maybe a new library.

   The only thing that's needed at run-time is to decode the bitvector in the metadata.

   There's also a use of `computeCriticalArgsFor` on the wrappers, but as I pointed out in a dedicated comment, that's not necessary.

aleclearmind:
NOTETOSELF: review from `InlineHelpersPolicy: define `revng.inline.policy`, included.

lib/HeadersGeneration/HelpersToHeader.cpp
aleclearmind: I don't understand, but if this goes away, point me to the new commit to review.

lib/HeadersGeneration/HelpersToHeader.cpp
andrealinux1: I removed the abbreviations in both `[HelpersToHeader: remove `_PACKED` handling]` and `[Clifter: remove abbreviation]`.

lib/HeadersGeneration/HelpersToHeader.cpp
andrealinux1: In the `[HelpersToHeader: remove _PACKED handling]` commit, I removed the special handling of `_PACKED` on the old backed, so everything is uniform, I left the padding at the end.

lib/FunctionIsolation/InlineHelpersPolicy.cpp
```diff
@@ -0,0 +1,83 @@
...
+         Idx = CriticalArgs.find_next(Idx)) {
+      auto *Constant = ConstantInt::get(Type::getInt32Ty(Context), Idx);
+      Operands.push_back(ConstantAsMetadata::get(Constant));
+    }
+  }
+
+  F.setMetadata(MetadataKey, MDTuple::get(Context, Operands));
+}
+
+Policy read(const Function &F) {
```
aleclearmind: 1. `deserialize`
2. When you do metadata heavy (de-)serialization, use `QuickMetadata`.

lib/FunctionIsolation/InlineHelpersPolicy.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include "revng/Support/Assert.h"
+
+using namespace llvm;
+
+namespace InlineHelpersPolicy {
+
+static constexpr StringLiteral AlwaysTag = "always";
+static constexpr StringLiteral ConditionalTag = "conditional";
+
+void attach(Function &F, const BitVector &CriticalArgs) {
```
aleclearmind: `serialize`

include/revng/FunctionIsolation/InlineHelpersPolicy.h
```diff
@@ -0,0 +1,39 @@
...
+};
+
+struct Policy {
+  Kind Tag;
+  llvm::BitVector CriticalArgs;
+};
+
+// Attach `!revng.inline.policy` to `F` based on `CriticalArgs`. An empty
+// `CriticalArgs` produces the `"always"` tag; a non-empty one produces
+// `"conditional"` followed by one `i32` ConstantInt per set bit.
```
aleclearmind: I'd just attach a `iKKK` where `KKK` is the number of arguments.
For instance, if you have 7 arguments, use a `i7`. Each bit represents an argument.

If `0`, no arguments are critical and means "always".

lib/FunctionIsolation/InlineHelpersStaticPolicy.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+    for (Function &F : M) {
+      if (F.getSection() != InlineHelpersSection) {
+        continue;
+      }
+
+      // We classify each helper as potentially inlinable following three
+      // criteria, which are short-circuited to avoid useless computation
+
+      // 1. Helpers in a cyclic SCC on the callgraph, cannot be unlined without
```
aleclearmind: 1. Drop numbers, they go out of sync
2. Outline the part setting `NeverInline` into a `bool mustNotInline(...)` function. Early return whenever a no-inlining conditions is met and `revng_log` right away.

lib/FunctionIsolation/InlineHelpersStaticPolicy.cpp
aleclearmind: 1. `MarkNonInlineableHelpers`?
2. Docs

lib/FunctionIsolation/InlineHelpersStaticPolicy.cpp
```diff
@@ -0,0 +1,117 @@
...
+    CallGraph CG(M);
+    for (auto It = scc_begin(&CG), End = scc_end(&CG); It != End; ++It) {
+      if (It.hasCycle()) {
+        for (auto &Node : *It) {
+          if (const Function *F = Node->getFunction()) {
+            Recursive.insert(F);
+          }
+        }
+      }
+    }
```
aleclearmind: Drop

lib/FunctionIsolation/InlineHelpers.cpp
```diff
@@ -3,65 +3,141 @@
...
+    return true;
+  }
+
+  // Build a per critical-argument breakdown: each entry records whether the
+  // actual value at this call site is a compile-time constant. A single
+  // `NON-CONST` entry is enough to veto the inline, but we walk the entire
+  // set anyway so the log can show the full picture for every call.
+  const BitVector &Critical = P.CriticalArgs;
+  SmallString<128> ArgBreakdown;
+  raw_svector_ostream OS(ArgBreakdown);
```
aleclearmind: You're computing the debug string even if the debugger is not active, which is bad.

```c++
revng_log(Log, "Analyzing arguments")
LoggerIndent Indent(Log);

for (int Idx = Critical.find_first(); Idx >= 0;
       Idx = Critical.find_next(Idx)) {
   // ...
   revng_log(Log, "Argument " << Idx << ": ...");
   // ...
}
```

lib/FunctionIsolation/InlineHelpers.cpp
```diff
@@ -105,10 +181,31 @@ static void dropDebugOrPseudoInst(Function *F) {
...
-  doInline(F);
+
+  if (F->isDeclaration()) {
+    return;
+  }
+
+  revng_log(Log, "======== analyzing " << F->getName() << " ========");
+
+  // Dump the IR of the function before the inlining takes place
+  if (Log.isEnabled() and not DumpBeforeDir.empty()) {
```
aleclearmind: Drop `Log.isEnabled()`. Never use `Logger`s for anything other than printing to loggers.

lib/FunctionIsolation/InlineHelpers.cpp
```diff
@@ -105,10 +181,31 @@ static void dropDebugOrPseudoInst(Function *F) {
...
 void InlineHelpers::run(Function *F) {
-  // Since all the helpers have been already inlined the `doInline` function can
-  // be called only once to inline all the helper functions.
-  doInline(F);
+
+  if (F->isDeclaration()) {
+    return;
+  }
+
+  revng_log(Log, "======== analyzing " << F->getName() << " ========");
```
aleclearmind: Never to formatting with `===`. Just `"Analyzing " << F->getName()`.

To separate, indent.

lib/FunctionIsolation/InlineHelpers.cpp
```diff
@@ -105,10 +181,31 @@ static void dropDebugOrPseudoInst(Function *F) {
...
+  // Dump the IR of the function before the inlining takes place
+  if (Log.isEnabled() and not DumpBeforeDir.empty()) {
+    dumpFunctionToDir(*F, DumpBeforeDir);
+  }
+
+  // Build time `slim-down-helpers-module` pre-flattens every intra helper call
+  // edge, but `promote-csvs` creates a wrapper at runtime, hiding the real call
+  // to the underlying helper. In this way, we traverse the call edge and inline
+  // also the underlying wrapper. Termination should be guaranteed because
+  // helpers in any SCC cycle are filtered out.
```
aleclearmind: Just say:
```
// `slim-down-helpers-module` already full inlines helpers, but we might still have wrappers created by...
```

lib/FunctionIsolation/InlineHelpers.cpp
```diff
@@ -129,6 +226,22 @@ void InlineHelpersPass::linkRequiredHelpers(llvm::Module &M) {
                "QEMU helpers of multiple architectures were linked inside the "
                "module");
 
+  // Collect the names of revng_inline helpers that lack a body in `M` and
+  // therefore need to be cloned from the per-architecture to-inline bitcode.
+  SmallVector<StringRef, 32> MissingBodies;
```
aleclearmind: All helpers miss the body.
This is just the list of helpers to inline.

lib/FunctionIsolation/InlineHelpers.cpp
```diff
@@ -167,9 +278,9 @@ bool InlineHelpersPass::runOnModule(llvm::Module &M) {
       Isolated.push_back(&F);
 
   llvm::Task T(Isolated.size(), "Inline helpers");
+  InlineHelpers IH;
   for (Function *F : Isolated) {
     T.advance(F->getName());
-    InlineHelpers IH(*F->getParent());
```
aleclearmind: You sure there's no state in `InlineHelpers` that, by being reused, might break the inlining?

tests/unit/llvm_lit_tests/InlineHelpers.ll
```diff
@@ -0,0 +1,73 @@
...
+; because all its calls were inlined away.
+; CHECK:       declare {{.*}}@helper_with_critical
+; CHECK:       declare {{.*}}@helper_no_critical
+
+; The dynamic-op call site must survive; the constant-op one and
+; the no-critical-args helper must have been inlined away.
+; CHECK-LABEL: define i64 @my_isolated_function(i64 %v, i32 %op)
+; CHECK-NOT:   call i64 @helper_with_critical(i64 %v, i32 0)
+; CHECK-NOT:   call i64 @helper_no_critical
+; CHECK:       call i64 @helper_with_critical(i64 %v, i32 %op)
```
aleclearmind: Whenever possible, `CHECK`s directives must be close to the corresponding part in the source code.

python/revng/internal/pipeline.yml
```diff
@@ -171,6 +171,9 @@ branches:
           passes:
             - remove-exceptional-functions
             - strip-dead-debug-info
+            # Verifier: from this point on, lifted (Isolated) functions
+            # must contain no `getelementptr` instructions.
+            - assert-no-gep
```
aleclearmind: 1. Just use the "isolated" term, not lifted.
2. Why here and not earlier?
3. I'd do a pass only if you expect to schedule it multiple times. Otherwise, if this could in principle stay in a pass we implemented, I'd put it there. Possibly under a `if (VerifyLog.isEnabled())` (only exception to use a logger for non-logging purposes).

share/revng/test/tests/floating-point/floating-point.c.filecheck
```diff
@@ -0,0 +1,84 @@
...
+CHECK:       }
+
+CHECK-LABEL: @div_float32(
+CHECK:       @float32_div(
+CHECK:       }
+
+// Comparison
+
+CHECK-LABEL: @compare_float64(
+CHECK:       @float64_{{(compare|lt)}}{{(_quiet)?}}(
```
aleclearmind: What's this? I expect a simple, clear, predictable call. Not a regexp.

share/revng/test/tests/floating-point/floating-point.c.filecheck
```diff
@@ -0,0 +1,84 @@
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+// These directives are matched against the `enforce-abi` artifact of the
+// `floating-point.c` test in the `revng-qa` repository. Each isolated function
+// must contain the softfloat call that implements its floating-point operation.
+// The closing-brace line after each call keeps every match scoped to the body
+// of the function its label opened.
+//
```
aleclearmind, line 8: Drop

lib/HelperArgumentsAnalysis/FixHelpers.cpp
andrealinux1: I extracted in `[FixHelpers: switch on env-relative offsets]` these changes.

lib/Lift/VariableManager.cpp
andrealinux1: I added a comment with an explanation for this in the new `[FixHelpers: switch on env-relative offsets]` commit grouping all the changes regarding `env`.

share/doc/revng/developer-manual/qemu-helpers.md
andrealinux1: In `[docs: tested docs for QEMU helper functions]`:

1. I moved the `qemu-helpers.md` doc file in `user-manual/key-concepts/`, and referenced it in the `index.md` of that section.
Still, I don't find it super pertinent to have the `qemu-helpers.md` in `key-concepts`, but it was the only reasonable way to mention in the an plausible index.

Let me know how you want to restructure this in a more sensible way.

2. Regarding cross-referencing it from other docs, there are no places in the current doc where we talk about helpers. We have some places in the _legacy_ doc (the one which is under `doc/`...), but I don't think we want to update that one.

aleclearmind: 1. Create `developer-manual/` and put it there for now.
   Also create an entry in the index right after "User's manual" for "Developer's manual" and link it there
2. OK, ignore it.

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
andrealinux1: 1. ACK
2. ACK.
3. I dropped the option in favor of your other suggestions in `[MarkInlineHelpersUpTo: introduce pass]`.

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
andrealinux1: 1. ACK, changed the term.
2. Restructured the check in `[MarkInlineHelpersUpTo: introduce pass]`.

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
andrealinux1: Implemented the changes and rephrased also the commit message to explicitly reflect this idea in `[MarkInlineHelpersUpTo: introduce pass]`

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
andrealinux1: Dropped in `[MarkInlineHelpersUpTo: introduce pass]`

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
andrealinux1: Corrected this in `[MarkInlineHelpersUpTo: introduce pass]`.

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
andrealinux1: I tried to change all these cases without the curly braces, let's hope this stick in my mind.

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
andrealinux1: Changed this in `[MarkInlineHelpersUpTo: introduce pass]`

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
andrealinux1: Removed the numbering in `[MarkInlineHelpersUpTo: introduce pass]`.

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
andrealinux1: Added your proposed indentation in `[MarkInlineHelpersUpTo: introduce pass]`

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
andrealinux1: 1. Merged these in `[MarkInlineHelpersUpTo: introduce pass]`.
2. Dropped it.

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
andrealinux1: Restructured this to make use of `OnceQueue` in `[MarkInlineHelpersUpTo: introduce pass]`

lib/HelperArgumentsAnalysis/TagSoftfloatHelpers.cpp
andrealinux1: I added some a doxygen comment above the pass in `[MarkInlineHelpersUpTo: introduce pass]`.

CMakeLists.txt
andrealinux1: It's a bugfix, so that the helpers modules are rebuilt whenever there is a change in the library code.

Added a new specific commit for this `[cmake: fix helper-annotation pipeline dependency]`.

lib/FunctionIsolation/PromoteCSVs.cpp
andrealinux1: ~~Yes, I changed this in `[PromoteCSVs: compute revng.inline.policy]` so that we always mark an `wrapper` as inlinable just on the basis of the attribute of the underlying helper, so we inline wrappers for inlanable helpers.~~

Done.

CMakeLists.txt
andrealinux1: In `[MarkInlineHelpersUpTo: introduce pass]` I've added an extensive comment explaining what each pass of the pipelines does.

share/revng/test/tests/floating-point/floating-point.c.filecheck
andrealinux1: Dropped in `[tests: add floating-point end-to-end tests]`

lib/FunctionIsolation/InlineHelpers.cpp
andrealinux1: Yeah, now `InlineHelpers` does not have any state, previously, its constructor computed the set of recursive helpers (callgraph in a SCC), but this criterion has now been moved away in another pass.

Anyway, I restored in `[InlineHelpers: Use revng.inline.policy metadata]` the previous form with the constructor called just before the `run` method, so things are more clear.

tests/unit/llvm_lit_tests/TagSoftfloatHelpers.ll
andrealinux1: In `[MarkInlineHelpersUpTo: introduce pass]` I_ should have removed all the non-essential metadatas from the test file, leaving just the minimal structure to define the file path.

tests/unit/llvm_lit_tests/TagSoftfloatHelpers.ll
andrealinux1: I dropped all the non essential debug info in `[MarkInlineHelpersUpTo: introduce pass]`.

include/revng/FunctionIsolation/InlineHelpersAnalysis.h
andrealinux1: Changed this in `[DetectInlinableHelpers: introduce pass]`.

CMakeLists.txt
aleclearmind: 2. Can you use just `revngHelperArgumentsAnalysis` instead of `$<TARGET_FILE:revngHelperArgumentsAnalysis>`?

include/revng/FunctionIsolation/InlineHelpersAnalysis.h
andrealinux1: I reworked this definitions and expanded them in `[DetectInlinableHelpers: introduce pass]`, I'll leave this open so you can take another look at the new form.

include/revng/FunctionIsolation/InlineHelpersAnalysis.h
andrealinux1: Removed this class of abbreviations in `[DetectInlinableHelpers: introduce pass]`.

lib/FunctionIsolation/InlineHelpersAnalysis.cpp
andrealinux1: Reworked this is as `isPointerToConstantGlobal` in `[DetectInlinableHelpers: introduce pass]`.

include/revng/FunctionIsolation/InlineHelpersAnalysis.h
andrealinux1: I expanded the doxygen doc of this class in `[DetectInlinableHelpers: introduce pass]`. I'll leave this open so you can take another look at the new form.

lib/FunctionIsolation/InlineHelpersAnalysis.cpp
andrealinux1: ~~Nope, this was crearly wrong, corrected in `[DetectInlinableHelpers: introduce pass]`.~~

Done.

lib/FunctionIsolation/InlineHelpersAnalysis.cpp
andrealinux1: ~~In `[DetectInlinableHelpers: introduce pass]` I now collect both the needed `Value`s in the same iteration loop.~~

Done.

lib/FunctionIsolation/InlineHelpersAnalysis.cpp
andrealinux1: ~~Changed this in `[DetectInlinableHelpers: introduce pass]`, `OnceQueue` is really usefull!~~

Done.

lib/FunctionIsolation/InlineHelpersAnalysis.cpp
andrealinux1: ~~`Constants.h` is needed, I removed the other ones in `[DetectInlinableHelpers: introduce pass]`.~~

Done.

lib/FunctionIsolation/InlineHelpersStaticPolicy.cpp
andrealinux1: This pass is now named `DetectInlinableHelpers`, and it also contains the definition of the data structure used to contain the `inline` policy serialization stuff.

lib/FunctionIsolation/InlineHelpersAnalysis.cpp
andrealinux1: ~~In `[DetectInlinableHelpers: introduce pass]`:~~
~~1. Restructured the loop taking your suggestions on using early `return` and `if`...~~
~~2. I added a comment for each case explaining the logic behind it.~~

Done, let me know if ok.

CMakeLists.txt
andrealinux1: 2. Yes the targets now only depend on the target names, the `TARGET_FILE` has already been dropped.

share/doc/revng/developer-manual/qemu-helpers.md
andrealinux1: 1. I think that what you are referring to as "the index", is indeed the `mkdocs.yml`, not the `index.md`. I did update it per your instructions in `[docs: tested docs for QEMU helper functions]`.

However, in order to make this documentation really easily reachable from the web interface, I added a new entry in the `index.md`, referring the new `Developer's Manual` category, and adding just the `qemu-helpers.md` reference.

You can take a look at it in the same commit, let me know if ok or you want changes.

2. ACK.

lib/FunctionIsolation/InlineHelpersPolicy.cpp
andrealinux1: Done.

lib/FunctionIsolation/InlineHelpersPolicy.cpp
andrealinux1: Done.

lib/FunctionIsolation/InlineHelpersStaticPolicy.cpp
andrealinux1: Done.

lib/FunctionIsolation/InlineHelpersStaticPolicy.cpp
andrealinux1: In `[DetectInlinableHelpers: introduce pass]`:
1. The pass is now called `DetectInlinableHelpersPass`, as the cpp. Let me know if you want a further refactor.
2. Done.

lib/FunctionIsolation/InlineHelpersStaticPolicy.cpp
andrealinux1: Done.

lib/FunctionIsolation/InlineHelpers.cpp
andrealinux1: Done.

lib/FunctionIsolation/InlineHelpers.cpp
andrealinux1: Done.

lib/FunctionIsolation/InlineHelpers.cpp
andrealinux1: Done.

lib/FunctionIsolation/InlineHelpers.cpp
andrealinux1: Done.

tests/unit/llvm_lit_tests/InlineHelpers.ll
andrealinux1: Done.

python/revng/internal/pipeline.yml
andrealinux1: In `[Canonicalize: add assert-no-gep verifier]`:
1. Done.
2. Moved it just after `inline-helpers` pass. Added a comment.
3. When we took a look at this, we said to assert no `gep`s in a pass, cause it could have been handy in the future to enforce this invariant of the IR explicitly (instead of relying on the fact that downstream passes, like `MakeModelGEP` abort on finding them). However, sic stantibus rebus, the only pass that could inline `gep`s is indeed `inline-helpers`, so if you prefer, I can move this at the end of it.

include/revng/FunctionIsolation/InlineHelpersPolicy.h
andrealinux1: I implemented in `[DetectInlinableHelpers: introduce pass]` what you suggested, however, a note on this:
- If we have a function with the first arguments that need to be constant, we need to represent it with an `i2` -> `10`.
- LLVM IR printer, uses the signed form to print that, so the metadata is rendered as `!{i2 -2}`. Not the nicest thing I've seen. (An example in the [unit tests](https://github.com/andrealinux1/revng/blob/dfea37891529751202337fa19adda130180b8878/tests/unit/llvm_lit_tests/DetectInlinableHelpers.ll#L127)
- However this is only a problem when looking at the serialized version of the module, programatically this is not evident (we can just `.setBit(idx)`...).
- A way to take your suggestion, is using `iK + 1`, so the most significant bit is always `0` and we never render negative integers, or decide to go with `i64` directly.

andrealinux1:
> I'm confused by how various parts are organized:
> 
>     1. Two namespaces for rather simple things (`InlineHelpersPolicy`, `InlineHelpersAnalysis`). Can we merge them?
>        If you want to have multiple small related files, create a new subdirectory, but beware not to use `../` in includes then.
> 
>     2. I'm not sure why you put `InlineHelpersAnalysis`, `InlineHelpersPolicy` and `InlineHelpersStaticPolicy` under `librevngFunctionIsolation`, these are all things that happen at build-time, so they belong more under `HelperArgumentsAnalysis`, or maybe a new library.
>        The only thing that's needed at run-time is to decode the bitvector in the metadata.
>        There's also a use of `computeCriticalArgsFor` on the wrappers, but as I pointed out in a dedicated comment, that's not necessary.

@aleclearmind  Regarding this broader comment:
1. We now have a `revngBuildTimeHelperAnalysis` library, which contains the things we need at build time to analyze and tag the helpers.
2. The previous split in different files and namespaces for the `InlineHelperAnalysis` and `InlineHelpersPolicy` has gone, everying is now under `DetectInlinableHelpers.h/cpp`.
3. To decode the metadata, we import the header and link against the `revngBuildTimeHelperAnalysis` library.

share/revng/test/tests/floating-point/floating-point.c.filecheck
andrealinux1: - The regexp is there in order to have a single `CHECK` line for the different architectures, which relies on different softfloat functions to implement the operations.
- An alternative to this would be to use `FileCheck` prefixes, like `CHECK-X86_64`, `CHECK-ARM64` etc... in order to avoid the regexp and have a clearer match, at the cost of a bit more verbose `.filecheck` file and driver.
- Would you prefer this approach?

share/doc/revng/developer-manual/qemu-helpers.md
aleclearmind: Very reasonable.

CMakeLists.txt
```diff
@@ -234,12 +234,31 @@ foreach(
...
+  # * `cpu-loop-exit` rewrites the CPU loop-exit pattern into something
+  #   `analyze-helper-arguments` can understand.
+  # * `analyze-helper-arguments` discovers, for every helper, which arguments
+  #   alias CSVs at the call site.
+  # * `loop-unroll` unrolls loops so the subsequent `fix-helpers` pass can
+  #   rewrite the accesses contained in bounded loops bodies.
+  # * `fix-helpers` rewrites helper bodies to access CSVs via the env pointer,
+  #   switching on env-relative offsets.
+  # * `detect-inlinable-helpers` excludes not viable helpers from inlining, and
+  #   performs the `ConstantOperands` analysis on the helpers bodies.
```
aleclearmind: > performs the `ConstantOperands` analysis on the helpers bodies

Maybe I'd leave this out, `ConstantOperands` is not something well known.

Good job!

tests/unit/llvm_lit_tests/TagSoftfloatHelpers.ll
aleclearmind: *massages nipples*

lib/HeadersGeneration/HelpersToHeader.cpp
```diff
@@ -76,9 +75,9 @@ static void printLLVMTypeDeclaration(const llvm::StructType *S,
     revng_assert(PreviousOffsetInBits % 8 == 0);
     uint64_t Alignment = StructLayout->getAlignment().value();
     uint64_t ByteSize = PreviousOffsetInBits / 8;
-    uint64_t NumUnalignedBytes = ByteSize % Alignment;
-    uint64_t AlignedByteSize = NumUnalignedBytes ?
-                                 (ByteSize - NumUnalignedBytes + Alignment) :
+    uint64_t AlignedBytesNumber = ByteSize % Alignment;
+    uint64_t AlignedByteSize = AlignedBytesNumber ?
+                                 (ByteSize - AlignedBytesNumber + Alignment) :
```
pfez: This rename is wrong. It's `Unaligned` not `Aligned`.

Also, I've seen LLVM has helper functions to deal with alignment without having to spell out this stuff by hand. Look for `alignTo` in LLVM and try and see if you can reuse that in all the places where we do this alignment by hand.

CMakeLists.txt
```diff
@@ -234,14 +234,33 @@ foreach(
...
+  #   downstream analyses see the smallest possible IR.
+  # * `cpu-loop-exit` rewrites the CPU loop-exit pattern into something
+  #   `analyze-helper-arguments` can understand.
+  # * `analyze-helper-arguments` discovers, for every helper, which arguments
+  #   alias CSVs at the call site.
+  # * `loop-unroll` unrolls loops so the subsequent `fix-helpers` pass can
+  #   rewrite the accesses contained in bounded loops bodies.
+  # * `fix-helpers` rewrites helper bodies to access CSVs via the env pointer,
+  #   switching on env-relative offsets.
+  # * `detect-inlinable-helpers` excludes not viable helpers from inlining, and
```
aleclearmind: `detect-`*un*`inlinable-helpers`?

include/revng/FunctionIsolation/InlineHelpersPolicy.h
aleclearmind: 1. > If we have a function with the first arguments that need to be constant, we need to represent it with an i2 -> 10.

   Why?
2. .
3. .
4. I'm OK with `iK + 1`.

python/revng/internal/pipeline.yml
aleclearmind: 3. The point is that I don't want a pass named `assert-no-gep`, I'd like `assert-we-respect-$FANCY_INVARIANT_NAME_OF_THIS_STAGE_OF_THE_PIPELINE`, but we're not ready for that and I don't want to spiral into more `assert-no-xxx` passess.
   Do exactly what `PostLiftVerifyPass`, which is the closest thing.

lib/BuildTimeHelperAnalysis/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+revng_add_analyses_library_internal(
+  revngBuildTimeHelperAnalysis DetectInlinableHelpers.cpp
```
aleclearmind: `revngHelperInliningAnalyses`

lib/FunctionIsolation/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ llvm_map_components_to_libnames(LLVM_LIBRARIES TransformUtils)
 target_link_libraries(
   revngFunctionIsolation
   revngABI
+  revngBuildTimeHelperAnalysis
```
aleclearmind: I'd prefer not to link this library at run-time at all.

Let's move `deserialize` into `IRHelpers.cpp` as `deserializeInliningPolicy`.

It's not nice, but I prefer to double down on not linking runtime code to exclusively build-time code if we can, at the cost of putting more stuff in `IRHelpers.{h,cpp}`.

lib/HeadersGeneration/HelpersToHeader.cpp
andrealinux1: 1. In `[HelpersToHeader: remove `_PACKED` handling]` and `[Clifter: remove abbreviation]` i fixed the typo, thanks for spotting it.!
2. Coming

andrealinux1: In `[HelpersToHeader: remove _PACKED handling]` and `[ Clifter: remove abbreviation and use alignTo]` we now make use of `llvm::alignTo` to compute the alignement.

CMakeLists.txt
andrealinux1: Done in `[MarkInlineHelpersUpTo: introduce pass]`

CMakeLists.txt
andrealinux1: I don't super dig it, but changed this in `[DetectUninlinableHelpers: introduce pass`]

lib/BuildTimeHelperAnalysis/CMakeLists.txt
andrealinux1: Done

lib/Lift/VariableManager.cpp
aleclearmind: Are you 100% this is related to GVN? I'd leave out references to specific passes unless you're 100% sure.

include/revng/FunctionIsolation/InlineHelpersPolicy.h
andrealinux1: Done in `[DetectUninlinableHelpers: introduce pass]`

lib/Lift/VariableManager.cpp
andrealinux1: Removed reference to GVN

lib/FunctionIsolation/CMakeLists.txt
andrealinux1: Done now in `[DetectUninlinableHelpers: introduce pass]`

python/revng/internal/pipeline.yml
andrealinux1: Ok, what you suggested is now in `[InlineHelpers: add post-inlining no-GEP verifier]`

share/revng/test/tests/floating-point/floating-point.c.filecheck
andrealinux1: So, we discussed about the following 3 points which are now implemented in [`tests: add floating-point end-to-end test`]:
1) Improve the formatting of the `CHECK` blocks so it is immediately clear that we are matching function blocks and `call`s to the softfloat helpers.
2) To expand the regexps used to match the different alternatives softfloat `call`s we match for each architecture, so that it is immediately clear, looking at this file, which are the softfloat `call`s possible alternatives (e.g. `call {{.*}}@{{(float64_compare|float64_compare_quiet|float64_lt)}}(`).
3) To match the `0` scale factor passed as argument for the softfloat  `scalbn` alternatives (e.g., `float64_to_int32_scalbn\([^,]+, [^,]+, i32 noundef 0)`).

- An additional note:
4) When testing the `convert_int32_t_to_float64` (and a sibling one), we match for both `\(int32_to_float64\(|int64_to_float64`. Why is that? The point is that for `s390x` and `aarch64` the QEMU translator sign-extend the 32 bit arguments to 64 bit before actually calling the softfloat implementation, which then results to be the 64 bit variant.

- If you want to make things more explicit:
5) We could employ `FileCheck` prefixes to make explicit what we match for the different architectures in the lines where we have the `|` alternatives (e.g., instead of having `CHECK: call {{.*}}@{{(float64_compare|float64_compare_quiet|float64_lt)}}(` we have a different `CHECK-AARCH64:` line just checking the corresponding alternative).

aleclearmind: 4. Add a comment

python/revng/internal/pipeline.yml
aleclearmind: It has to be *exactly* the same.

```c++
class Post...VerifyPass {
```

lib/Clifter/Clifter.cpp
```diff
@@ -332,11 +332,7 @@ class ClifterImpl final : public Clifter {
       revng_assert(PreviousOffsetInBits % 8 == 0);
       uint64_t Alignment = StructLayout->getAlignment().value();
       uint64_t ByteSize = PreviousOffsetInBits / 8;
-      uint64_t NumUnalignedBytes = ByteSize % Alignment;
-      uint64_t AlignedByteSize = NumUnalignedBytes ?
-                                   (ByteSize - NumUnalignedBytes + Alignment) :
-                                   ByteSize;
-      revng_assert(AlignedByteSize == StructByteSize);
+      revng_assert(llvm::alignTo(ByteSize, Alignment) == StructByteSize);
```
pfez, line 335: Nice. There is another occurrence in `HelpersToHeader.cpp`.

lib/HeadersGeneration/HelpersToHeader.cpp
pfez: Thanks. I'm closing this thread. I opened another one separately because there's still another instance I know of in the codebase of a place that could benefit from `alignTo`.

lib/Clifter/Clifter.cpp
andrealinux1: Right and isn't this already ok and addressed in the same way? https://github.com/andrealinux1/revng/blob/eff2330e0223d85bd4052fcf64482109c0d9cb86/lib/HeadersGeneration/HelpersToHeader.cpp#L78

share/revng/test/tests/floating-point/floating-point.c.filecheck
andrealinux1: Done in `[tests: add floating-point end-to-end tests]`

include/revng/ABI/Definition.h
```diff
@@ -99,14 +99,8 @@ class Definition : public generated::Definition {
...
-    if (Offset == 0)
-      return 0;
-
     revng_assert(llvm::isPowerOf2_64(Alignment));
-    if (Offset % Alignment != 0)
-      return Offset + Alignment - Offset % Alignment;
-
-    return Offset;
+    return llvm::alignTo(Offset, Alignment);
   }
```
andrealinux1: @pfez In a comment you closed you mention of another place that could benefit from the usage of `alignTo`, but I cannot find the new thread that you mention having opened. I guess is this the place? Is the fix ok?

python/revng/internal/pipeline.yml
andrealinux1: Ah ok I thought you meant _take inspiration from_...

In `[ InlineHelpers: add post-inlining no-GEP verifier]` I now introduce the ` PostInlineHelpersVerifyPass` which checks for no GEPs, completely identical to `PostLiftVerifyPass` in the structure and intent.

lib/Clifter/Clifter.cpp
pfez: Yeah it was that, but for some reason I could not see it when I reviewed this.

include/revng/ABI/Definition.h
pfez: It wasn't this one, it was the one in `HelpersToHeader.cpp` that I couldn't see.
But this is a good place one to replace too.

In fact, this whole method can be dropped and replaced with `llvm::alignToPowerOf2`, which also embeds the assertion.

Not sure if this in the scope of the PR though. Let's hear @aleclearmind and maybe you can leave this out.

aleclearmind: I've no opinions, I'd like to get this merged. You guys decide.

andrealinux1: Ok, I'm crafting the version dropping the method and replacing the callsites with `feature/detect-abi-inline-helpers`, and then @pfez decides if including it or not

andrealinux1: Ok @pfez `[ABI: use llvm::alignToPowerOf2]` should contain what you wanted

aleclearmind:
Merged in 1af2a5a. Thanks.

PR #582, opened 2026-05-12
Clift local variable scope tightening

vasama-work:


lib/Canonicalize/DeferAllocas.cpp
```diff
@@ -0,0 +1,63 @@
...
+bool DeferAllocasPass::runOnFunction(llvm::Function &F) {
+  llvm::DominatorTree DT;
+  DT.recalculate(F);
+
+  bool Modified = false;
+  for (llvm::BasicBlock &BB : F) {
+    // Iterate manually to avoid invalidating the iterator if and when the
+    // alloca instruction is moved out of the basic block.
+    for (auto I = BB.begin(), E = BB.end(); I != E;) {
+      auto Alloca = llvm::dyn_cast<llvm::AllocaInst>(&*I++);
```
pfez: Either use `llvm::make_early_inc_range` or collect a `std::vector<std::pair<AllocaInst *, BasicBlock*>>` and only move the alloca while you're done iterating.

lib/CliftTransforms/TightenVariableScopes.cpp
```diff
@@ -0,0 +1,166 @@
...
+    llvm::MapVector<LocalVariableOp, LocalVariableLocation> Locals;
+
+    auto WalkCallback = [&](mlir::Operation *Op, unsigned OpNestingLevel) {
+      // For each operand of the operation, we check if it is a local variable.
+      for (mlir::Value Operand : Op->getOperands()) {
+        auto Local = Operand.getDefiningOp<LocalVariableOp>();
+
+        // If the defining op is not a local variable, we can skip it.
+        if (not Local)
+          continue;
```
pfez, line 113: We should also skip the stack frame variable. We always want it to stay where it is. In practice I don't expect it to be moved around often, but still we want to put a hard stop to make sure this never happens even if it could.

lib/Clifter/Clifter.cpp
```diff
@@ -1741,6 +1742,11 @@ class ClifterImpl::FunctionClifter {
         auto [Iterator, Inserted] = AllocaMapping.try_emplace(A, Op);
         revng_assert(Inserted);
 
+        // Create a clift.require operation to indicate the position dominating
+        // all uses of this local variable. This prevents the local variable
+        // from being later moved past this point during scope tightening.
+        Builder.create<RequireOp>(Loc, Op);
```
pfez, line 1738: Actually, forget my last comment on special casing the stack frame when tightening the scope of local variables. The right thing to do is to make sure the stack frame variable is always required here.

vasama-work: `defer-allocas` does not move the stack frame alloca, which means that its `RequireOp` is inserted in the basic block where the alloca is, i.e. the entry block (unless for some reason it was already in some other block, but I don't think we ever do that).

lib/Canonicalize/DeferAllocas.cpp
vasama-work: I'll use `llvm::make_early_inc_range`.

PR #584, opened 2026-05-14
cmake: only specify python options where needed

mrjackv:


aleclearmind:
Merged in 358291a. Thanks.

PR #586, opened 2026-05-19
Clift local variable initialiser hoisting

vasama-work:


include/revng/Clift/CliftOpInterfacesBasic.td
```diff
@@ -52,6 +52,16 @@ def Clift_StatementOpInterface : Clift_OpInterface<"StatementOpInterface"> {
         return ConcreteOp::template hasTrait<clift::NoFallthrough>();
       }]
     >,
+    InterfaceMethod<
+      /*desc=*/"Returns a value more directly representing the entity or "
+               "expression to which the specified block argument refers, or "
+               "null if not applicable.",
+      "mlir::Value", "translateBlockArgument", (ins "mlir::BlockArgument":$Argument),
```
pfez: How about `getBlockArgumentAsValue`?

Also, for the cases where this is useful as of now, it seems to me that we always want it to be a `LocalVariableOp`. No?

If so, I think we should aim for a name with a narrower meaning, and we should just return `clift::LocalVariableOp`.

lib/Clift/Clift.cpp
```diff
@@ -616,6 +616,10 @@ mlir::LogicalResult DoWhileOp::verify() {
 
 //===-------------------------------- ForOp -------------------------------===//
 
+mlir::Value ForOp::translateBlockArgument(mlir::BlockArgument Argument) {
+  return getOnlyOp<LocalVariableOp>(getInitializer());
```
pfez, line 620: I find confusing that here you return the initializer, while the other place where we have this (`LocalVariableOp`) returns the local variable itself.

include/revng/Clift/CliftOpInterfacesBasic.td
pfez: Also, why is this in `StatementOpInterface`? I'd make this a separate interface. Currently we only plan to use it for `ForOp` and `LocalVariableOp`. For all the other `StatementOp`s it's not clear to me it does even make sense.

lib/Clift/Clift.cpp
```diff
@@ -898,8 +898,80 @@ mlir::LogicalResult IfOp::verify() {
 
 //===--------------------------- LocalVariableOp --------------------------===//
 
+mlir::Value
+LocalVariableOp::translateBlockArgument(mlir::BlockArgument Argument) {
+  return *this;
```
pfez, line 903: You see? This returns a `LocalVariableOp`, not its initializer, which is incoherent with what we do on `ForOp`.

lib/Clift/Clift.cpp
```diff
@@ -898,8 +898,80 @@ mlir::LogicalResult IfOp::verify() {
...
+                                " argument.";
+      }
+
+      if (R.getArgument(0).getType() != getType()) {
+        return emitOpError() << getOperationName()
+                             << " initializer region argument type must match "
+                                " the local variable type.";
+      }
+    }
+
```
pfez, line 974: From what I read here, it seems to me that adding this argument opens up the door to various ways to pass in other arguments that are not `self`, which was the original motivation for this addition.

I don't envision the use of this argument for anything but `self`, so please verify that. It's a pretty strict condition, but I don't want this to be abused (by users or by some of our optimizations) to pass in other block arguments like e.g. globals, or function arguments.

lib/CliftTransforms/VariableInitializerHoisting.cpp
```diff
@@ -0,0 +1,92 @@
...
+namespace {
+
+static bool visitLocalVariable(LocalVariableOp Local, bool HoistInitializer) {
+  mlir::Region &Initializer = Local.getInitializer();
+  if (not Initializer.empty())
+    return false;
+
+  auto Statement = getNextOp<ExpressionStatementOp>(Local);
+  if (not Statement)
+    return false;
```
pfez: This only hoists the initializer if the very same statement after the `LocalVariableOp` is an assignment that initializes the variable.

It would not catch things like:
```c
int f(int a) {
  int x;
  int y;
  x = a;
  // ....
}
```

I think in practice this is fine, because we expect local variable scope tightening to run before this, and that would rewrite my snippet as:

```c
int f(int a) {
  int y;
  int x;
  x = a;
  //...
}
```

which would make hoist it properly.

But it's worth a comment here to explain it's deliberately not handled for this reason.

pfez: Also, in principle I would like our pipeline to rewrite both the following snippets (1) and (2) into snippet (3). Would that work in both cases? I don't think so. I think that if variable scope tightening produces (2) we succeed, if it produces (1) we fail.

```c
// (1)
int f() {
  int x;
  int y;
  x = y;
  // ...
}

// (2)
int f() {
  int y;
  int x;
  x = y;
  // ...
}

// (3)
int f() {
  int y;
  int x = y;
  // ...
}
```

In any case, it's fine if we only handle it opportunistically, since fixing it properly would probably require to establish an ordering between how variable are assigned to others, and we could probably even have loops.

But I wanted to spend a couple of words on this to make sure my understanding is correct.

lib/CliftTransforms/VariableInitializerHoisting.cpp
```diff
@@ -0,0 +1,92 @@
...
+    // The assignment operation is no longer needed.
+    Assignment->erase();
+
+    // The expression statement operation is no longer needed.
+    Statement->erase();
+
+    mlir::Value Value = Local;
+    for (mlir::OpOperand &Operand : Value.getUses()) {
+      if (Initializer.args_empty())
+        Initializer.addArgument(Local.getType(), Local->getLoc());
```
pfez: Why is this in the loop? It's not using `Operand` at all.

lib/CliftTransforms/VariableInitializerHoisting.cpp
```diff
@@ -0,0 +1,92 @@
...
+    Yield->setOperand(0, Assignment.getRhs());
+
+    // The assignment operation is no longer needed.
+    Assignment->erase();
+
+    // The expression statement operation is no longer needed.
+    Statement->erase();
+
+    mlir::Value Value = Local;
+    for (mlir::OpOperand &Operand : Value.getUses()) {
```
pfez: You're iterating on uses of the local variable but, at this point in the execution of the program, some uses could be inside the initializer region and that region doesn't have an argument yet.
It's better if you add the argument to the initializer region before this loop, and replace all uses of the local variable in that region with a use of the block argument still before this loop.

lib/CliftTransforms/VariableInitializerHoisting.cpp
```diff
@@ -0,0 +1,92 @@
...
+      if (Initializer.args_empty())
+        Initializer.addArgument(Local.getType(), Local->getLoc());
+
+      if (Local->isAncestor(Operand.getOwner()))
+        Operand.set(Initializer.getArgument(0));
```
pfez: Ah this is actually what this thing is doing.

Well don't do that. Use [`replaceAllUsesInRegionWith`](https://mlir.llvm.org/doxygen/namespacemlir.html#a15a5d204dd78ac563241168e6b2e40eb).

lib/CliftTransforms/VariableInitializerHoisting.cpp
```diff
@@ -0,0 +1,92 @@
...
+    }
```
pfez: Basically I think this whole loop could be replaced with.
```c++
Initializer.addArgument(Local.getType(), Local->getLoc());
replaceAllUsesInRegionWith(Local, Initializer.getArgument(0), Initializer);
```

lib/CliftTransforms/VariableInitializerHoisting.cpp
vasama-work: Ah yes. I looked for this, but couldn't find it. Must have looked in the wrong place. Thanks.

include/revng/Clift/CliftOpInterfacesBasic.td
vasama-work: > `getBlockArgumentAsValue`

`BlockArgument` is-a `Value`, so that naming seems rather misleading.

> Also, for the cases where this is useful as of now, it seems to me that we always want it to be a LocalVariableOp. No?

Yes.

> If so, I think we should aim for a name with a narrower meaning, and we should just return clift::LocalVariableOp.

We can do that if you insist. It doesn't seem particularly valuable to me though.

> Also, why is this in StatementOpInterface? I'd make this a separate interface.

There's no value in a separate interface. Only statements have interesting block arguments. Not all statements have block arguments, and for those that don't, you would never call this function. I don't think we'll ever need to distinguish operations based on whether they translate block arguments.

lib/Clift/Clift.cpp
vasama-work: The whole point of this interface function is to do different things for different operations. Otherwise we could just use `getDefiningOp()->getResult(0)`.

In the case of local variable initialiser arguments, we translate the argument to the local variable value itself.

In the case of the for loop arguments, we translate the argument to the value of the local variable defined in its initialiser region. The for loop operation itself has no results.

lib/Clift/Clift.cpp
vasama-work: Please elaborate. There's nothing to verify. Block arguments are never "passed". They come out of nowhere. They are placeholders for some operation-specific idea of a value. We define what that value means (through `translateBlockArgument`).

lib/CliftTransforms/VariableInitializerHoisting.cpp
vasama-work: The only annoying bit is that we first need to know if there are any users within that region...

lib/CliftTransforms/VariableInitializerHoisting.cpp
vasama-work: If you want, I can change hoisting to look for assignments instead and to scan backwards to see if the assigned variable is in the set of immediately preceding local variables.

lib/Clift/Clift.cpp
pfez: Right. Nevermind then. I keep to find confusing that we have block arguments that are not associated to an operand of the operations that owns the region where the block is. That's what caused my misunderstanding.

All good then.

lib/CliftTransforms/VariableInitializerHoisting.cpp
pfez: True.

I think you have two options here:
1. add a block argument, call `replaceAllUsesInRegionWith`, and then if the argument is unused remove it
2. walk through the initializer expression before moving that into the initializer region of the local variable, looking for uses of the local variable itself. if you find some, you add the block argument and replace all uses in the regions, otherwise you don't.

lib/CliftTransforms/VariableInitializerHoisting.cpp
pfez: Could work. But explain. How do you envision that would handle the following two cases?
```c
// (A)
int x;
int y;
x = something();
y = something_else();
// (B)
int w;
int z;
z = something();
w = something_else();
```

In case A we want to be able to fold both, in the second only `z`.
But perhaps handling A is a non-goal, because variable scope tightening would already rewrite it as:
```c
// (A)
int x;
x = something();
int y;
y = something_else();
```

vasama-work: Hoisting would not and should not handle either of these cases. This is the responsibility of scope tightening.

include/revng/Clift/CliftOpInterfacesBasic.td
pfez: > `BlockArgument` is-a `Value`, so that naming seems rather misleading

True.


> > If so, I think we should aim for a name with a narrower meaning, and we should just return clift::LocalVariableOp.

> We can do that if you insist. It doesn't seem particularly valuable to me though.

It seems valuable to me, because the `translateBlockArgument` name is odd, it's not clear what it does.

We could make it return a `LocalVariableOp` and call the method `getInitializedVariable`. Which would be obviously self for `LocalVariableOp`, and it would be much easier to understand for `ForOp`.

If we call it like this, I think the meaning of the interface method is much clearer. And it becomes less scary to leave it around on all statements. Bonus point: if in the future we want to support `while(int x = 100; x--)` we don't need more machinery in the dialect.

> There's no value in a separate interface. Only statements have interesting block arguments.

Not sure, I was thinking about initializers for global variables. They are not statements, but the same reasoning we do for `LocalVariableOp` applies to them I think.

vasama-work: > I was thinking about initializers for global variables.

Fair point. Also a perfect example of why we should not return `LocalVariableOp`... 😄

pfez: Fair. Then we should have a trait or an interface for operations that declare local variables. And then both local and global variables should implement this. And then `getInitializedVariable` should return that instead of `LocalVariableOp`.

vasama-work: Global variables don't declare local variables though.

lib/CliftTransforms/VariableInitializerHoisting.cpp
vasama-work: I went with solution 1.

lib/CliftTransforms/VariableInitializerHoisting.cpp
vasama-work: I changed it to accept any number of uninitialized local variables between the assigned local and the assignment.

pfez: Ok. But I see that you're moving the local variable down before the initializing assignment. This can potentially change the order of local variable in some cases. I think it's better to move the initializing assignment up right after the local variable declaration. In this way you have the guarantee that this transformation never changes the order of local variable declarations.

include/revng/Clift/CliftOpInterfacesBasic.td
pfez: Symbols. Ugh. We need to discuss this live. I'll ping you on mattermost.

lib/CliftTransforms/VariableInitializerHoisting.cpp
vasama-work: That doesn't work in this case:
```c
int x;
int y;
x = y;
```
Anyway, reordering local variables never changes semantics, so it's safer to do that.

pfez: I wasn't worried about semantics, I was thinking about surprising effects for users, like having `var_3` declared before `var_2`. But on a second thought it's a bogus concern because the variable numbers in the names get assigned after this transformation takes place. And if we're using user-defined names that's not even a concern.

Let's keep it like this.

include/revng/Clift/CliftOpInterfacesBasic.td
vasama-work: `translateBlockArgument` was renamed to `getBlockArgumentVariable`. It still returns `mlir::Value`, but it is now guaranteed that the value is non-null and is defined by a local variable operation.

include/revng/Clift/CliftOpInterfacesBasic.td
```diff
@@ -52,6 +52,16 @@ def Clift_StatementOpInterface : Clift_OpInterface<"StatementOpInterface"> {
...
+    InterfaceMethod<
+      /*desc=*/"Returns a local variable operation referred to by the "
+               "specified block argument. The specified block argument must be "
+               "defined by a region of this statement operation.",
+      "mlir::Value", "getBlockArgumentVariable", (ins "mlir::BlockArgument":$Argument),
+      /*methodBody=*/[{}],
+      /*defaultImplementation=*/ [{
+        return nullptr;
+      }]
+    >,
```
pfez, line 64: We've agreed that this should always return a `LocalVariableOp` even if the prototype returns a `mlir::Value`.

Can we add a `verify` to this interface that checks that `if(auto Var = getBlockArgumentVariable())` then `isa<LocalVariableOp>(Var->getDefiningOp())`?

lib/CliftTransforms/VariableInitializerHoisting.cpp
```diff
@@ -0,0 +1,123 @@
...
+  for (auto I = Local->getIterator(); I != Statement->getIterator(); ++I) {
+    if (I == Block->end())
+      return false;
+
+    auto IntercedingLocal = mlir::dyn_cast<LocalVariableOp>(&*I);
+    if (not IntercedingLocal or not IntercedingLocal.getInitializer().empty())
+      return false;
+
+    HasIntercedingLocals = true;
+  }
```
pfez, line 58: It just occurred to me that we should also consider and skip over `MakeLabelOp`s.

include/revng/Clift/CliftOpInterfacesBasic.td
pfez: Closing this since this is settled.

lib/Clift/Clift.cpp
pfez: Closing this since this is settled.

lib/Clift/Clift.cpp
pfez: Closing this since this is settled.

lib/CliftTransforms/VariableInitializerHoisting.cpp
vasama-work: And `RequireOp`s, and `AssignLabel`s... I'd rather not. This relies on scope tightening that already solved all other potential issues than the variable declaration ordering. So skipping over anything is pointless because it will never happen in the pipeline.

include/revng/Clift/CliftOpInterfacesBasic.td
vasama-work: `getBlockArgumentVariable()` requires an argument... This sounds kind of complicated.

PR #590, opened 2026-06-03
Inline QEMU helpers in `detect-abi` phase

andrealinux1:


CMakeLists.txt
```diff
@@ -257,9 +257,10 @@ foreach(
             revngHelperInliningAnalyses
     COMMAND
       ./bin/revng opt -mark-inline-helpers-up-to -sroa -instsimplify
-      -cpu-loop-exit -analyze-helper-arguments -loop-unroll -fix-helpers
-      -fix-helpers-architecture="${ARCH}" -dce -detect-uninlinable-helpers
-      "${LIBTCG_HELPERS}" -o "${LIBTCG_HELPERS_FULL}")
+      -cpu-loop-exit -analyze-helper-arguments -loop-unroll -unroll-count=4
```
aleclearmind: What was the ratio for this number? What did you experience, specifically?

include/revng/FunctionIsolation/LinkRequiredHelpers.h
```diff
@@ -0,0 +1,23 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "llvm/IR/Module.h"
+#include "llvm/Pass.h"
+
+/// Link the bodies of every `revng_inline`-tagged helper that's still a
+/// declaration in `M`, cloning them from the per-architecture
+/// `libtcg-helpers-to-inline-<arch>.bc` shipped with revng.
+void linkRequiredHelpers(llvm::Module &M);
```
aleclearmind: This does not need to be public

include/revng/InlineHelpers/DeleteHelperBodies.h
aleclearmind, line 1: > InlineHelpers: split into link / inline / delete

The commit message talks about implementation functions (`runInlineHelpers`, ...) instead of the passes.

lib/FunctionIsolation/LinkRequiredHelpers.cpp
```diff
@@ -0,0 +1,93 @@
...
+}
+
+char LinkRequiredHelpersPass::ID = 0;
+
+using Register = RegisterPass<LinkRequiredHelpersPass>;
+static Register
+  X("link-required-helpers", "Link Required Helpers Pass", true, true);
+
+bool LinkRequiredHelpersPass::runOnModule(llvm::Module &M) {
+  linkRequiredHelpers(M);
```
aleclearmind: Inline this call, same for the other functions implementing the core of the passes.

lib/FunctionIsolation/LinkRequiredHelpers.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+  auto ClonedHelpersModule = ::cloneFiltered(*HelpersModule, ToClone);
+  linkModules(std::move(ClonedHelpersModule), M);
+}
+
+char LinkRequiredHelpersPass::ID = 0;
+
+using Register = RegisterPass<LinkRequiredHelpersPass>;
+static Register
+  X("link-required-helpers", "Link Required Helpers Pass", true, true);
```
aleclearmind: `link-helpers-to-inline` and `LinkHelpersToInline`.

We link helpers in other parts of the code, the name would be misleading.

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -1116,6 +1129,10 @@ llvm::Error DetectABI::run(Model &Model,
   revng::pipes::CFGMap CFGs("");
   ControlFlowGraphCache FMC(CFGs);
 
+  // Link helper bodies once at the start of the analysis, to avoid relinking
+  // for every `inline-helpers` call
+  linkRequiredHelpers(Module);
```
aleclearmind: This is already done here:

```diff
    Manager.add(new CollectFunctionsFromCalleesWrapperPass());
    Manager.add(new ControlFlowGraphCachePass(CFGs));
+   Manager.add(new LinkRequiredHelpersPass());
    Manager.add(new efa::DetectABIPass());
    Manager.add(new CollectFunctionsFromUnusedAddressesWrapperPass());
```

lib/FunctionIsolation/CMakeLists.txt
```diff
@@ -2,15 +2,22 @@
...
+# Small library exposing the inline-helpers chain (link/inline/delete) both as
+# LLVM passes and as free functions.
+revng_add_analyses_library_internal(
+  revngInlineHelpers DeleteHelperBodies.cpp InlineHelpers.cpp
+  LinkRequiredHelpers.cpp PostInlineHelpersVerifyPass.cpp)
+
+llvm_map_components_to_libnames(LLVM_INLINE_HELPERS_LIBRARIES TransformUtils)
+
+target_link_libraries(revngInlineHelpers revngBasicAnalyses revngModel
+                      revngSupport ${LLVM_INLINE_HELPERS_LIBRARIES})
```
aleclearmind: No, we need to have a 1-to-1 relationship between directories (`revng/include/xxx` and `lib/xxx`) and libraries.

Why do you need this library?
If you just want to group, you can make a subdirectory (as long as you don't do `#include "../"`).

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -462,6 +465,16 @@ void DetectABI::analyzeABI() {
...
 
+  // Inline `revng_inline`-tagged helper calls into each outlined stub
+  // so the subsequent ABI dataflow analysis can observe the helpers'
+  // register reads and writes directly.
+  // Helpers whose per-call critical arguments are not LLVM constants at
+  // this point survive un-inlined and are consumed as opaque calls by
+  // the analysis.
+  Task.advance("Inline helpers into outlined stubs");
+  for (auto &[Entry, OutlinedFn] : Functions)
+    inlineHelpersInto(*OutlinedFn->Function);
```
aleclearmind: I'd prefer not to expose raw functions replicating the behavior of passes.

What prevents you from running a module pass?

If you're trying to avoid inlining helpers into `root` we could exclude it by checking the corresponding tag.

Maybe what you did ends up being reasonable, but let's discuss because I don't want these "double" functionality exposed unless there's a reason.

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -83,6 +83,14 @@ static opt<ABIOpt> ABIEnforcement("abi-enforcement-level",
                                                     "found.")),
                                   init(ABIOpt::FullABIEnforcement));
 
+static opt<std::string> DumpAfterInlineIR("detect-abi-after-inline-ir-dir",
```
aleclearmind: `--detect-abi-dump-post-helpers-inlining`?

Never create directories unless you really really really really really need to.
Just dump the whole module. A bit more data but less files around.

You can use `dumpModule`.

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -483,6 +484,17 @@ void DetectABI::analyzeABI() {
   for (auto &[Entry, OutlinedFn] : Functions)
     inlineHelpersInto(*OutlinedFn->Function);
 
+  // When `--debug-names` is set, rename each outlined stub from the
+  // metaaddress-based identifier to its source-symbol
+  // name as recorded in the model.
+  model::CNameBuilder NameBuilder = *Binary;
```
aleclearmind: 1. `model::CNameBuilder NameBuilder(*Binary)`
2. `// TODO: we should probably not be using CNameBuilder here`
3. Does `DebugNames` uses `CNameBuilder` as well elsewhere?

share/revng/test/configuration/revng/detect-abi-after-inline.yml
```diff
@@ -0,0 +1,23 @@
...
+commands:
+  - type: revng.floating-point-detect-abi-after-inline
+    from:
+      - type: revng-qa.compiled-with-debug-info
+        filter: for-floating-point and !i386
+    suffix: /
+    command: |-
+      revng2 quick analyze --analyses=parse-binary,detect-abi "$INPUT" -- --debug-names --detect-abi-after-inline-ir-dir="$OUTPUT" > /dev/null ;
+      PREFIXES=CHECK ;
+      file "$INPUT" | grep -q "ELF 64-bit" && PREFIXES=$$PREFIXES,CHECK-64 ;
```
aleclearmind: Do not rely on `file`, it's not a dependency we have.

Easiest thing you can do is check `$QEMU_NAME`:

```
revng-qa $ git grep QEMU_
share/revng/test/configuration/revng-qa/architectures.yml:      QEMU_NAME: arm
share/revng/test/configuration/revng-qa/architectures.yml:      QEMU_NAME: x86_64
share/revng/test/configuration/revng-qa/architectures.yml:      QEMU_NAME: mips
share/revng/test/configuration/revng-qa/architectures.yml:      QEMU_NAME: mipsel
share/revng/test/configuration/revng-qa/architectures.yml:      QEMU_NAME: i386
share/revng/test/configuration/revng-qa/architectures.yml:      QEMU_NAME: s390x
share/revng/test/configuration/revng-qa/architectures.yml:      QEMU_NAME: aarch64
```

share/revng/test/configuration/revng/detect-abi-after-inline.yml
```diff
@@ -0,0 +1,23 @@
...
+# architecture.
+# We pass the `--debug-names` flag to easily anchor the `CHECK` directives for
+# multiple function names.
+# The CHECK-64 prefix is used for 64-bit arch only tests.
+
+commands:
+  - type: revng.floating-point-detect-abi-after-inline
+    from:
+      - type: revng-qa.compiled-with-debug-info
+        filter: for-floating-point and !i386
```
aleclearmind: `!i386`?

share/revng/test/configuration/revng/detect-abi-after-inline.yml
```diff
@@ -0,0 +1,23 @@
...
+    suffix: /
+    command: |-
+      revng2 quick analyze --analyses=parse-binary,detect-abi "$INPUT" -- --debug-names --detect-abi-after-inline-ir-dir="$OUTPUT" > /dev/null ;
```
aleclearmind: Split on multiple lines for readability

share/revng/test/configuration/revng/detect-abi-after-inline.yml
aleclearmind, line 1: Merge this file with the pre-existing one and give it a name related to inlining helpers.
They test the same thing, more or less, just in different place.

It also makes it more obvious that we're using the same ground truth.

CMakeLists.txt
andrealinux1: - I basically found the minimum unroll loop value which enabled us to correctly unroll all the for loops we need to correctly optimize the `x86_64` helpers (I'll put a refresher down here):
```
void helper_addsd(CPUX86State *env, Reg *d, Reg *v, Reg *s) {
    d->ZMM_D(0) = float64_add(v->ZMM_D(0), s->ZMM_D(0), &env->sse_status);
    for (i = 1; i < 2; i++) {
        d->ZMM_Q(i) = v->ZMM_Q(i);  // copy upper 64 bits unchanged
    }
}
```
- The idea was to save space for the `.bc` helpers files (and build time).
- However, I ran the numbers and the saving wrt just leaving the default is minimal (207.0 MB with unroll=4 vs 211.4 MB with unroll=default), so we could drop it, or even better:
- We could unroll only in `x86` archs, the other archs do not need the trick above, and that could save us space/build time.
- If you are open to this, I can run the numbers and tell you more about the savings, otherwise I'll just drop this commit and use the default.

aleclearmind: > We could unroll only in x86 archs, the other archs do not need the trick above, and that could save us space/build time.

Let's do this and clearly articulate why we do it in a comment.

share/revng/test/configuration/revng/detect-abi-after-inline.yml
andrealinux1: We said we don't want to focus on `i386` until we work on `x87`, the same is done in the floating point tests [here](https://github.com/revng/revng/blob/2bc640a5c26474fa94679881d432ac1813d7ea3f/share/revng/test/configuration/revng/floating-point.yml#L19)

CMakeLists.txt
andrealinux1: ACK

include/revng/InlineHelpers/DeleteHelperBodies.h
andrealinux1: Commit `[InlineHelpers: split into link / inline / delete]` now mention the flags for invoking the passes.

lib/FunctionIsolation/LinkRequiredHelpers.cpp
andrealinux1: Done

share/revng/test/configuration/revng/detect-abi-after-inline.yml
aleclearmind: OK, add a comment

lib/EarlyFunctionAnalysis/DetectABI.cpp
andrealinux1: Are you sure? The code extract you pasted is in the `run` method of `DetectABIAnalysis` which IIUC is used by the legacy pipeline.

The code you commented instead is in the `DetectABI::run` method used in the new pypeline.

- Indeed, if you look around, you see the same "passes" replicated, e.g. `    Manager.add(new CollectFunctionsFromCalleesWrapperPass());` and `collectFunctionsFromCallees(Module, GCBI, Binary);` etc etc...

- If you look at the commit history, you can see that c1217dcfc introduces the new things for pypeline including this.

lib/EarlyFunctionAnalysis/DetectABI.cpp
andrealinux1: Mmm, there are a couple of problems with the approach you propose.

1) Running the module pass here means instantiating a function pass manager inside `DetectABI`, since we need to perform the inlining once we outlined all the interesting stubs. Are you suggesting that?

2) The second problem is that now, the `inline-helpers` pass inlines the helpers only in the `Isolated` functions. Here instead, we need to inline into these so called outlined stubs. If the only way to call the inlining is to call the module pass, how can we control this behavior? An argument on the pass which either inlines only in the isolated, in all except `root`? Seems very complicated to me.

CMakeLists.txt
andrealinux1: Ok, in `[InlineHelpers: limit `loop-unroll` to x86]` I added both in the commit message and as a comment the explanation of this.

Unrolling only for x86 saves us ~58 MB on the final size of the helpers, on a total of 364 MB (~14%). Not much, but better than nothing.

lib/FunctionIsolation/CMakeLists.txt
andrealinux1: Eh.
1) The point is that now `DetectABI`, which is in `revngEarlyFunctionAnalysis`, needs to inline. And this means that if the `inline` stuff lives inside `revngFuntionIsolation`, we have a dependency there `revngEarlyFunctionAnalysis` -> `revngFunctionIsolation` (`->` means _depends from_.
2) However, `revngFunctionIsolation` -> `revngEarlyFunctionAnalysis`, because function isolation uses the results of `EFA`.
3) We have a cycle, so we need to split the `inline` stuff in a separate library.
4) In `[DetectABI: inline helpers]` I added the new `lib/InlineHelpers` dir for this.
5) Please note that the split of `inline` into 3 components is done in the previous `[InlineHelpers: split into link / inline / delete]` commit, but I kept the move to the new dir in the commit introducing the dependency in `DetectABI` to better represent the fact that we need to move it once we introduce this new dependency. Let me know if you don't like this and want the new dir already with the split commit.

aleclearmind: When you answer, first state what you did. Your answer is confusing if you don't tell me that you did move `revngInlineHelpers` in its own directory. Don't assume I have context and diff what change since the last iteration.

If you need to break the dependency cycle, the current solution seems fine.

lib/EarlyFunctionAnalysis/DetectABI.cpp
aleclearmind: > is used by the legacy pipeline

Ah OK.

lib/EarlyFunctionAnalysis/DetectABI.cpp
aleclearmind: 1. Yes. I see no problems there.
2. A feature is *either* a pass *or* a bunch of helper functions. It should to be a pass if it makes sense to run it from `opt`, which seems to be the case here (agree?).

   We do not offer a feature as a pass and a helper function, never. It's a hack.

   If we agree this is a pass, you need to be able to articulate what it does. To me, this pass inlines helper functions into isolated functions, which you can easily spot via the `FunctionTags::Isolated` tag.

   Alternatively, you can state that this pass inlines helpers in all functions *except* those marked with the `FunctionTags::Root` tag. IIRC functions produced by `detect-abi` are not marked with `FunctionTags::Isolated`, so probably this route is easier, but you need to check if the cloned root has `FunctionTags::Root`.

   Figure out something along these lines.

CMakeLists.txt
```diff
@@ -246,20 +246,33 @@ foreach(
...
   # * `fix-helpers` rewrites helper bodies to access CSVs via the env pointer,
   #   switching on env-relative offsets.
   # * `detect-uninlinable-helpers` excludes not viable helpers from inlining.
+
+  # Only the x86 softfloat helpers iterate over the vector-register lanes with
+  # bounded `for` loops (e.g. the packed `helper_*p[sd]` operate on
+  # `XMM/ZMM_D(i)`). `fix-helpers` can only rewrite the per-lane CSV accesses
+  # once those loops are unrolled. This step is not needed on other archs, so we
+  # avoid it to save space on the generated libtcg helpers and build time.
+  set(LOOP_UNROLL_PASSES "")
```
aleclearmind: `EXTRA_HELPER_PASSES`, we might need more customization in the future.

share/revng/test/configuration/revng/detect-abi-after-inline.yml
andrealinux1: Done

lib/FunctionIsolation/CMakeLists.txt
andrealinux1: Done

lib/EarlyFunctionAnalysis/DetectABI.cpp
andrealinux1: Done

share/revng/test/configuration/revng/detect-abi-after-inline.yml
andrealinux1: Done

share/revng/test/configuration/revng/detect-abi-after-inline.yml
andrealinux1: I did this in `[tests: filecheck IR after detect-abi inlining]`.

I don't super like the new naming, the `floating point` attribute kinda get lost, and wha's strange is that now in `revng-qa` we have the `for-floating-point` target which builds the binary we use for these tests, which are now called `inline-helpers`. Should we also change the correspondent part in `revng-qa`?

lib/EarlyFunctionAnalysis/DetectABI.cpp
andrealinux1: Done.

lib/EarlyFunctionAnalysis/DetectABI.cpp
andrealinux1: In `[DetectABI: name outlined stubs after symbols]` I took the liberty of dropping the `CNameBuilder` useges (even the pre-existing one), in favor of just using `llvmName()` as customary in other parts of the pipeline.

CMakeLists.txt
andrealinux1: Done.

share/revng/test/configuration/revng/detect-abi-after-inline.yml
andrealinux1: I used `$POINTER_SIZE` defined in revng-qa in `architectures.yml` ok?

lib/EarlyFunctionAnalysis/DetectABI.cpp
Cvelth: Do the names reach any user-facing artifacts? If so, they should probably use name builder. If not - who cares :)

share/revng/test/configuration/revng/detect-abi-after-inline.yml
aleclearmind: Works!

lib/EarlyFunctionAnalysis/DetectABI.cpp
aleclearmind: Just debugging/testing.

share/revng/test/configuration/revng/inline-helpers.yml
```diff
@@ -0,0 +1,48 @@
...
+      revng2 -C "$$PROJECT" project init "$INPUT" -- --debug-names ;
+      revng2 -C "$$PROJECT" project artifact enforce-abi \
+        --format yaml -o "$$PROJECT/out.yml" -- --debug-names ;
+      PREFIXES=CHECK ;
+      [ "$POINTER_SIZE" = "64" ] && PREFIXES=$$PREFIXES,CHECK-64 ;
+      yq -r 'to_entries[].value' "$$PROJECT/out.yml"
+        | while IFS= read -r b64 ; do
+            printf '%s' "$$b64" | base64 -d | revng opt -S /dev/stdin -o - ;
+          done
+        | sed 's,@local_,@,g'
```
aleclearmind: Never make these hacks. We need a tool, ask @mrjackv.

share/revng/test/configuration/revng/detect-abi-after-inline.yml
aleclearmind: Here we're testing inlining floating pointer helpers.

Let's make everything coherent:

* `for-floating-point-helpers-inlining`
* `revng.test-floating-point-helpers-inlining-in-enforce-abi`
* `revng.test-floating-point-helpers-inlining-in-detect-abi`

share/revng/test/configuration/revng/inline-helpers.yml
```diff
@@ -0,0 +1,48 @@
...
+#
+
+# End-to-end tests that verify the emergence of the softfloat calls, on the same
+# `for-floating-point` ground truth, at two points of the pipeline:
+#   * `revng.revng.inline-helpers`: in the `enforce-abi` artifact;
+#   * `revng.inline-helpers-in-detect-abi`: right after helper inlining inside
+#     `DetectABI::analyzeABI()`.
+# We pass the `--debug-names` flag to easily anchor the `CHECK` directives for
+# multiple function names.
+# We also strip the `local_` prefix to have a more coherent FileCheck file.
```
aleclearmind: Move close to where you do this

share/revng/test/configuration/revng/inline-helpers.yml
mrjackv: @andrealinux1 Look at how things are done in `simplify-switch.yml` (with tar), it should make things much simpler to handle

lib/EarlyFunctionAnalysis/DetectABI.cpp
andrealinux1: IIUC the new approach is ok, tell me if not

share/revng/test/configuration/revng/inline-helpers.yml
andrealinux1: Done

share/revng/test/configuration/revng/inline-helpers.yml
andrealinux1: @mrjackv In `[tests: filecheck IR after detect-abi inlining]` I now mirror what you pointed to in `simplify-switch.yml`, let me know if it is ok for you or if it needs further changes.

mrjackv: Looks good in principle, but the previous version checked each module separately, if that's what you want no issue with it.

andrealinux1: Yeah but since we can concatenate the dumped modules is more handy in this way, it is also more similar to the test in the following block were we dump all the module where _virtually_ all the functions are concatenated in the module

share/revng/test/configuration/revng/detect-abi-after-inline.yml
andrealinux1: Ok, I think you can take another look to `[tests: filecheck IR after detect-abi inlining]` and also in the [companion revng-qa PR](https://github.com/revng/revng-qa/)

share/revng/test/configuration/revng/floating-point-helpers-inlining.yml
aleclearmind, line 1: Rename to `floating-point-helpers-inlining.yml`

andrealinux1: Done

lib/FunctionIsolation/LinkRequiredHelpers.cpp
andrealinux1: Now done in `[InlineHelpers: split into link / inline / delete]`

include/revng/FunctionIsolation/LinkRequiredHelpers.h
andrealinux1: This and all the similar _public_ functions have been dropped, we only interact through passes.

lib/EarlyFunctionAnalysis/DetectABI.cpp
andrealinux1: Ok, I took the approach you suggest last, meaning we DON'T inline into functions with the `FunctionTags::Root`, but also not into functions with `FunctionTags::Helper`.

This is to avoid we ever inline an helper into another helper, and the second helper is later inlined into an isolated function.

Given this change, we now don't have anymore _public_ exposed free function, but we just call the three phases as passes.

aleclearmind: > also not into functions with FunctionTags::Helper

1. Is this "just in case" or did you have a real problem?
   Consider that `Helper` are those that start with `helper_` IIRC, not all of the QEMU functions.
2. > DON'T inline into functions with the FunctionTags::Root

   Did you check if the root cloned by `detect-abi` has that tag?
   Overall, double check we don't inline things in root in `detect-abi`.

andrealinux1: As we agreed, I now add the `Isolated` tag in `[DetectABI: inline helpers]` to the `DetectABI` we want to inline into, and reverted the changes on `inline-helpers` so it just inlines into `Isolated` tagged functions.

andrealinux1:
Reminder that this PR has a [sister PR in revng-qa](https://github.com/revng/revng-qa/pull/36)

lib/EarlyFunctionAnalysis/DetectABI.cpp
aleclearmind: OK, if you have verified that we don't inline in root or other weird places, mark this as solved.

aleclearmind: Ping

share/revng/test/configuration/revng/inline-helpers.yml
aleclearmind: Rebase your branch on top of `mrjackv`'s `feature/merge-ir-cmd`, which has already been reviewed and will become part of your branch.
Adopt the `revng merge llvm` tool.

lib/EarlyFunctionAnalysis/DetectABI.cpp
andrealinux1: Yes this is ok, I didn't see the comments

share/revng/test/configuration/revng/inline-helpers.yml
andrealinux1: Ok, in `[tests: filecheck IR after detect-abi inlining]` we now use the new shiny `reng merge tool`

aleclearmind:
Merged in 27bd08e. Thanks.

PR #592, opened 2026-06-10
Fix handling of `import-from-c` tests various failure modes

pfez:
`import-from-c` test members could pass even when the analysis crashed: the crash trace ended up in the log, but the test still exited 0.

The root cause is in the test command of `share/revng/test/configuration/revng/import-from-c.yml`. The test harness joins the command block into a single shell line, and the diagnostic block of the error-comparison branch was missing its statement terminators: the whole block collapsed into a single `echo` whose arguments included `cat`, the file paths — and `exit 1`. The mismatch branch therefore printed its diagnostics and fell through with exit 0: the error comparison could never fail the test.

Three weaker spots additionally conspired to hide crashes:

- The exit code of `revng2 pipeline run-analysis` was only consumed as a boolean, so a crash (134 = SIGABRT, 139 = SIGSEGV, ...) was indistinguishable from a graceful analysis failure (an uncaught Python `RuntimeError`, which always exits 1).
- Success-shaped members have no `expected-error.txt`, so on failure `tail` of the missing file yielded an empty string that compared equal to the empty `RuntimeError:` payload extracted from a crash trace.
- Nothing required a `RuntimeError:` marker to be present in stderr at all.

One commit per problem:

1. Terminate the diagnostic statements so `exit 1` executes.
2. `cat` the error file instead of echoing its path.
3. Capture the exit code: only exit code 1 may enter the expected-error path; anything else dumps stderr and fails.
4. Fail when the analysis fails but the test expects success.
5. Fail with a dedicated diagnostic when stderr carries no `RuntimeError:` payload.
6. Document the new failure modes.

### DIFF
--- a/share/revng/test/configuration/revng/import-from-c.yml
+++ b/share/revng/test/configuration/revng/import-from-c.yml
@@ -60,6 +60,10 @@ commands:
   #
   # otherwise, the error is checked against `${INPUT}/expected-error.txt` with
   # its license header stripped.
+  #
+  # Any other outcome fails the test: a crash (exit code other than 0 or 1),
+  # a failure in a test that has no `expected-error.txt`, or a failure whose
+  # stderr carries no `RuntimeError:` payload.
   - type: revng.import-from-c
     from:
       - type: source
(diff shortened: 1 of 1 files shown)


aleclearmind:
Make a single commit, too many commits with long explanations to read.

share/revng/test/configuration/revng/import-from-c.yml
```diff
@@ -70,22 +74,43 @@ commands:
...
-           "${INPUT}/input-model.yml"
-           -o "${OUTPUT}/model.yml"
-           2> "${OUTPUT}/error.txt"; then
+      EXIT_CODE=0;
+      revng2 pipeline run-analysis import-from-c
+        -c "$$(echo -e "$$CONFIGURATION")"
+        "${INPUT}/input-model.yml"
+        -o "${OUTPUT}/model.yml"
+        2> "${OUTPUT}/error.txt" || EXIT_CODE=$$?;
+
```
aleclearmind, line 83: No need for `EXIT_CODE=0` + `|| EXIT_CODE=$$?`.

Just do.

```
revng2 pipeline run-analysis ...
EXIT_CODE=$$?
```

mrjackv: The entire script is run with `set -e`, the `|| EXIT_CODE=$$?` is needed otherwise the script will fail at the `revng2 pipeline run-analysis` line

aleclearmind: Right

mrjackv (approved):
LGTM, only thing I will say is that maybe this needs to be converted to a script, there's a lot going on and having `if`s with `;` is really weird to look at

pfez:
Done!

aleclearmind:
> LGTM, only thing I will say is that maybe this needs to be converted to a script, there's a lot going on and having ifs with ; is really weird to look at

The `;` problem will soon be solved with the new `revng-qa` stuff

PR #593, opened 2026-06-11
Port last tests to pypeline

mrjackv:


aleclearmind:
1. Fix commit message capitalization
2. Port this to `feature/merge-ir-cmd`. Make sure we use `revng merge llvm` in all tests where it makes sense.
3. Close this

PR #594, opened 2026-06-11
Add `revng merge llvm` command

mrjackv:


python/revng/internal/cli/_commands/merge-llvm-ir.py
```diff
@@ -0,0 +1,99 @@
...
+        return dict(TarDictionary(input_))
+
+    # No good format found, assume it's a single plain file
+    input_.seek(0)
+    return {"input": input_.read()}
+
+
+class MergeLLVMIRCommand(Command):
+    def __init__(self):
+        super().__init__(("merge-llvm-ir",), "Merge multiple LLVM IR files into a single one")
```
aleclearmind: Either `revng merge llvm` or, if we want to be smart, we can have a single `revng merge` command that will handle MLIR as well.

python/revng/internal/cli/_commands/merge-llvm-ir.py
```diff
@@ -0,0 +1,99 @@
...
+            return self.handle_dict_input(data, options)
+        elif len(args.inputs) == 1:
+            with open(args.inputs[0], "rb") as f:
+                data = read_single_file(f)
+                return self.handle_dict_input(data, options)
+        else:
+            return run(build_command_with_loads("merge-llvm-ir", args.inputs, options), options)
+
+    def handle_dict_input(self, data: Mapping[str, bytes], options: Options):
+        process = popen(
```
aleclearmind: I think we have a more appropriate way to run commands than `popen`.

tools/merge-llvm-ir/Main.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+      if (OutputModule != nullptr)
+        linkFunctionModules(std::move(Module), OutputModule);
+      else
+        OutputModule = std::move(Module);
+    }
+  }
+
+  revng_assert(OutputModule != nullptr);
+  revng::forceVerify(OutputModule.get());
```
aleclearmind: Never use `forceVerify` directly, it's for debugging purposes only.

tools/merge-llvm-ir/Main.cpp
aleclearmind, line 1: What does this program do? Comment.

tools/merge-llvm-ir/Main.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+static llvm::ExitOnError AbortOnError;
+
+int main(int argc, char *argv[]) {
+  revng::InitRevng X(argc, argv, "", { &MainCategory });
+
+  llvm::LLVMContext Context;
+  std::unique_ptr<llvm::Module> OutputModule;
+
+  if (options::TarInput) {
```
aleclearmind: I'm not too much of a fan of these two-layered (python + C++) smarties in handling various inputs.

It's not straightforward to articulate what one layers does and how it's combined with the next one.

You could create a socket for each tar entry and feed them as multiple files, so this tool does not need to know about tars.

Alternatively, this tool could know about tars *and* YAML and we can drop the Python part.

mrjackv: > You could create a socket for each tar entry and feed them as multiple files

Yes, but then you would need to encode in some way when one files finishes and the other one begins, ... so you would end up with poor man's tar

> Alternatively, this tool could know about tars and YAML and we can drop the Python part.

Couple of considerations:
* We still need python to do `build_command_with_loads`
* Doing things in just C++ is not feasible, you cannot reliably determine if an input file is tar or yaml without inevitable getting an unrecoverable assertion

python/revng/internal/cli/_commands/merge-llvm-ir.py
mrjackv: While it might be possible for the binary formats, the text-only ones do not have a magic number that can be used. Unless you meant to have a `--llvm` and `--mlir` switch.

python/revng/internal/cli/_commands/merge-llvm-ir.py
mrjackv: ~~It needs to feed the tar via stdin, it cannot work with the other wrappers (e.g. `run`)~~
It's still needed because `stdout` needs to be redirected, which `run` does not do

python/revng/internal/cli/_commands/merge-llvm-ir.py
mrjackv: Went with `revng merge llvm`

tools/merge-llvm-ir/Main.cpp
mrjackv: Must have been from me debugging, switched to `revng::verify`

tools/merge-llvm-ir/Main.cpp
mrjackv: As this C++ program has been heavily simplified, now the body is very small and trivial to read

tools/merge-llvm-ir/Main.cpp
mrjackv: As discussed IRL, the C++ part has been simplified and the python part does all the heavy lifting

python/revng/internal/cli/_commands/merge-llvm-ir.py
aleclearmind: As a policy, we *never* use `popen` directly.

We *always* use an helper with a well-defined semantics that *does not* allow you to specify arbitrary kwargs for popen.

In fact `revng.internal.cli.support.popen` should be `_popen`.

Define a new helper with a clear semantic.

mrjackv: Done

aleclearmind:
This will likely get merged by #590.

mrjackv:
> This will likely get merged by #590.

Should I close this PR?

aleclearmind:
No, let's wait for the merge to take place so we don't forget about this.

aleclearmind:
Subsumed by #590.

PR #595, opened 2026-06-17
Improve handling of 32-bit architectures

aleclearmind:


include/revng/Support/IRBuilder.h
```diff
@@ -120,6 +142,47 @@ class IRBuilder : public detail::LLVMBuilderBase {
...
+    if (T->isPointerTy()) {
+      auto Size = DL.getPointerSizeInBits(T->getPointerAddressSpace());
+      return CreateAlloca(this->getIntNTy(Size));
+    }
+
+    if (T->isIntegerTy())
+      return CreateAlloca(T);
+
+    auto Size = DL.getTypeStoreSize(T).getFixedValue();
+    return CreateAlloca(ArrayType::get(this->getInt8Ty(), Size));
```
pfez, line 155: Why not `CreateAlloca(this->getInt8Ty(), Size)`?

include/revng/Support/IRBuilder.h
```diff
@@ -4,11 +4,26 @@
...
 #include "llvm/IR/IRBuilder.h"
+#include "llvm/IR/Instructions.h"
+#include "llvm/IR/Module.h"
 
+#include "revng/Support/Assert.h"
 #include "revng/Support/Debug.h"
 
+inline llvm::Type *getVariableType(const llvm::Value *Variable) {
+  if (auto *Alloca = llvm::dyn_cast<llvm::AllocaInst>(Variable))
+    return Alloca->getAllocatedType();
```
pfez: You should also check for `Alloca->getArraySize` and wrap the allocated type into an array if the alloca is allocating more than 1 element with the given allocated type.

lib/Canonicalize/SwitchToStatements.cpp
```diff
@@ -78,7 +78,7 @@ template<bool IsLegacy>
 using AssignType = std::conditional_t<IsLegacy, CallInst, StoreInst>;
 
 template<bool IsLegacy>
-using CopyType = std::conditional_t<IsLegacy, CallInst, LoadInst>;
```
pfez, line 81: What? This shouldn't be necessary.

include/revng/Support/IRBuilder.h
aleclearmind: I'd avoid using the second `alloca` argument, now that I know it doesn't have to be constant.

include/revng/Support/IRBuilder.h
aleclearmind: I think we should stop using the second argument and assert we don't.
Dynamic `alloca` are unnecessarily evil.

pfez: We do have assertions e.g. in the Clifter that the second argument is either missing or constant. But if you want to make this stricter it's fine.

lib/Canonicalize/SwitchToStatements.cpp
aleclearmind: It isn't indeed.

aleclearmind: Fixed.

include/revng/Support/IRBuilder.h
aleclearmind: I banned them in `Ban dynamic-size allocas, use ArrayType instead`

pfez: Works!

aleclearmind:
Merged in c2ce667.

PR #596, opened 2026-06-22
Clift type refinement

vasama-work:


include/revng/Clift/Clift.td
```diff
@@ -732,9 +732,7 @@ def Clift_UndefOp : Clift_ExpressionOp<"undef"> {
   }];
 }
 
-def Clift_YieldOp : Clift_Op<"yield", [Clift_ExpressionOpInterface,
-                                       Terminator,
-                                       NoRegionArguments]> {
```
pfez, line 756: Why? This breaks the invariant that expression regions only contain operations that implement `ExpressionOpInterface`. Unless you have strong reasons for this I think this should be reverted.

lib/Clift/Clift.cpp
```diff
@@ -1527,7 +1527,9 @@ mlir::LogicalResult AccessOp::verify() {
 //===----------------------------- SubscriptOp ----------------------------===//
 
 mlir::LogicalResult SubscriptOp::verify() {
-  auto PointerT = mlir::dyn_cast<PointerType>(getPointer().getType());
+  auto PointerT =
+    clift::unwrapped_dyn_cast<PointerType>(getPointer().getType());
+
```
pfez: The original stance was that a bitcast should be required, and then that it's silenced by implicit cast elision. What changed? If the problem is simply that implicit cast elision doesn't handle this explicitly, please do add support for this in implicit cast elision.

lib/Clift/Clift.cpp
```diff
@@ -1322,17 +1322,17 @@ static void printCliftPointerArithmeticOpTypes(mlir::OpAsmPrinter &Printer,
...
 static mlir::LogicalResult verifyPointerArithmeticOp(mlir::Operation *Op) {
-  mlir::Type LhsT = unwrapTypedefs(Op->getOperand(0).getType());
-  mlir::Type RhsT = unwrapTypedefs(Op->getOperand(1).getType());
+  mlir::Type LhsT = Op->getOperand(0).getType();
+  mlir::Type RhsT = Op->getOperand(1).getType();
 
-  auto LhsPT = mlir::dyn_cast<PointerType>(LhsT);
-  auto RhsPT = mlir::dyn_cast<PointerType>(RhsT);
+  auto LhsPT = clift::unwrapped_dyn_cast<PointerType>(LhsT);
+  auto RhsPT = clift::unwrapped_dyn_cast<PointerType>(RhsT);
```
pfez, line 1323: Noice!

lib/CliftTransforms/FieldAccessReplacement.cpp
```diff
@@ -236,23 +230,10 @@ bool Replacement::replace(ExpressionOpInterface PointerToReplace,
...
-      mlir::Type FieldType = Type.getFields()[Index].getType();
-      CurrentValue = Builder.create<AccessOp>(PointerToReplaceLoc,
-                                              FieldType,
-                                              CurrentValue,
-                                              IsIndirect,
-                                              Index);
-      break;
-    }
-
-    case FieldAccessInfo::Kind::Union: {
```
pfez: Nice, this slipped through some refactoring.

lib/CliftTransforms/EmitFieldAccesses/BestTraversal.cpp
```diff
@@ -745,7 +745,7 @@ BestTraversalChooser::computeBestTraversal(ExpressionOpInterface
   // In case we end up with a `BestTraversal` which does not actually traverse
   // any `struct` field or `array` element, we avoid the rewriting altogether,
   // and we leave the explicit pointer arithmetic access in `clift`
-  if (BestTraversal->depth() == 0) {
```
pfez: This looks pretty suspicious. Can you describe better the problem and set up a reproducer for me? I pretty strongly against merging with this change.

pfez: This looks pretty suspicious. Can you describe better the problem and set up a reproducer for me? I pretty strongly against merging with this change.

lib/CliftTransforms/EmitFieldAccesses/FieldAccessReplacement.cpp
```diff
@@ -384,20 +384,20 @@ bool Replacement::replace(ExpressionOpInterface PointerToReplace,
...
-  }
+  auto AsType = [&](mlir::Type RequestedType) -> mlir::Value {
+    mlir::Value Value = CurrentValue;
+    if (Value.getType() != RequestedType) {
+      Value = Builder.create<BitCastOp>(PointerToReplaceLoc,
+                                        RequestedType,
+                                        Value);
+    }
+    return Value;
+  };
```
pfez: Nice, but don't capture `&`. Capture `PointerToReplaceLoc` if you want, but `RequestedType` and `CurrentValue` should be arguments.

So in call sites you read: `AsType(CurrentValue, RequestedType)`.

lib/CliftTransforms/EmitFieldAccesses/FieldAccessReplacement.cpp
```diff
@@ -304,42 +310,28 @@ bool Replacement::replace(ExpressionOpInterface PointerToReplace,
...
-      // constructed above
-      CurrentValue = Builder.create<SubscriptOp>(PointerToReplaceLoc,
-                                                 CurrentValue,
-                                                 IndexValue);
+      // Finally, we emit a pointer addition of `CurrentValue` and the
+      // `IndexValue` created above.
+      CurrentValue = Builder.create<PtrAddOp>(PointerToReplaceLoc,
+                                              CurrentValue.getType(),
+                                              CurrentValue,
+                                              IndexValue);
```
pfez: Ah nice, you went the extra mile and did the `PtrAdd` thing too.

I believe this doesn't require updates to tests because we already run `-optimize-expressions` already in those tests. Ok.

lib/CliftTransforms/EmitFieldAccesses/FieldAccessReplacement.cpp
```diff
@@ -273,50 +277,44 @@ bool Replacement::replace(ExpressionOpInterface PointerToReplace,
       // We declare all the possible components (constant and variable parts)
       // as uninitialized here, and later fill only the components that we
       // need to emit.
-      mlir::Value FixedIndexValue = {};
-      mlir::Value DynamicIndexValue = {};
       mlir::Value IndexValue = {};
 
       // If present, we emit a new `mlir::Value` representing the constant
       // component of the `Index` access. If we do not have a `DynamicIndex`
       // component, we still emit a `imm 0` to represent the access to `[0]`.
```
pfez: I think this commit is stale. Update it and explain also why we special case the case for the constant index 0.

tools/clift-opt/Main.cpp
```diff
@@ -16,13 +15,42 @@
...
+
 static constexpr char ToolName[] = "Standalone optimizer driver\n";
 
 static void initializeCliftDialect(mlir::MLIRContext *Context,
                                    clift::CliftDialect *Dialect) {
   Dialect->setDefaultDataModel(CDataModel::getDefaultDataModel(8));
 }
 
+struct EmitHashPass : mlir::PassWrapper<EmitHashPass,
+                                        mlir::OperationPass<mlir::ModuleOp>> {
```
pfez: This is dead as of now. No test uses it.

The reason it was brought up was for testing that subsequent runs at fixed point of EFA and optimize expression did not produce IRs that were logically equivalent while being physically equivalent.

That is still something we want to check, but AFAICT is not tested yet. We need at least a test that uses this, otherwise leave this pass out in a separate feature branch we can revive if somebody needs this for debugging.

lib/CliftTransforms/TypeRefinementWithEFA.cpp
```diff
@@ -0,0 +1,113 @@
...
+      //P(NewFingerPrint);
+      if (NewFingerPrint == FingerPrint)
+        return mlir::success();
+
+      FingerPrint = NewFingerPrint;
+    }
+
+    //revng_abort("Type refinement with EFA max iterations reached.");
+    return mlir::failure();
+  }
```
pfez: Using the fingerprints here is expensive. I think we can do much better.
I think we can turn all of this in a cooperative worklist-based fixed point, where the two parts (`applyPatternsGreeditly` and `emitFieldAccesses`) ever needs to rescan the whole IR, and each one after the first iterations starts only from the list of generated changes by the other one.

The ingredients we need for this are.
0. Don't use a pass manager to run `emitFieldAccess`, invoke `emitFieldAccesses` directly.
1. For `emitFieldAccesses` you also need a new outparam, a `SmallVector` of the new operations it emitted, including their users and their operands (that may not be new). This can be used to feed the next iteration of `applyPatternsGreedily`.
2. We also need a way for `applyPatternsGreedily` to tell the next run of EFA what new ops it generated, including their users and operands. I believe you can use `GreedyRewriterConfig.setListener()` for this, so that every new operation generated (or modified) by the pattern rewriter gets registered in there. `FrontierCollector` can help, but I'm not sure it's exactly what you need please investigate and let me know what you find out.
3. We need a way for initializing `emitFieldAccesses` with the new ops collected by the last run of `applyPatternsGreedily`. It might involve some boilerplate, but I'm sure we can do this. I leave this up to you on the details.
4. Finally `emitFieldAccesses` is "stateful" in the sense that it has functions-wise data structure, the `TraversalInfoMap` that is costly to compute and are always guaranteed to never be invalidated by TR. This needs to be reafactored so that it's lifetime spans across all the fixed point iteration.

I still like having both `type-refinement` and `emit-field-accesses` individually as passes, because it's nice for unit testing them separately, but here we need to be more aggressive.

This extra performance is important, because in many cases I've seen where I expect big benefits from running EFA+TR at fixed point, the pattern will end up ping-ponging back and fort many times and doing only a small change at every ping pong, for many times in a row. If at every ping pong we compute all the hashes and re-run all the rewrite patterns on the whole function and all EFA on all functions we'll spend a lot of time for nothing.

include/revng/Clift/Clift.td
vasama-work: That invariant is not in itself useful. We're left with the invariant we already had: expression regions contain operations implementing `ExpressionOpInterface` and a single `YieldOp` terminating it. The reason for this change is that it just isn't useful to group `YieldOp` together with the expression operations. It's much more useful to group together the expression operations that actually share properties.

lib/Clift/Clift.cpp
vasama-work: I don't remember anything like this. Anyway, this is inconsistent with all other operations such as pointer arithmetic, indirection, member access, etc. We allow typedefs to be used everywhere else.

lib/CliftTransforms/TypeRefinementWithEFA.cpp
vasama-work: `GreedyRewriterConfig.setListener()` is not available in our version of LLVM.

I agree with you that the hash-based solution is not ideal, but since this is only a performance concern, perhaps we can go with this for now and figure out a feasible improved solution later? Otherwise I don't think we'll get this in this week.

lib/CliftTransforms/EmitFieldAccesses/FieldAccessReplacement.cpp
vasama-work: The relevant tests don't optimise and do need to be updated.

lib/CliftTransforms/EmitFieldAccesses/BestTraversal.cpp
vasama-work: Yeah, it's something of a temporary workaround, as the commit message indicates. I couldn't figure out a better solution yet though.

This example shows the problem:
```mlir
!generic64_t = !clift.int<generic 8>
!int64_t = !clift.int<signed 8>
!void = !clift.void
!test = !clift.func<"" as "test" : !void(!clift.ptr<8 to !int64_t>)>
module attributes {clift.module} {
  clift.func @test<!test>(%arg0: !clift.ptr<8 to !int64_t>) -> !void attributes {handle = ""} {
    clift.expr {
      %0 = clift.imm 1 : !generic64_t
      %1 = clift.ptr_add %arg0, %0 : (!clift.ptr<8 to !int64_t>,!generic64_t)
      clift.yield %1 : !clift.ptr<8 to !int64_t>
    }
  }
}
```

EFA replaces the expression with a new equivalent expression:
```mlir
!generic64_t = !clift.int<generic 8>
!int64_t = !clift.int<signed 8>
!void = !clift.void
!test = !clift.func<"" as "test" : !void(!clift.ptr<8 to !int64_t>)>
module attributes {clift.module} {
  clift.func @test<!test>(%arg0: !clift.ptr<8 to !int64_t>) -> !void attributes {handle = ""} {
    clift.expr {
      %0 = clift.imm 1 : !generic64_t
      %1 = clift.imm 1 : !generic64_t
      %2 = clift.ptr_add %arg0, %1 : (!clift.ptr<8 to !int64_t>,!generic64_t)
      %3 = clift.ptr_add %arg0, %0 : (!clift.ptr<8 to !int64_t>,!generic64_t)
      clift.yield %2 : !clift.ptr<8 to !int64_t>
    }
  }
}
```

Meaning that in this case, EFA is not physically idempotent.

include/revng/Clift/Clift.td
pfez: The invariant is useful for structure, and for ease of learning clift. Saying, expression regions only contain expressions that implement the `ExpressionOpInterface` is clear and simple. If you want to group other instructions that share other non-structural properties, use a separate interface for that. In any case, this change doesn't seem necessary for this PR and the changeset of this commit is a mixed bag: sometimes you remove code and make it more concise and clear, sometimes you add code and have to remember to match `YieldOp` explicitly. Let's leave this out.

lib/Clift/Clift.cpp
pfez: Yeah, this was probably part of the initial stance of being very strict about casts, so I wanted to double check. But since in C bitcasts back and forth from pointers to typedefs-of-pointers are not a thing, we've converged on not requiring them anywhere, which in the current design is probably for the best.

lib/CliftTransforms/EmitFieldAccesses/FieldAccessReplacement.cpp
pfez: Ah sad that it needs updating, but good that we have a test. Let's update it then.

lib/CliftTransforms/EmitFieldAccesses/FieldAccessReplacement.cpp
pfez: I meant "comment" not "commit"

lib/CliftTransforms/TypeRefinementWithEFA.cpp
pfez: Ok, but my expectations are that 1 and 4 are still very important, and to the best of my knowledge they should be actionable now.

Start by turning this thread in a `TODO` comment on the fixed-point implementation you have set up, noting down in the comment also what's the first LLVM version where we'll have `setListener`.

Then I'd be very happy if you can make 1 and 4 happen this week, provided all the rest is done.

include/revng/Clift/Clift.td
vasama-work: > The invariant is useful for structure, and for ease of learning clift. Saying, expression regions only contain expressions that implement the `ExpressionOpInterface` is clear and simple.

Sure it's simple, but in isolation it also falls somewhere between useless and actively harmful for understanding how expression regions work. To do anything useful relating to expression regions one must understand the role of `YieldOp` and how it is entirely different from the real expression operations.

> In any case, this change doesn't seem necessary for this PR and the changeset

No, it isn't necessary. It's possible there's some later assumption that is broken by the revert, but I'm not too worried about that.

> sometimes you remove code and make it more concise and clear, sometimes you add code and have to remember to match YieldOp explicitly

There's three such changes in this commit. I think the one in `ImportDescriptiveInfo.cpp` makes sense regardless of this change. After that it's one removal, one addition.

What you're not seeing here though is the dozens of uses of `ExpressionOpInterface` that required no changes, but where `YieldOp` never made sense, and where passing one would be a bug.

lib/CliftTransforms/EmitFieldAccesses/BestTraversal.cpp
pfez: Ah, I got what's the problem.

Some context first.

The part of `emitFieldAccesses` that decides what expressions are amenable for rewrite is designed to never try to rewrite field accesses that are already present. It treats them as hard barriers, because it assumes that whoever added them (likely it's previous iteration) already knew better. The reason why we did that was to avoid `emitFieldAccesses` to keep re-analyzing and re-writing things it had already taken a decision on.

Before your patches, the same hard stance was taken when finding subscript operations, but *not* for `clift.ptr_add`. This is because many instances of `ptr_add` come from LLVM's GEPs via the clifter that turns them into `ptr_add` so they *cannot* be considered strong decisions by `emitFieldAccesses`, otherwise it would never rewrite most things that we actually want it to improve.

Now, to the problem.

The fact that you've changed EFA to emit `ptr_add` instead of `subscript` breaks this in 2 ways:
1. performance-wise, because previous iterations of EFA don't have a way anymore to use subscripts to cast some decision in stone, so that following iterations don't have to rerun on them
2. idempotency-wise, because the very moment you start re-analyzing past decision it's very likely that you'll take the same decision all over again

As an aside, using subscripts as hard-stop to EFA action may leave some accuracy on the table, because in principle I think it's possible to write up some clift code where a following run on EFA after a run of TR can in principle take a different better decision. But we decided to leave this on the table on purpose, because my current understanding is that it happens very seldomly on revng-generated IR, and when it happens the results are a little worst but not ugly. And even if we wanted to support doing this properly we'd probably need to extend the dialect or (better) use  new discardable attributes. None of this is high priority.

So I think the best thing to do is to just drop this commit, and drop the commit that changes EFA to emit `ptr_add` instead of `subscript`. We still have to keep the commit with the workaround on "array accesses at offset zero" though.

include/revng/Clift/Clift.td
pfez: > What you're not seeing here though is the dozens of uses of ExpressionOpInterface that required no changes, but where YieldOp never made sense, and where passing one would be a bug.

Interesting, can I have some references?

vasama-work: Pretty much anywhere that it's used as a parameter. E.g:
https://github.com/revng/revng/blob/develop/lib/CliftTransforms/BestTraversal.h#L135
https://github.com/revng/revng/blob/develop/lib/CliftTransforms/Legalization.cpp#L281

There's also a bunch of places where `getDefiningOp<ExpressionOpInterface>()` is used, where you could never get `YieldOp`.

lib/CliftTransforms/EmitFieldAccesses/BestTraversal.cpp
vasama-work: Okay... That works for now, but only because we don't canonicalise `&p[i]` to `p + i` as we should. As soon as we implement that canonicalisation, this will break again.

lib/CliftTransforms/EmitFieldAccesses/FieldAccessReplacement.cpp
vasama-work: After reverting the `PtrAdd` change, there's a total of 16 tests that need to be updated. Looks to be mostly because address-of operations are inserted earlier.

lib/CliftTransforms/EmitFieldAccesses/BestTraversal.cpp
pfez: I don't see why you stay that we should, but even if it is true. Even if it is true, it's probably conditionally true depending on the fact that `p` is a decay from array or directly already a pointer. If it's a decay, I don't want the that to be canonicalized like it is. Which I think defuses the issue of interfering with EFA as well.

In any case let's keep subscripts in EFA for now. If it really turns out we need the canonicalization you suggest in the general case, that will spin up a design discussion that spans across EFA, canonicalization, and even TR at this point. And this is not the right moment to spin up this discussion. Perhaps add a TODO commentto EFA, in the part that emits subscripts, with these considerations. Thanks.

vasama-work: We should canonicalize somehow, and `p + i` is fewer operations than `&p[i]`. And yes, we should canonicalise expressions without considering their operands, meaning that we also canonicalise when `p` is a decay.

Now, this is canonicalisation and it only exists for the purposes of *producing a canonical form*. Canonicalisation is not beautification, and many canonical forms are arguably worse for the reader than non-canonical forms, but that's fine, because that's not the purpose of canonicalisation.

Then, at the end of the pipeline, when we're no longer concerned with maintaining a canonical form, we can rewrite `decay(a) + i` back to `&a[i]` if we like. And in many cases that's probably a better form.

include/revng/Clift/CliftOpInterfaces.h
```diff
@@ -63,6 +63,18 @@ LabelAssignmentOpInterface getLabelAssignmentOp(mlir::Value Label);
...
 
+enum class LvalueToRvalueConversion : uint8_t {
+  // The operand is not subject to l-value-to-r-value conversion.
+  No,
+
+  // The operand is subject to l-value-to-r-value conversion.
+  Yes,
+
+  // The operand is subject to l-value-to-r-value conversion if the result of
+  // it sparent operation is subject to l-value-to-r-value conversion.
```
pfez: Typo `s/it sparent/its parent`

pfez: Also, from how we use this, I don't think "parent" is what you mean. It seems more like "user" is what you mean.

vasama-work: I fixed the typo. Parent is correct. We're talking about an operand, so the parent is the operation to which that operand belongs.

pfez: Ok for the typo. But I'm a bit lost about the parent vs user argument.

`Transparent` is only used for direct `AccessOp` and from what I understand a direct `AccessOp` converts an one of its operands from lvalue to rvalue only one when the only "user" of the `AccessOp` forces it to. It's not clear what you mean by "parent" operation in this context. When reading the comment it makes you think at the operation owning the region where the operand is, which is not what you mean I guess. I think that I'm using the "user" naming here that is more LLVM-y. The proper naming in MLIR is probably "owner" operation: in MLIR terms IIUC the owner of an operand is the operation that uses the operand (hence the "user" in my initial wording). If this is the case I'd use "owner" instead of "parent" here. And clarifying in the comment that the "owner" of an MLIR operand means "user" in LLVM terms (if I haven't gotten this wrong) would really help future readers, since the specific terminology used by MLIR is not as well known as LLVM's among our devs yet.

vasama-work: That's fair. I'll change it to owner instead.

vasama-work: Oh, and ideally I would prefer not to have `Transparent` at all, and to instead split `AccessOp` into three variants with three different semantics.

pfez: I know your stance on `Transparent` but let's not split `AccessOp` for now.

lib/CliftTransforms/EmitFieldAccesses/BestTraversal.cpp
pfez: > we should canonicalize expressions without considering their operands

Fair point, in general.

> We should canonicalize somehow, and `p + i` is fewer operations than `&p[i]`.

I don't think that we must canonicalize this somehow just because it's fewer operations. We need to canonicalize it if we think that the 2 versions are 100% replaceable with each other in all the situations (except beautification which is a separate issue) so canonicalizing enables to write simpler optimizations for both cases because we can only handle the canonical one.

What I'm arguing for here is that subscripts carry somewhat stronger information, that is added only in a single point (EFA) and for that very reason it's wrong to consider them 100% replaceable with each other, and for that very same reason it's not ok to canonicalize either as the other.

The fact that subscripts carry a somewhat stronger information and are not 100% equivalent to seemingly-equivalent ptradds also matches 2 aspects of the current implementation.

* a) the fact that the only part of the clift pipeline that emits those subscripts (EFA) does it with a somewhat expensive non-local analysis
* b) the fact that we had to special case EFA out of emitting subscript zero because that would be canonicalized away. now that I think about it, I'm starting to reconsider that canonicalization shouldn't canonicalize away subscript zero, since the only cases where this happens are cases that were generated by EFA before the subscript-zero fix, and I don't expect that canonicalization to ever trigger at all. this would be more aligned to my stance I just stated that we subscript and ptradd are not 100% equivalent and one should not be canonicalized to the other. if anything we have to canonicalize `p + 0` to `p`, but that's beyond the point of this discussion.

Still all this is a broader design decision that goes beyond this PR.

I believe that for this PR we're on the same page that:
* EFA should emit subscripts, except for the subscript-zero corner case you've fixed
* we shouldn't canonicalize away subscripts (even if IIUC we currently do for subscript zero, which according to point b I think could be avoided)

If both these hold, all the rest works and I consider it good for this PR.

include/revng/Clift/Clift.td
pfez: > There's also a bunch of places where `getDefiningOp<ExpressionOpInterface>()` is used, where you could never get `YieldOp`.

That's not a big problem.

It seems to me that this all boils down to saying that `YieldOp` is the only expression operation that doesn't have a result value, which makes it an odd fellow.

In any case, this affects the dialect and it's snowballing into a design discussion. It may be worth having, but I don't want to rush it in since it doesn't solve any practical problems we're having now.

Please drop this commit from the PR, and keep it around for when you're back from vacations, when we'll deal with the fixed point and the few pieces we're leaving out for now. Please make a pad to recap the pieces we've dropped out for now and share it with me on mattermost.

lib/CliftTransforms/EmitFieldAccesses/BestTraversal.cpp
vasama-work: I agree that this is not relevant for this PR.

I do not agree with the assessment that subscripts carry extra information. I do think we should canonicalise them and that we should find a better way to fix EFA. We can discuss this later, together with the EFA/TR performance improvements that I suspect warrant some changes to EFA anyway.

include/revng/Clift/Clift.td
vasama-work: > It seems to me that this all boils down to saying that `YieldOp` is the only expression operation that doesn't have a result value, which makes it an odd fellow.

That's part of it. For example it would be useful to give `ExpressionOpInterface` a `getResult` member function and maybe an implicit `mlir::Value` conversion operator (all the concrete expression operations have this out of the box) and to stop using `E->getResult(0)` everywhere.

The other part is that `YieldOp` is kind of a weird Clift specific plumbing thing that only exists for MLIR reasons and doesn't actually represent any C expression the way the real expression operations do. It's just completely different from the real expression operations in all significant aspects.

lib/CliftTransforms/EmitFieldAccesses/FieldAccessReplacement.cpp
pfez: I see you've implemented this. Thanks.

lib/CliftTransforms/TypeRefinementWithEFA.cpp
```diff
@@ -0,0 +1,113 @@
...
+      auto H = std::bit_cast<std::array<uint8_t, 20>>(FP);
+      for (uint8_t X : H)
+        fprintf(stderr, "%02x", static_cast<unsigned>(X));
+      fprintf(stderr, "\n");
+
+      Function->walk([&](mlir::Operation *Sub) {
+        llvm::errs() << "  " << Sub << "\n";
+      });
+    };
+#endif
```
pfez: This `if 0` has to go, and the surrounding seem still very WIP and full of comments. Can you clean this up? I'm looking forward to the final form of fixed point that we can try to merge in this PR.

lib/CliftTransforms/TypeRefinement.cpp
```diff
@@ -337,6 +337,10 @@ struct VariableForwardRefinementPattern : mlir::OpRewritePattern<AssignOp> {
     if (not NewType)
       return mlir::failure();
 
+    // TODO: Implement array type refinement.
```
pfez: More than "implement" I think we still have to decide what to do. Not refining is quite sane as of now I think. Change the comment to say that we still have to decide what to do about them in type refinement please.

tools/clift-opt/Main.cpp
pfez: Is this still needed at all? According to our last discussion we can probably drop this.

vasama-work: No, I can drop it from the final one unless we want to add EFA unit tests to ensure physical idempotence.

share/revng/test/tests/analysis/EditCBody/run-edit-c-body
```diff
@@ -62,7 +62,7 @@ def insert_before(anchor, *directives):
 
 insert_before("var_0", "// RENAME: renamed_local", "// RETYPE: uint32_t")
 insert_before("label_0:", "// RENAME: renamed_goto_label")
-insert_before("var_2 = 0UL", "// A plain comment on a statement.")
+insert_before("var_2 = 0", "// A plain comment on a statement.")
```
pfez, line 65: Can you elaborate more on why is this needed and what problem it's solving?

python/revng/internal/pipeline.yml
```diff
@@ -458,10 +458,7 @@ branches:
             - loop-detection
             - optimize-statements
             - label-merging
-            - optimize-expressions
-            - emit-field-accesses
-            - canonicalize
-            - optimize-expressions
+            - refine-types-with-efa
```
pfez: We can call this `refine-types-and-accesses`

share/doc/revng/user-manual/tutorial/editing-types-in-c.md
```diff
@@ -120,7 +120,7 @@ Re-emitting `summary` makes the edit concrete: it takes an `account *` and reads
 $ revng project artifact emit-c summary | revng ptml
 _ABI(SystemV_x86_64)
 generic64_t summary(account *argument_0) {
-  return *(generic64_t *) &argument_0->balance + (int32_t) *(generic32_t *) &argument_0->id + (int32_t) *(generic32_t *) &argument_0->flags;
+  return argument_0->balance + argument_0->id + argument_0->flags;
```
pfez, line 123: Nice!

share/doc/revng/user-manual/tutorial/editing-types-in-c.md
```diff
@@ -44,7 +44,7 @@ Here is the decompiled `summary`:
 $ revng project artifact emit-c summary | revng ptml
 _ABI(SystemV_x86_64)
 generic64_t summary(struct_58 *argument_0) {
-  return argument_0->offset_8 + (int32_t) argument_0->offset_0 + (int32_t) argument_0->offset_16;
```
pfez, line 47: This is worse than the original. Why? What would be required to fix it?

share/doc/revng/user-manual/tutorial/comments-and-local-variables.md
```diff
@@ -140,7 +140,7 @@ The comment, the name and the type are now part of the model, so they show up in
 $ revng project artifact emit-c resolve | revng ptml | grep -E 'value =|mix the key'
   int64_t value = 0L;
   // only mix the key when it differs from the salt
-    value = raw(argument_0) ^ argument_0;
```
pfez, line 143: This is worse than the original. Why? What would be required to fix it?

share/revng/compile-flags-sanity.cfg
```diff
@@ -232,3 +232,6 @@
 # int32_t var_0 = 4294966128;
 # ```
 -Wno-constant-conversion
+
+# WIP: Add a better comment.
```
pfez: Please update this comment. We have agreed to keep this commit in.

lib/CliftTransforms/EmitFieldAccesses/BestTraversal.cpp
pfez: The discussion on subscripts vs ptradd was postponed. For now we'll keep emitting subscripts, with the exception of the case where all emitted subscripts are zero.

share/revng/test/tests/analysis/EditCBody/run-edit-c-body
vasama-work: Type refinement changed the type. The test assumes that the immediate always has type `unsigned long`, but now it doesn't.

share/doc/revng/user-manual/tutorial/editing-types-in-c.md
vasama-work: Unclear. I'll look into it...

lib/CliftTransforms/TypeRefinementWithEFA.cpp
pfez: The `if 0` is gone.

share/revng/test/tests/analysis/EditCBody/run-edit-c-body
pfez: This seems legit.

lib/CliftTransforms/TypeRefinement.cpp
```diff
@@ -0,0 +1,703 @@
...
+#include "revng/CliftTransforms/Passes.h.inc"
+} // namespace clift
+
+using namespace clift;
+
+//===--------------------------- Type comparison --------------------------===//
+
+namespace {
+
+template<size_t Size>
```
pfez: So `Size` is the number of bits. Let's call that `BitSize`.

lib/CliftTransforms/TypeRefinement.cpp
```diff
@@ -0,0 +1,703 @@
...
+class Bitset {
```
pfez, line 26: Any reason for now using `std::bitset`?

lib/CliftTransforms/TypeRefinement.cpp
```diff
@@ -0,0 +1,703 @@
...
+
+  Opaque,
+  Generic,
+  PointerOrNumber,
+  Pointer,
+  Number,
+  Signed,
+  Unsigned,
+  FloatingPoint,
+  Composite,
```
pfez: What's `Composite`? Please document it.

lib/CliftTransforms/TypeRefinement.cpp
```diff
@@ -0,0 +1,703 @@
...
+  if (auto U = mlir::dyn_cast<EnumType>(T))
+    return U.getUnderlyingType();
+
+  if (auto U = mlir::dyn_cast<TypedefType>(T))
+    return U.getUnderlyingType();
+
+  return nullptr;
+}
+
+/// Returns the fully unwrapped type and number of layers.
```
pfez, line 173: So a typedef counts as a layer, and an enum counts as a layer as well. Let me think if there are some consequences of this.

lib/CliftTransforms/TypeRefinement.cpp
```diff
@@ -0,0 +1,703 @@
...
+  static constexpr size_t TypeTableSize = TypeKindCount * TypeKindCount;
+
+  Bitset<TypeTableSize> Table = {};
+
+public:
+  [[nodiscard]] static bool ordered(TypeKind K1, TypeKind K2);
+
+private:
+  consteval TypeTable() {
+    // NOTE: Define more refined orderings before less refined ones.
```
pfez, line 75: Why? Is this required for semantic, or it's just a general not for future edits? Please clarify the comment.

tests/filecheck/clift/type-refinement/backward-add-refinement.mlir
```diff
@@ -0,0 +1,24 @@
...
+!f = !clift.func<"/model-type/1001" : !void(!generic32_t)>
+
+module attributes {clift.module} {
+  clift.func @f<!f>(%arg0 : !generic32_t) -> !void {
+    // CHECK: clift.expr {
+    clift.expr {
+      %0 = clift.add %arg0, %arg0 : !generic32_t
+      %1 = clift.bitcast %0 : !generic32_t -> !int32_t
+      clift.yield %1 : !int32_t
+    // CHECK: }
```
pfez, line 25: This isn't checking anything.

tests/filecheck/clift/type-refinement/backward-neg-refinement.mlir
```diff
@@ -0,0 +1,24 @@
...
+!f = !clift.func<"/model-type/1001" : !void(!generic32_t)>
+
+module attributes {clift.module} {
+  clift.func @f<!f>(%arg0 : !generic32_t) -> !void {
+    // CHECK: clift.expr {
+    clift.expr {
+      %0 = clift.neg %arg0 : !generic32_t
+      %1 = clift.bitcast %0 : !generic32_t -> !int32_t
+      clift.yield %1 : !int32_t
+    // CHECK: }
```
pfez, line 24: This isn't checking anything either.

tests/filecheck/clift/type-refinement/backward-ternary-refinement.mlir
```diff
@@ -0,0 +1,30 @@
...
+      // CHECK: %0 = clift.undef : !int32_t
+      %0 = clift.undef : !int32_t
+      // CHECK: %1 = clift.bitcast %arg0 : !generic32_t -> !int32_t
+      // CHECK: %2 = clift.bitcast %arg0 : !generic32_t -> !int32_t
+      // CHECK: %3 = clift.ternary %0, %1, %2 : (!int32_t,!int32_t)
+      %1 = clift.ternary %0, %arg0, %arg0 : (!int32_t, !generic32_t)
+      %2 = clift.bitcast %1 : !generic32_t -> !int32_t
+      // CHECK: clift.yield %3 : !int32_t
+      clift.yield %2 : !int32_t
+    // CHECK: }
```
pfez, line 27: I'm having second thoughts. This increased the number of bitcasts. What are your thought on this?

lib/CliftTransforms/TypeRefinementWithEFA.cpp
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  mlir::LogicalResult applyPatternRewrites(FunctionOp Function) {
+    mlir::GreedyRewriteConfig Config;
+    Config.maxIterations = mlir::GreedyRewriteConfig::kNoLimit;
+    return mlir::applyPatternsAndFoldGreedily(Function, Patterns, Config);
+  }
+
+  mlir::LogicalResult applyIteratively() {
+    static constexpr unsigned MaxIterations = 100;
```
pfez: 🤔 why 100?

lib/CliftTransforms/TypeRefinementWithEFA.cpp
pfez: IIRC we said that at the end we should turn EFA into a `RewritePattern` as well, and this will make this fixed point iteration basically a single invocation to `applyPatternsAndFoldGreedily`.

Can you do that in a commit at the end of the branch? If we converge on that we can then drop the hash entirely, together with this intermediate form.

lib/CliftTransforms/TypeRefinement.cpp
```diff
@@ -0,0 +1,703 @@
...
+    return mlir::failure();
+  }
+
+  mlir::LogicalResult
+  matchAndRewrite(ExpressionOpInterface Op,
+                  mlir::PatternRewriter &Rewriter) const override {
+    if (mlir::isa<NegOp, BitwiseNotOp, ShlOp, ShrOp, SarOp>(Op)) {
+      TypeConstraint Constraint = IsaConstraint<IntegralType>;
+
+      // TODO: Allow refinement and fix mismatched signedness in legalisation.
```
pfez: Is this comment obsolete?

pfez: There's a bunch of other comments all equal to this. From our discussions I remember a decision to fix this in legalization.

lib/CliftTransforms/TypeRefinement.cpp
```diff
@@ -0,0 +1,703 @@
...
+}
+
+using TypeConstraint = llvm::function_ref<bool(mlir::Type)>;
+
+static constexpr auto NoConstraint = [](mlir::Type) { return true; };
+static constexpr auto SignedConstraint = [](mlir::Type T) { return isSigned(T); };
+static constexpr auto UnsignedConstraint = [](mlir::Type T) { return isUnsigned(T); };
+
+template<typename TypeKind>
+static constexpr auto IsaConstraint = [](mlir::Type T) { return clift::unwrapped_isa<TypeKind>(T); };
```
pfez: Make some functions please.

lib/CliftTransforms/TypeRefinement.cpp
vasama-work: It's not constexpr.

lib/CliftTransforms/TypeRefinement.cpp
vasama-work: Would you prefer something like this?

```cpp
static bool noConstraint(mlir::Type) { return true; }

template<bool (&Function)(mlir::Type)>
static constexpr TypeConstraint wrap = [](mlir::Type Type) -> bool {
  return Function(Type);
};

TypeConstraint Constraint = wrap<noConstraint>;
```

tests/filecheck/clift/type-refinement/backward-ternary-refinement.mlir
vasama-work: It's a unit test devoid of any larger context. It's not really representative of anything.

lib/CliftTransforms/TypeRefinementWithEFA.cpp
vasama-work: Had to pick something.

share/revng/compile-flags-sanity.cfg
vasama-work: Done.

lib/CliftTransforms/TypeRefinement.cpp
vasama-work: `Size` is arguably the most appropriate name, following stdlib conventions, but I have changed it to `BitCount`.

lib/CliftTransforms/TypeRefinement.cpp
vasama-work: Done.
```cpp
    // NOTE: Define more refined orderings before less refined ones.
    //
    //       This is simply because of how orderTransitive is defined. If the
    //       less refined ordering is defined first, the transitive propagation
    //       of inherited relations is not applied correctly.
```

lib/CliftTransforms/TypeRefinement.cpp
vasama-work: Yes, legalisation now takes care of this. I removed the relevant code.

lib/CliftTransforms/TypeRefinement.cpp
pfez: Ah sad, that's just from C++23? Ok then. But put a comment that we can replace it with `std::bitset` when that's `constexpr`.

lib/CliftTransforms/TypeRefinement.cpp
vasama-work: I renamed it to `UserDefined`.

lib/CliftTransforms/TypeRefinementWithEFA.cpp
pfez: How wild would it be to leave this unlimited? I think it should converge nicely pretty quickly, no?

vasama-work: In case it does not converge, we get an infinite loop. Better to fail gracefully in my opinion. Anyway, I think the discussion is moot, since the plan is to convert EFA into a rewrite pattern.

lib/CliftTransforms/TypeRefinement.cpp
```diff
@@ -0,0 +1,671 @@
...
+      return mlir::failure();
+
+    mlir::Type NewType = Cast.getResult().getType();
+
+    // For now, arrays not being value types, cannot be refined.
+    if (clift::unwrapped_isa<ArrayType>(NewType))
+      return mlir::failure();
+
+    if (not (compareTypeRefinement(NewType, Local.getType()) > 0))
+      return mlir::failure();
```
pfez: This whole check is done in a number of places, varying `NewType` and `Local.getType()`. Refactor it in a free function.

lib/CliftTransforms/TypeRefinement.cpp
```diff
@@ -0,0 +1,671 @@
...
+                  mlir::PatternRewriter &Rewriter) const override {
+    auto Op = Cast.getValue().getDefiningOp<ExpressionOpInterface>();
+    if (not Op)
+      return mlir::failure();
+
+    mlir::Type NewType = Cast.getResult().getType();
+
+    // For now, arrays not being value types, cannot be refined.
+    if (clift::unwrapped_isa<ArrayType>(NewType))
+      return mlir::failure();
```
pfez, line 593: I think this can be refactored as well. If you push it down into the various `matchAndRewrite*` methods, it will reach the other check on `ArrayType` and it boils down to the same check I proposed to refactor above.

lib/CliftTransforms/TypeRefinementWithEFA.cpp
pfez: > I think the discussion is moot, since the plan is to convert EFA into a rewrite pattern.

Makes sense. I'll leave this thread open until the `RewritePattern` is ready.

lib/CliftTransforms/TypeRefinementWithEFA.cpp
vasama-work: EFA is now a pattern rewrite.

lib/CliftTransforms/TypeRefinement.cpp
vasama-work: Done.

lib/CliftTransforms/TypeRefinement.cpp
vasama-work: I deduplicated the second check using a function named `isNotMoreRefined`, but for the array thing, I don't think there's any benefit from using a separate function.

include/revng/CliftTransforms/ExpressionPostProcessing.td
```diff
@@ -0,0 +1,21 @@
...
+include "revng/Clift/Clift.td"
+include "revng/CliftTransforms/ExpressionHelpers.td"
+
+// 0xFFFFFFFF -> -1
+// Constraint: the immediate value top bit and at least one other bit is set.
+def NegativeImmediatePattern
+  : Pat<(Clift_ImmediateOp:$res $value),
+        (Clift_NegOp (Clift_ImmediateOp (makeInteger (NativeCodeCall<[{ negateIntegerValue($0.getValue()) }]> $value)),
+                                        (returnType $res))),
+        [(Constraint<CPred<[{ isNegativeImmediateValue($0.getValue()) }]>> $value)]>;
```
pfez, line 19: Ok but IIUC this comes from a bunch of quick small improvements we asked to implement. I think that this makes sense only if the modulus of the negative immediate is small: we want to se `-1` but not, say, `-12312124321`. IIRC we said up to `-64` or `-512` as a start.

tests/filecheck/clift/expression-rewrites/subscript-zero.mlir
```diff
@@ -1,66 +0,0 @@
-//
-// This file is distributed under the MIT License. See LICENSE.md for details.
-//
-
-// RUN: %root/bin/revng clift-opt %s --optimize-expressions | FileCheck %s
```
pfez, line 5: remember to fixup this commit

tests/filecheck/clift/backend/array.mlir
```diff
@@ -50,7 +50,8 @@ module attributes {clift.module} {
       %q = clift.indirection %p_array : !int32_t$1$p
       %p = clift.decay %q : !int32_t$1 -> !int32_t$p
       %i = clift.imm 0 : !int32_t
-      %comma = clift.comma %i, %i : !int32_t, !int32_t
+      %j = clift.imm 0 : !int32_t
+      %comma = clift.comma %i, %j : !int32_t, !int32_t
```
pfez, line 54: I'm confused. This commit claims to "Improve expression region verification", but then it also changes a bunch of tests.

lib/CliftTransforms/Expressions.cpp
```diff
@@ -162,6 +236,8 @@ void clift::populateWithExpressionOptimizationPatterns(mlir::RewritePatternSet
   populateWithBooleanNegationPatterns(Set);
   populateWithCastCanonicalizations(Set);
 
+  Set.add<TypePunnedReadPattern>(Set.getContext());
```
pfez, line 278: Commenting here, but I mean the whole commit. Fix the commit message. I want to keep this in.

pfez: We also decided some time ago to do the same for type punned writes. It would be great if we manage to put it in this PR as well.

lib/CliftTransforms/TypeRefinementWithEFA.cpp
```diff
@@ -0,0 +1,56 @@
...
+  mlir::LogicalResult initialize(mlir::MLIRContext *Context) override {
+    mlir::RewritePatternSet Set(Context);
+
+    populateWithEmitFieldAccessesPatterns(Set);
+    populateWithTypeRefinementPatterns(Set);
+    populateWithExpressionOptimizationPatterns(Set);
+
+    Patterns = mlir::FrozenRewritePatternSet(std::move(Set),
+                                             disabledPatterns,
+                                             enabledPatterns);
```
pfez, line 38: What are the `disabledPatterns` and the `enabledPatterns` again? Why do we need to provide them? Are we enabling or disabling anything in particular here?

lib/CliftTransforms/TypeRefinementWithEFA.cpp
```diff
@@ -0,0 +1,56 @@
...
+    return mlir::success();
+  }
+
+  void runOnOperation() override {
+    mlir::GreedyRewriteConfig Config;
+    Config.maxIterations = mlir::GreedyRewriteConfig::kNoLimit;
+    if (mlir::applyPatternsAndFoldGreedily(getOperation(), Patterns, Config)
+          .failed())
+      signalPassFailure();
+  }
```
pfez, line 51: Oh finally!

lib/CliftTransforms/EmitFieldAccesses.cpp
```diff
@@ -0,0 +1,111 @@
...
+  void initialize() { setDebugName("emit-field-accesses"); }
+
+  mlir::LogicalResult
+  matchAndRewrite(ExpressionOpInterface Op,
+                  mlir::PatternRewriter &Rewriter) const override {
+    std::optional<TraversalInfoMap> LocalTraversalCache;
+    TraversalInfoMap *TraversalCache = ThreadTraversalCache;
+
+    if (TraversalCache == nullptr)
+      TraversalCache = &LocalTraversalCache.emplace();
```
pfez, line 44: Can't we just let the constructor provide a reference to a cache? In this way it's the caller that controls the policy of the lifetime for the cache.

lib/CliftTransforms/TypeRefinement.cpp
pfez: I meant a single function that embedded both the array check and what's now `isNotMoreRefined`. Something like `cannotBeRefined`. Otherwise it's so easy to forget one or the other.

include/revng/CliftTransforms/ExpressionPostProcessing.td
vasama-work: Why would we want `(int) 2166572391u` over `-2128394905`?

Certainly there's an argument to be made that we should only apply this rewrite for signed immediates.

lib/CliftTransforms/TypeRefinementWithEFA.cpp
vasama-work: These are for controlling the pass from the command line. Useful for tests and debugging.

tests/filecheck/clift/backend/array.mlir
vasama-work: This is that old thing about expression regions forming a tree, not a DAG. It was just never verified until now. Some of the tests violated that rule, so they had to be fixed.

lib/CliftTransforms/EmitFieldAccesses.cpp
vasama-work: This doesn't work when the pattern set is built on pass construction. There's other alternatives for caching if you're not keen on this. This is just the simplest solution.

lib/CliftTransforms/TypeRefinementWithEFA.cpp
pfez: I get it. But in practice, we aren't using them now, no?

lib/CliftTransforms/EmitFieldAccesses.cpp
pfez: What are the the other alternatives you have in mind?

include/revng/CliftTransforms/ExpressionPostProcessing.td
pfez: There's more. For example, for operands of bitwise operators you never want `-something`. And you always want to print them in hexadecimal. All of this makes `NegativeImmediatePattern` less trivial than it seems at first sight. Let's take that out of this PR for now. I don't think it's needed by anything else, is it?

lib/CliftTransforms/TypeRefinement.cpp
pfez: `SignedConstraint` and `UnsignedConstaint` aren't used, so just drop them.

For the rest, just this works.
```c++
static constexpr bool noConstraint(mlir::Type) {
  return true;
};

template<typename TypeKind>
static constexpr bool isaConstraint(mlir::Type T) {
  return clift::unwrapped_isa<TypeKind>(T);
};
````

Comment (vasama-work) [reply lib/CliftTransforms/TypeRefinement.cpp]: They need to be objects, not functions.

File: lib/CliftTransforms/TypeRefinementWithEFA.cpp (hunk 37 above)
Comment (vasama-work) [reply lib/CliftTransforms/TypeRefinementWithEFA.cpp:38]: For this particular pass, no.

Somewhat related, perhaps instead of a separate pass, we should just include the type refinement rewrites in the optimize-expressions pass.

File: lib/CliftTransforms/EmitFieldAccesses.cpp (hunk 39 above)
Comment (vasama-work) [reply lib/CliftTransforms/EmitFieldAccesses.cpp:44]: One option would be to create some attribute to store the cache and to attach it to the function.

Another option would be to abuse attributes and the existing thread safe attribute uniqueing mechanism to do context-wide caching.

File: include/revng/CliftTransforms/ExpressionPostProcessing.td (hunk 33 above)
Comment (vasama-work) [reply include/revng/CliftTransforms/ExpressionPostProcessing.td:19]: No, this isn't needed by anything else.

> And you always want to print them in hexadecimal.

I don't agree. `x & 1` is better than `x & 0x1`.

Anyway, bitwise operators is an entirely separate issue and is likely best solved through some other means. The ratio of negative immediates used in bitwise operators vs others is likely low. Don't let perfect be the enemy of good. This just restores the previous status quo of negative immediates after the other PR removed the emission of negative immediates from the backend. We can think about how best to solve the bitwise operator operand issue later.

That said, by all means we can constrain this to signed immediates.

Comment (vasama-work) [reply include/revng/CliftTransforms/ExpressionPostProcessing.td:19]: > Ok but IIUC this comes from a bunch of quick small improvements we asked to implement.

Not exactly. This change was a long time coming, and was also helpful for the arbitrary precision immediates needed by REP 79. As noted above, this rewrite restores the previous status quo.

File: lib/CliftTransforms/TypeRefinement.cpp (hunk 30 above)
Comment (pfez) [reply lib/CliftTransforms/TypeRefinement.cpp]: I don't understand: the following works and pass tests.

```diff
diff --git a/lib/CliftTransforms/TypeRefinement.cpp b/lib/CliftTransforms/TypeRefinement.cpp
index 2562a5ad2..9d2de9f2a 100644
--- a/lib/CliftTransforms/TypeRefinement.cpp
+++ b/lib/CliftTransforms/TypeRefinement.cpp
@@ -330,17 +330,14 @@ static bool isNotMoreRefined(mlir::Type LHS, mlir::Type RHS) {

 using TypeConstraint = llvm::function_ref<bool(mlir::Type)>;

-static constexpr auto NoConstraint = [](mlir::Type) { return true; };
-static constexpr auto SignedConstraint = [](mlir::Type T) {
-  return isSigned(T);
-};
-static constexpr auto UnsignedConstraint = [](mlir::Type T) {
-  return isUnsigned(T);
+static constexpr bool noConstraint(mlir::Type) {
+  return true;
 };

 template<typename TypeKind>
-static constexpr auto IsaConstraint =
-  [](mlir::Type T) { return clift::unwrapped_isa<TypeKind>(T); };
+static constexpr bool isaConstraint(mlir::Type T) {
+  return clift::unwrapped_isa<TypeKind>(T);
+};

 struct VariableForwardRefinementPattern : mlir::OpRewritePattern<AssignOp> {
   using OpRewritePattern::OpRewritePattern;
@@ -412,7 +409,7 @@ struct ExpressionForwardRefinementPattern
                                  unsigned OperandIndex,
                                  mlir::PatternRewriter &Rewriter,
                                  TypeConstraint Constraint =
-                                   NoConstraint) const {
+                                   noConstraint) const {
     mlir::Type RefinedType = getInputType(Op, OperandIndex);

     if (not RefinedType)
@@ -441,7 +438,7 @@ struct ExpressionForwardRefinementPattern
                                   unsigned OperandIndex2,
                                   mlir::PatternRewriter &Rewriter,
                                   TypeConstraint Constraint =
-                                    NoConstraint) const {
+                                    noConstraint) const {
     mlir::Type T1 = getInputType(Op, OperandIndex1);
     mlir::Type T2 = getInputType(Op, OperandIndex2);

@@ -481,7 +478,7 @@ struct ExpressionForwardRefinementPattern
   mlir::LogicalResult
   matchAndRewriteComparison(ExpressionOpInterface Op,
                             mlir::PatternRewriter &Rewriter,
-                            TypeConstraint Constraint = NoConstraint) const {
+                            TypeConstraint Constraint = noConstraint) const {
     mlir::Type T1 = getInputType(Op, 0);
     mlir::Type T2 = getInputType(Op, 1);

@@ -528,7 +525,7 @@ struct ExpressionForwardRefinementPattern
       return matchAndRewriteUnaryExpression(Op,
                                             /*OperandIndex=*/0,
                                             Rewriter,
-                                            IsaConstraint<IntegralType>);
+                                            isaConstraint<IntegralType>);
     }

     if (mlir::isa<AddOp,
@@ -545,7 +542,7 @@ struct ExpressionForwardRefinementPattern
                                              /*OperandIndex1=*/0,
                                              /*OperandIndex2=*/1,
                                              Rewriter,
-                                             IsaConstraint<IntegralType>);
+                                             isaConstraint<IntegralType>);
     }

     if (mlir::isa<TernaryOp>(Op))
```

vasama-work: Ah, I believe there was a previous iteration where function objects were indeed needed to avoid pointer decay, but it seems that is not the case now. I'll change these to functions then. 👍

lib/CliftTransforms/TypeRefinement.cpp
vasama-work: I would rather keep them separate. In fact, I think I would like to get rid of this whole array check and turn them into value types.

pfez: That's 2 separate issues. Turning the check into value types makes sense, but that doesn't mean I agree splitting them. Having a single reused function is very useful when you read the code. If you keep them separate you keep wondering if there is some place that differs. If the function is the same that can't happen and it much better. So let's do that.

lib/CliftTransforms/Expressions.cpp
vasama-work: Done.

lib/CliftTransforms/TypeRefinementWithEFA.cpp
pfez: > For this particular pass, no.

Allright.

> perhaps instead of a separate pass, we should just include the type refinement rewrites in the optimize-expressions pass

In principle, it could very well be. But I'll keep them separate for now.

include/revng/CliftTransforms/ExpressionPostProcessing.td
pfez: Allright then.

lib/CliftTransforms/EmitFieldAccesses.cpp
pfez: After our discussion on mattermost, I'll keep this as it is.

PR #597, opened 2026-06-23
Improve memory profiling performance

mrjackv:


include/revng/InlineHelpers/LinkHelpersToInline.h
```diff
@@ -14,6 +14,11 @@ class LinkHelpersToInlinePass : public llvm::ModulePass {
 public:
   static char ID;
 
+  // This map stores the loaded helper module(s) to be inlined. Generally state
+  // in a pass is frowned upon, but since this is only ever read it's simpler to
+  // store it at the pass level.
```
aleclearmind: Use doxygen

lib/Support/Progress.cpp
```diff
@@ -30,6 +30,33 @@ static opt<bool> HighResolutionMemory("high-resolution-memory-tracing",
                                            "for memory traces"),
                                       init(false));
 
+template<StrictSpecializationOf<std::chrono::duration> T>
+class DurationParser : public llvm::cl::parser<T> {
```
aleclearmind, line 34: Let's keep it simple, just use a `uint64_t`. I don't see a need for `DurationParser`.

include/revng/InlineHelpers/LinkHelpersToInline.h
```diff
@@ -14,6 +14,11 @@ class LinkHelpersToInlinePass : public llvm::ModulePass {
...
+  llvm::StringMap<std::unique_ptr<llvm::Module>> HelpersModules;
```
aleclearmind, line 20: I was 100% we were already doing this 🤔

lib/Support/Progress.cpp
mrjackv: This is to avoid having to create a `std::chrono::duration` every time `emitMemoryUsage` is called. Doing it in the constructor is not viable because `--trace=... --memory-profiler-interval=XXX` would not work since clopts are processed in order

include/revng/InlineHelpers/LinkHelpersToInline.h
mrjackv: We did, but then gus' PR shifted things around and this part got lost

lib/Support/Progress.cpp
```diff
@@ -14,12 +14,48 @@ extern "C" {
...
+public:
+  using llvm::cl::parser<T>::parser;
+
+  bool parse(llvm::cl::Option &Option,
+             llvm::StringRef ArgName,
+             const llvm::StringRef ArgValue,
+             T &Val) {
+    uint64_t IntegerValue;
```
aleclearmind: Always zero-initialize scalars.

mrjackv: Done

include/revng/InlineHelpers/LinkHelpersToInline.h
mrjackv: Done

aleclearmind:
Merged in 97f20b7. Thanks.

PR #603, opened 2026-07-01
EmitFieldAccesses: PA::`verify` soft-fail

andrealinux1:
`computePointerArithmetic` asserted on a `PointerArithmetic` that did not verify, aborting on a negative `Stride` or a negative `BaseOffset`.

Return `nullopt` to keep the raw pointer arithmetic instead and transform the failure into a soft-fail.

`verify()` now also rejects negative `BaseOffset`s.

### DIFF
--- a/lib/CliftTransforms/PointerArithmetic.cpp
+++ b/lib/CliftTransforms/PointerArithmetic.cpp
@@ -36,10 +36,18 @@ bool PointerArithmetic::isAddress() const {
 
 bool PointerArithmetic::verify() const {
 
+  // A negative `BaseOffset` would point before the pointed-to object, so it
+  // cannot correspond to a field or array element access
+  if (Offset.BaseOffset.isNegative()) {
+    return false;
+  }
+
   const auto &LinearCombination = Offset.LinearCombination;
 
   // Check that strides are strictly positive. Negative or null strides do not
-  // make sense
+  // make sense.
+  // TODO: currently negative `Stride`s fail the verify, we may need to support
+  //       them in the future.
   for (const auto &Term : LinearCombination) {
     if (Term.Stride.isNegative() or Term.Stride.isZero()) {
       return false;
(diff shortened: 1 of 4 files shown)


lib/CliftTransforms/PointerArithmetic.cpp
```diff
@@ -231,6 +239,15 @@ PointerArithmeticBuilder::computePointerArithmetic(ExpressionOpInterface
     return std::nullopt;
   }
 
+  // If the traversal failed, or produced a `PointerArithmetic` that is not in
+  // canonical form, we cannot rewrite the expression, so we leave it as raw
```
pfez: It's not "canonical". It's the only valid form of `PointerArithmetic` we support. "Canonical" makes it seem that there are also non canonical forms that may be valid. Let's not use "canonical" for this anywhere please.

lib/CliftTransforms/PointerArithmetic.h
```diff
@@ -66,7 +66,8 @@ struct PointerArithmetic {
   /// Check if this is an address (has base pointer) arithmetic
   bool isAddress() const;
 
-  /// Verify the invariants for the strides contained in the PointerArithmetic
+  /// Verify the `PointerArithmetic` is in canonical form: a non-negative
```
pfez: Ditto: on "canonical".

lib/CliftTransforms/PointerArithmetic.cpp
pfez: If you fix this "canonical" wording this is good to merge.

andrealinux1: Removed references to "canonical" and changed with "valid"

lib/CliftTransforms/PointerArithmetic.h
andrealinux1: As above

PR #604, opened 2026-07-01
Mass testing report fixes 2

mrjackv:


python/revng/internal/cli/_commands/mass_testing/generate_report/stacktrace.py
```diff
@@ -23,6 +24,24 @@
 )
 
 
+def _percentile(data: Iterable[float], p: float):
```
aleclearmind, line 32:
```python
from statistics import quantiles
```

mrjackv: This function also works for sub-unit percentiles (e.g. `99.7`), also it's more precise than `quantile`, because it uses linear interpolation between the two closest ranks

aleclearmind: If you think this is important, OK, but add a comment

mrjackv: Done

aleclearmind:
Merged in 55b7f2d. Thanks.

PR #605, opened 2026-07-03
CLI improvements

aleclearmind:


aleclearmind:
@mrjackv This PR is still a bit WIP but I'd like your review on:

* ``--trace`: ensure we emit valid JSON`
* `lit.cfg.py: do not leak temporary directory`
* `Pipeline: add --overwrite to project init`
* `Pipeline: default artifact output to auto format`
* `Pipeline: detect on-disk changes`
* `CLI: support specifying objects by name`

I also left some comments on parts I have doubts about.

python/revng/pypeline/storage/local_provider_migrations/v3.sql
```diff
@@ -0,0 +1,15 @@
...
+-- set_model. It lets us diff against the on-disk model and purge only the
+-- affected objects when the model file is edited behind our back, instead of
+-- discarding all the caches. The caches are cleared here because they cannot
+-- be reconciled with a baseline that did not exist before this migration.
+
+DELETE FROM objects;
+DELETE FROM dependencies;
+DELETE FROM custom_dependencies;
+
+ALTER TABLE project ADD COLUMN last_seen_model BLOB;
```
aleclearmind, line 15: @mrjackv Is `BLOB` OK?

python/revng/pypeline/storage/local_provider.py
```diff
@@ -411,51 +410,6 @@ def _check_version_and_pipeline_hash(self, pipeline_hash: str):
                     (version, pipeline_hash),
                 )
 
-    def _check_model(self):
```
aleclearmind, line 414: Given the other changes, I think it's safe to drop `_check_model`.

What's your take?

python/revng/pypeline/storage/local_provider_migrations/v3.sql
mrjackv: Yes, `BLOB` is the correct type

python/revng/pypeline/storage/local_provider.py
mrjackv: Considering that the logic has morally moved to `unprocessed_model_changes` it's ok

python/revng/pypeline/storage/local_provider.py
```diff
@@ -650,28 +604,73 @@ def get_model(self) -> tuple[Model, int]:
...
         with self._cursor() as cursor:
             cursor.execute(
-                "UPDATE project SET model_hash = ?, model_mtime = ? WHERE id is 0",
-                (compute_hash(model_bytes), self._model_path.stat().st_mtime),
+                "UPDATE project SET model_hash = ?, model_mtime = ?, last_seen_model = ?"
+                " WHERE id is 0",
+                (compute_hash(model_bytes), self._model_path.stat().st_mtime, model_bytes),
             )
             self.epoch += 1
             self._write_metadata(cursor, _MetadataUpdate.MODEL_SAVE)
```
mrjackv, line 686: I think this should all be moved inside the if (or better, have `if disk_bytes == model_bytes: return`. Updating the mtime when the model hasn't changed is wasteful.

python/revng/pypeline/model.py
```diff
@@ -28,6 +28,10 @@ def paths(self) -> ModelPathSet:
     def serialize(self) -> bytes:
         raise NotImplementedError()
 
+    def empty(self) -> bool:
+        """True if the diff represents no change."""
+        return len(self.paths()) == 0
```
mrjackv: This materializes the paths, which could be big. If possible it'd be better to have `ModelDiff.size()` and then have at the nanobind level `__len__`

python/revng/pypeline/pipeline.py
```diff
@@ -615,13 +615,44 @@ def run_analysis(
...
+    ) -> tuple[Model, int]:
+        """
+        Load the model from the storage provider, reconciling any change that
+        happened to the backing model behind the provider's back (e.g. the
+        on-disk model file was edited). When such a change is detected, the
+        diff is used to purge only the affected objects from the caches
+        (including custom invalidation) instead of discarding everything.
+        """
+        diff = storage_provider.unprocessed_model_changes()
+        if not diff.empty():
```
mrjackv: In relation to the previous comment, `if len(diff) == 0:` is more pythonic

include/revng/PipeboxCommon/Model.h
```diff
@@ -66,6 +66,39 @@ class Model {
     revng_abort();
   }
 
+  std::vector<std::string> aliases(const ObjectID &Object) const {
```
mrjackv, line 76: This misses C-style names, e.g. `function_0x402a87_Code_x86_64`
Just making sure it's intentional.

python/revng/pypeline/cli/utils.py
```diff
@@ -264,7 +264,30 @@ def list_objects_for_container(
...
+
+def deserialize_object(model: ReadOnlyModel, kind: Kind, obj: str) -> ObjectID:
+    """
+    Deserialize a single object identifier. If the string is not a valid
+    ObjectID, fall back to resolving it as an alias through the model.
+    """
+    obj_id_type = get_singleton(ObjectID)  # type: ignore[type-abstract]
+    try:
+        return obj_id_type.deserialize(obj)
+    except RuntimeError:
```
mrjackv: Maybe `resolve_alias` should also handle plain objectIDs, as it avoids two round-trips in the worst case.

python/revng/internal/cli/_commands/process_docs_yaml/analyses.md.tpl
```diff
@@ -16,7 +16,7 @@ This page reports the list of all the analyses currently featured in rev.ng.
 ## <a id="/analyses/{{analysis.Name}}"></a>`{{analysis.Name}}` analysis
```
mrjackv: Putting it here since I can't comment on the whole commit: do note that the `revng process-docs-yaml` still uses the old pipeline doc (`share/revng/pipelines/revng-pipelines.yml`), it should switch to the new one (`python/revng/internal/pipeline.yml`)

mrjackv: Additional note: `artifacts-and-analyses.md` still uses the old pipeline terms

share/doc/revng/references/cli/revng2-common.md
```diff
@@ -0,0 +1,123 @@
...
+With `--debug DIR`, a directory where you can easily reproduce each step of the invocation is created. Each pipe and analysis is run as a separate [`revng2 pipeline run-pipe`](revng2-pipeline-run-pipe.md) / [`run-analysis`](revng2-pipeline-run-analysis.md) subcommand, and its inputs and outputs are recorded under `DIR`. `DIR` gets one numbered subdirectory per executed step, in execution order:
+
+```
+0000-import-files/
+0001-collect-cfg/
+0002-process-assembly/
+0003-yield-assembly/
+```
+
+Each step subdirectory contains:
```
mrjackv: This is slightly inexact, the layout described below is for pipes, analyses are slightly different (e.g. don't have an `output/` dir)

share/doc/revng/references/cli/revng2-project-analyze.md
```diff
@@ -4,45 +4,105 @@
...
-The documentation describes [what an analysis is](../../user-manual/key-concepts/artifacts-and-analyses.md#analyses), the [available analyses and their options](../analyses.md).
+Runs the [analysis](../analyses.md) (or [analysis list](../pipeline.md#analysis-lists)) `ANALYSIS` on the project in the current directory, updating its model in place.
+
+Run without arguments to list all the available analyses.
+
+The documentation describes [what an analysis is](../../user-manual/key-concepts/artifacts-and-analyses.md#analyses).
+
+The project must have been created with [`revng2 project init`](revng2-project-init.md) first, so that a `model.yml` is present in the working directory.
 
 OPTIONS
```
mrjackv, line 27: Here the `--*-objects` option is missing. While it's not so used, it should be there for completeness' sake.

share/doc/revng/references/cli/revng2-project-artifact.md
```diff
@@ -9,46 +9,66 @@ NAME
...
-Produces the artifact `ARTIFACT`.
+This command is a subcommand of [`revng2 project`](revng2-project.md).
 
-If no `OBJECT` is specified then the command will produce all available objects.
+Produces the artifact `ARTIFACT` from the project in the current directory.
 
-Run `revng2 project artifact --help` to list the available artifacts.
+Run without arguments to list all the available artifacts.
+
+`OBJECTS` is a single comma-separated list of object ids to produce (for example `0x401000:Code_x86_64,0x401120:Code_x86_64`). If omitted, all the available objects are produced.
```
mrjackv: Here we should state that function/type names are also ok

mkdocs.yml
```diff
@@ -35,11 +35,21 @@ nav:
       - Artifacts: references/artifacts.md
       - Analyses: references/analyses.md
       - CLI:
-          - revng-common: references/cli/revng-common.md
```
mrjackv, line 38: Mkdocs suport having index pages, we could have this be a bit more hierarchical, e.g.:
```yaml
- CLI:
  - revng2:
    - references/cli/revng2.md
    - revng2-project:
      - references/cli/revng2-project.md
      - revng2-project-init: references/cli/revng2-project-init.md
      ...
```

python/revng/pypeline/container.py
```diff
@@ -310,6 +318,13 @@ def to_bytes(
                             info.size = len(data_bytes)
                             tar.addfile(tarinfo=info, fileobj=io.BytesIO(data_bytes))
                     return byte_stream.getvalue()
+            case ContainerFormat.AUTO:
+                # A lone object is emitted raw, without any YAML/base64 wrapping;
+                # zero or multiple objects fall back to YAML.
+                contents = list(self.to_dict(objects).values())
```
mrjackv: This serializes things twice, which is not good if the binary has a lot of functions. Maybe leave a TODO to have `Container._effective_objects(objects: ObjectSet)` which uses `Container.objects()` before doing any serialization

python/revng/pypeline/storage/local_provider.py
aleclearmind: I didn't like the side-effect (of an otherwise read-only method) either, but it's true that it saves from reading the model next time.

python/revng/pypeline/cli/utils.py
aleclearmind: I wanted this to be a fallback, in case one calls a function `/functions/...` or `0x1000...`.

Still, `resolve_alias` could prioritize it on its own, so it's probably good to do what you suggest.

python/revng/pypeline/model.py
aleclearmind: Fixed in `Pipeline: detect on-disk changes`

python/revng/pypeline/pipeline.py
aleclearmind: I went if with just `if diff:`

include/revng/PipeboxCommon/Model.h
aleclearmind: `CLI: support specifying objects by name` now also considers `automaticName`s.

python/revng/pypeline/cli/utils.py
aleclearmind: `CLI: support specifying objects by name` handles everything internally, we don't first try `.deserialize` on the caller side.

python/revng/internal/cli/_commands/process_docs_yaml/analyses.md.tpl
aleclearmind: Fixed, along with a general overhaul of the docs, in `Docs: generate reference docs from pipeline.yml`

python/revng/pypeline/pipeline.py
mrjackv: Ok

python/revng/internal/cli/_commands/process_docs_yaml/analyses.md.tpl
aleclearmind: Addressed in `Docs: document the revng2 CLI, retire legacy pages`

share/doc/revng/references/cli/revng2-common.md
aleclearmind: Addressed in `Docs: revise revng2 CLI reference pages and nav`

share/doc/revng/references/cli/revng2-project-analyze.md
aleclearmind: Addressed in `Docs: revise revng2 CLI reference pages and nav`

share/doc/revng/references/cli/revng2-project-artifact.md
aleclearmind: Addressed in `Docs: revise revng2 CLI reference pages and nav`

mkdocs.yml
aleclearmind: Addressed in `Docs: revise revng2 CLI reference pages and nav`

python/revng/pypeline/container.py
aleclearmind: Take a look now, I now rely on the number of objects in `objects`.

mrjackv: Looks good

python/revng/pypeline/storage/local_provider.py
aleclearmind: Did what you suggested in `revng2: if single object, unwrap it`.

mrjackv: Wrong thread

aleclearmind: Wrong commit references, I meant `pypeline: detect on-disk changes`

python/revng/pypeline/container.py
```diff
@@ -310,6 +318,14 @@ def to_bytes(
                             info.size = len(data_bytes)
                             tar.addfile(tarinfo=info, fileobj=io.BytesIO(data_bytes))
                     return byte_stream.getvalue()
+            case ContainerFormat.AUTO:
+                # A lone object is emitted raw, without any YAML/base64 wrapping;
+                # zero or multiple objects fall back to YAML.
+                target = self.objects() if objects is None else objects
+                if len(target) == 1:
+                    buffer, _ = self.serialize(target).values()
```
mrjackv:
```diff
- buffer, _ = self.serialize(target).values()
+ buffer = list(self.serialize(target).values())[0]
```

include/revng/PTML/CTokenEmitter.h
```diff
@@ -245,12 +245,18 @@ class CTokenEmitter {
     bool &IsEmittingComment;
 
     CommentKind Kind;
+    // Text emitted right after the `//` of each line of a line comment (empty
+    // by default). Used to space the body away from the marker; the doxygen
+    // `/ ` prefix is added by CDoxygenEmitter on top instead.
+    llvm::StringRef LinePrefix;
```
Cvelth: This patchset looks simple on the first glance, but the more I look into it, the less I like it.

(I have a bunch of separate comments, but they all boil out to, pretty much, the same thing, so I'm dropping them and merging everything I have to say into a single wall of text (tm) here).

---

First of all, I'm not sure this is the correct level to even do this in the first place. The comment even mentions `CDoxygenEmitter` by name. Which exists to solve this *exact* problem in the doxygen comment space. This screams inconsistency: we do the very same thing in two different places using two different levels of abstraction.

The situation, as far as I understand it, is:
- We want `/ ` prefix in doxygen and ` ` prefix in non-doxygen comments (` * ` and `` for block mode).
- We want (?, I guess) to share as much code as we can by pulling the common `//`-part into its own emitter.
- From what I remember, the original reasoning is that this is the smallest thing that is *required* to prove `CTokenEmitter` always emits valid C (miss-formed comments are not possible) - everything else (spacing and whatnot) is fluff on top and should come somewhere further down the emitter chain - hence the complicated templated emitter inheritance system.

---

Everything but the ` ` prefix in non-doxygen line comments are handled on `develop` already. For the missing piece, the LLM (I'm assuming) that authored this snippet just added the extra field in here without any regard to how it fits into the picture (which, btw, definitely needs better docs!).

So we're ending up with a field that is unused in 3/4 of the potential cases, and only ever gets a *hard-coded* value in those cases it is in fact used.

---

The correct thing to do here, to my eyes at least, is to rename the current `CDoxygenEmitter` into something more generic (`CCommentEmitter`?) and let it handle all the cases we require. It's just a matter of adding more "modes" of operation to it after all.

Note that it already contains all of the non-trivial the logic we're duplicating here, like the
```cpp
PTML.emit(IsEmptyLine ? LinePrefix.rtrim() : LinePrefix);
```
thing (see the function at `DoxygenEmitter.h` line 108 and its callees).
https://github.com/revng/revng/blob/7ccb5b34efd293950517b7453ab3e5e4e98b5737/include/revng/PTML/DoxygenEmitter.h#L108

---

Another option (which, I personally prefer and tried to fight `@vasama-work` and `@fez` really hard for, a couple of months back) is to get rid of the comment logic fragmentation and to have a *single* comment-emitter-thing that takes care of all our needs, instead of fragmenting it across three classes (one templated).

The problem with that is it requires either putting the entire thing inside `CTokenEmitter` or opening holes in its interface to allow for custom logic injection. And both of them are *extremely* defensive about the `CTokenEmitter` interface and complexity (for a good reason, but still, forces us to have workarounds like these)!

---

P. S. I wish I could have just redirected this review to `@vasama-work` (who has much better understanding of the new emitter ecosystem), but alas he's on vacation.

P. P. S. Have you shown this code to `@fez`? I don't think this would pass his review either.

P. P. P. S. What happens when this `LinePrefix` is ever set for the doxygen emitter? Wouldn't we get a prefix like `// / `?

aleclearmind: I dropped this in favor of a branch adopting `clang-format`.

python/revng/internal/cli/revng2.py
```diff
@@ -325,6 +325,11 @@ async def async_part_of_command(
     type=click.Path(exists=True, dir_okay=False, path_type=Path),
```
mrjackv, line 326: Some commits have `(cherry picked from commit XXXXXXX)` in the commit message, you might want to remove it

python/revng/pypeline/storage/local_provider.py
```diff
@@ -650,28 +604,82 @@ def get_model(self) -> tuple[Model, int]:
...
+        if stored_mtime == model_mtime:
+            return get_singleton(ModelDiff)(), None  # type: ignore[type-abstract]
+
+        current_bytes = self._model_path.read_bytes()
+        if compute_hash(current_bytes) == stored_hash:
+            # Only the mtime moved, the content is unchanged: refresh the mtime
+            # so we can skip hashing next time.
+            with self._cursor() as cursor:
+                cursor.execute("UPDATE project SET model_mtime = ? WHERE id is 0", (model_mtime,))
+            return get_singleton(ModelDiff)(), None  # type: ignore[type-abstract]
```
mrjackv: I would hoist a `ModelDiffType = get_singleton(ModelDiff)`

python/revng/internal/cli/revng2.py
```diff
@@ -33,6 +33,7 @@
 from revng.pypeline.cli.wrappers import WRAPPER_REGISTRY, WrappablePypeCommand, WrapperOption
 from revng.pypeline.cli.wrappers import exec_with_wrapper, exec_wrapper_if_needed
 from revng.pypeline.container import ContainerDeclaration, ContainerFormat
+from revng.pypeline.engine import EngineFeature, engine_for
```
mrjackv: I don't know if `Engine` is the correct term here. Engine sounds like something that does stuff (e.g. graph layout), here is more of an adapter between two implementations. I'd go for something like `revng.pypeline.cli.backend.Backend`

python/revng/pypeline/cli/project/daemon.py
```diff
@@ -9,13 +9,25 @@
...
 @pass_context
 def run_daemon(ctx: ClickContext, production: bool):
     """Start the HTTP daemon."""
+    engine = engine_for(
+        ctx.obj.storage_provider_url,
+        pipeline=ctx.obj.pipeline,
+        base_directory=ctx.obj.base_directory,
+        cache_dir=ctx.obj.cache_dir,
+    )
+    if EngineFeature.DAEMON in engine.features:
```
mrjackv: This seems like fake abstraction, I'd just do `if isinstance(engine, DaemonEngineSession)`

python/revng/pypeline/engine/engine.py
```diff
@@ -0,0 +1,150 @@
...
+        configuration: PipelineConfiguration,
+    ) -> tuple[Model, InvalidatedObjects]:
+        """Run an analysis list and return the new model and invalidations."""
+
+    @abstractmethod
+    def put_files(self, files: list[FileStorageEntry]) -> list[str]:
+        """Store input files, returning their hashes."""
+
+
+class Engine(ABC):
```
mrjackv: The naming is not great IMHO, you have `Engine` and the `EngineSession`, I'd stick with what's used everywhere else and have `XXX` and `XXXFactory`.
On top of that `Session` usually means that there is state involved, however here both classes basically have no state.

python/revng/pypeline/engine/engine.py
```diff
@@ -0,0 +1,150 @@
...
+    DEBUG = auto()
+    # Compute is delegated to a running daemon over HTTP rather than performed
+    # in-process.
+    DAEMON = auto()
+    # `--invalidations` can report the objects an analysis invalidated. A remote
+    # engine that only returns a model diff cannot.
+    INVALIDATIONS = auto()
+
+
+class EngineSession(ABC):
```
mrjackv: There should be a type annotation that states that the `features` classvar exists

lib/EditCBody/Statements.h
```diff
@@ -0,0 +1,55 @@
...
+
+namespace mlir {
+class Operation;
+} // namespace mlir
+
+namespace revng::editcbody {
+
+/// The kind of a statement, coarse enough to be shared between the Clang AST
+/// and the Clift representation. Blocks (`{ }`) are transparent and produce no
+/// node.
```
pfez: > Blocks (`{ }`) are transparent and produce no node.

Can you please elaborate?

include/revng/EditCBody/EditCBodyAnalysis.h
```diff
@@ -0,0 +1,48 @@
...
+///   statement, located by the addresses of the instructions that make it up.
+///   It can be placed before any statement that carries at least one address
+///   (a computed expression, a `return`, or the condition of an `if` or a
+///   loop); a statement with no address, such as a synthesized `break`, cannot
+///   be commented.
+///
+/// - a `RENAME: <name>` and/or `RETYPE: <type>` comment renames and/or retypes
+///   a local variable, recorded as a `LocalVariable` located by the addresses
+///   of the instructions that use it. It can only be placed before the
+///   variable's declaration.
```
pfez, line 37: Do you expect this to be able to handle a comment that has many of these at the same time?

Like:
```c
// A counter
// RETYPE: int32_t

// RENAME: counter
generic32_t var;
```

python/revng/pypeline/cli/backend/daemon_backend.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def _get(self, path: str, **kwargs) -> Response:
+        return self._request("GET", path, **kwargs)
+
+    def _post(self, path: str, **kwargs) -> Response:
+        return self._request("POST", path, **kwargs)
+
+    def _request(self, method: str, path: str, **kwargs) -> Response:
+        try:
+            return self._http.request(method, self._url(path), **kwargs)
```
mrjackv, line 155: Here there should be a `.raise_for_status` to catch all the non-200 responses

lib/PipeboxCommon/Python/_pipebox.cpp
```diff
@@ -229,7 +229,7 @@ NB_MODULE(_pipebox, m) {
   nanobind::class_<Model>(m, "Model", ModelBaseClass)
     .def(nanobind::init<>())
     .def_ro_static("identifier", &"revng")
-    .def_static("model_name", []() { return nanobind::str("model.yml"); })
```
mrjackv, line 230: This is a big change, I'm sure it will also require giving a look at revng-ui and hitman to make sure there aren't any hardcoded strings sitting around

python/revng/pypeline/cli/pipeline/run_analysis.py
```diff
@@ -70,13 +70,14 @@ def _build_analysis_command(self, analysis_name: str):
         config = getattr(
             analysis_type,
             "configuration_help",
-            f'Configuration for the analysis "{analysis_name}".',
+            f'Path to a YAML file with the configuration for the analysis "{analysis_name}".',
```
mrjackv: This is kind of important,  we should write somewhere that all configurations are in YAML

python/revng/project/cli_project.py
```diff
@@ -79,10 +87,14 @@ def _analyze_impl(
...
-            temp_file.seek(0)
-            model = Binary.deserialize(temp_file, self)
+        try:
+            with TemporaryFile("w+") as temp_file:
+                self._cli_helper.run(args, stdout=temp_file)
+                temp_file.seek(0)
+                model = Binary.deserialize(temp_file, self)
+        finally:
+            for config_file in config_files:
+                config_file.close()
```
mrjackv: I think `contextlib.ExitStack` is better here

python/revng/internal/cli/revng2.py
aleclearmind: As I mentioned, I still need to reword the commit messages.

lib/PipeboxCommon/Python/_pipebox.cpp
aleclearmind: Go ahead and report back

python/revng/pypeline/cli/pipeline/run_analysis.py
aleclearmind: Ah no, I'll `s/YAML//`

include/revng/EditCBody/EditCBodyAnalysis.h
aleclearmind: Yes, that's the idea.

lib/EditCBody/Statements.h
aleclearmind: I'll drop it, but I think the meaning was like "we don't associated a statement to parentheses, not even if they are empty".

lib/EditCBody/Statements.h
```diff
@@ -0,0 +1,55 @@
...
+enum class StatementKind {
+  Declaration,
```
pfez: I would call this `LocalVariableDeclaration`.

lib/EditCBody/Statements.h
```diff
@@ -0,0 +1,55 @@
...
+  Expression,
+  Return,
+  If,
+  While,
+  DoWhile,
+  For,
+  Switch,
+  Goto,
+  Break,
```
pfez, line 30: You don't distinguish between breaking from loops and breaking from `switch`. Perhaps this is not necessary, if clang doesn't distinguish either (which I don't recall).

This is a note-to-self to check what clang does, and see if there are consequences if it does it differently.

lib/EditCBody/CliftFlatten.cpp
```diff
@@ -0,0 +1,134 @@
...
+    return StatementKind::DoWhile;
+  if (mlir::isa<ForOp>(Op))
+    return StatementKind::For;
+  if (mlir::isa<SwitchOp>(Op))
+    return StatementKind::Switch;
+  // break_to and continue_to are macros for goto, so the C parser sees a plain
+  // goto for all three. The break closing a fallthrough switch case is a real
+  // break, but it is synthesized by the C backend and handled as a marker.
+  if (mlir::isa<GotoOp, BreakToOp, ContinueToOp>(Op))
+    return StatementKind::Goto;
```
pfez: It seems to me that you never emit `StatementKind::Break` nor `StatementKind::Continue`. The proper way to do it would be to check the parent loop for the `BreakOp` and `ContinueOp`.

Then, if the last operation of the parent loop's body is an `AssignLabelOp` that assigns the label the `ContinueOp` jumps to you can emit a `StatementKind::Continue`.

And for `break`, you have to check the next operation in the parent loop's parent block, and if that is an `AssignLabelOp` that assigns the same label of the `BreakOp`, you can emit `StatementKind::Break`.

Perhaps none of this is really important nor interesting, since it's rare that you want to add comments on those. But still, take this into consideration. If you decide you don't care about this then it may be the case that `StatementKind::Break` nor `StatementKind::Continue` are really needed.

lib/EditCBody/ClangParse.cpp
```diff
@@ -0,0 +1,391 @@
...
+    } else if (isa<ReturnStmt>(Statement)) {
+      push(StatementKind::Return, Statement);
+    } else if (isa<DeclStmt>(Statement)) {
+      push(StatementKind::Declaration, Statement);
+    } else if (isa<GotoStmt>(Statement)) {
+      push(StatementKind::Goto, Statement);
+    } else if (isa<BreakStmt>(Statement)) {
+      push(StatementKind::Break, Statement);
+    } else if (isa<ContinueStmt>(Statement)) {
+      push(StatementKind::Continue, Statement);
```
pfez, line 277: Ah, see. These mismatch with clift, when clift has a `BreakToOp` that gets emitted as a regular `break` and not as `break_to`. Same holds for `continue_to`.

lib/EditCBody/ClangParse.cpp
```diff
@@ -0,0 +1,391 @@
...
+    } else if (isa<Expr>(Statement)) {
+      push(StatementKind::Expression, Statement);
+    } else {
+      Output.Error = "unsupported C statement in the provided C code";
```
pfez: Would be nice to print the offending AST node too.

lib/EditCBody/ClangParse.cpp
```diff
@@ -0,0 +1,391 @@
...
+    return std::nullopt;
+  auto Index = MaybeHeaderPath->rfind('/');
+  if (Index == std::string::npos)
+    return std::nullopt;
+
+  return MaybeHeaderPath->substr(0, Index);
+}
+
+std::vector<std::string> getCompilationFlags() {
+  StringRef CompileFlagsPath = "share/revng/compile-flags.cfg";
```
pfez: This is a bit strict, since it contains a bunch of warnings that we use for recompilation tests but that in general may not always be guaranteed to never trigger.

Sure, if the warnings trigger it is considered a bug (or at least a quality issue) on our size, but I don't think we want overly strict warnings to affect the analysis that is based on this.

I think we either have to split `compile-flags.cfg` in two parts (the "core" and the "warnings"), or that in this specific invocation to the compiler you have to override warnings, turning them all off.

lib/EditCBody/ClangParse.cpp
```diff
@@ -0,0 +1,391 @@
...
+        if (not isAllWhitespace(Buffer.substr(Entry.End, Boundary - Entry.End)))
+          break;
+        if (not startsOwnLine(Buffer, Entry.Begin))
+          break;
+
+        Entry.Consumed = true;
+        Collected.push_back(Entry.Raw);
+        Boundary = Entry.Begin;
+      }
+      std::reverse(Collected.begin(), Collected.end());
```
pfez: Why `reverse`? Is this because if you append first the comments that show up later in the file you don't invalidate the positions of the others that come before?

lib/EditCBody/ClangParse.cpp
```diff
@@ -0,0 +1,391 @@
...
+
+        RawEntry &Entry = Entries[Best];
+        if (not isAllWhitespace(Buffer.substr(Entry.End, Boundary - Entry.End)))
+          break;
+        if (not startsOwnLine(Buffer, Entry.Begin))
+          break;
+
+        Entry.Consumed = true;
+        Collected.push_back(Entry.Raw);
+        Boundary = Entry.Begin;
```
pfez, line 337: Shouldn't this be before `if (isAllWhitespace ... `?

lib/EditCBody/ClangParse.cpp
```diff
@@ -0,0 +1,391 @@
...
+static Logger Log("edit-c-body-clang");
+
+namespace {
+
+//
+// Comment parsing
+//
+
+/// Strip the `//` or `/* */` markers off a raw comment, returning its body.
+std::string commentBody(StringRef Raw) {
```
pfez, line 52: This isn't considering doxygen comments with leading `///`. Is that on purpose?

lib/EditCBody/ClangParse.cpp
```diff
@@ -0,0 +1,391 @@
...
+  if (Raw.starts_with("/*")) {
+    Raw = Raw.drop_front(2);
+    if (Raw.ends_with("*/"))
+      Raw = Raw.drop_back(2);
+    return Raw.trim().str();
+  }
+  return Raw.trim().str();
+}
+
+bool isAllWhitespace(StringRef Text) {
```
pfez: Drop this and use `llvm::isSpace` instead

pfez: `llvm::all_of(S, llvm::isSpace)`

lib/EditCBody/ClangParse.cpp
```diff
@@ -0,0 +1,391 @@
...
+
+/// Whether the comment starting at `Begin` is the first thing on its line, i.e.
+/// only whitespace precedes it. A leading comment must start its own line; this
+/// rules out a comment that trails a statement on the same line, which belongs
+/// to that statement rather than to the one that follows.
+bool startsOwnLine(StringRef Buffer, unsigned Begin) {
+  size_t LineStart = Buffer.substr(0, Begin).rfind('\n');
+  LineStart = (LineStart == StringRef::npos) ? 0 : LineStart + 1;
+  return isAllWhitespace(Buffer.substr(LineStart, Begin - LineStart));
+}
```
pfez, line 130: I'm confused. The comment above this function makes me think that you want to handle also statements that trail a statement, while my current understanding is that we don't want that, and the code doesn't support that. Can you clarify?

lib/EditCBody/ClangParse.cpp
```diff
@@ -0,0 +1,391 @@
...
+  /// directly, without producing a node of its own.
+  RecursiveCoroutine<void> flattenStatement(const clang::Stmt *Statement) {
+    using namespace clang;
+
+    if (Statement == nullptr or isa<NullStmt>(Statement)) {
+      rc_return;
+    } else if (auto *Compound = dyn_cast<CompoundStmt>(Statement)) {
+      for (const Stmt *Child : Compound->body())
+        rc_recur flattenStatement(Child);
+    } else if (auto *Label = dyn_cast<LabelStmt>(Statement)) {
```
pfez, line 240: Putting this here for a lack of a better place. We've talked a lot about `RENAME` for local variables, but I think we could use that for label names as well. How much additional work do you think it would be necessary for that?

lib/EditCBody/Statements.h
aleclearmind: Distinguishing `goto` from `break` is more a sanity check that we're not miscounting, more than an actual semantic issue.

What matters is that we index statements/labels/declarations coherently between clift and clang.

I don't we'd gain much in distinguishing one of type of break from the other.

lib/EditCBody/CliftFlatten.cpp
aleclearmind: As I said in the other comment, this is more like a sanity check.
To be honest, we could map `break`, `continue`, `break_to`, `continue_to` and `goto` to `StatementKind::Goto` and call it a day, or maybe `NonLocalJump`.

> since it's rare that you want to add comments on [goto/continue/break]

Assigning a `StatementKind` is not about being able or not to assign a comment, just making sure we're not matching, say, an expression statement on clift to a break on the clang AST.

Again, mostly sanity check, assertions.

lib/EditCBody/ClangParse.cpp
aleclearmind: I think we want this to get to the user, I'd prefer not to dump the AST in its face.
Maybe we can just dump the node type name?

lib/EditCBody/ClangParse.cpp
aleclearmind: This is also what we do in `import-from-c`. Maybe we should factor these together.

I'll look into the core vs warnings thing.

lib/EditCBody/EditCBodyAnalysis.cpp
```diff
@@ -0,0 +1,357 @@
...
+  Variable.Name() = NewName.has_value() ? *NewName :
+                                          LocalVariable.getName().str();
+  if (NewTypeName.has_value()) {
+    model::UpcastableType Type = resolveTypeName(*NewTypeName, Binary);
+    if (Type.isEmpty())
+      return revng::createError("unknown type in `RETYPE`: " + *NewTypeName);
+    Variable.Type() = std::move(Type);
+  }
+  Variable.Location() = std::move(Location);
+  return Variable;
```
pfez, line 144: This doesn't handle the case where the variable originally had a defaulted name and  `NewName` is the same.

lib/EditCBody/ClangParse.cpp
aleclearmind: Usually one doesn't put doxygen comments in the function body.

I don't think we want to handle that.

lib/EditCBody/EditCBodyAnalysis.cpp
aleclearmind: Why would you do that? That would be like doing that in the model, which is not smart but valid.

I don't think it makes much sense to handle that case in particular

lib/EditCBody/EditCBodyAnalysis.cpp
```diff
@@ -0,0 +1,357 @@
...
+    }
+    if (MaybeEdits->Variable.has_value())
+      NewVariables.push_back(std::move(*MaybeEdits->Variable));
+  }
+
+  // Replace the function's comments with the imported ones.
+  model::Function &ModelFunction = *FunctionIterator;
+  ModelFunction.Comments().clear();
+  for (model::StatementComment &Comment : NewComments)
+    ModelFunction.Comments().insert(std::move(Comment));
```
pfez, line 356: Let me get this straight: here we are clearing out comments and only adding new ones. In practice that also allows editing comments in place, because they are reparsed and inserted anew. But that's not ideal from an invalidation standpoint, isn't it? On the other hand, if even a single comment is added or modified, the whole body of the function will have to be invalidation, so that's not really a big deal.

I'm wondering what will happen when we have stack frame type inlining though. What about the stack field pre existing comments? It's not clear to me if with the current machinery they would mistakenly parsed and attached to the stack variable as well.

lib/EditCBody/EditCBodyAnalysis.cpp
```diff
@@ -0,0 +1,357 @@
...
+
+  // Apply the variable edits, replacing any variable already located at the
+  // same set of addresses.
+  for (model::LocalVariable &Variable : NewVariables) {
+    ModelFunction.LocalVariables().erase_if([&](const auto &Existing) {
+      return Existing.Location() == Variable.Location();
+    });
+    ModelFunction.LocalVariables().insert(std::move(Variable));
+  }
```
pfez, line 365: Nice.

lib/EditCBody/EditCBodyAnalysis.cpp
aleclearmind: If we re-parse the comments as they were, there's no invalidation.

Invalidation is computed on diffing before and after, not edits to the model.
We track reads to the model, not writes.

python/revng/internal/cli/revng2.py
aleclearmind: Fixed.

python/revng/pypeline/storage/local_provider.py
aleclearmind: Done.

python/revng/internal/cli/revng2.py
aleclearmind: Please review if you like the new naming.

lib/EditCBody/EditCBodyAnalysis.cpp
```diff
@@ -0,0 +1,357 @@
...
+  return std::move(*MaybeFile);
+}
+
+//
+// Local variable helpers
+//
+
+/// Resolve a C type name (as used in the decompiled output) to a model type: a
+/// primitive, a named type definition, or either with trailing `*`s. Returns an
+/// empty type if the name cannot be resolved.
```
pfez: What about arrays and pointers to arrays? In both those case you have trailing `[]`. Which aren't handled here.

python/revng/pypeline/cli/project/daemon.py
aleclearmind: Done.

python/revng/pypeline/engine/engine.py
aleclearmind: Please review if you like the new naming.

python/revng/pypeline/engine/engine.py
aleclearmind: Done, see if it's what you meant.

lib/EditCBody/EditCBodyAnalysis.cpp
pfez: Also, what about `pointer32_t(T)`?

python/revng/pypeline/cli/backend/daemon_backend.py
aleclearmind: See `DaemonBackendFactory` if satisfies you.

python/revng/pypeline/cli/pipeline/run_analysis.py
aleclearmind: Fixed

python/revng/project/cli_project.py
aleclearmind: See if you like it

lib/EditCBody/EditCBodyAnalysis.cpp
```diff
@@ -0,0 +1,357 @@
...
+
+  model::UpcastableType Result = model::PrimitiveType::fromCName(Name);
+  if (Result.isEmpty())
+    Result = model::getTypeDefinitionByNameOrID(Model, Name);
+
+  if (Result.isEmpty())
+    return Result;
+
+  uint64_t PointerSize = //
+    model::Architecture::getPointerSize(Model.Architecture());
```
pfez: So you're always using the pointer size of the input binary to size a pointer here? Why not the target pointer size?

python/revng/internal/cli/revng2.py
mrjackv: Looks good

python/revng/pypeline/engine/engine.py
mrjackv: Yes, it is

python/revng/project/cli_project.py
mrjackv: Looks good

lib/EditCBody/EditCBodyAnalysis.cpp
aleclearmind: Ah, eh, I didn't think about it.
I reused the logic we use in `import-from-c`, but there we have the clang AST.

Options:

1. see if we can parse "on demand" a string as a C type with clang;
2. try to parse C types by hand (and cry);
3. decide to parse a more reasonable way to express C types, like `model::Type::toDebugString` (sad).

Ideas.

python/revng/pypeline/cli/backend/daemon_backend.py
mrjackv: ~~I'm not seeing any change~~

NVM, it was a little bit below, OK

lib/EditCBody/ClangParse.cpp
aleclearmind: Comments are collected nearest-first (walking upward from the statement), so `Collected` is in reverse source order.

lib/EditCBody/EditCBodyAnalysis.cpp
aleclearmind: We can turn

```c
// RETYPE: lol ***
```

into

```c
// RETYPE: lol ***
typedef lol ***special_typedef_123;
```

So clang parses it, something like that.

lib/EditCBody/ClangParse.cpp
aleclearmind: No, the whitespace check validates the gap between the candidate comment's end and the current boundary before accepting it, advancing the boundary first would check an empty range and wrongly accept non-adjacent comments.

lib/EditCBody/Statements.h
aleclearmind: Dropped

lib/PipeboxCommon/Python/_pipebox.cpp
aleclearmind: I also renamed the `import-from-c` analysis to `edit-c-type`.
Can you make a PR for that?

lib/EditCBody/Statements.h
aleclearmind: Done

lib/EditCBody/ClangParse.cpp
aleclearmind: I introduced `Introduce ClangToModel` which splits off `compile-flags.cfg` some sanity checks into `compile-flags-sanity.cfg`.

lib/EditCBody/ClangParse.cpp
aleclearmind: Done

lib/EditCBody/ClangParse.cpp
aleclearmind: We don't handle trailing comments.

This is now documented and the comment has been fixed.

lib/EditCBody/ClangParse.cpp
aleclearmind: Done and tested.

See `Introduce the `edit-c-body` analysis`:

```
share/revng/test/configuration/revng/edit-c-body.yml
share/revng/test/tests/analysis/EditCBody/run-edit-c-body
```

Also see [revng-qa#37](https://github.com/revng/revng-qa/pull/37).

lib/EditCBody/EditCBodyAnalysis.cpp
aleclearmind: I ended up using `void __revng_retype_<index>(<type>);`, which should handle all cases nicely, arrays included.

See `__revng_retype_` in `Introduce the `edit-c-body` analysis`.

lib/EditCBody/EditCBodyAnalysis.cpp
aleclearmind: Unfortunately I can't see the original context of this, but if it's about `RETYPE`, I think this is no longer relevant, since we now use the same logic we use in `edit-c-type` (was: `import-from-c`).

lib/EditCBody/Statements.h
aleclearmind: Answering here for all goto-related comments: the clift backend *does not* currently emit `break` nor `continue`.

I left a TODO about supporting those once it does.

lib/PipeboxCommon/Python/_pipebox.cpp
mrjackv: https://rev.ng/gitlab/revng-private/revng-ui/-/merge_requests/58

lib/EditCBody/ClangParse.cpp
pfez: I agree that we don't want to handle that
I'm just wondering what's gonna happen if someone does.
And my understanding is that we'd just strip 2 of the 3 leading `/`s and leave the third in place, which isn't great but is probably fine.
This is worth adding a comment to the code, so that whenever we look back at this we'll now this was on purpose and not something that slipped through review.

lib/EditCBody/ClangParse.cpp
pfez: Ok for what concerns `startsOwnLine`. I still have to review rest of the machinery with the retype probes, but I'm closing this thread since they're not related to this.

lib/EditCBody/ClangParse.cpp
pfez: Nice! I'll leave this thread open until I'm done reviewing that as well.

lib/EditCBody/EditCBodyAnalysis.cpp
pfez: Fair enough. I had a scenario in mind, but now that I think about it only applies to comments, and for those there is no problem.

lib/EditCBody/Statements.h
pfez: Yeah the fact that the clift backend doesn't emit those yet it's a sad leftover. Ok for leaving this as a TODO. I'm going to close also the other related threads.

lib/EditCBody/ClangParse.cpp
pfez: Yes, I was thinking about providing information to the user as well.

I'm definitely biased but I think clang's AST is pretty readable. But if you don't like it the node kind is still better than nothing.

In the end, if the user hits this path I can only see 2 scenarios:
1. they've edited the C code as well, not just adding comments, which is not supported
2. revng itself generates C code for which we don't support properly reparsing

In neither case the user has something actionable to do with the error message: in the first case they should just stop editing C on their own, in the second case it's our fault.

aleclearmind: > I'm definitely biased but I think clang's AST is pretty readable.

There's pointers, we definitely don't want to send those to the user. It's also a security liability (could help bypassing ASLR).

I'm printing the AST node name right now, which is at least a little bit more informative.

lib/ClangToModel/CompileFlags.cpp
```diff
@@ -0,0 +1,80 @@
...
+  revng_assert(MaybeBuffer);
+
+  llvm::SmallVector<StringRef, 0> Lines;
+  MaybeBuffer->get()->getBuffer().split(Lines, '\n');
+  for (StringRef &Line : Lines)
+    if (Line.size() > 0 and Line[0] == '-')
+      Result.push_back(Line.str());
+
+  return Result;
+}
```
pfez: At this point, instead of reinventing the parsing of the configuration file, can't we just pass `--config=FilePath` to the underlying invocation of `clang`?

lib/EditCBody/ClangParse.cpp
aleclearmind: Added a comment.

lib/EditCBody/ClangParse.cpp
pfez: Pretty good. I'm closing this.

lib/ClangToModel/CompileFlags.cpp
aleclearmind: Good idea. Done.

lib/EditCBody/EditCBodyAnalysis.cpp
pfez: Nice! Closing this thread.

lib/EditCBody/ClangParse.cpp
pfez: Funny that you put the "we ignore doxygen comments" comment in a doxygen comment.

Anyway, good!

aleclearmind:
Merged in 4d7f0ac. Thanks.

PR #608, opened 2026-07-13
SVG: emit a *very* basic css with the graph

Cvelth:


lib/Yield/SVG.cpp
```diff
@@ -276,6 +276,71 @@ concept NodeExporter = requires(CallableType &&Callable, const NodeType &Node) {
   { Callable(Node) } -> std::convertible_to<std::string>;
 };
 
+constexpr llvm::StringRef getDefaultStyle() {
+  return "\n@import url('https://fonts.googleapis.com/css2?family="
+         "Source+Code+Pro&amp;display=swap');\n"
```
aleclearmind, line 281: Inline in the next line?

Cvelth: This is a workaround to our line length checker. This line is 90 characters long, so we need to somehow split it in the source without affecting the output.

Cvelth: If we only put the second half of it into the "raw" part of the literal, it looks even more awkward, IMO.

aleclearmind:
Merged in 5dbdfbf. Thanks.

PR #609, opened 2026-07-13
NameBuilder: allow known primitive aliases

Cvelth:


include/revng/ABI/abi-internal-schema.yml
```diff
@@ -102,9 +102,31 @@ definitions:
...
     key:
       - ABI
 
+  - name: PrimitiveAlias
+    type: struct
+    doc: |
+      Attaches a specific primitive type to an arbitrary name (e.g. `size_t`).
+    fields:
+      - name: Name
+        type: string
```
aleclearmind, line 198: I was more thinking about an `enum`, but after all this makes sense as well.

lib/Model/NameBuilder.cpp
```diff
@@ -597,6 +599,41 @@ model::CNameBuilder::isNameStaticallyReserved(llvm::StringRef Name) const {
...
+bool CNB::isKnownPrimitiveAlias(const model::TypedefDefinition &Typedef,
+                                llvm::StringRef Name) const {
+  revng_assert(!Name.empty());
+
+  // DANGER!!!! This is a cyclic dependency! We're in `libModel`, but
+  // this is using `libABI`.
+  //
+  // WIP: (IMPORTANT) I had to remove `-Wl,--no-undefined` from the linker
+  //       invocation locally to be able to test this out. We most likely have
+  //       to move `abi::Definition` into `revngModel` for this to work properly
```
aleclearmind: > I had to remove `-Wl,--no-undefined`

Illegal!

> We most likely have to move `abi::Definition` into `revngModel`

We could, but also `CNameBuilder` does not really belong to the model, I think.
This is C-specific, it could be closer to the backend, or in its own library.

tests/filecheck/clift/name-import/typedefs/known-primitive-aliases.mlir.yml
aleclearmind, line 1: Add a comment somewhere what you mean by "standard" vs "non-standard".

include/revng/ABI/abi-internal-schema.yml
Cvelth: I kind of also think this should be an enum. The reason I didn't do it is because string is faster to implement :P

But since you agree with me, let's go with a dedicated enum.

lib/Model/NameBuilder.cpp
Cvelth: 1. Yep, pretty much why I added a huge "DANGER!!!" as well as a "WIP" to make sure this won't get merged :P

2. That's, honestly, why I didn't move `abi::Definition` in yet. There are options. Do I still think moving `abi::Definition` in is the best course of action? Yes, we want that anyway - to let users define their custom ABIs in the model. Did I want to discuss it first? Also yes. One way or the other, it's your call whether I should move `Definition` or `NameBuilder`. We, most likely, want to move both eventually anyway.

tests/filecheck/clift/name-import/typedefs/known-primitive-aliases.mlir.yml
Cvelth: Actually, in hind-sight. These are awful names.

The tests meant to be a "`size_t` preserved" test (`standard.mlir`) and "`size_t` discarded" (`non-standard.mlir`") test.

Any ideas? You're better at naming stuff than I am :P

include/revng/ABI/abi-internal-schema.yml
aleclearmind: To be honest a string is fine. I'd just assert it's in a whitelist.

If you go with an enum, make sure I can write `size_t` and not `SIZE_T` or the like.

lib/Model/NameBuilder.cpp
aleclearmind: 2. Whatever takes less work now. Let's try to speedrun these small changes.

tests/filecheck/clift/name-import/typedefs/known-primitive-aliases.mlir.yml
aleclearmind: I'd do a single test, `size_t` preserved and `ptrdiff_t` discarded.
Call it `primitive-aliases.mlir` and add a comment about why the latter is not preserved.

include/revng/ABI/abi-internal-schema.yml
Cvelth: I ended up back-tracking on this and leaving the string in.

lib/Model/NameBuilder.cpp
Cvelth: Done.

tests/filecheck/clift/name-import/typedefs/known-primitive-aliases.mlir.yml
Cvelth: Done (`known-primitive-aliases.mlir`).

include/revng/Model/ABI/CType.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "revng/Model/ABI/Generated/Early/CType.h"
+
+#include "revng/Model/ABI/Generated/Late/CType.h"
```
aleclearmind, line 9: This has to be at the bottom of the file.

include/revng/Model/ABI/ScalarKind.h
```diff
@@ -0,0 +1,13 @@
...
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "revng/Model/ABI/Generated/Early/ScalarKind.h"
+
+#include "revng/Model/ABI/Generated/Late/ScalarKind.h"
+
+namespace model {
+namespace ScalarKind = abi::ScalarKind;
+} // namespace model
```
aleclearmind: Let's not move this into `model::` yet.

For good or for bad we don't have a strict relationship between libraries and namespaces.
Let's keep the namespace `abi::`, it will go into `model::` once we merge.

include/revng/Model/ABI/schema.yml
```diff
@@ -13,94 +13,166 @@ definitions:
     doc: for the documentation see `references/abi-definition.md`
     fields:
       - name: ABI
+        doc: The name of the ABI this definition describes.
         type: model::ABI::Values
 
       - name: ArgumentsArePositionBased
+        doc: |
+          See `revng/share/doc/revng/references/abi-definition.md` for details.
```
aleclearmind, line 21: Let's drop these and leave a TODO at the top of the file stating to move the "reference" part of `abi-definition.md` in here.

tests/filecheck/clift/name-import/header/known-primitive-aliases.mlir
aleclearmind, line 1: I think we can drop this file entirely and turn the model into the filecheck file.

```
RUN: revng2 ... artifact emit-type-and-global-header | FileCheck
```

Or you can use `pre-backend-clift` or something like that if you want to keep testing clift output.

I understand you want to test the individual pipe, but having two input files is just more maintenance burden for little return.

include/revng/Model/ABI/schema.yml
Cvelth: The reason I added these is because we made docs mandatory in the model. On everything. But duplicating stuff felt wrong, hence a comment like that.

tests/filecheck/clift/name-import/header/known-primitive-aliases.mlir
Cvelth: IMO it improves readability, but I don't have a strong stance, one file works just as well.

Cvelth: Actually, wait, no, on a fresh look, all the similar tests (`name-import` directory) have a separate model.

And there's no better place to put this check, IMO.

aleclearmind: Many of those tests likely predate clift emission of headers. All of those that do not check anything against a function body should be convert to have no MLIR input.

Leave a TODO about migrating them to check against `pre-backend-clift` without hardcoding MLIR.

Cvelth: No-no, `name-import` is mostly about types! *some* of them have bodies (i.e. labels and local variables), but most are about the type system!

aleclearmind: What do you mean?

What I'm saying is that, for instance, `enum.mlir` could easily be tested using a new artifact between `import-descriptive-info` and `emit-type-and-global-header`.

No MLIR input, just a model input + filecheck on MLIR.

> Many of those tests likely predate clift emission of headers

What I meant by this is that the test is not how I just described it because, before we switched to emit headers from clift, the only way to emit an `enum` in clift was to use it in the body of a function, but this is no longer the case.

Therefore `enum.mlir` (and probably others), which just tests if we applied the name `e_0` correctly, could be easily ported to what I described above.

Are we on the same page?
If so, leave a TODO about this and proceed as you did.

include/revng/Model/ABI/CType.h
Cvelth: This is not an issue anymore as the alias is gone.

include/revng/Model/ABI/ScalarKind.h
Cvelth: Done.

tests/filecheck/clift/name-import/header/known-primitive-aliases.mlir
Cvelth: I ended up splitting the two "kinds" of tests into `function-body` (those that cannot function without clift) and `header` (those that can be fully generated from the model).

I also left TODOs and a brief explanation of what's going on in the `model.yml` (the file that will be split into pieces). Please take a look.

---

I only see two potential issues with making these tests model-only:
1. We will be testing the entirety of the clift-header branch of the pipeline instead of the single pipe (and `@pfez` has always been adamant about tests covering as narrow of an area as they can (probably because of old backend PTSD :D)).
2. The fact that there isn't (correct me if it was introduced recently or something) a clean way to run `revng2 quick artifact` by passing *only* a model in. And doing non-trivial temporary directory manipulation in a `FileCheck` test doesn't sound particularly exciting to me.

That being said, none of that is a concern of this PR, let's get this in already!

> new artifact between `import-descriptive-info` and `emit-type-and-global-header`

Just for the context, if we ever refer to this thread again in the future, that artifact already exists for a while, it's called `type-and-global-header-clift`.

Cvelth:
@aleclearmind, pushed it, sorry about the delay. The only major changes are:
1. `Model: move ABIDefinition from abi::` dropped the namespace alias and adjusted all the corresponding users accordingly.
2. `pypeline: fix a mypy error` - I turned your patch into a commit.
3. two `name-import tests:` commits - regrouping and todos I mentioned in the related thread already.

aleclearmind:
Merged in 6a9f840. Thanks.

PR #610, opened 2026-07-20
VSCode Electron

mrjackv:


CMakeLists.txt
```diff
@@ -359,12 +359,12 @@ set(PYTHON_GENERATED_MODEL_PATH revng/model/_generated.py)
 #
 enable_testing()
 
-#
-# Find Python using the FindPython module. We request two components: *
-# Interpreter: needed for some variables, such as $Python_SITELIB * Development:
-# needed to compile nanobind module and friends, provides the
-# $Python_INCLUDE_DIRS and $Python_LIBRARIES variables
-#
+# ~~~
```
aleclearmind, line 362: 1. How did you come up with this trick?
2. Use 

```cmake
# cmake-format: off
# set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address -shared-libasan")
# cmake-format: on
```

lib/PipeboxCommon/Python/_pipebox.cpp
```diff
@@ -107,10 +107,14 @@ NB_MODULE(_pipebox, m) {
...
 
+    // This map stores the SIG_DFL and SIG_IGN for later
+    std::map<int, sighandler_t> EphemeralSavedSignals;
     // Save the signal pointers for later
     for (int SigNumber : Signals) {
       sighandler_t Handler = signal(SigNumber, SIG_DFL);
-      if (Handler != SIG_ERR && Handler != NULL)
+      if (Handler == SIG_DFL or Handler == SIG_IGN)
+        EphemeralSavedSignals[SigNumber] = Handler;
+      else if (Handler != SIG_ERR)
```
aleclearmind, line 120: The commit message describes the *what* but not the *why*.

What problem are you solving here?
I'm also not sure what's "ephemeral" about those signals.

python/revng/pypeline/storage/local_provider.py
```diff
@@ -278,7 +284,10 @@ def _create_provider(self, base_directory: Path, cache_dir: str, pipeline_hash:
         if self.inline:
             cache_path = (directory / ".cache").resolve()
             cache_path.mkdir(parents=True, exist_ok=True)
-            db_path = cache_path / "data.sqlite"
+            if self.multiproject and project_id is not None:
+                db_path = cache_path / f"{project_id}-data.sqlite"
```
aleclearmind, line 301: Mmh, are you sure we don't want to give an extra push and make that sqlite DB multi-project aware?

What was the ratio here?

python/revng/pypeline/storage/local_provider.py
```diff
@@ -251,10 +253,14 @@ def init(self, directory: Path):
...
 
-    def _create_provider(self, base_directory: Path, cache_dir: str, pipeline_hash: str):
+    def _create_provider(
+        self, project_id: str | None, base_directory: Path, cache_dir: str, pipeline_hash: str
+    ):
         # Figure out how the model should be name
         model_type = get_singleton(Model)  # type: ignore [type-abstract]
         model_name = model_type.model_name()
+        if self.multiproject and project_id is not None:
+            model_name = f"{project_id}-{model_name}"
```
aleclearmind: Would this be `1231-revng.yml`?

I'd prefer `revng-1231.yml`.

CMakeLists.txt
mrjackv: 1. It's the official way to avoid cmake-format from formatting a comment - [link](https://cmake-format.readthedocs.io/en/latest/configopts.html?highlight=preformat%20fences#fence-pattern)
2. re-reading the docs I think this is the proper way rather than what you propose

lib/PipeboxCommon/Python/_pipebox.cpp
mrjackv: Basically, before the commit we were storing `0x1` (`SIG_IGN`) in `SavedSignals` and then trying to call it, leading to a SIGSEGV. After looking at the code there were actually two issues:
1. We were not restoring `SIG_DFL` and `SIG_IGN` after running `RevngInit` even though the it was requested we did
2. We were storing `SIG_IGN` into SavedSignals to be later called, causing the SIGSEGV

python/revng/pypeline/storage/local_provider.py
mrjackv: > Would this be 1231-revng.yml?

Yes

> I'd prefer revng-1231.yml.

This would mean making assumptions about what `model_name()` returns, for all the pipeline know it could be `asdf龜亀` (without an extension)

CMakeLists.txt
aleclearmind: OK, can you fix `cmake-format: off` then?

lib/PipeboxCommon/Python/_pipebox.cpp
aleclearmind: Better explain this change in the commit message.

Instead of ephemeral we could say `NoNeedToWrapSignals`?

python/revng/pypeline/storage/local_provider.py
aleclearmind: 1. OK, but I still don't like this. Leave a TODO about having `model_name` return a name + extension pair.
2. Apart from the above, will this corrupt finding the model by recursively exploring parent directories?

mrjackv: 1. ACK
2. The `model_name` variable set there is the one used by the recursive model search, so it's guaranteed to be consistent with the existing expectations

CMakeLists.txt
mrjackv: Sure

lib/PipeboxCommon/Python/_pipebox.cpp
mrjackv: `SignalsWithPlaceholder`?

python/revng/pypeline/storage/local_provider.py
mrjackv: The main annoyance I see with it is that if you want to delete a single project's data you can't just delete the correct `.sqlite` file, modulo sprinkling the `ProjectID` column in all the tables.

lib/PipeboxCommon/Python/_pipebox.cpp
aleclearmind: `IgnoredOrDefaultSignals`

python/revng/pypeline/storage/local_provider.py
aleclearmind: I don't know, it seems cleaner to me. But let's not obsesses over this mostly debug-only feature.

Leave a TODO suggesting the possibility.

CMakeLists.txt
mrjackv: Done

lib/PipeboxCommon/Python/_pipebox.cpp
mrjackv: Done, also fixed the git commit message

python/revng/pypeline/storage/local_provider.py
mrjackv: 1. Found a solution, now it does `revng-1234.yml`

python/revng/pypeline/storage/local_provider.py
mrjackv: Done

python/revng/pypeline/storage/local_provider.py
aleclearmind: How?

mrjackv: `model_name.rsplit(".", 1)` with a few checks around

include/revng/PipeboxCommon/Model.h
```diff
@@ -177,7 +178,9 @@ class Model {
     return Out;
   }
 
-  static llvm::Expected<Model> deserialize(llvm::ArrayRef<uint8_t> Input) {
+  static llvm::Expected<Model>
+  deserialize(llvm::ArrayRef<uint8_t> Input,
+              std::optional<std::string> Path = std::nullopt) {
```
aleclearmind: Avoid default arguments in C++

python/revng/pypeline/cli/project/daemon.py
```diff
@@ -37,9 +65,30 @@ def run_daemon(ctx: ClickContext, production: bool):
         base_directory=ctx.obj.base_directory,
     )
 
+    hypercorn_config = ctx.obj.hypercorn_configuration
+    daemon_socket_path = None
+    socket_in_use = True
+    if hypercorn_config.bind is Config._bind:
```
aleclearmind, line 71: What does this line mean? Comment

python/revng/pypeline/cli/project/daemon.py
```diff
@@ -37,9 +65,30 @@ def run_daemon(ctx: ClickContext, production: bool):
...
+        daemon_socket_path = asyncio.run(daemon.socket_path())
```
aleclearmind: `asyncio.run` is a very bad pattern. I'd be prone to ban it entirely.

At very least, put a comment.

Can we have `storage_provider_factory` provide a non-async method to obtain this?

python/revng/pypeline/cli/project/daemon.py
```diff
@@ -37,9 +65,30 @@ def run_daemon(ctx: ClickContext, production: bool):
...
+        if daemon_socket_path is not None:
+            socket_in_use = unix_socket_has_server(daemon_socket_path)
+            hypercorn_config.bind = [f"unix:{daemon_socket_path}"]
```
aleclearmind: No escaping needed?

python/revng/pypeline/cli/project/daemon.py
```diff
@@ -13,11 +19,33 @@
...
+            sock.connect(path)
+            return True
+        except OSError:
+            return False
+
+
 @click.command(cls=WrappablePypeCommand)
 @hypercorn_command()
+@click.option(
+    "--socket-path",
```
aleclearmind: This really sounds like the user providing the path to the socket.

`--socket-location-file`? `--socket-path-file`?

Same with the argument name.

python/revng/pypeline/cli/project/daemon.py
```diff
@@ -37,9 +65,30 @@ def run_daemon(ctx: ClickContext, production: bool):
...
+        daemon_socket_path = asyncio.run(daemon.socket_path())
+        if daemon_socket_path is not None:
+            socket_in_use = unix_socket_has_server(daemon_socket_path)
+            hypercorn_config.bind = [f"unix:{daemon_socket_path}"]
+
+    if socket_path is not None:
+        assert len(hypercorn_config.bind) == 1 and hypercorn_config.bind[0].startswith("unix:")
+        bind_path = Path(hypercorn_config.bind[0].removeprefix("unix:")).resolve()
+        with open(socket_path, "w") as f:
+            f.write(str(bind_path))
```
aleclearmind: Extremely confusing given the name `socket_path`.

Comment.

python/revng/pypeline/cli/project/daemon.py
```diff
@@ -37,9 +65,30 @@ def run_daemon(ctx: ClickContext, production: bool):
         base_directory=ctx.obj.base_directory,
     )
 
+    hypercorn_config = ctx.obj.hypercorn_configuration
+    daemon_socket_path = None
```
aleclearmind, line 67: Reading this code I'm confused about:

1. What happens if the socket exists but there's no one on the other side.
   I'd expect to unlink and recreate it. Are we not doing it because we can just reuse it? If so, comment.
2. What happens if there `daemon_socket_path` stays `None`.
   What and where is defined the *other* default? Or are we erroring out?

python/revng/pypeline/cli/project/daemon.py
mrjackv: It means that the user hasn't specified `--bind` in the command line
Edit: added comment

python/revng/pypeline/cli/project/daemon.py
mrjackv: No, it's `storage_provider` that knows where the model is, and to get it you have to go through `async get`. Obviously I can add extra non-async methods, but for simplicity's sake I'd keep the `asyncio.run` in this instance.
Edit: added comment

python/revng/pypeline/cli/project/daemon.py
mrjackv: No, it's a plain path

python/revng/pypeline/cli/project/daemon.py
mrjackv: 1. `socket_in_use` will return `False`. I'll add code to take care of doing `os.unlink` in case it exists before running the server.
2. The default, dictated by hypercorn, is `127.0.0.1:8000`. If `daemon_socket_path` is `None` that one is used. Do note that this will never happen in normal CLI usage, it can only happen if someone specifies `--storage-provider-url` which is a strong indicator that it's doing developer stuff.

include/revng/PipeboxCommon/Model.h
mrjackv: Dropped

python/revng/pypeline/cli/project/daemon.py
mrjackv: Renamed to `--socket-location-file`

python/revng/pypeline/cli/project/daemon.py
mrjackv: Done

python/revng/pypeline/cli/project/daemon.py
mrjackv: 1. I've added the code, so if `socket_in_use` is false and the file exists it will be unlinked before running the server process

python/revng/pypeline/cli/project/daemon.py
aleclearmind: We're spinning up an event loop and fetching+parsing+discarding the model just to know where we should put the model, in case there's such a place. AFAIU we could even be fetching the model from remote.

Seems a bit overkill when we could just have `self.storage_provider_factory.model_path(self.base_directory)` since `StorageProviderFactory` is already storage-provider-specific.

Or am I missing something?

python/revng/pypeline/cli/project/daemon.py
aleclearmind: Sometimes these `unix:` paths have the option to encode additional information with some separator.

But if this is not the case, OK.

python/revng/pypeline/cli/project/daemon.py
aleclearmind: 1. > I'll add code to take care of doing os.unlink in case it exists before running the server.

   OK, nice.
2. > The default, dictated by hypercorn, is 127.0.0.1:8000.

   Please add a comment about this default we'll use in case our default is not appropriate.

python/revng/pypeline/cli/project/daemon.py
mrjackv: This is not the case here

python/revng/pypeline/cli/project/daemon.py
mrjackv: No, I've added `StorageProviderFactory.model_path`

python/revng/pypeline/cli/project/daemon.py
mrjackv: 2. Done

aleclearmind:
Merged in ad0f32e. Thanks.

PR #613, opened 2026-07-24
Adopt clang-format

aleclearmind:
Companion PRs:

* [llvm-project](https://github.com/revng/llvm-project/pull/53)


tests/filecheck/clift/backend/cast.mlir
```diff
@@ -34,7 +34,7 @@ module attributes {clift.module} {
       clift.yield %1 : !uint32_t
     }
 
-    // CHECK: (uint32_t) &var_0;
+    // CHECK: (uint32_t) & var_0;
```
pfez, line 37: 🤔 Personal taste but this is really not nice to read. I think the default should be the old behavior.

include/revng/PTML/CTokenEmitter.h
```diff
@@ -257,12 +257,8 @@ class CTokenEmitter {
...
     ~CommentEmitter();
 
-    [[nodiscard]] auto makeTagInitializer(llvm::StringRef Tag) {
-      return PTML.makeTagInitializer(Tag);
-    }
-
-    [[nodiscard]] PTMLTagEmitter initializeOpenTag(llvm::StringRef Tag) {
-      return PTML.initializeOpenTag(Tag);
+    [[nodiscard]] PTMLTagEmitter openTag(llvm::StringRef Tag) {
+      return PTML.openTag(Tag);
```
pfez: I presume this is what you mean in the commit message when you say "drop superfluous abstractions" (BTW there's a typo in the commit message: superflous). Can you elaborate on why you think they're now superfluous? I understand that there's never a situation where you want to keep `makeTagInitializer` separate from `initializeOpenTag` now, but are you sure we won't need it anymore?

Also, even if you collapse these two, I want you to keep the `initializeOpenTag` name, because `openTag` suggests that after calling this the tag is open, while it has only initialized the open tag, and you still have to call `finalizeOpenTag`. Only after `finalizeOpenTag` the tag is really open. So keeping the `initialize` vs `finalize` naming is much better I think.

include/revng/PTML/DoxygenEmitter.h
```diff
@@ -21,10 +21,7 @@ struct DoxygenCommentConfiguration {
 /// creation only C comments are emitted (using CDoxygenEmitter via
 /// CTokenEmitter::CommentEmitter), future use for assembly comments is planned.
 template<PTMLEmitter EmitterT>
-class DoxygenEmitter : IndentingEmitter<EmitterT> {
-private:
-  using BaseEmitter = IndentingEmitter<EmitterT>;
-
+class DoxygenEmitter : protected EmitterT {
```
pfez, line 24: I'm so happy that the indenting emitter is going away!

include/revng/PTML/PTMLEmitter.h
```diff
@@ -101,45 +65,39 @@ class PTMLTagEmitter {
...
-  /// Convenience function for PTMLTagEmitter initialization.
-  ///
-  /// \note It is important that this function remains a simple wrapper around
-  ///       the makeTagInitializer and does nothing more than that. This ensures
-  ///       that delayed PTMLTagEmitter initialization remains valid.
-  [[nodiscard]] PTMLTagEmitter initializeOpenTag(llvm::StringRef Tag) {
-    return makeTagInitializer(Tag);
+  /// Opens a PTML tag, returning the RAII PTMLTagEmitter guarding it.
+  [[nodiscard]] PTMLTagEmitter openTag(llvm::StringRef Tag) {
+    return PTMLTagEmitter(*this, Tag);
```
pfez: Same reasoning about the `initializeOpenTag` vs `openTag` here.

tests/filecheck/clift/backend/cast.mlir
aleclearmind: I investigated this. This happens only because this is statement expression that won't happen in real code.

include/revng/PTML/CTokenEmitter.h
aleclearmind: > superfluous abstractions

More like `IndentingEmitter`.

> I understand that there's never a situation where you want to keep makeTagInitializer separate from initializeOpenTag now, but are you sure we won't need it anymore?

Mmmh, dunno, I don't think these are so hard to re-introduce, in case.

> Also, even if you collapse these two, I want you to keep the initializeOpenTag name, because openTag suggests that after calling this the tag is open, while it has only initialized the open tag, and you still have to call finalizeOpenTag. Only after finalizeOpenTag the tag is really open. So keeping the initialize vs finalize naming is much better I think.

Will review this idea.

include/revng/PTML/DoxygenEmitter.h
aleclearmind: `\o/`

include/revng/PTML/PTMLEmitter.h
```diff
@@ -76,17 +38,19 @@ concept PTMLEmitter = //
...
 
 public:
-  explicit PTMLTagEmitter(detail::PTMLEmitterBase &ParentEmitter,
+  explicit PTMLTagEmitter(PTMLStreamEmitter &ParentEmitter,
                           llvm::StringRef Tag);
 
   PTMLTagEmitter(const PTMLTagEmitter &) = delete;
   PTMLTagEmitter &operator=(const PTMLTagEmitter &) = delete;
 
+  PTMLTagEmitter(PTMLTagEmitter &&Other);
```
pfez: Move constructor? Do we need it? And if so, why don't we have a move assignment as well?

lib/PTML/PTMLEmitter.cpp
```diff
@@ -71,52 +70,67 @@ emitEscaped(std::type_identity_t<EmitterT> &Emitter, llvm::StringRef String) {
...
+  ParentEmitter(Other.ParentEmitter),
+  Tag(Other.Tag),
+  IsEmittingOpenTag(Other.IsEmittingOpenTag) {
+  // Take over the parent's back-reference if this tag is the current open one.
+  if (ParentEmitter != nullptr
+      and ParentEmitter->CurrentOpenTagEmitter == &Other)
+    ParentEmitter->CurrentOpenTagEmitter = this;
+
+  // Leave the moved-from emitter inert, so its destructor does nothing.
+  Other.ParentEmitter = nullptr;
```
pfez: For good measure, I'd clear the `Tag` and the `IsEmittingOpenTag` as well.

lib/CliftEmitC/CEmissionPasses.cpp
```diff
@@ -49,10 +49,12 @@ struct CEmissionPass : BaseT<CEmissionPass<BaseT, Impl>> {
     mlir::ModuleOp Module = Base::getOperation();
 
     auto Tagging = static_cast<ptml::Tagging>(Base::EmitTags.getValue());
-    ptml::CTokenEmitter Emitter(File->os(), Tagging);
+    ptml::CTokenEmitter Emitter(Tagging);
 
     if (not Impl(Module, Emitter))
       return Base::signalPassFailure();
+
+    File->os() << Emitter.extract();
```
pfez, line 57: 🤔 is this for preventing any write to `File->os()` in case of errors?

aleclearmind: This is... to write the result in the stream.

Please be more specific.

lib/PTML/CTokenEmitter.cpp
```diff
@@ -316,6 +318,50 @@ static StringEscape getStringEscape(char Character, char NextCharacter) {
...
+  if (DisableCReformatting)
+    return Buffer;
+
+  if (Tags == Tagging::Enabled) {
+    // The document is PTML: reformat it using the metadata built while
+    // emitting, so its tags are preserved without re-lexing it.
+    return PTML.metadata().reformat(Buffer);
+  }
+
+  return ptml::reformatC(Buffer);
```
pfez, line 356: Make sure you always also clean the `Buffer` before returning.

lib/PTML/CTokenEmitter.cpp
```diff
@@ -316,6 +318,50 @@ static StringEscape getStringEscape(char Character, char NextCharacter) {
 
 //===---------------------------- CTokenEmitter ---------------------------===//
 
+static ptml::MetadataBuilding shouldBuildMetadata(ptml::Tagging Tags) {
+  // Metadata only exists to reposition whitespace in the tagged document; the
+  // plain-C path reformats without it.
+  return Tags == ptml::Tagging::Enabled ? ptml::MetadataBuilding::Enabled :
+                                          ptml::MetadataBuilding::Disabled;
+}
```
pfez: I'd put this as `inline` function in `PTMLEmitter.h`.

lib/PTML/CTokenEmitter.cpp
pfez: And possibly reinitialize `BufferStream` as well.

include/revng/PTMLCFormat/Format.h
```diff
@@ -0,0 +1,67 @@
...
+
+/// Anchors a run of source characters that map one-to-one onto consecutive PTML
+/// bytes: the character at SourceOffset sits at PTMLOffset, and every following
+/// character advances both offsets by one, until the next anchor. Anchors are
+/// recorded only where this correspondence breaks (a tag, or an escaped
+/// character), so there is no entry per character.
+struct PTMLSourceAnchor {
+  size_t SourceOffset = 0;
+  size_t PTMLOffset = 0;
+};
```
pfez: I find this confusing for a few reasons:
1. the comment talks about "next anchor", but it's not clear what the next anchor is, here we're only looking at a single anchor
2. I don't understand what's an "anchor" in the void. it seems a mapping from an offset in the "source file" (I presume this is a file that comes from PTML after stripping) and a PTML file. Please clarify this, and perhaps come up with a better name. I can't suggest a better name off the top of my head, because the exact meaning of this is still foggy to me

include/revng/PTMLCFormat/Format.h
```diff
@@ -0,0 +1,67 @@
...
+#include <string>
+#include <vector>
+
+#include "llvm/ADT/StringRef.h"
+
+namespace ptml {
+
+/// A single element tag of a PTML document, identified by its byte range in it.
+/// IsClosing distinguishes end tags (`</...>`) from start tags.
+struct PTMLMarkup {
```
pfez: `PTMLMarkupRange` perhaps?

include/revng/PTMLCFormat/Format.h
pfez: Answers to these questions become clearer reading the comments of the rest of the file, but still the "anchor" name remains foggy to me. It seems more a "mapping" or something.

lib/PTMLCFormat/Format.cpp
```diff
@@ -0,0 +1,265 @@
...
+  for (const clang::tooling::Replacement &Replacement : Replacements) {
+    size_t Offset = Replacement.getOffset(); // -> 3
+    size_t Length = Replacement.getLength(); // -> 2
+
+    // Map the replaced Source range onto a PTML byte range. Whitespace is never
+    // escaped nor placed inside a tag, so each replaced byte maps one-to-one:
+    // Begin is the PTML offset of the first replaced byte, End one past the
+    // last. In between lie only the replaced whitespace bytes and whole tags.
+    size_t Begin = ptmlOffset(Offset); // -> 8
+    size_t End = Length == 0 ? Begin : ptmlOffset(Offset + Length - 1) + 1;
```
pfez, line 221: Do you really need the ternary? Wouldn't the following work as well?
```c
size_t End = ptmlOffset(Offset + Length - 1) + 1;
```

lib/PTMLCFormat/Format.cpp
```diff
@@ -0,0 +1,265 @@
...
+    // Nudge the range past any tags flush against it, so the new whitespace
+    // lands between elements. Both are no-ops here: no tag ends at 8, and the
+    // tag starting at 16 is opening, not closing.
+    Begin = expandLeftPastOpeningTags(Begin, EndToMarkup); // -> 8
+    End = expandRightPastClosingTags(End, StartToMarkup); // -> 16
+
+    // Whitespace-only edits never overlap: mapped to PTML they stay ordered,
+    // and the nudges above only ever meet flush ranges, never cross them, so
+    // this edit always begins at or after the previous one ended.
+    revng_assert(Begin >= Position);
```
pfez, line 234: It's probably sane to also `revng_assert` that `Replacement.getReplacementText` only contains whitespace.

lib/PTMLCFormat/Format.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+} // namespace
+
+namespace ptml {
+
+size_t PTMLMetadata::ptmlOffset(size_t SourceOffset) const {
+  // SourceMap is sorted by SourceOffset and its first anchor is always at
+  // offset 0, so there is an anchor at or before any queried offset. Find the
+  // last such anchor and extend its one-to-one run up to the query.
+  auto Iterator = std::upper_bound(SourceMap.begin(),
```
pfez: You've got `llvm::upper_bound`.

include/revng/PTMLCFormat/Format.h
```diff
@@ -0,0 +1,67 @@
...
+  size_t PTMLOffset = 0;
+};
+
+/// Describes a PTML document in terms of the C source underlying it and how it
+/// maps back onto the markup. This is everything reformat needs to map the
+/// edits produced on the recovered source back onto the tagged document.
+///
+/// It is built by the emitter as it writes the document, so reformatting never
+/// has to parse the PTML back.
+struct PTMLMetadata {
```
pfez: "Metadata" is a loaded term. What kind of metadata is this? A few ideas: `PTMLToSourceMetadata`, `PTMLReformatMetadata`.

The second makes it less surprising that we have a `reformat` method.

include/revng/PTML/PTMLEmitter.h
```diff
@@ -81,16 +85,37 @@ concept PTMLEmitter = //
...
-    StreamEmitter(OS), EmitTags(Tags == Tagging::Enabled) {}
+  explicit PTMLStreamEmitter(llvm::raw_ostream &OS,
+                             Tagging Tags,
+                             MetadataBuilding BuildMetadata =
+                               MetadataBuilding::Disabled) :
+    StreamEmitter(OS),
+    EmitTags(Tags == Tagging::Enabled),
+    BuildMetadata(BuildMetadata == MetadataBuilding::Enabled) {
+    revng_assert(EmitTags or not this->BuildMetadata,
+                 "Building metadata requires emitting tags.");
```
pfez: This smells like the two `enum`s should be merged.
You have something like a `enum TaggingPolicy` with three values: `NoTagging`, `PlainTagging`, `TaggingWithMetadata`.

lib/PTML/PTMLEmitter.cpp
```diff
@@ -101,8 +104,15 @@ PTMLTagEmitter::~PTMLTagEmitter() {
...
 
+  size_t EndTagBegin = ParentEmitter->BuildMetadata ? ParentEmitter->OS.tell() :
+                                                      0;
+
   if (ParentEmitter->EmitTags)
     ParentEmitter->OS << '<' << '/' << Tag << '>';
+
+  if (ParentEmitter->BuildMetadata) {
+    ParentEmitter->recordCloseTag(EndTagBegin, ParentEmitter->OS.tell());
+  }
```
pfez: I'd push the check down inside the body of `recordCloseTag`, or assert in `recordCloseTag` that `BuildMetadata` is on. The latter makes the code less brittle, the former makes it more readable and always correct.

Same logic for `recordOpenTag`.

pfez: The former, pushing down the checks to `recordCloseTag` also enable that method to take fewer arguments: `EndTagBegin` can be computed inside its body, and `Parentemitter->OS.tell()` as well.

Same holds for `recordOpenTag`.

include/revng/PTMLCFormat/Format.h
```diff
@@ -0,0 +1,67 @@
...
+  /// The document's tags, in document order.
+  std::vector<PTMLMarkup> Markup;
+
+  /// Reformats a PTML document with clang-format, preserving all of its tags.
+  /// PTML must be the document this metadata was built from: the recovered
+  /// source is formatted and the resulting whitespace-only edits are mapped
+  /// back onto the tags. E.g. "<span>foo</span> <span>(</span>", whose
+  /// recovered source "foo (" clang-format rewrites to "foo(", becomes
+  /// "<span>foo</span><span>(</span>".
+  std::string reformat(llvm::StringRef PTML) const;
```
pfez, line 59: It's sad that this is C specific, while all the rest is source-language agnostic. The metadata should be decoupled from the reformatting. And the source-agnostic metadata can be put in `include/revng/PTML`. This can turn out to be useful if we ever want to automatically reformat different outputs than just C.

pfez: If you strip this out, then the obvious suggestion for how to rename `PTMLMetadata` becomes `PTMLToSourceMetadata`, since the formatting has been moved elsewhere.

lib/CliftEmitC/CEmissionPasses.cpp
pfez: Yeah, I didn't have full undersanding of the rest of the patch. Makes sense now.

include/revng/PTMLCFormat/Format.h
aleclearmind: I don't find anchor such a bad name. An "anchor" in HTML parlance is something you can jump to via a URL fragment `#anchor-name`.

This is not exactly that, but it recalls it.

If we call it mapping it sounds weird that you go from one "mapping" to the next "mapping". An anchor is a sort of invisible delimiter, placeholder, bookmark that maps plain text to PTML.

Any better names?

include/revng/PTMLCFormat/Format.h
aleclearmind: Mmmh, that doesn't capture `IsClosing`.

`PTMLTag`?

include/revng/PTMLCFormat/Format.h
aleclearmind: I'll go with `PTMLCReformattableDocument`

include/revng/PTMLCFormat/Format.h
aleclearmind: > It's sad that this is C specific

I wouldn't overengineer this.
The only scope of this thing is C as of now and for the foreseeable future.
The library name is PTML*C*Format.

include/revng/PTMLCFormat/Format.h
pfez: `PTMLTagRange`?

include/revng/PTML/CTokenEmitter.h
aleclearmind: After reviewing the changes, I decided to switch back to how things were (i.e., preserving `makeTagInitializer` and others).

I still have `extract`, which can be called only once (it asserts otherwise).
`CTokenEmitter` was not "reusable" before, making it reusable requires you e.g., to re-initialize the `MainTag`, which is some suffering we can avoid.

include/revng/PTML/PTMLEmitter.h
aleclearmind: I kept the old naming.

include/revng/PTML/PTMLEmitter.h
aleclearmind: This is gone.

lib/PTML/PTMLEmitter.cpp
aleclearmind: Gone.

lib/PTML/CTokenEmitter.cpp
aleclearmind: I don't want to make this reusable. I'm now asserting you must not call `extract` twice.

Let's continue discussion [on this thread](https://github.com/revng/revng/pull/613#discussion_r3646280669).

lib/PTML/CTokenEmitter.cpp
aleclearmind: Done

lib/PTMLCFormat/Format.cpp
aleclearmind: Pure insertion has length 0, you might call `ptmlOffset(-1)`, not nice.

lib/PTMLCFormat/Format.cpp
aleclearmind: Done

lib/PTMLCFormat/Format.cpp
aleclearmind: Done

include/revng/PTMLCFormat/Format.h
aleclearmind: Done

include/revng/PTMLCFormat/Format.h
aleclearmind: Done

include/revng/PTML/PTMLEmitter.h
aleclearmind: Done

lib/PTML/PTMLEmitter.cpp
aleclearmind: Done

include/revng/PTMLCFormat/Format.h
pfez: Yeah, I meant that a part of it could stay out of `PTMLCFormat` and live in plain `PTML`. Anyway, this is the only output format for which we do auto-formatting for now, so there's no need to separate the two. We'll consider that if we have the need to format other things in the future.

include/revng/PTML/PTMLEmitter.h
pfez: Nice, but you still have `ptml::Tagging`. I think the two can be conflated everywhere. But perhaps that's a bit too extreme. There are places where you never want to emit markup and those still work as well with simply `ptml::Tagging`.

include/revng/PTML/CTokenEmitter.h
pfez: Ok I guess. But now it can go out of scope without anyone `extract()`ing anything. I would make a destructor to check for that.

lib/PTML/CTokenEmitter.cpp
pfez: Ok but I'll wait to close this thread until the other is ready to be closed as well.

include/revng/PTML/PTMLEmitter.h
aleclearmind: This would imply exposing to the `CTokenEmitter` constructor whether to build metadata or not.

But `CTokenEmitter` must handle it on its own since it unconditionally reformats the code and needs to decide whether to build the metadata or not based on whether you want PTML or plain text.

include/revng/PTMLCFormat/Format.h
pfez: Yeah, "mapping" is not great either. Naming is hard.
With the HTML explanation it makes kinda sense but it's still a bit off.
Perhaps something very simple like `OffsetPair` with two fields: `PTML` and `Source`. It gets straight to the point without relying on the name vaguely referring to HTML.

pfez: `PTMLSourceTie`

pfez: I don't consider this a blocker though.

tests/filecheck/clift/backend/cast.mlir
pfez: This doesn't clarify. What guarantees do we have it doesn't happen?

include/revng/PTML/CTokenEmitter.h
aleclearmind: I wanted to enforce that, but there are places where there's some non-local failure decision being taken and therefore you don't want to extract, specifically:

```c++
    auto Tagging = static_cast<ptml::Tagging>(Base::EmitTags.getValue());
    ptml::CTokenEmitter Emitter(Tagging);

    if (not Impl(Module, Emitter)) {
      // --> WE SHOULD FORCIBLY extract HERE IN ORDER NOT TO ABORT <--
      return Base::signalPassFailure();
    }

    File->os() << Emitter.extract();
```

include/revng/PTMLCFormat/Format.h
aleclearmind: > Perhaps something very simple like OffsetPair

Well it's not incorrect but you erased the additional semantics of the data structure which enables to articulate comments such as the one above, which seems useful to me.

`PTMLSyncPoint`?

pfez: The original semantic is in the `SourceMap` not in the pair of offset themselves.

`PTMLSourceSyncPoint`?

include/revng/PTML/CTokenEmitter.h
pfez: I see. No further comments your honor.

tests/filecheck/clift/backend/cast.mlir
aleclearmind:
```
$ cat /tmp/asd.c
```
```c
void f(void) {
  (uint32_t)&var_0;
  (my_struct_t)&var_0;
  (unsigned)&var_0;
  (int)&var_0;
  g((uint32_t)&var_0);
  return (uint32_t)&var_0;
  int y = (uint32_t)&var_0;
}
```
```
$ clang-format < /tmp/asd.c
```
```c
void f(void) {
  (uint32_t) & var_0;
  (my_struct_t) & var_0;
  (unsigned)&var_0;
  (int)&var_0;
  g((uint32_t)&var_0);
  return (uint32_t)&var_0;
  int y = (uint32_t)&var_0;
}
```

I'd prefer not to fix `clang-format` bugs on things we shouldn't emit in the first place (a statement that's just a cast of a `&` expression).

include/revng/PTMLCFormat/Format.h
pfez: Or even just `OffsetSyncPoint`, with two fields `PTML` and `Source`.

tests/filecheck/clift/backend/cast.mlir
pfez: Oh wow it only happens in the void. 

```c
int y = (uint32_t)&var_0;
```

I think this is something we can totally end up emitting if the stack address is taken for some reason and `var_0` is the stack frame variable.
But this is one of the cases where it would be fine.

Ok then :shrug:

aleclearmind: > we can totally end up emitting if the stack address is taken for some reason

I think we'd emit something like `escape(&stack)`. Anyway, just `&var_0;` and `(void)&var_0;` work.
`clang-format` thinks `(uint32_t)` is an expression and `&` is a bitwise and.

include/revng/PTMLCFormat/Format.h
aleclearmind: Went for `OffsetSyncPoint`.

pfez:
Merged in 8c6b56d24d7c10e37124545c2fce3e297f4047d0

PR #614, opened 2026-07-24
Inline stack frame

Cvelth:


include/revng/CliftEmitC/Configuration.h
```diff
@@ -31,4 +31,12 @@ struct TypeEmitterConfiguration {
...
   /// is not affected because it explicitly handles `_STARTS_AT` attribute).
   bool ExplicitPadding = true;
+
+  /// When emit the stack variable, emit the *definition* of its struct instead
+  /// of a declaration.
+  ///
+  /// Note that the inlined definition is emitted regardless of the presence of
+  /// the global definition for the same type, which means the produced C is
+  /// intentionally not syntactically valid as a single translation unit.
+  bool InlineStackFrame = false;
```
pfez: `InlineStackFrameType` please

include/revng/Clift/Clift.td
```diff
@@ -547,7 +547,8 @@ def Clift_LocalVariableOp
                                                       ["getBlockArgumentVariable"]>]> {
 
   let arguments = (ins DefaultValuedStrAttr<StrAttr, "">:$name,
-                       DefaultValuedStrAttr<StrAttr, "">:$handle);
+                       DefaultValuedStrAttr<StrAttr, "">:$handle,
+                       DefaultValuedAttr<BoolAttr, "false">:$stack_frame);
```
pfez: We said to make this a discardable attribute.

pfez: Also we should add to model-aware clift verification a check that there is at most one local variable with this attribute in every function.

Cvelth: 1. My bad.
2. Will do!

lib/CliftEmitC/CBackend.cpp
```diff
@@ -1176,6 +1176,19 @@ class CliftToCEmitter : CEmitter {
...
+    // NOTE: placing this here and not inside `emitLocalVariableDeclaration` to
+    //       make sure it is emitted *before* the comment attached to
+    //       the variable.
+    if (auto S = mlir::dyn_cast<LocalVariableOp>(Op)) {
+      if (Configuration.InlineStackFrame and S.getStackFrame()) {
+        auto Class = mlir::dyn_cast<ClassType>(S.getResult().getType());
+        revng_assert(Class and Class.getHandle() != Configuration.TypeToOmit);
+
+        TypeDefinitionEmitter Emitter(Tokens, DataModel, Configuration);
+        Emitter.emitTypeDefinition(Class);
```
pfez: Should we emit type declaration as well?
I'm invoking @aleclearmind on this: do you want A or B?

```c
// A
typedef struct stack_type stack_type; // declaration, currently skipped
stuct stack_type {
  int32_t stack_slot;
};
stack_type stack;

// B - this is the current behavior implemented in this PR
stuct stack_type {
  int32_t stack_slot;
};
stack_type stack;
// ^ this technically does not compile because there is no typedef for stack_type in this version
//   but who cares, when you inline the stack frame type you're giving up on recompilability
```

lib/CliftEmitC/CEmissionPasses.cpp
pfez, line 93: I want to see `InlineStackFrame` initialized explicitly here as well.

tests/filecheck/clift/backend/stack.mlir
```diff
@@ -0,0 +1,190 @@
...
+  // INLINE:       /// Second occurrence of e inside the same union references the already-inlined enum by name.
+  // INLINE:       e choice_again;
+  // INLINE:       /// e_t is a typedef of y, but when the enum is inlined, we silently ignore typedefs, printing the real enum name.
+  // INLINE:       e choice_t;
+  // INLINE:     };
+  // INLINE:     uint8_t padding_at_80[8];
+  // INLINE:     /// Second occurrence of z must reuse the first inline definition by name.
+  // INLINE:     z another_z _STARTS_AT(88);
+  // INLINE:   };
+  // INLINE:   my_stack var_1;
```
pfez: Nice that you have a complex test. But please make a few simpler tests, that incrementally build up complexity until reaching this.

lib/CliftEmitC/TypeDefinitionEmitter.cpp
```diff
@@ -188,11 +254,25 @@ void TypeDefinitionEmitter::emitClassDefinition(clift::ClassType Class) {
...
+
+          const auto &ToInline = InlineTarget->DefinitionToInline;
+          auto Defined = mlir::cast<clift::DefinedType>(ToInline);
+          if (markInlinedIfNotAlready(Defined)) {
+            emitTypeDefinition(Defined);
+            continue;
+          }
+        }
+      }
+
```
pfez: 🤔 note to self. It seems very brittle to put this here. The `TypeDefinitionEmitter` is instantiated in many places where we've decided we'll never emit types inline (like the `type-and-global-header`). I'd like to have this logic in a separate kind of type emitter, but that isn't really feasible without duplicating almost all the current `TypeDefinitionEmitter`, which is a showstopper. On the other hand having it guarded under the `InlineStackFrame` configuration flag is not terrible. I'll think about this to see if I have a solution.

include/revng/CliftEmitC/TypeDefinitionEmitter.h
```diff
@@ -16,6 +16,17 @@ class TypeDefinitionEmitter : public CEmitter {
 private:
   TypeEmitterConfiguration Configuration;
 
+  /// Set of type handles whose definition has already been inlined into the
+  /// current stack-frame local.
+  struct StringRefHash {
+    size_t operator()(llvm::StringRef S) const { return llvm::hash_value(S); }
+  };
+  std::unordered_set<llvm::StringRef, StringRefHash> InlinedTypes;
+  bool markInlinedIfNotAlready(clift::DefinedType Type) {
```
pfez: Drop `IfNotAlready` from the name please.

lib/CliftEmitC/TypeDefinitionEmitter.cpp
```diff
@@ -148,6 +149,71 @@ void TypeDefinitionEmitter::emitPaddingField(clift::ClassType Class,
...
+  while (true) {
+    // Preserve constness.
+    IsConst = IsConst || clift::isConst(Type);
+    Type = clift::removeConst(Type);
+
+    // Typedefs are discarded though (constness is preserved above).
+    if (auto Typedef = mlir::dyn_cast<clift::TypedefType>(Type)) {
+      Type = Typedef.getUnderlyingType();
+      continue;
+    }
```
pfez: !!!

Wait, we've never agreed on this.

What I had in mind was to peek through typedefs, and if the transitive underlying type is also an aggregate, then you can emit the `typedef` inline, but without discarding the `typedef` entirely. Remember that typedefs have strong identities in the model and in clift. We can't just ignore them.

pfez: There's other comments I have on this function, but this alone implies quite some restructuring, so I'll keep them for myself until we've settled this.

lib/CliftEmitC/CBackend.cpp
aleclearmind: I my mind it was just:

```c
struct {
  int32_t stack_slot;
} stack;
```

I think the guideline should be: the code must compile, *except if* there are references to the type of the stack frame anywhere other than `model::Function::StackFrameType`.
If this doesn't work, we can refine, but I think this is the direction.

Other than that, whatever is the most compact.

lib/CliftEmitC/TypeDefinitionEmitter.cpp
Cvelth: No, this is not much of a restructuring! Note that this only affect the `Declaration` part of it. I kind of expected this reaction - hence why `Definition` and `Declaration` parts are fully separate.

As such, please flush your other comments, in case those have a wider impact.

Cvelth: And since you want to emit typedefs, it's a matter of treating them similarly to how the array are currently treated.

include/revng/CliftEmitC/Configuration.h
Cvelth: Done.

include/revng/Clift/Clift.td
Cvelth: Done and done.

lib/CliftEmitC/CBackend.cpp
Cvelth: Done.

lib/CliftEmitC/CEmissionPasses.cpp
Cvelth: Done.

tests/filecheck/clift/backend/stack.mlir
Cvelth: Done, please take a fresh look.

lib/CliftEmitC/TypeDefinitionEmitter.cpp
Cvelth: This is not relevant anymore, closing the thread.

lib/CliftEmitC/TypeDefinitionEmitter.cpp
Cvelth: This is not relevant anymore, since we no longer recur, but I'm still curious what other comments you had!

If you have them noted somewhere, I'd appreciate if you flushed them.

Obviously if you still need to spell them out AND they are not relevant anymore, no need to spend time on this, it's mostly just my curiosity.

include/revng/CliftEmitC/TypeDefinitionEmitter.h
Cvelth: This is not relevant anymore, closing the thread.

lib/CliftEmitC/TypeDefinitionEmitter.cpp
```diff
@@ -340,28 +341,27 @@ void TypeDefinitionEmitter::emitEnumDefinition(clift::EnumType Enum) {
       }
     }
   }
-
-  Tokens.emitPunctuator(ptml::CTokenEmitter::Punctuator::Semicolon);
-  Tokens.emitNewline();
```
pfez, line 345: The commit message says "suppress `;\n` coming from `emitClassDefinition`." but here the offending semicolon was in `emitEnumDefinition`, and the one in `emitClassDefinition` is still there. How so?

pfez: Also, split the commit in 2: one that makes the methods private, one that fixes the semicolon.

lib/CliftEmitC/TypeDefinitionEmitter.cpp
```diff
@@ -249,8 +250,18 @@ void TypeDefinitionEmitter::emitClassDefinition(clift::ClassType Class) {
       emitPaddingField(Class, PreviousOffset, Class.getObjectSize());
   }
 
-  Tokens.emitPunctuator(ptml::CTokenEmitter::Punctuator::Semicolon);
-  Tokens.emitNewline();
```
pfez, line 252: And the semicolon from `emitClassDefinition` is removed here, in another commit. Please fix.

lib/CliftEmitC/TypeDefinitionEmitter.cpp
Cvelth: 1. Whoops, looks like a mess up when reordering the commits. Good catch!
2. Will do!

include/revng/CliftEmitC/TypeDefinitionEmitter.h
```diff
@@ -39,12 +39,21 @@ class TypeDefinitionEmitter : public CEmitter {
...
   /// The trailing `;` is NOT emitted here; it is the caller's responsibility.
-  void emitClassDefinition(clift::ClassType StructOrUnion);
+  ///
+  /// Note that this method is only public for a specific user (and we cannot
+  /// easily expose it only to a "child" emitter because `CTokenEmitter` uses
+  /// composition). Treat it as if it was protected and use `emitTypeDefinition`
+  /// in the general case.
+  void
+  emitClassDefinition(clift::ClassType StructOrUnion,
+                      std::optional<DeclaratorInfo> Declarator = std::nullopt);
```
pfez: Having this `public` is a time bomb before someone misuses it.
Leave this `private`.
Make a `static` member function `emitInlineClassDefinition`, with a non-nullable `Declarator`. And inside the body of that you can instantiate a new standalone `TypeDefinitionEmitter`, and call this.
If you do that, you don't even need to add the `InlineStackFrameType` to the type emitter configuration, becuse the new `static` member function always does its thing properly. You only have to thread it down the C backend so it can invoke `emitInlineClassDefinition`, but you hardly need the full `TypeEmitterConfiguration` there. Only the boolean `InlineStackFrameType`.

lib/CliftEmitC/CBackend.cpp
```diff
@@ -56,7 +58,17 @@ class CliftToCEmitter : CEmitter {
   // Ambient precedence of the current expression.
   OperatorPrecedence CurrentPrecedence = {};
 
+  // Configuration controlling optional emission behaviour for this function,
+  // most notably whether the stack-frame struct definition should be inlined
+  // at the top of the function body.
+  TypeEmitterConfiguration Configuration;
```
pfez: See my other thread on `emitClassDefinition` for this, but basically you don't need to thread all or this up to here. The `InlineStackFrameType` boolean is enough. Let's restrict ourselves to things we actually use.

Cvelth: Wait-wait-wait! No! Passing the full configuration is the correct thing to do! Note that we are now emitting struct definitions! Which relies on the said configuration! Currently only `ExplicitPadding` is relevant, but I'd assume there will be more in the future!

lib/CliftEmitC/Headers.cpp
```diff
@@ -291,6 +291,7 @@ void emitHelperHeader(ptml::CTokenEmitter &Tokens,
   TypeEmitterConfiguration Configuration = {
     .EmitMaximumEnumValue = true,
     .ExplicitPadding = true,
+    .InlineStackFrameType = false,
```
pfez: If we do the changes I've suggested on `emitInlineClassDefinition`, `InlineStackFrameType` never becomes a parameter of the type definition emitter, it's only a parameter that can be activated calling that `static` member function, on individual calls, which is the strict subset of funcitonality we need in the C backend. If you do that, you don't need any of the changes in this commit, because there's simply no way to set the type configuration emitters wrong.

lib/CliftEmitC/CBackend.cpp
Cvelth: P. S. I can see the world where `InlineStackFrameType` is *not* a member of `TypeEmitterConfiguration` (now that it's not recursive anymore), but the configuration itself has to stay!

include/revng/CliftEmitC/TypeDefinitionEmitter.h
Cvelth: Most of this makes sense. The only doubt I have is about the static method.

What is the advantage of 
```cpp
TypeDefinitionEmitter::emitInlineClassDefinition(whatever)
```
over
```cpp
TypeDefinitionEmitter Emitter(config_and_whatever);
Emitter.emitInlineClassDefinition(whatever);
```
?

Cvelth: If anything, I'd get rid of `InlineStackFrameType` boolean in the first place! And just do the right thing (tm) inside `emitInlineClassDefinition` only!

Cvelth: At `TypeDefinitionEmitter` level I mean. Obviously `CBackend` needs it.

include/revng/CliftEmitC/TypeDefinitionEmitter.h
```diff
@@ -40,10 +40,18 @@ class TypeDefinitionEmitter : public CEmitter {
...
+  /// Emit the full definition of a class type, optionally binding it to
+  /// a single variable declarator (`struct foo { ... } var_1`).
   ///
   /// The trailing `;` is NOT emitted here; it is the caller's responsibility.
-  void emitClassDefinition(clift::ClassType StructOrUnion);
+  ///
+  /// Note that this method is only public for a specific user (and we cannot
+  /// easily expose it only to a "child" emitter because `CTokenEmitter` uses
+  /// composition). Treat it as if it was protected and use `emitTypeDefinition`
+  /// in the general case.
```
pfez: This comment is outdated

Cvelth: Good catch!

lib/CliftEmitC/TypeDefinitionEmitter.cpp
```diff
@@ -148,7 +148,9 @@ void TypeDefinitionEmitter::emitPaddingField(clift::ClassType Class,
   Tokens.emitNewline();
 }
 
-void TypeDefinitionEmitter::emitClassDefinition(clift::ClassType Class) {
+void TypeDefinitionEmitter::emitClassDefinition(clift::ClassType Class,
+                                                std::optional<DeclaratorInfo>
+                                                  Declarator) {
```
pfez: Drop the `Declarator` argument from here, and only put it in `emitInlineClassDefinition`

lib/CliftEmitC/TypeDefinitionEmitter.cpp
```diff
@@ -247,6 +249,19 @@ void TypeDefinitionEmitter::emitClassDefinition(clift::ClassType Class) {
...
+    Tokens.emitIdentifier(Declarator->Identifier,
+                          Declarator->Location,
+                          Declarator->Kind,
+                          ptml::CTokenEmitter::IdentifierKind::Definition);
+    if (Declarator->CAttributes) {
+      emitCAttributes(Declarator->CAttributes,
+                      /* SpaceBefore = */ true,
+                      /* SpaceAfter = */ false);
+    }
+  }
```
pfez: This should live in `emitInlineClassDefinition`

include/revng/CliftEmitC/CBackend.h
```diff
@@ -8,6 +8,15 @@
...
 #include "revng/Clift/Clift.h"
+#include "revng/CliftEmitC/Configuration.h"
 #include "revng/PTML/CTokenEmitter.h"
 
-void decompile(clift::FunctionOp Function, ptml::CTokenEmitter &Emitter);
+/// Backend-specific configuration options.
+struct CBackendConfiguration {
+  /// Configuration for the type emitter.
+  TypeEmitterConfiguration TypeEmitter = {};
+};
```
pfez, line 21: We don't need this. The `TypeEmitterConfiguration` is only ever used once, and it's always defaulted. Inline it in the only user.

include/revng/CliftEmitC/CBackend.h
```diff
@@ -15,6 +15,9 @@
 struct CBackendConfiguration {
   /// Configuration for the type emitter.
   TypeEmitterConfiguration TypeEmitter = {};
+
+  /// Should stack frame types be inlined?
+  bool InlineStackFrameType = false;
```
pfez, line 20: This can be a standalone field of the parent class, without the need for `CBackendConfiguration`.

lib/CliftEmitC/CBackend.cpp
```diff
@@ -797,13 +798,39 @@ class CliftToCEmitter : CEmitter {
...
+          if (auto S = clift::unwrapped_dyn_cast<clift::StructType>(Type)) {
+            // emit the *definition* of the struct/union type inline instead of
+            // a regular declaration, so the C output reads as
+            // `struct _PACKED ... my_stack { ... } var_1;`.
+            //
+            // Note that this skips all the typedefs that might be there before
+            // the struct.
+            TypeDefinitionEmitter Emitter(Tokens,
+                                          DataModel,
+                                          Configuration.TypeEmitter);
```
pfez, line 824: No need to take this from the configuration. Just default it.

lib/CliftEmitC/TypeDefinitionEmitter.cpp
Cvelth: That would mean duplicating the body, would it not?!

lib/CliftEmitC/CBackend.cpp
```diff
@@ -797,13 +798,39 @@ class CliftToCEmitter : CEmitter {
...
+    mlir::Type Type = Var.getResult().getType();
+    DeclaratorInfo Declarator{
+      .Identifier = Var.getName(),
+      .Location = Var.getHandle(),
+      .CAttributes = getDeclarationOpCAttributes(Var),
+      .Kind = CTE::EntityKind::LocalVariable,
+    };
+
+    bool DefinitionEmitted = false;
+    if (Configuration.InlineStackFrameType) {
```
pfez, line 810: And if you drop the `Configuration` you just do `if (InlineStackFrameType)` here.

lib/CliftEmitC/TypeDefinitionEmitter.cpp
Cvelth: Ah, wait! I get it, good idea!

include/revng/CliftEmitC/CBackend.h
pfez: nevermind

lib/CliftEmitC/CBackend.cpp
pfez: nevermind

lib/CliftEmitC/CBackend.cpp
pfez: nevermind

include/revng/CliftEmitC/CBackend.h
pfez: nevermind

lib/CliftEmitC/Headers.cpp
pfez: This comment is outdated.

include/revng/CliftEmitC/TypeDefinitionEmitter.h
pfez: This has been settled.

lib/CliftEmitC/CBackend.cpp
pfez: Fixed.

include/revng/CliftEmitC/TypeDefinitionEmitter.h
Cvelth: (Closing this)

include/revng/CliftEmitC/TypeDefinitionEmitter.h
Cvelth: Done.

lib/CliftEmitC/TypeDefinitionEmitter.cpp
Cvelth: Done.

lib/CliftEmitC/TypeDefinitionEmitter.cpp
Cvelth: Done.

pfez:
Merged in b7d8ca11db499d0a22c04d2a4d2d7d0466054fbe thanks

PR #615, opened 2026-07-24
Drop old pipeline

mrjackv:
Sister PRs:
* [orchestra](https://github.com/revng/orchestra/pull/211)

~~Depends on https://github.com/revng/revng/pull/605~~


python/revng/pypeline/pyproject.toml
```diff
@@ -12,7 +12,7 @@ dependencies = [
     "psuitl",
     "xdg",
     "starlette",
-    "uvicorn"
+    "hypercorn"
```
aleclearmind, line 15: > The `uvicorn` dependency was switched out to hypercorn in commit
https://github.com/revng/revng/commit/653f92514b53f6f4c40fc5a6b896f5b649731cd7.

Reference commits by name.

python/revng/pypeline/cli/pipeline/run_pipe.py
```diff
@@ -161,7 +161,7 @@ def build_pipe_command(
             " advanced invalidation entry."
         ),
     )
-    @container_format_options(ContainerFormat.YAML)
+    @container_format_options(ContainerFormat.AUTO)
```
aleclearmind, line 169: This should be in my branch

include/revng/Ranks/IRHelpers.h
```diff
@@ -0,0 +1,72 @@
...
+
+namespace detail {
+
+using InstructionLocation = pipeline::Location<
+  decltype(revng::ranks::Instruction)>;
+
+}
+
+std::optional<
+  detail::InstructionLocation> inline getLocation(const llvm::Instruction *I) {
```
aleclearmind: You can't return a type in `detail::`!

include/revng/DataLayoutAnalysis/DLAPass.h
aleclearmind, line 1: > Inline code
>
> Inline code that was previously shared between the two pipe
implementations, since now there's only the implementation in the new
pipeline.

This commit message makes no sense if you don't look at the patch.

lib/EarlyFunctionAnalysis/DetectABI.cpp
```diff
@@ -970,6 +969,34 @@ DetectABI::computePreservedRegisters(const CSVSet &ClobberedRegisters) const {
   return Result;
 }
 
+static Logger FunctionFromCalleesLog("functions-from-callees-collection");
+static void collectFunctionsFromCallees(Module &M,
```
aleclearmind, line 974: Blank line

lib/PromoteStackPointer/CMakeLists.txt
```diff
@@ -6,9 +6,9 @@ revng_add_analyses_library_internal(
   revngPromoteStackPointer
   CleanupStackSizeMarkersPass.cpp
   ComputeStackAccessesBoundsPass.cpp
-  DetectStackSizePass.cpp
+  DetectStackSize.cpp
   InstrumentStackAccessesPass.cpp
-  InjectStackSizeProbesAtCallSitesPass.cpp
+  InjectStackSizeProbesAtCallSites.cpp
```
aleclearmind, line 11: Are you normalizing file names?
If so, what's the new policy, so we can enforce it from now on.

lib/FunctionIsolation/EnforceABI.cpp
```diff
@@ -452,6 +452,8 @@ CallInst *EnforceABI::generateCall(revng::IRBuilder &Builder,
 namespace revng::pypeline::piperuns {
 
 // TODO: merge ::EnforceABI into EnforceABI once we dismiss the old pipeline
+// WIP: EnforceABI operates on each function `llvm::Module`, it cannot be
+//      inlined
```
aleclearmind: Can you elaborate? Maybe we can discuss this in person.

mrjackv: Let's take `EnforceABI` as an example, if you look at its functions, e.g. `createPrologue`, it uses the `llvm::Module &M` private property. Since each functions lives in its own module, you cannot trivially inline the logic of `EnforceABI` (the implementation class) inside `EnforceABI` (the pipe) because they operate at two different lifetimes:
* The pipe lives for the duration of the run, e.g. if there are 3 functions it lives for all of them
* The implementation lives only for the duration of a single function

lib/PromoteStackPointer/CMakeLists.txt
mrjackv: Basically, if a TU had `*Pass.cpp` but now, in the new pipeline, it turned into a pipe, it should not have `*Pass.cpp`

include/revng/Ranks/IRHelpers.h
mrjackv: Is it OK to have the `using` outside of `detail`? The problem otherwise is that check-conventions complains

aleclearmind: > Is it OK to have using outside of detail?

Works

python/revng/pypeline/pyproject.toml
mrjackv: Done

include/revng/Ranks/IRHelpers.h
mrjackv: Done

lib/EarlyFunctionAnalysis/DetectABI.cpp
mrjackv: Added

include/revng/DataLayoutAnalysis/DLAPass.h
mrjackv: I've reworded it to better explain what's going on. Let me know if it looks good now.

python/revng/pypeline/cli/pipeline/run_pipe.py
aleclearmind: It is now on my branch

mrjackv: Dropped commit

lib/FunctionIsolation/EnforceABI.cpp
mrjackv: Simplified `EnforceABI` and `PromoteCSVs`, kept the classes as we discussed IRL.

PR #616, opened 2026-07-28
faster `lit`

mrjackv:
Replace the `lit` program with a simplified version.
I've tried a couple of times and it reduces the *wall time* to run unit tests by ~20%

### DIFF
--- a/tests/filecheck/simplit
+++ b/tests/filecheck/simplit
@@ -1,0 +1,93 @@
+#!/usr/bin/env python3
+
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+# This script is an optimized version of the `lit` command-line tool from LLVM,
+# for the execution of a single test. It executes faster than the stock lit for
+# two reasons:
+# * It does not auto-discover the `lit.cfg.py` file. Normally lit does this by
+#   traversing the parent directories of the test file, and this causes a lot of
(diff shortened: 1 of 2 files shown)


tests/filecheck/simplit
aleclearmind, line 1: 1. Always write on top of a script what it does!
2. How is this different from regular `lit`? What makes it faster? What are we leaving out?
3. Rename to `simplit` `:P`
4. Do you have (even rough) data on performance improvement? Reporting it at least in the commit message would be nice.

mrjackv: 1. Ok
2. Lit wastes a bunch of time doing these two things:
   * It has a whole bunch of logic related to figuring out where the config of a test is, it looks at the parents directory until it finds a file that matches. In the script we provide the path directly.
   * It sets up a `multiprocessing.Pool` to run multiple tests in parallel. This involves doing `fork` and is quite costly considering we run just one test. We run the `_execute` function directly without spawning other processes.
3. Ok
4. Roughly from ~60s in unit tests to ~45s (-25%)

aleclearmind: OK, write all these things in the script header.

mrjackv: Done

aleclearmind:
Merged in f759573. Thanks.

PR #617, opened 2026-08-03
Fix `bash ignore=` in `test_docs`

pfez:
I used `~` as `sed` command delimiter for regexp replacement, because with the `§` symbol you originally suggested `sed` complains for it not being single-character.

### DIFF
--- a/python/revng/internal/cli/_commands/test_docs/__init__.py
+++ b/python/revng/internal/cli/_commands/test_docs/__init__.py
@@ -175,14 +175,27 @@ class BashDoctest(Doctest):
                 if heredoc_terminator is None:
                     next_line_is_command = line.endswith("\\")
             else:
-                if not ignore_regexp or not re.match(".*(" + ignore_regexp + ").*", line):
+                if not ignore_regexp:
                     self.expected_output += line + "\n"
+                else:
+                    self.expected_output += re.sub(ignore_regexp, "IGNORED", line) + "\n"
 
         if silent:
             self.script += " ) >& /dev/null\n"
 
         if ignore_regexp:
-            self.script += f" ) |& grep -vE '{ignore_regexp}'\n"
+            # TODO: in the future we could replace only certain parts of the regexp using named
+            #       groups: `struct_(?P<ignore>[0-9]+)` would turn `a struct_1234 b` into
+            #       `a struct_IGNORED b` instead of `a IGNORED b`, as this does.
+            filter_script = """
+import re
+import sys
+pattern = re.compile(sys.argv[1])
+sys.stdout.writelines(pattern.sub("IGNORED", line) for line in sys.stdin)
+"""
+            self.script += (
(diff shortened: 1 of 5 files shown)


share/doc/revng/developer-manual/qemu-helpers.md
```diff
@@ -561,7 +561,7 @@ Running `-simplifycfg` eliminates the `unreachable` blocks, turning the error co
 These `llvm.assume` calls are later removed by the `remove-llvmassume-calls` pass (which runs as part of the `segregate-stack-accesses` step).
 Adding `-dce` cleans up the remaining dead instructions:
 
-```bash ignore="%[0-9]|[0-9]:"
+```bash ignore="%[0-9]+|[0-9]+:"
```
aleclearmind, line 564: Unfortunately this does not work:

```
$ echo 'a %234v' | sed 's|%[0-9]+|IGNORED|'
a %234v
$ echo 'a %234v' | sed 's|%[0-9]\+|IGNORED|'
a IGNOREDv
```

I think we should stop trying to use distinct regex implementations.
See my other comment for a solution.

share/doc/revng/developer-manual/qemu-helpers.md
aleclearmind, line 1: Typo in commit message

python/revng/internal/cli/_commands/test_docs/__init__.py
```diff
@@ -175,14 +176,16 @@ def process(self, code, extra=""):
...
                     self.expected_output += line + "\n"
+                else:
+                    self.expected_output += re.sub(ignore_regexp, "IGNORED", line) + "\n"
 
         if silent:
             self.script += " ) >& /dev/null\n"
 
         if ignore_regexp:
-            self.script += f" ) |& grep -vE '{ignore_regexp}'\n"
+            self.script += f" ) |& sed -E 's~{ignore_regexp}~IGNORED~g'\n"
```
aleclearmind:
```python
        if ignore_regexp:                     
            filter_script = """
import re
import sys                                                 
pattern = re.compile(sys.argv[1])             
sys.stdout.writelines(pattern.sub("IGNORED", line) for line in sys.stdin)                     
"""                                                                                           
            self.script += (                                                                  
                f" ) |& python3 -c {shlex.quote(filter_script)}"                              
                f" {shlex.quote(ignore_regexp)}\n"                                            
            )
```

Maybe leave a TODO about the fact that in future we could replace only certain parts of the regexp using named groups: `struct_(?P<ignore>[0-9]+)` turning `a struct_1234 b` into `a struct_IGNORED b` instead of `a IGNORED b` as this would be doing.

But not important right now.

share/doc/revng/developer-manual/qemu-helpers.md
pfez: Fixed

python/revng/internal/cli/_commands/test_docs/__init__.py
pfez: Done

PR #620, opened 2026-08-07
Terminal Branch Complement Hoisting on switch

pfez:


include/revng/Clift/CliftOpTraits.h
```diff
@@ -25,6 +25,23 @@ class NoFallthrough
...
+/// region ends in. Mixed is used for a branch whose sub-regions are each
+/// non-fallthrough but disagree on how (e.g. one returns while another breaks).
+enum NoFallthroughKind {
+  FallsThrough = 0,
+  Continue,
+  Break,
+  Goto,
+  Return,
+  Mixed,
+};
```
vasama-work: Use a scoped enum or place it in its own namespace.

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
```diff
@@ -48,71 +51,166 @@ static unsigned approximateRegionWeight(mlir::Region &R) {
...
-                          If->getBlock(),
-                          std::next(If->getIterator()));
+                          Block,
+                          Op->getBlock(),
+                          std::next(Op->getIterator()));
       });
 
-      Rewriter.eraseBlock(ElseBlock);
+      if (isDroppableBranch(Branch, Target))
+        Rewriter.eraseBlock(Block);
```
vasama-work: I would move the `If` declaration out of the if statement introducer and inline `isDroppableBranch` as `If or Target == 0`.

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
```diff
@@ -48,71 +51,166 @@ static unsigned approximateRegionWeight(mlir::Region &R) {
...
+    if (CandidateWeight <= *BestWeight) {
+      Best = I;
+      BestWeight = CandidateWeight;
+    }
+  }
 
-  bool ThenFallthrough = not isIndirectlyNoFallthrough(If.getThen());
-  bool ElseFallthrough = not isIndirectlyNoFallthrough(If.getElse());
+  return Best;
+}
```
vasama-work: This loop is very ugly and hard to reason about. I wonder if there's a better way to do this?

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
```diff
@@ -48,71 +51,166 @@ static unsigned approximateRegionWeight(mlir::Region &R) {
...
+    return std::nullopt;
 
-  // Hoist the region with the lower weight (or else if equal).
-  return static_cast<HoistingTarget>(ElseWeight <= ThenWeight);
+  // With a single fall-through branch, all others are non-fallthrough, so it
+  // can be hoisted (as long as that makes progress).
+  if (Fallthrough.size() == 1) {
+    unsigned Target = Fallthrough.front();
+    if (hoistMakesProgress(Branch, Target, Regions[Target]))
+      return Target;
```
vasama-work: I would get rid of `hoistMakesProgress` and instead do `if (hasEmptyBlock(Regions[Target]) or mlir::isa<IfOp>(Branch) and Target == 0)`.

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
```diff
@@ -48,71 +51,166 @@ static unsigned approximateRegionWeight(mlir::Region &R) {
...
+    return true;
+
+  // getBranchRegions() of a switch is [default, cases...]; only the default may
+  // be dropped.
+  return Target == 0;
 }
 
 struct TerminalBranchComplementHoistingPattern
-  : mlir::OpRewritePattern<clift::IfOp> {
+  : mlir::OpInterfaceRewritePattern<BranchOpInterface> {
```
vasama-work: Instead of using `BranchOpInterface`, I think I would template it for `IfOp` and `SwitchOp`, since you're doing things very differently for the two.

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
```diff
@@ -244,6 +244,37 @@ struct TerminalBranchComplementHoistingPattern
   }
 };
 
+// A branch operation whose branch regions are all empty (no statements at all)
+// does nothing but evaluate its condition, so it reduces to an expression
+// statement that yields the condition. This covers an if with two empty
+// branches and a switch whose cases and default are all empty, and it also
+// cleans up the empty ifs that complement hoisting can leave behind.
+struct DegenerateBranchPattern
```
vasama-work: Did you actually see this happening in a real binary? And if so, was it only after terminal branch complement hoisting? How is that possible? And if not, this should be in statement optimisations.

vasama-work:
Splitting [Clift: extend terminal branch hoisting to switch](https://github.com/revng/revng/pull/620/changes/6b052ac03dedd0041a587a496ac7bdcf1e449419) and [Clift: pick hoisted branch by precedence](https://github.com/revng/revng/pull/620/changes/3378d2ce52bca2831bc02d419e5e0ba989ead7f3) in two pointlessly makes reviewing harder. Is this an artifact of multiple AI prompts? IIRC you mentioned that this was AI-generated?

pfez:
> Is this an artifact of multiple AI prompts? IIRC you mentioned that this was AI-generated?

No it's an artifact of the fact that initially I wanted to experiment different policies for precedence, and I was thinking about having a single commit for each policy, but then I rapidly converged to the current one. Anyway, now that I converged I can probably squash the two commits.

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
pfez: I haven't seen this in a real binary. I just thought this could start showing up and I added a transform. But it's probably not called for. I'll take this out of the PR.

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
pfez: What do you think is hard to reason about? It's 20 lines of code.

vasama-work: I'm thinking perhaps it could be done in a more readable way using `std::ranges::min`.

include/revng/Clift/CliftOpTraits.h
pfez: Done

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
pfez: Done

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
pfez: Done

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
pfez: done

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
pfez: Dropped

lib/CliftTransforms/TerminalBranchComplementHoisting.cpp
pfez: Done, have another look if you think it's less ugly this way.

vasama-work: Yes, that's much nicer.

pfez:
Merged in 102eaed2eeea72814f287b52ba02e84f84ff103d

PR #621, opened 2026-08-19
Pre type refinement

vasama-work:


lib/CliftTransforms/EmitFieldAccesses/FieldAccessReplacement.cpp
```diff
@@ -291,38 +295,49 @@ bool Replacement::replace(ExpressionOpInterface PointerToReplace,
       // (or a bare `imm` when the index has no variable part, e.g. a `[0]` or
       // `[n]` access), then add each scaled variable component.
       mlir::Value IndexValue = {};
+
+      auto AddTerm = [&](mlir::Value Term) {
```
pfez: Use explicit captures.

PR #622, opened 2026-08-21
Clift wide integers (REP 79)

vasama-work:


lib/CliftTransforms/Legalization.cpp
```diff
@@ -461,6 +461,24 @@ mlir::LogicalResult clift::legalizeForC(clift::FunctionOp Function) {
...
+  Set.add<ArithmeticSignPattern<SCmpLtOp, true, 0, 1>>();
+  Set.add<ArithmeticSignPattern<UCmpLtOp, false, 0, 1>>();
+  Set.add<ArithmeticSignPattern<SCmpGtOp, true, 0, 1>>();
+  Set.add<ArithmeticSignPattern<UCmpGtOp, false, 0, 1>>();
+  Set.add<ArithmeticSignPattern<SCmpLeOp, true, 0, 1>>();
+  Set.add<ArithmeticSignPattern<UCmpLeOp, false, 0, 1>>();
+  Set.add<ArithmeticSignPattern<SCmpGeOp, true, 0, 1>>();
+  Set.add<ArithmeticSignPattern<UCmpGeOp, false, 0, 1>>();
+
+  Set.nextPriorityLevel();
```
pfez: Why deprioritize this?

vasama-work: We want to do this after
1. applying promotions, because those may change the operation type from unsigned to signed, and
2. marking intrinsics, because those have explicit semantics and don't require matching the operand.

I can add some comments on the priority levels.

pfez: > I can add some comments on the priority levels.

Please do

vasama-work: Done. See the new commits at the end.

tools/clift-opt/Main.cpp
```diff
@@ -20,7 +20,12 @@ static constexpr char ToolName[] = "Standalone optimizer driver\n";
 
 static void initializeCliftDialect(mlir::MLIRContext *Context,
                                    clift::CliftDialect *Dialect) {
-  Dialect->setDefaultDataModel(CDataModel::getDefaultDataModel(8));
+  auto DataModel = CDataModel::getDefaultDataModel(8);
+
+  // Enable 128-bit integer support:
+  DataModel.ExtendedIntegerSizeMask |= 128 / 8;
```
pfez, line 26: Why is this necessary?

vasama-work: It isn't necessary, but for the default target we might as well enable 128-bit integer support. This only affects manual users of `clift-opt` and those unit tests that do not explicitly specify a data model.

share/revng/include/primitive-types.h
```diff
@@ -184,9 +175,8 @@ typedef long double float128_t;
 #if defined(__FLT128_MIN__)
 typedef _Float128 float128_t;
 #else
-
 typedef struct {
-  char data[16];
```
pfez, line 189: These should stay `char` arrays.

share/revng/include/primitive-types.h
```diff
@@ -198,20 +188,6 @@ static_assert(sizeof(float80_t) == 10, "");
...
-#ifdef LEGACY_BACKEND
-#define pointer16_t(T) __typeof__((T *){ 0 })
-#define pointer32_t(T) __typeof__((T *){ 0 })
-#define pointer64_t(T) __typeof__((T *){ 0 })
-#else
-#define pointer16_t(T) uint16_t
-#define pointer32_t(T) uint32_t
-#define pointer64_t(T) uint64_t
-#endif
-
```
pfez, line 214: thank god

share/revng/include/primitive-types.h
pfez: And all the other for `float*_t` as well.

share/revng/include/runtime-library.h
```diff
@@ -10,143 +10,146 @@
...
+#define __eq(x, y) (__operator_cmp(ecmp, x, y) == 1)
+#define __ne(x, y) (__operator_cmp(ecmp, x, y) == 0)
+#define __slt(x, y) (__operator_cmp(scmp, x, y) < 0)
+#define __ult(x, y) (__operator_cmp(ucmp, x, y) < 0)
+#define __sgt(x, y) (__operator_cmp(scmp, x, y) > 0)
+#define __ugt(x, y) (__operator_cmp(ucmp, x, y) > 0)
+#define __sle(x, y) (__operator_cmp(scmp, x, y) <= 0)
+#define __ule(x, y) (__operator_cmp(ucmp, x, y) <= 0)
+#define __sge(x, y) (__operator_cmp(scmp, x, y) >= 0)
+#define __uge(x, y) (__operator_cmp(ucmp, x, y) >= 0)
```
pfez: Once we've settled on the prefix, this PR can be merged.

pfez: I've discussed this with @aleclearmind 

Let's use the `rr_` prefix, instead of `__`. Also, remember to reserve that in the `NameBuilder`.

vasama-work: Done.

PR #624, opened 2026-08-31
Clift boolean type

vasama-work:


lib/CliftTransforms/Legalization.cpp
```diff
@@ -631,6 +628,19 @@ mlir::LogicalResult clift::legalizeForC(clift::FunctionOp Function) {
...
+  // opportunities for immediate canonicalisations.
+  {
+    mlir::RewritePatternSet Set(Context);
+
+    Set.add<ImmediateCastPattern>(Context, DataModel);
+
+    auto Patterns = mlir::FrozenRewritePatternSet(std::move(Set));
+    if (mlir::applyPatternsAndFoldGreedily(Function, Patterns).failed())
+      return mlir::failure();
+  }
```
pfez: Can you add a unit test for this?

lib/CliftTransforms/ImplicitCastElision.cpp
```diff
@@ -163,6 +163,15 @@ class ImplicitCastElider {
...
   }
 
+  bool isNullPointerConstant(mlir::Value Value) {
+    if (auto Immediate = Value.getDefiningOp<ImmediateOp>()) {
+      return Immediate.getValue().isZero()
+             and mlir::isa<IntegerType>(Value.getType());
+    }
+
+    return false;
+  }
```
pfez: This helper could go at the dialect level.

include/revng/Clift/CliftTypes.td
```diff
@@ -24,6 +24,14 @@ def Clift_LabelType : Clift_Type<"Label", "label"> {
   }];
 }
 
+def Clift_BoolType : Clift_Type<"Bool", "bool"> {
+  let summary = "Clift boolean type";
+
+  let description = [{
+    1-bit boolean type. Has no model equivalent.
```
pfez: Explain that it can only appear in expressions, never as a subtype of other types.
Also, make sure we have a unit test that expectedly fails if it's used as subtype.

lib/CliftEmitC/CBackend.cpp
```diff
@@ -657,19 +657,19 @@ class CliftToCEmitter : CEmitter {
...
         .Emit = &CliftToCEmitter::emitImmediateExpression,
       };
     }
 
+    if (auto Null = mlir::dyn_cast<NullOp>(E.getOperation())) {
+      return {
+        .Precedence = OperatorPrecedence::Primary,
+        .Emit = &CliftToCEmitter::emitNullPointerConstant,
+      };
+    }
```
pfez: Can we have a unit test for the backend for this?

lib/CliftTransforms/Legalization.cpp
```diff
@@ -167,6 +167,15 @@ struct PointerResizePattern : mlir::OpRewritePattern<OpT> {
...
 };
 
+struct ResizeNullPattern : PointerResizePattern<NullOp> {
+  using PointerResizePattern::PointerResizePattern;
+
+  mlir::LogicalResult
+  matchAndRewrite(NullOp Op, mlir::PatternRewriter &Rewriter) const override {
+    return replacePointerResult(Rewriter, Op);
+  }
+};
```
pfez, line 177: Missing unit test

lib/CliftEmitC/CBackend.cpp
```diff
@@ -421,6 +421,7 @@ class CliftToCEmitter : CEmitter {
       auto Inner = Cast.getValue().getDefiningOp<CastOpInterface>();
       if (not Inner or not isHiddenCast(Inner))
         break;
+      Cast = Inner;
```
pfez: I already merged this because it was causing timeouts in mass-testing. But can you add a unit test please?

include/revng/Clift/CliftC.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+//
+// This file is distributed under the MIT License. See LICENSE.md for details.
+//
+
+#include "revng/Clift/Clift.h"
+
+namespace clift::c {
```
pfez: Drop the `::c`

lib/CliftEmitC/CBackend.cpp
```diff
@@ -426,7 +426,7 @@ class CliftToCEmitter : CEmitter {
   }
 
   static bool isHiddenCast(CastOpInterface Cast) {
-    return mlir::isa<DecayOp>(Cast) or Cast->hasAttr("clift.implicit");
```
pfez, line 414: At this point, do we even need `DecayOp`? Can't that be folded directly into the `ImplicitCastOp`?

include/revng/CliftTransforms/Expressions.td
```diff
@@ -12,6 +12,48 @@ include "revng/CliftTransforms/ExpressionHelpers.td"
...
+// sext(bext(x)) -> bext(x)
+def SignExtendBoolExtendPattern
+  : Pat<(Clift_SignExtendOp (Clift_BoolExtendOp $arg)),
+        (Clift_BoolExtendOp $arg)>;
+
+// truncate(bext(x)) -> bext(x)
+def TruncateBoolExtendPattern
+  : Pat<(Clift_TruncateOp (Clift_BoolExtendOp $arg)),
+        (Clift_BoolExtendOp $arg)>;
+
```
pfez: Shouldn't these be part of cast canonicalization?

lib/CliftEmitC/CBackend.cpp
vasama-work: 1. `ImplicitCastOp` is much more permissive. It's much better for verification to keep `DecayOp` with its more explicit semantics and stricter verification until it is elided.
2. Not all decays can be elided, so some still make it to the backend where they are emitted as C-style casts, while implicit casts are ignored.

include/revng/Clift/CliftC.h
vasama-work: I would like to make some distinction for utilities that model C semantics and are only intended for use by the backend.

lib/CliftEmitC/CBackend.cpp
pfez: 2. Example please? If that's the case it makes sense to keep them separate.

include/revng/Clift/CliftC.h
pfez: I see, but we currently have only this one. It makes sense to start namespacing when we have a few things that would live in the new namespace. Also, this is not a hard rule, but we try to see introducing new namespaces as some kind of project wide decisions. So take this out for now please.

lib/CliftEmitC/CBackend.cpp
vasama-work: Big difference here:
```c
uint8_t a[8];
bitcast(some_struct, a);
bitcast(some_struct, (uint8_t *) a);
```
There's probably other cases too, but this is the most obvious one.

lib/CliftTransforms/ImplicitCastElision.cpp
vasama-work: This is very C specific, and I want to reduce the scope of such things.

lib/CliftEmitC/CBackend.cpp
vasama-work: This function was removed entirely by a subsequent commit.

include/revng/CliftTransforms/Expressions.td
vasama-work: They could be, but would still have to be in a separate set. In the first legalisation step we want to apply the existing cast collapsing rewrites, but not these, so as not to break boolean extension canonicalisation.

lib/CliftTransforms/Legalization.cpp
vasama-work: Done. Also added a new commit to take apply immediate canonicalizations during legalization.

include/revng/Clift/CliftTypes.td
vasama-work: Done.

lib/CliftTransforms/Legalization.cpp
vasama-work: I added one.

include/revng/Clift/CliftC.h
vasama-work: Done. In place of the namespace, I renamed the functions to say `InC`.

include/revng/CliftTransforms/Expressions.td
vasama-work: I ended up moving these into cast canonicalizations and instead delaying boolean extension legalization.

lib/CliftTransforms/Legalization.cpp
```diff
@@ -626,18 +624,26 @@ mlir::LogicalResult clift::legalizeForC(clift::FunctionOp Function) {
...
+  //   This should only be done after promotions and pointer resizing, because
+  //   those rewrites may surface opportunities for immediate canonicalisations.
+  // * Canonicalize boolean extensions.
+  //   This should only be done without cast canonicalization rewrites, as those
+  //   would undo the boolean extension canonicalization rewrites.
   {
     mlir::RewritePatternSet Set(Context);
 
     Set.add<ImmediateCastPattern>(Context, DataModel);
+    Set.add<BooleanCanonicalizationPattern>(Context, DataModel);
```
pfez, line 646: This part doesn't have to do with immediates. It's about boolean canonicalization. Split it apart from this commit, and perhaps squash it in another relevant commit, if possible.

vasama-work: Yes, that makes sense. I split it into a separate commit.

pfez: This is done.

PR #627, opened 2026-09-10
Emit strings in clift

aleclearmind:
This is a branch to emit strings in clift.

Right now we only emit char strings, wide strings are out of scope.


tests/filecheck/clift/passes/implicit-cast-elision-string.mlir
```diff
@@ -0,0 +1,45 @@
...
+!f = !clift.func<
+  "/type-definition/1001-CABIFunctionDefinition" : !uchar$const$ptr()
+>
+
+module attributes {clift.module} {
+  // C: const uint8_t *fun_0x40001001(void) {
+  clift.func @fun_0x40001001<!f>() attributes {
+    handle = "/function/0x40001001:Code_x86_64"
+  } {
+    // C: return (const uint8_t *) "hello";
```
aleclearmind, line 30: @vasama-work ideas on how to omit this cast?

vasama-work: The function return type must be `char const*`.

include/revng/Clift/LocationAddresses.h
```diff
@@ -23,27 +23,30 @@ namespace clift {
 /// expression down to the `clift.use` of the segment, accumulating the offset
 /// on the way. Anything else, in particular an index that is not a constant,
 /// makes the address unknown.
-inline std::optional<MetaAddress> getSegmentAddress(mlir::Value V) {
+///
+/// \return an invalid address when \p V reaches no segment, or reaches one at
+///         an offset that is not known here.
+inline MetaAddress getSegmentAddress(mlir::Value V) {
```
vasama-work, line 29: I guess you'll want to squash this commit (`Clift: report a missing segment address as invalid`) together with the one that introduced this function.

lib/CliftPipes/EmitStrings.cpp
```diff
@@ -0,0 +1,160 @@
...
+    Dead->erase();
+    Dead = Next;
+  }
+}
+
+} // namespace
+
+namespace revng::pypeline::piperuns {
+
+void EmitStrings::runOnCliftFunction(const model::Function &Function,
```
vasama-work: I would prefer to make this a pattern rewrite.

lib/CliftPipes/EmitStrings.cpp
```diff
@@ -0,0 +1,160 @@
...
+                                     clift::FunctionOp MLIRFunction) {
+  // The accesses are collected up front, since replacing one erases the
+  // operations it was built out of. Both `a.b` and `a->b` can name a string,
+  // so the walk goes through the interface the two share.
```
vasama-work: Is that true? In what scenario would `a->b` name a string?

lib/CliftPipes/EmitStrings.cpp
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        auto Pointer = PointerType::get(Character, PointerSize);
+        mlir::Value Value = Builder.create<DecayOp>(Loc, Pointer, Literal);
+        Value = Builder.create<BitCastOp>(Loc, Decay.getType(), Value);
+
+        Decay.getResult().replaceAllUsesWith(Value);
+        Decay.erase();
+      }
+    }
```
vasama-work, line 127: Clift expressions have exactly one use (`getOnlyUse`).

lib/CliftPipes/EmitStrings.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+    eraseAccessChain(Operation);
+  }
+}
```
vasama-work, line 164: So overall, as noted above, I would make this a pattern rewrite.

Match `DirectAccessOp`. I don't think there's any need to match `IndirectAccessOp`.

There's a single user who expects an lvalue of type `T[N]`. In line with the direction of the rest of the dialect, it might make sense to relax the `StringOp` requirements and allow it to have any integral type (of the appropriate size). Then you can just emit a `StringOp` of the desired type and replace that one user directly. You will want to erase the access op though, so as to avoid matching it again. Finally the type of the string literal can be fixed in legalization to match the C target.

tests/filecheck/clift/passes/implicit-cast-elision-string.mlir
aleclearmind: What do you mean? The return type is `const uint8_t *`. We don't have a `char` type.

include/revng/Clift/LocationAddresses.h
aleclearmind: Right, it's not merged yet.

aleclearmind: Gone, I also fixed the branch I wanted you to review to the relevant commits.

lib/CliftPipes/EmitStrings.cpp
aleclearmind: > So overall, as noted above, I would make this a pattern rewrite.
>
> Match DirectAccessOp. I don't think there's any need to match IndirectAccessOp.

:+1:

> [...] it might make sense to relax the StringOp requirements [...]
> Finally the type of the string literal can be fixed in legalization to match the C target.

1. By "the string literal" you mean recreating the literal with the right type or adding a cast?
2. In any case, the type of a `char` string is not something I currently envision as being target-dependent. I'd stick to `number8_t` for now. Once we introduce `wchar_t` we'll need target-specific information and it might make sense to treat `char` strings the same, but for now it seems like we'd be introducing bells and whistles, while lacking the real use case.

   Strings need love due to `wchar_t` and so on, but for now I'd keep it simple.

tests/filecheck/clift/passes/implicit-cast-elision-string.mlir
vasama-work: I mean that in order to hide the cast, the type of the expression must match the return type. The type of `"hello"` is `char const[6]`, and it decays to `char const*`. In order to get rid of the cast, we must ensure that the function return type is `char const*`. Yes, we don't have the ability to do that yet. Once we have more primitive types, we'll be able to emit one of them (chosen depending on the C target) as `char` and then the cast can be elided.

lib/CliftPipes/EmitStrings.cpp
vasama-work: 1. I mean that here you could emit a string literal with any type, and then in legalization we insert a cast in front of it, because no matter what we pretend the type of the string literal is in Clift, its type will always be `char const[N]` in C. `number8_t` will never be equivalent to `char`, and so a conversion is required.

   Eventually, once we have more primitive types and are able to represent `char`, we should also recreate the string literal at that point and change its type to `char const[N]`.
3. Indeed, the type of the string literal is not dependent on the target (it is always `char const[N]`), but the properties of `char` are, and so which of our primitive types we are able to map onto `char` varies.

tests/filecheck/clift/passes/implicit-cast-elision-string.mlir
aleclearmind: So you're saying there's currently no way to elide that cast without corrupting semantic?
Can you make an example of how removing the cast could corrupt the semantic?
I think we can assume `int8_t` is equivalent to `signed char`.
I really want to get rid of that cast.

> Once we have more primitive types

There's currently no plan to do that.

lib/CliftPipes/EmitStrings.cpp
aleclearmind: 1. OK, so add a cast, not recreate.
2. Right now we assume `char` is a signed 8-bit integer `int8_t`, basically.
   Let's go by this assumption for now.
   Therefore, at this stage, I don't think relaxing type of a string + pulling in legalization is useful.

tests/filecheck/clift/passes/implicit-cast-elision-string.mlir
vasama-work: It's not a matter of changing semantics, it's a matter of type checking. `uint8_t` and `char` are distinct types, so the conversion changes the pointer type:
```c
uint8_t const* f(void) {
  return "hello"; // error: incompatible pointer types
}
```

> I think we can assume `int8_t` is equivalent to `signed char`.

Yes, very likely, but it doesn't matter. `char` is distinct from `signed char`, `unsigned char` and from every other type.

> There's currently no plan to do that.

No concrete plan, yes? But even in this PR you've been talking about `wchar_t` support we've talked about `text8_t` and friends. That is what I'm referring to.

lib/CliftPipes/EmitStrings.cpp
vasama-work: 1. For now, yes.
2. The issue is that `char` is distinct from `int8_t`.

   Yes, we can also insert the cast in this pipe without involving legalization.

tests/filecheck/clift/passes/implicit-cast-elision-string.mlir
aleclearmind: Thanks for providing an example. Always bring examples right away, they bring your point across much more effectively.

I see a couple of solutions:

1. `-Wno-pointer-sign`
2. Wrap strings into a macro casting to a `stdint.h` type.
3. If target is C23, we can do
   ```c
   uint8_t const *f(void) {
       return u8"hello";
   }
   ```

Possibly we can go with 1) and have a strict mode doing 2).
`char` is a royally cursed type, that gets "different signedness" warnings against both `unsigned char` and `signed char`.
I really don't want this cursedness to creep into clift.
In general, I don't want `int`, `char` or any other of those into rev.ng, for reasons similar to those highlighted in my comment on `bool`/REP59 on Mattermost.
We need to devise a sane subset of C. We're not a compiler.

Monday I'll run this through Pietro.

> you've been talking about wchar_t support we've talked about

I'm talking about finding ways to exploit it *in emission*. For instance, if we have a UTF16 string in a `uint16_t` array and the wide strings in the target ABI are UTF16 as well, we can emit it as a wide string literal. I'm not implying we want to explicitly add to clift or the model's type system `wchar_t`.

vasama-work: 1. Yes, we can look into having a non-strict mode where `-Wpointer-sign` is disabled, but that represents an entirely new kind of implicit cast elision that we are not doing currently.
2. Sure, that's easy enough.
3. Yes, possibly, depending on encoding.

> I really don't want this cursedness to creep into clift.
> In general, I don't want int, char or any other of those into rev.ng, for reasons similar to those highlighted in my comment on bool/REP59 on Mattermost.

I totally agree. I have never proposed anything of the sort.

What I am saying we can do is have some type (e.g. by leveraging primitive kinds) to indicate that a value contains a signed UTF-8 code unit for example. For the sake of the argument, let's say we emit that type as `signed_utf8_t`, and we define it as a typedef for `int8_t`.

However, if and only if the chosen C target has signed 8-bit char and the narrow encoding is UTF-8, then we could (depending on user's configuration) emit `char` instead of `signed_utf8_t`, being that the two are equivalent.

Naturally, this isn't a complete design, but it is a direction that avoids all of the weirdness of standard C types, both in the model and in Clift, while still allowing us to take full advantage of them and to elide casts even in the strictest of cast-preserving modes.

aleclearmind: > Yes, we can look into having a non-strict mode where -Wpointer-sign is disabled, but that represents an entirely new kind of implicit cast elision that we are not doing currently.

To be clear, I think the current problem is in Clift (not in C): we're not suppressing the cast from `number8_t const *` (decay'd type of `"asd"`) to `uint8_t const *`, isn't it?
Right now, we don't have `char`.

> I totally agree. I have never proposed anything of the sort.

Nice! `:)`

We can better discuss this later on, in the `wchar_t` context.
For the time being I'd really like to handle simple strings and have no ugly casts.

vasama-work: > To be clear, I think the current problem is in Clift (not in C): we're not suppressing the cast from number8_t const * (decay'd type of "asd") to uint8_t const *, isn't it?

The status quo is that we do not elide any pointer casts with incomplatible types. The casts we do elide are:
1. `T*` → `T const*`
2. `T*` → `void*`
3. `T*` → `U*` where either of `T` and `U` may be typedefs, but the underlying type is the same.

We can certainly look into extending implicit cast elision, but that doesn't require anything special in the string handling code, and will affect casts more generally.

aleclearmind: > We can certainly look into extending implicit cast elision

Yeah, I'm just trying to get rid of annoying casts that pop up in front of me.

Maybe we should suppress casts between things that in C we know to be the same type (e.g., `uint8_t *` and `generic8_t *`) and then, on top of this (kinda unrelated), switch strings from `number8_t[]` to `int8_t[]`.

vasama-work: > Maybe we should suppress casts between things that in C we know to be the same type (e.g., `uint8_t *` and `generic8_t *`)

I thought we already do this but perhaps not? I can look into it. Certainly that's the intention.

> on top of this (kinda unrelated), switch strings from `number8_t[]` to `int8_t[]`.

Well as I said in another thread, I think we should relax it and allow any integer type of the appropriate width and then fix it in legalisation. In any case, changing it to `int8_t` doesn't solve the problem, because whatever type we give it in Clift cannot change the fact that it is, and always will be, `char` in C.

lib/CliftPipes/EmitStrings.cpp
aleclearmind: Right now `emit-field-accesses` leaves us with `(&segment)->str`.

`getSegmentAddress` catches both `(&segment)->str` and `segment.str`. It also catches pointer arithmetic.

lib/CliftPipes/EmitStrings.cpp
aleclearmind: Please take a look.

lib/CliftPipes/EmitStrings.cpp
vasama-work: `(&segment)->str` is rewritten to `segment.str` in expression optimisation.

lib/CliftPipes/EmitStrings.cpp
```diff
@@ -0,0 +1,166 @@
...
+  uint64_t PointerSize;
+
+  EmitStringsPattern(mlir::MLIRContext *Context,
+                     RawBinaryView &BinaryView,
+                     uint64_t PointerSize) :
+    OpInterfaceRewritePattern(Context),
+    BinaryView(BinaryView),
+    PointerSize(PointerSize) {}
+
+  void initialize() { setDebugName("emit-strings"); }
```
vasama-work, line 68: There's no harm in having this, but also no benefit.

lib/CliftPipes/EmitStrings.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+  mlir::LogicalResult
+  matchAndRewrite(AccessOpInterface Access,
+                  mlir::PatternRewriter &Rewriter) const override {
+    mlir::Operation *Operation = Access.getOperation();
```
vasama-work, line 73: You can also just do `Access->whatever()` and avoid introducing an additional variable.

lib/CliftPipes/EmitStrings.cpp
```diff
@@ -0,0 +1,166 @@
...
+    mlir::Value Result = Operation->getResult(0);
```
vasama-work, line 74: You can also avoid introducing this variable. `Access` is implicitly convertible to `mlir::Value`, so later you can simply pass `Access` to `getSegmentAddress`. But if you want to be more explicit, use `AccessOpInterface::getResult()` instead of `mlir::Operation::getResult(unsigned index)`.

lib/CliftPipes/EmitStrings.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+    unsigned CharacterSize = getCharacterSize(Type);
+    if (CharacterSize == 0)
+      return mlir::failure();
+
+    // Right now we only support number8_t strings
+    if (CharacterSize != 1) {
+      revng_log(Log, "Ignoring a string of " << CharacterSize << "-byte chars");
+      return mlir::failure();
+    }
```
vasama-work, line 85: The first check is redundant.

lib/CliftPipes/EmitStrings.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+  auto Character = mlir::dyn_cast<IntegerType>(Array.getElementType());
+  if (not Character or Character.getKind() != IntegerKind::Unsigned)
+    return 0;
+
+  if (Character.getSize() != 1 and Character.getSize() != 2)
+    return 0;
+
+  return Character.getSize();
+}
```
vasama-work, line 37: This function is only used once, and the rewrite relies on invariants established within this function, so I would prefer just inlining it in the rewrite pattern.

lib/CliftPipes/EmitStrings.cpp
```diff
@@ -0,0 +1,166 @@
...
+      Rewriter.replaceOp(Operation, Cast);
+      return mlir::success();
+    }
+
+    auto Pointer = PointerType::get(Character, PointerSize);
+    mlir::Value Value = Rewriter.create<DecayOp>(Loc, Pointer, Literal);
+    Rewriter.replaceOpWithNewOp<BitCastOp>(Decay, Decay.getType(), Value);
+    Rewriter.eraseOp(Operation);
+
+    return mlir::success();
```
vasama-work, line 144: I'm strongly of the opinion that you should relax `StringOp` constraints and simply emit a `StringOp` matching the type of the replaced access here. Then introduce any necessary casts during legalisation.

include/revng/Clift/LocationAddresses.h
```diff
@@ -4,6 +4,8 @@
 // This file is distributed under the MIT License. See LICENSE.md for details.
 //
 
+#include <optional>
+
```
vasama-work: This include is not used.

lib/CliftPipes/EmitStrings.cpp
aleclearmind: Dropped, please take a look

lib/CliftPipes/EmitStrings.cpp
aleclearmind: Dropped, please take a look

lib/CliftPipes/EmitStrings.cpp
aleclearmind: Dropped, please take a look

lib/CliftPipes/EmitStrings.cpp
aleclearmind: Dropped, please take a look

lib/CliftPipes/EmitStrings.cpp
aleclearmind: Done, please take a look

PR #628, opened 2026-09-11
Clift quickies

vasama-work:


python/revng/internal/pipeline.yml
```diff
@@ -453,21 +453,38 @@ branches:
...
+            # Fixed-point optimization phase. In this phase, fixed point passes
+            # are used to simplify statements and expressions. These transforms
+            # rely on canonical forms that may not necessarily match the most
+            # beautiful form. Later phases transform once canonical forms into
+            # more readable ones.
             - optimize-statements
             - label-merging
             - refine-types-and-accesses
+
+            # Post processing phase
```
pfez: A couple more words on this? What's its purpose? I think we said this is for transforms that are meant to turn Clift into a form such that C code generated from it is "better" to look at, according to the very opinionated definition of "better" decided by `revng`. In the future I expect steps in this phase to start exposing user configurations.

python/revng/internal/pipeline.yml
```diff
@@ -453,21 +453,38 @@ branches:
...
             - terminal-branch-complement-hoisting
+
+            # Backend phase. In this phase, in preparation for the emission of
+            # C code from the Clift IR, various previously held invariants are
+            # broken, while new ones are introduced. Previously canonical forms
+            # would break semantics of the emitted C code, and so a high level
+            # of care must be taken when applying rewrites in this phase.
             - tighten-variable-scopes
             - hoist-variable-initializers
             - promote-break-continue
```
pfez, line 498: I think these 3 belong to post-processing.

python/revng/internal/pipeline.yml
```diff
@@ -453,21 +453,38 @@ branches:
...
             - c-legalization
             - elide-implicit-casts
             - eliminate-boolean-casts
             - deduce-immediate-radices
```
pfez, line 510: Maybe this belongs to post-processing as well?

python/revng/internal/pipeline.yml
```diff
@@ -453,21 +453,38 @@ branches:
...
+
+            # Post processing phase
             - post-process-expressions
             - terminal-branch-complement-hoisting
+
+            # Backend phase. In this phase, in preparation for the emission of
+            # C code from the Clift IR, various previously held invariants are
+            # broken, while new ones are introduced. Previously canonical forms
+            # would break semantics of the emitted C code, and so a high level
+            # of care must be taken when applying rewrites in this phase.
```
pfez: This sounds scary, but does not clarify. 
I would call this Legalization phase and reword the comment as follows.
> Legalization phase.
> C code emitted directly from Clift in the form enforced by the previous phases is not guaranteed to preserve semantics.
> This phase legalizes Clift to guarantee that the generated C code is semantic preserving, and performs a few late stage optimizations to ensure elision of casts injected by legalization.

lib/CliftTransforms/Statements.cpp
```diff
@@ -270,6 +270,55 @@ struct BranchEqualizationPattern : StatementRegionRewritePattern {
   }
 };
 
+struct JumpDuplicationPattern : StatementRegionRewritePattern {
```
pfez, line 273: Unit test this please, with if and switch, and in particular I also want an example with a switch with only some cases falling through.

tests/filecheck/clift/statement-rewrites/do-while-conversion-with-empty-then.mlir
```diff
@@ -1,52 +0,0 @@
-//
-// This file is distributed under the MIT License. See LICENSE.md for details.
-//
-
-// RUN: %root/bin/revng clift-opt %s --optimize-statements=enable-patterns=do-while-conversion --canonicalize | FileCheck %s
```
pfez, line 5: Why did you drop this?

python/revng/internal/pipeline.yml
```diff
@@ -453,21 +453,43 @@ branches:
...
             - optimize-statements
             - label-merging
             - refine-types-and-accesses
+
+            # Post processing phase
             - post-process-expressions
-            - terminal-branch-complement-hoisting
+            - hoist-unique-fallthrough-branches
+            - promote-while-conditions
+            - promote-do-while-conditions
```
pfez, line 485: My understanding is that these two are the only places where we promote a `while(true)` to better looking loops with actual conditions.

I was under the impression that these were going to end up as part of `optimize-statements`. Can you motivate exactly why you've opted for putting them in post-processing? Not that I'm against it, but I want to understand:
1. would it be possible to run them as part of `optimize-statements`?
2. if no, why?
3. if yes, would the end results be better or worse? or perhaps the same but they would require all transforms in `optimize-statements` to match a wider set of patterns for loops?

share/revng/include/runtime-library.h
```diff
@@ -41,6 +41,8 @@ void rr_inc_impl(void *in_out, size_t size);
 void rr_dec_impl(void *in_out, size_t size);
 
 int rr_test_impl(const void *in, size_t size);
```
pfez, line 43: `rr_test_impl` is not necessary anymore, since the backend asserts we never emit `TestOp`. Please drop it.

pfez: Together with `rr_test` and `rr_operator_test`.

share/revng/include/runtime-library.h
```diff
@@ -142,6 +151,7 @@ int rr_ucmp_impl(const void *lhs, const void *rhs, size_t size);
 #define rr_post_dec(x) rr_operator_post_inc(dec, x)
 
 #define rr_test(x) rr_operator_test(x)
+#define rr_bext(type, x) rr_operator_bext(type, x)
```
pfez: Can't this just be `rr_operator_zext(type, x)`? So we can avoid `rr_operator_bext` and `rr_operator_bext_impl`.

share/revng/include/runtime-library.h
vasama-work: We do emit `TestOp` when it is an intrinsic. The backend asserts we do not emit `TestOp` as a cast.

tests/filecheck/clift/statement-rewrites/do-while-conversion-with-empty-then.mlir
vasama-work: IIRC it was identical to another test.

python/revng/internal/pipeline.yml
vasama-work: > I was under the impression that these were going to end up as part of `optimize-statements`.

No. We talked about this. We need to do it after unique fallthrough branch hoisting and we can't do that in `optimize-statements`, because it produces a non-canonical form that is undone by branch equalisation.

python/revng/internal/pipeline.yml
vasama-work: No, definitely not. This is part of the backend. It just adds guidance attributes to existing operations. Immediates are modified in the previous backend phases, so we can't do it any earlier.

python/revng/internal/pipeline.yml
vasama-work: It's fair to say that they're not exactly part of the backend, but if you don't want them there, then I would instead introduce a new "scary post-processing phase". The reason being that it's at this point we start breaking previously held invariants.

E.g. if you ran `optimize-expressions` again after the current post-processing phase, nothing would break. You'd just get uglier code. On the other hand, if you ran it after these passes, you might get a crash or broken semantics.

python/revng/internal/pipeline.yml
vasama-work: It is very intentionally scary, because there are a lot of implicit invariants introduced in this phase, and changing things here should be scary.

Legalisation is just one part of the backend, so I don't know why we would call it the legalisation phase.

include/revng/Clift/CliftAttributes.td
```diff
@@ -188,4 +188,32 @@ def Clift_TypedefAttr
   let genVerifyDecl = 1;
 }
 
+def Clift_SwitchCaseArrayAttr : Clift_Attr<"SwitchCaseArray", "switch"> {
```
pfez: I'd got for `"switch_case"` or `"switch_case_array"`.
It's not very important, since it almost never shows up except for when `verify` fails, but still, I think it makes the output less confusing when `verify` fails, which is good for debugging.

lib/Clift/CliftAttributes.cpp
```diff
@@ -1012,6 +1012,105 @@ mlir::LogicalResult clift::CAttributeAttr::verify(EmitErrorType EmitError,
...
+  llvm::SmallSet<uint64_t, 16> Set;
+  unsigned PrevIndex = -1;
+
+  for (const auto &[Index, Value] : Cases) {
+    if (Index != PrevIndex and Index != PrevIndex + 1)
+      return EmitError() << "case region indices must be ordered and with no "
+                            "gaps";
+
+    if (not Set.insert(Value).second)
+      return EmitError() << "case values must be unique";
```
pfez: Print the duplicate case please.

lib/Clift/CliftAttributes.cpp
```diff
@@ -1012,6 +1012,105 @@ mlir::LogicalResult clift::CAttributeAttr::verify(EmitErrorType EmitError,
...
+mlir::LogicalResult
+SwitchCaseArrayAttr::verify(EmitErrorType EmitError,
+                            llvm::ArrayRef<SwitchCase> Cases) {
+  llvm::SmallSet<uint64_t, 16> Set;
+  unsigned PrevIndex = -1;
+
+  for (const auto &[Index, Value] : Cases) {
+    if (Index != PrevIndex and Index != PrevIndex + 1)
+      return EmitError() << "case region indices must be ordered and with no "
+                            "gaps";
```
pfez: Print the 2 indices involved in the problem.

python/revng/internal/pipeline.yml
pfez: Oh now I remember. Please document this in `pipeline.yml`.

share/revng/include/runtime-library.h
pfez: But when it's an intrinsic its emission in C should be transparent, so my expectation is that `rr_test` should not be emitted in C anyway. The `TestOp` semantic in C is just implicit and then you don't need to emit the call to `rr_test`. If that's not the case, let's talk about because this is unexpected.

python/revng/internal/pipeline.yml
pfez: I'm starting to think Legalization and Backend should be 2 separate phases.

* Legalization is about guaranteeing that C code generation from Clift doesn't break semantics.
* Backend is about making a few final eye-candy changes to Clift, without breaking Legalization, on pattern that only become visible after Legalization.

vasama-work: Legalisation only servers the emission of C, which is the purview of the backend. And anyway, legalisation composes one pass, so there's little point in treating it as its own phase. The concept of phases is just for documentation purposes anyhow.

share/revng/include/runtime-library.h
vasama-work: The request was that we do not emit `bool`. That has not changed. But if we're boolean testing `int256_t` that is actually a struct, we have to call a runtime library function. That function is called `rr_test`.

vasama-work:
```c
int256_t i;

if (i) // error: anonymous struct is not a scalar type
  ;

if (rr_test(i)) // ok
  ;
```

python/revng/internal/pipeline.yml
vasama-work: Done.

python/revng/internal/pipeline.yml
vasama-work: I moved these to a separate phase.

python/revng/internal/pipeline.yml
vasama-work: Done.

include/revng/Clift/CliftAttributes.td
vasama-work: I renamed it to `#clift.switch_case_array`.

lib/Clift/CliftAttributes.cpp
vasama-work: Done.

lib/Clift/CliftAttributes.cpp
vasama-work: Done.

share/revng/include/runtime-library.h
vasama-work: Do you mean emit `rr_zext` in place of `rr_bext`, or implement `rr_bext` in terms of `rr_operator_cast`?

I think either should work, so perhaps we can avoid `rr_bext` entirely and just emit `rr_zext`...

python/revng/internal/pipeline.yml
pfez: I definitely don't want them as part of the Backend phase. And even with the change I proposed to split Legalization and Backend in two separate phases, these still don't seem like they are good candidates to be part of the Legalization phase either.

Unfortunately "scary post-processing phase" is not a good enough explanation.

"regular" and "scary" seem to have the same goal: going away from the optimization-friendly canonical form to make the C code generated look better.
And they also achieve that goal in the same way: applying one-off transforms, once.
And they also share some constraints that prevent such transforms to be part of the "fixed-point optimization phase": most of these transforms (all?) would undo other transforms that are already in "fixed-point optimization" (and that are fundamental for making it converge to an optimized form), so they just can't be part of "fixed-point" because they would make the fixed-point spiral into an infinite loop.

If we want to split them "regular" and "scary" post-processing in two separate phases we need to be able to articulate what they do differently. So let's try to nail down the properties that can help us tell "regular post-processing" apart from "scary post-processing".
You said that re-running `optimize-expressions` again after "regular post-processing" is fine (no crashes, no brokeen semantics, just not great quality), while running it after the "scary post-processing" might break things. This is a start, but we need to be more precise in defining what's going on here. 

So my question is: why? What are "scary" transforms doing that "regular" aren't?

My hunch is that the "scary" transforms aren't doing anything particularly scary on their own. But they are putting Clift into a more complex shape. And once these more complex shapes appear, whatever transform is executed afterwards will have to deal with them. Transforms in "fixed-point" and in "regular post-processing" are currently not equipped to do it without crashing and/or breaking semantic. And equipping them to do it would just make them more complex and harder to maintain, if that's even possible at all, for no real benefit.

From a quick look:
1. `tighten-variable-scopes` is moving `LocalVariableOp`s away from the entry of the function, down into some nested scopes. The "scary" property is that after doing so, any later transform can't hoist or sink any scope without the risk of moving around a local variable declaration as well, with implications at least for visibility and possibly for semantics.
2. `hoist-variable-initializers` is folding assignment expression into the initializer of `LocalVariableOp`s. So obviously it has to run right after `tighten-variable-scopes` to have any chance to do something meaningful without breaking semantic. What it does in and on itself is not scary at all, this is just an ordering constraint with `tighten-variable-scopes`. The only slightly "scary" property is that after this runs, whatever runs next has to deal with `LocalVariableOp`s with non-empty initializer.
3. `promote-break-continue` is clearing the `MakeLabelOp` from `break_to` and `continue_to` if they are "well behaved". The "scary" here is that whatever runs after this has to take into account that moving around scopes that contain a label-less `break` or `continue` make break semantic if the label is not re-populated.

In practice, these boil down to 3 properties or perhaps "implicit invariants" that are assumed in "fixed-point optimization phase" and in "regular post-processing", that "scary post-processing" breaks piece by piece.

With the same numbering as above, these "implicit invariants" are:
1. all `LocalVariableOp` are at the beginning of the function
2. no `LocalVariableOp` has a non-empty initializer; I think that this is less strong than the others; like I said it's more an ordering requirement
3. no label-less `break`/`continue`.

These are at the same time invariants that are assumed by "fixed-point" and that passes in "fixed-point" must preserve. And in fact they are preserving these. No "fixed-point" and no "regular post-processing" transform moves `LocalVariableOp` around, nor populates the intializers, nor clears the labels in `break`/`continue`.

So instead of having ideas such as "implicit invariants" that "scary pos-tprocessing" breaks, we should spell out these invariants clearly (I tried, with the 3 points above). Are you aware of other such invariants that I've missed? If so, I'm all ears.

Then we can have a `verifyFixedPointInvariants` helper function, and call it (guarded by `VerifyLog`) at the beginning of each pass that belongs to the "fixed point optimization" phase. And possibly at the beginning of "regular post proceessing passes" too.
It would be nice to have a bunch of different kinds of MLIR passes, that only differ in the kind of `verify` they run on execution. Perhaps it's possible to do that with the passes we define. Can you look into this?

python/revng/internal/pipeline.yml
pfez: Another comment on "regular post processing".
After my investigation on "scary post processing", I'm definitely convinced that we want to keep the 2 phases separate.
But "post processing" as a term is really showing its limits, because it doesn't explain what we're post-processing Clift for.

What about "c-idiom-matching"?

python/revng/internal/pipeline.yml
pfez: Thinking a bit more about this, all the "scary" transforms, in particular 1. and 3. share a common property I think: they make transforms that are only valid with a given specific nesting of all the regions. 1. works only if all the uses of the local variables are nested in a certain way. 3. works only if all the parent regions of the `break`/`continue` are nested in a certain way. So whenever we execute any of the "scary" transforms, we're giving up any opportunity to rearrange the nesting of the scopes later.

So perhaps this "scary" phase is "context sensitive C idiom matching"? While the "regular post processing" is "context free C idiom matching"?

vasama-work: I don't like "C idiom matching", since this has little to do with C and we're not actually matching any idioms as such. But I do like context free and context sensitive. So I would rather call it context free post-processing and context sensitive post-processing. Unless we come up with a better name, but personally I think post-processing is fine. The name is only useful for referring to this thing, but either way you have to go look at it to understand it. If anything, a less descriptive name can be better because it's less likely to produce faulty assumptions.

python/revng/internal/pipeline.yml
pfez: As I said in comment in another thread, I'm thinking that we should clarity the "implicit invariants" and have dedicated `verify` methods for them. So it's not just for documentation. It's at the same a tool for reasoning about what we're doing, about why we're doing it, and at this point I hope also a tool for practical verification.

Given that we will now split the "scary post processing" out of Backend (for reasons I explained in the other thread), we can then leave the "backend phase" as backend and not call it legalization. In this sense legalization will remain a part of the backend phase. The description I added in the first message of this thread applies even better now.

python/revng/internal/pipeline.yml
pfez: Good but not enough.
1. What you call "Usual clift invariants" should be written somewhere in code. I would call them "strict invariants" or "context-free invariants" at this point, because they are only relaxed only by the "scary post-processing", which is now going to be called "context-aware post-processing" (or something like it). It's important to state in the comment that these are invariants that are strictly stricter than `verify`, and they are not strictly speaking a canonical form for clift in MLIR terms. They are simply the assumptions that all context-free transforms can assume and need to enforce. We still have room for improving the terminology here, but "usual clift invariants" or "implicit invariants" is not good enough, and "post processing phase 1/2" isn't enough as well.
2. You don't say anywhere the reason why the entire "post processing phase 1" isn't folded into the "fixed point optimization". This is very important. The reason AFAIU is that most of these transforms undo stuff in "fixed-point" for readability reasons, so folding them into "fixed-point" would turn it into a non-converging infinite loop.

python/revng/internal/pipeline.yml
pfez: Ok, agreed for context-free and context-sensitive. "C idiom matching" is not perfect, but it gives the sense that we're matching patterns in Clift that can be transformed into shapes that result in C code that is better to look at, for our very opinionated definition of better. Post-processing is really vague, but as long as we document in `pipeline.yml` that those 2 phases collect one-off steps that are meant to make the output C code look nicer, we can keep post-processing, since I can't come up with a better name :)

share/revng/include/runtime-library.h
pfez: I meant the second. But if we can get away without `rr_bext` I'm all for it!

python/revng/internal/pipeline.yml
pfez: Turning the "context free invariants" into code and `verify`, and making sure that passes before "context-aware preprocessing" do use that additional `verify` at least under `VerifyLog` is still very important though. Please investigate this, with an eye on how to prevent mistakes and on sane defaults for the future. Perhaps all our passes should verify that except for those that explicitly opt out? And then you opt out only in the 6 passes from context aware post-processing to the backend?

tests/filecheck/clift/statement-rewrites/do-while-conversion-with-empty-then.mlir
pfez: But now you have re-enabled it?

vasama-work: No. I deleted it because it added nothing.

share/revng/include/runtime-library.h
pfez: This went away. Nice!

python/revng/internal/pipeline.yml
pfez: We decided offline to not do anything about verification. The invariant on scoping of local variables in already caught by regular `verify`. And the invariant about break/continue would be the only additional invariant, and we are exploring other solutions for that, that are outside the scope of this PR.

The "scary post processing" has been split out of Backend into a dedicate "context-aware post processing phase".

python/revng/internal/pipeline.yml
pfez: We've settled the naming issue. Closing this.

lib/CliftTransforms/Statements.cpp
pfez: I think this is still missing.

python/revng/internal/pipeline.yml
pfez: Item 2. here hasn't been applied. The current comments still don't mention the fact that those passes can't be put into the fixed point phase because they would turn that into an infinite loop.

lib/CliftTransforms/Statements.cpp
vasama-work: I added tests.

python/revng/internal/pipeline.yml
vasama-work: I added a note.

PR #630, opened 2026-09-18
BinaryImporter: tolerate a PE/COFF Load Config Table RVA outside any section

### LINKED ISSUE #598: RVA 0x2a0 for load config table not found on some .exe files
trying to analyze Heroes3.exe from Heroes of Might and Magic III Complete Eddition (from GoG)
```
revng analyze import-binary Heroes3.exe -o Heroes3.yml
RVA 0x2a0 for load config table not found
```

this exe
https://drive.google.com/file/d/1X_x3THm5zxohMZtryR4MoZx0h_MS4rs0/view?usp=drive_link

sjkim1127:
## Why

Fixes #598. `revng analyze import-binary` fails outright on some real-world PE/COFF binaries (e.g. `Heroes3.exe` from the issue) with:

```
RVA 0x2a0 for load config table not found
```

## Root cause

This is not a bug in revng's `PECOFFImporter`. The error is raised by `llvm::object::createBinary` itself, before revng's own importer code ever runs.

`COFFObjectFile::initialize()` (in `llvm/lib/Object/COFFObjectFile.cpp`) eagerly resolves several data directories while constructing the object file, including the Load Config Table. If a directory's RVA doesn't fall inside any section, `getRvaPtr` returns a hard `parse_failed` error (`"RVA 0x... for <directory> not found"`), which `COFFObjectFile::initialize()` propagates — `ignoreStrippedErrors` only swallows the *different*, already-handled case of an RVA that's inside a section but past `SizeOfRawData`. So `createBinary` fails and the file can't be opened at all, even though everything else about it may be fine.

This happens in practice with binaries from older/non-standard linkers that leave a stale RVA in the load config directory instead of zeroing it when the directory isn't actually used.

revng builds against whatever LLVM `find_package(LLVM)` finds, so patching this specific check into a private LLVM fork isn't something this PR does — the fix below is a revng-local workaround that avoids needing one.

## Fix

`lib/Model/Importer/Binary/COFFDataDirectorySanitizer.h` adds `sanitizeCOFFLoadConfigDirectory()`, called from `BinaryImporter.cpp` before `object::createBinary`. It parses the raw buffer by hand (DOS header → PE signature → COFF header → optional header → load config data directory entry → section table, all with explicit bounds checks) and checks whether the Load Config Table's RVA falls inside `[VirtualAddress, VirtualAddress + VirtualSize)` of some section — mirroring the exact check `COFFObjectFile::getRvaPtr` performs. If it doesn't, we zero out that one directory entry's RVA/Size in a patched copy of the buffer (i.e., treat it as absent) and log what we dropped via `-debug=binary-importer`. Well-formed binaries are left byte-for-byte untouched (function returns `nullptr`).

**Scope**: this intentionally only covers the Load Config Table, not the other directories `COFFObjectFile` resolves eagerly (import, delay-import, export, base relocation, debug, TLS), even though they share the exact same failure mode. Unlike load config, revng's `PECOFFImporter` actively consumes the import/delay-import directories to build the model's imported functions — silently dropping those would change import information in the resulting model rather than just tolerating a stale, unused directory. If a real binary turns up needing one of the others, I'd rather extend this with a concrete repro than guess at the scope now.

## Validation

- Added `tests/unit/COFFDataDirectorySanitizer.cpp` (registered in `tests/unit/CMakeLists.txt` as `test_coff_data_directory_sanitizer`), which builds a minimal synthetic PE32 in memory and checks: a Load Config RVA outside the only section makes `createBinary` fail before the patch and succeed after it; a valid Load Config RVA and a binary with no load config directory at all are both left untouched by the sanitizer; a handful of non-PE/truncated buffers don't crash it.
- I don't have a full revng/Orchestra build in the environment I used, so I could not run this test through revng's actual CMake/CTest setup, or run `revng analyze import-binary` end-to-end against the reporter's `Heroes3.exe`. What I *did* verify: I compiled the header and this exact test file standalone against Homebrew's LLVM 22 and Boost 1.92 (same `llvm::object`/`llvm::COFF` struct layouts, different LLVM version than whatever revng pins) — all 4 test cases pass, including under ASan+UBSan, and the standalone build is clean under `-Wall -Wextra -Wpedantic`.
- Also fuzzed the bounds-checking logic outside the checked-in test by truncating the malformed synthetic PE at every length from 0 to 700 bytes and confirming no crashes under ASan+UBSan.
- I did not download/attach the reporter's `Heroes3.exe` (proprietary game binary behind a Google Drive link).

## Limitations / what's left

- **Not yet run through revng's actual build/CTest**, and not yet confirmed against the reporter's real `Heroes3.exe` — I'd like to do both (or have a maintainer do so) before this comes out of draft.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


aleclearmind:
Hi, we didn't manage to triage the original issue.

There have been some changes around that area recently, so I want to take a moment to review it before considering this PR.

In case you closed this PR due to the lack of answer: we're off on the weekend `:)`.

sjkim1127:
Thanks for reopening the PR and for the clarification — no worries about the weekend :)

I’ll leave it open as a draft while you triage #598 and review the recent changes in this area. I’ve also pushed a small follow-up commit adding the optional-header bounds check, a PE32+ regression test, and a malformed `SizeOfOptionalHeader` test.

I’ll hold off on further changes until you’ve had a chance to review. Thanks for taking the time!

PR #631, opened 2026-09-21
SegmentReferences: fix replacement of PHINodes

pfez:
Before this commit, EmitSegmentReferences was building invalid IR when replacing a PHI, because the insertion point of the replacement was the top of the PHI's own block.
This commit fixes the problem emitting the replacement at the end of the associated incoming block.

Whenever the incoming edge of the PHI is a back edge, replacements emitted like this end up in a block the traversal has not reached yet. So we keep track of when this happens to avoid considering the replacements emitted in this way for further replacement.


lib/SegmentReferences/EmitSegmentReferences.cpp
```diff
@@ -43,8 +51,21 @@ class EmitSegmentReferences {
   }
 
   void run(Function &F) {
+    ToSkip.clear();
```
aleclearmind: Move `ToSkip` to be a local variable and forward it to the only user out of `run` (`handleOperand`).

lib/SegmentReferences/EmitSegmentReferences.cpp
```diff
@@ -27,6 +29,12 @@ class EmitSegmentReferences {
   IntegerType *IntPtrType = nullptr;
   llvm::DenseMap<const model::Segment *, Function *> SegmentGlobals;
 
+  /// Replacements emitted into a block the traversal has not reached yet, which
+  /// must not be mistaken for original instructions once it gets there. Entries
+  /// are removed as they are encountered, so this only ever holds the ones
+  /// still pending a visit.
+  llvm::SmallPtrSet<const Instruction *, 16> ToSkip;
```
aleclearmind: This comment is not very useful, it basically implies details of places where it's used. I'd drop it.

lib/SegmentReferences/EmitSegmentReferences.cpp
```diff
@@ -94,16 +117,39 @@ class EmitSegmentReferences {
...
+
+    // Only a `PHINode`'s replacement goes into a block other than the one being
+    // visited, so only that one can be met again by the traversal. Everything
+    // else is emitted immediately before `I`, which the iteration over this
+    // block has already moved past, and recording it would leave an entry in
+    // `ToSkip` that is never matched and so never erased.
+    if (Phi != nullptr)
+      for (llvm::Value *V : Emitted)
+        if (auto *Emission = dyn_cast<Instruction>(V))
+          ToSkip.insert(Emission);
```
aleclearmind: I don't think injecting code and marking it as "to be skipped" is the right approach here.

The right approach is probably to split the `run` method into a part where we collect the things to patch and one part where we apply the changes.
It's going to be more invasive but it's the right way.

lib/SegmentReferences/EmitSegmentReferences.cpp
pfez: We don't have `ToSkip` anymore.

lib/SegmentReferences/EmitSegmentReferences.cpp
pfez: We don't have `ToSkip` anymore, nor the comment.

lib/SegmentReferences/EmitSegmentReferences.cpp
pfez: Done, see, if you like it.

lib/SegmentReferences/EmitSegmentReferences.cpp
```diff
@@ -43,6 +51,20 @@ class EmitSegmentReferences {
...
+    // looked at for replacement, which is something we want to avoid.
+    llvm::SmallVector<Replacement, 16> Replacements;
+    collect(F, Replacements);
+
+    for (const Replacement &TheReplacement : Replacements)
+      apply(TheReplacement);
+  }
+
+private:
+  void collect(Function &F, llvm::SmallVectorImpl<Replacement> &Replacements) {
```
aleclearmind: `Replacements` should be a return value

lib/SegmentReferences/EmitSegmentReferences.cpp
```diff
@@ -57,53 +79,68 @@ class EmitSegmentReferences {
...
-      return nullptr;
+      return;
 
     // Check if the address is mapped
     auto [Segment, Offset] = Binary.getSegmentFor(MaybeAddress);
     if (Segment == nullptr)
-      return nullptr;
+      return;
+
+    Replacements.push_back({ &Operand, Segment, Offset });
```
aleclearmind: Drop `Replacements` argument and `return { &Operand, Segment, Offset }`.

lib/SegmentReferences/EmitSegmentReferences.cpp
```diff
@@ -43,6 +51,20 @@ class EmitSegmentReferences {
   }
 
   void run(Function &F) {
+    // Decide everything before emitting anything. If we don't do this, the
+    // replacement of a PHI operand goes into the incoming block, which the
+    // visit may not have reached yet. In that case, when the visit finally
+    // reaches the incoming block, the replacement instruction would also be
+    // looked at for replacement, which is something we want to avoid.
```
aleclearmind: I'd drop this comment, or move it to the commit message.

lib/SegmentReferences/EmitSegmentReferences.cpp
pfez: Done. See if you like it.

aleclearmind: Nice.

PR #632, opened 2026-09-21
Preserve generic-vector register dependencies before ABI analysis

wakamex:
GPT-6-Astra-Max says:

> QEMU generic-vector helpers take void pointers into CPU state, so the existing typed-pointer analysis misses their register reads and writes. In the included 20-byte equality reproducer, the decompiled output drops the left-hand input loads, returns a constant comparison result, and reports an extra XMM return value.
> 
> I record CPU-state offsets when creating CSVs and specialize the shipped helper bodies before ABI analysis. Their actual loads and stores become scalar CSV accesses, preserving aliases and partial overlap without duplicating QEMU's vector operations or descriptor format. Unsupported calls retain their original operands and produce a diagnostic. Repeated specializations share the work.
> 
> The standalone 46-byte assembly regression checks recovered register arguments, return values, and comparisons in decompiled C. The LLVM regressions cover equality, byte addition, destination aliases, partial overlap, tail clearing, scalar operands, and unsupported inputs.
> 
> I built the engine from source with Orchestra and passed all 65 selected CTest cases: 42 unit-test groups, 22 LLVM pass regressions, and the new whole-ELF regression. With the same engine and the new pass disabled, both the argument/return and decompiled-C checks fail. I also compared the scalarized functions against the actual built QEMU helpers: seven standalone cases with 256 inputs each, totaling 1,792 calls and 143,360 mapped CPU-state bytes, with zero mismatches. Those executable comparisons check the transformation's correctness. I haven't run the full 490-test suite or the installation target.
> 
> This handles constant CPU-state pointers whose accesses specialize to mapped integer CSVs on little-endian storage. Calls that do not meet those conditions keep the existing helper behavior and emit a diagnostic. The end-to-end regression covers x86-64.


wakamex:
I asked my friend to describe the rationale:

The fix makes a vector helper’s register reads and writes visible before rev.ng reconstructs function arguments and return values.

The bug comes from two different representations of the same state. Lifted instructions access LLVM variables representing CPU registers. QEMU’s generic-vector helpers receive untyped pointers into its CPU-state structure. In the lifted code, those pointers appear as numeric offsets. Rev.ng wasn’t connecting those offsets to the register variables, so it could discard values that the helper actually needed.

That breaks the dependency chain:

`input pointer -> loaded bytes -> vector comparison -> return value`

The repair has four steps:

1. Record where register variables live. When the lifter creates a CPU-state variable, it now attaches its byte offset within QEMU’s CPU state. That provides an exact mapping from a helper’s offsets to the corresponding variables. It uses the actual layout supplied by the lifter. Offset recording ([lib/Lift/VariableManager.cpp:556](https://github.com/revng/revng/pull/632/changes#diff-376b1c588c7e49520a3c8b2649dfd3853e2438e775159e83edaf85b64aae5c48R556))
2. Specialize the actual QEMU helper. The pass imports its implementation and substitutes the known register offsets and constant parameters, such as the vector-size descriptor. LLVM then simplifies branches and unrolls loops where possible. The input bytes remain runtime values. This lets QEMU’s implementation define equality, addition, tail clearing, and other operations. Specialization ([lib/BasicAnalyses/MaterializeGVecCalls.cpp:75](https://github.com/revng/revng/pull/632/changes#diff-e0fbd6a67502d0afc5061aee594e5a8d2dea0159e1ad7d5e11d6a1ecc7fc76bbR75))
3. Replace CPU-state memory accesses with explicit register accesses. Once each memory access has a known offset, the pass translates it into loads and stores of the mapped register variables. Byte masks and shifts handle accesses covering only part of a variable. The comparison’s dependencies are now visible to ordinary dataflow analysis. Access rewriting ([lib/BasicAnalyses/MaterializeGVecCalls.cpp:188](https://github.com/revng/revng/pull/632/changes#diff-e0fbd6a67502d0afc5061aee594e5a8d2dea0159e1ad7d5e11d6a1ecc7fc76bbR188))
4. Insert the result before ABI analysis. The scalar operations replace the original helper call immediately after lifting. Argument and return recovery therefore sees the complete dependency chain. Pipeline placement ([python/revng/internal/pipeline.yml:137](https://github.com/revng/revng/pull/632/changes#diff-4aeadb67380fc8f53952a8d51d316c8bf08a832c95748a2ee7d8827be510a752R145))

Aliasing needs particular care. Source and destination may refer to the same register or overlap partially. During specialization, all pointers share one symbolic CPU-state base, and the rewriting preserves access order. Cleanup also removes artificial reads of old destination values when the helper completely overwrites them; otherwise those reads could create spurious function arguments.

The pass imports only needed helpers and caches repeated specializations. If it cannot resolve a call safely, it preserves the call and emits a diagnostic.

For the equality reproducer, exposing those dependencies restores comparisons using both input pointers and eliminates the extra XMM return value.

wakamex:
eli20:

Think of CPU registers as temporary variables.

The original program effectively does:

Put bytes from input A into register X.
Put bytes from input B into register Y.
Compare X and Y.
Return whether they match.

But rev.ng represents that comparison as a call to a QEMU helper. The helper receives numeric locations inside the CPU’s saved state, and rev.ng wasn’t connecting those locations to X and Y.

Consequently, it could decide that the earlier loads were unused and discard them. That’s how a real comparison became a constant-zero result.

Our fix:

1. Records which CPU-state locations correspond to which register variables.
2. Opens up the helper’s actual implementation and turns its hidden register accesses into explicit reads, comparisons, and writes.
3. Does this before rev.ng decides which values are needed.

Now rev.ng can follow the entire chain from both inputs to the comparison result.

wakamex:
eli12:

Imagine a recipe says:

1. Put apples in bowl A.
2. Put oranges in bowl B.
3. Run the mixing machine.

Someone simplifying the recipe doesn’t realize the machine uses those bowls. They cross out steps 1 and 2 as unnecessary. Now the recipe is broken.

That’s what happened here: the decompiler couldn’t see that a helper function used certain CPU registers, so it could remove the instructions that filled them.

Our fix opens up the helper’s instructions and shows exactly which registers it uses. The decompiler can then keep the needed values and produce the correct comparison.

wakamex:
eli5:

The computer put two things in boxes, then asked a helper to compare them.

Our tool couldn’t see inside the helper, so it thought filling the boxes was pointless and erased that step.

We made the helper’s work visible. Now the tool knows the boxes are needed and keeps them filled.

wakamex:
closed in favor of problem description in #633

